PR #30, opened 2015-02-12
459830 Fix jdk.dio.policy for SPI device access

bittailor:
Fix for the [bug 459830 - Error in the jdk.dio.policy shiped with Kura, that denies SPI access](https://bugs.eclipse.org/bugs/show_bug.cgi?id=459830)

Replace jdk.dio.spi.SPIPermission with jdk.dio.spibus.SPIPermission which is the correct fully qualified name for the SPI permission.


LucaDazi:
Thanks for the bugfix!

PR #44, opened 2015-06-26
Improvements to configuration service, web and SSL Manager

MMaiero:


dwoodard1:
This is a pull request from Matteo Maiero. Matteo has been nominated to be a Kura committer and is awaiting PMC approval. We are accepting this request even with the invalid signature. 

--Dave

PR #48, opened 2015-07-06
Java8 Compact 2 compliant code

MMaiero:
Updated code to be compliant with Java 8 Compact 2.
Updated poms in order to create uncompressed jars.
Reviewed bundles startup order.
Code refactoring.


dwoodard1:
Allowing pull request as Matteo is now a committer for Kura. He will begin making commits directly to repository.

--Dave

PR #64, opened 2015-10-16
Expose the "control prefix" as a setting to allow users to change the…

kartben:
… MQTT namespace used by CloudService

Signed-off-by: Benjamin Cabé <benjamin@eclipse.org>


dwoodard1:
Hi Benjamin,

Did you also change CloudServiceOptions.java? If not, the the control prefix is still hard coded to "$EDC". I can make the change, just wanted to make sure I wasn't missing something.

Thanks,
--Dave

kartben:
Mmmh not sure? What's hard coded is the default value for the control
prefix. The code is CloudServiceOptions was pretty much ready to allow for
the control prefix to be overloaded via the properties map given when
constructing the CSOptions class (a remains from the past where this used
to be a setting, maybe?)

dwoodard1:
Yes, you are correct, forgot we had the option to override the default.

--Dave

PR #66, opened 2015-10-26
Fix bug 480434 - Multiple bluetooth le devices connection problem;

gavinying:
The existing code limits only one bluetooth device to connect by using a fixed thread pool  (each device consumes two threads), considering the number of actual connected bluetooth device could not be very big in practical use case, so use a cached thread pool should be OK.


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=480434

pierantoniomerlino:
The current implementation of the bluetooth service uses a fixed thread pool and, indeed, this limits to one the maximum number of simultaneous connections. Changing to a cached thread pool fix the problem.

PR #67, opened 2015-10-26
Trace log messages use common string formatting instead of concatenation

darugnaa:
Cleaned up a little the TRACE log messages. Instead of string concatenation, the regular string formatting is used (passing variables to be logged in the trace method call).
This should also be a small performance gain when a higher log level is used (which is the default setting).
I did a quick build and no compile time errors showed up.
I am thinking about improving also DEBUG and INFO levels formatting, if you think it's worth.


cdealti:
Looks good to me, thanks! As a rule, if you make changes to any bundle you should also uptick the the version of the bundle in ```distrib/config/kura.build.properties``` otherwise the build of the ```distrib``` project (and hence the installer) will use the last released version of the bundle.

Also note that changes affecting many bundles have implications on the size of the upgrader (it contains only the bundles whose version has changed since the last release).

darugnaa:
Thank you, I took notice. Do you wish me to uptick the versions in this pull request too?
I thought about submitting other two pull requests, one for DEBUG and one for INFO in the future, since they are quite big as they involve lots of files.

PR #68, opened 2015-10-28
Add ethtool dependency to kura nn version;

gavinying:
A patch for bug 480606. Tested on Raspberry Pi 2. 

Signed-off-by: Ying Shaodong <helloysd@gmail.com>

### DIFF
--- a/kura/distrib/pom.xml
+++ b/kura/distrib/pom.xml
@@ -742,7 +742,7 @@
 		<profile>
 			<id>beaglebone-nn</id>
 			<properties>
-				<project.raspbian.dependencies.nn>dos2unix, unzip, telnet</project.raspbian.dependencies.nn>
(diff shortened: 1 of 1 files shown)


cdealti:
+1 for me. Thanks.
@ibinshtok since the ```org.eclipse.kura.linux.net``` is installed and ACTIVE even in the ```nn``` installer and the services it provides could be used by client bundles other then the ```org.eclipse.kura.net.admin``` bundle, all the networking-related Debian packages should be installed.

gavinying:
@cdealti if network serivces are fully used, especially working as a router, I guess kura with network management could be the choice. However, in most of cases, we just use the device as a network client, so I think there's no need to install the other packages (bind9, isc-dhcp-server and hostapd) for nn version, installing ethtool is just to eliminate the exception under the current setting.

PR #92, opened 2016-01-13
Added BLE beacon scanning functionality

lmarshall-eth:
Added BluetoothAdapter.startBeaconScan method, which begins an ongoing scan for iBeacon devices.
Each received packet fitting the iBeacon advertising format is returned via the BluetoothBeaconScanListener interface.

'hcidump' is required in order for this functionality to work correctly, as we need to get at the iBeacon data which hcitool doesn't return. It could also be made to work with tcpdump.

Signed-off-by: Lee Marshall <lee.marshall@eurotech.com>


lee-marshall-eurotech:
Have now signed the CLA - not sure what I have to do to 'refresh' this pull request.

PR #95, opened 2016-01-26
[doc] install mToolkit from update site

kartben:
Change mToolkit instructions and indicate to use regular update site instead of dropins-based install.


kartben:
ping?? 
FWIW I think the Getting started need some more love. The quick start for Raspi and BBB both have wrong URL and filenames in the "Target platform installation" section.

cdealti:
Hi @kartben. This will be merged asap. We are a little busy with the Kura 1.4.0 release.

Issue #88, opened 2016-01-06
Pure Java Kura


dwoodard1:
This bug was originally filed at https://bugs.eclipse.org/bugs/show_bug.cgi?id=481529.

Original Message:

This bug will track a feature request to create a minimal version of Kura with no native code dependencies. At present, the SystemService in Kura depends on native libraries to start. This dependency restricts the possibility of running Kura on very resource constrained devices. This request will remove native library dependency on core Kura bundles.

Thanks,
--Dave

Does "Pure Java Kura" mean that it must be OS independent? This would also require removing references to OS system files (e.g. /etc/) or spawning external processes via Runtime.exec() or similar.

dwoodard1:
I think the idea is to remove dependencies that can't be controlled with Java, specifically the native libraries. Most of the OS system file references are due to the networking bundles. If someone doesn't need networking support, they are still gated by dependencies on native libraries.

More importantly, I believe the core of Kura should start without loading native libraries. If a service is needed that depends on native code, it would be up to the user to compile and build Kura to meet their needs.

--Dave

dwoodard1:
@cdealti Should we close this? I am not sure if removing the udev dependency is realistic considering our backlog of work.

PR #107, opened 2016-03-15
Kura Windows

malikm:
Another attempt signed with my own blood !


kartben:
It looks like you haven't signed a CLA in the first place, or have you?
https://projects.eclipse.org/user/206/cla

malikm:
Yes, I did sign it and also a validation check says it’s OK. Don’t understand why it won’t accept it.
However now when I check https://projects.eclipse.org/user/206/cla I get Access Denied and the validation page says it’s unavailable.
What’s this mess ?

Regards,
Martin

kartben:
Sorry, I did send a wrong link. You do have a CLA in place indeed, sorry I should have directly checked at https://projects.eclipse.org/user/cla/validate

I think we'll need webmaster help here. My guess is that trying to do a pull-request for a branch that contains commits from authors other than you is causing an issue. In general, you may want to rebase your work before doing a PR anyways, in order to "hide" the WIP steps.

malikm:
Well, our previous attempts showed other errors too, like other contributors not having CLA signed, or problems with signed-off-by. Those should be resolved by now, so I think there must be some other problem. How can we contact the webmaster ?

kartben:
I've opened https://bugs.eclipse.org/bugs/show_bug.cgi?id=489685 for you, and I've cc'd you. Thanks for your patience (and pretty excited to see support for Windows coming to Kura btw!)

malikm:
Thanks for your help Benjamin !
Originally we had a full Windows fork with all the changes/commits during development, but since we were not aware of the signed-off-by requirement, we needed to re-do all our commits from scratch in the hope that it will finally work ;-)

kartben:
Yay. CLA validation passes!
@dwoodard1 and team, this is good to be submitted in a CQ (contrib is >1000 LOC). Looks like this will bring some build time dependency on the NSIS tool.

malikm:
Great, thanks guys !

cdealti:
Nice! @dwoodard1 can you submit the CQ for this contribution?

kartben:
FWWI any Kura committer can create a CQ :-) Let me know if you have any
questions, I'd be happy to help.

On Thu, Mar 17, 2016 at 9:37 AM, Cristiano De Alti <notifications@github.com

cdealti:
Hi Mark an Martin,
  Thanks for your nice contribution.
I'm moving forward filing the CQ for this pull request.

My understanding is that at a minimum I need to file a CQ only for your original work, i.e. no third party libraries or executable are redistributed with this Windows port.

Also it looks to me that the native libraries included in this PR as original work only use the Win32 API. @kartben I don't expect any issue linking against the Microsoft system libraries, can you confirm?

The above native libraries are precompiled and distributed in binary form.

However, the build scripts of this Windows port can **optionally** generate an ``NSIS`` installer if the ``NSIS`` system is present on the build machine. If the ``NSIS`` system is not found on the building machine (e.g. Hudson) the build script just assembles the Kura installation directory.

I believe that the producing an installer is the right way to go on Windows but I'm not sure if we are allowed to install the ``NSIS`` system on the Hudson CI. @kartben Can you kindly check this?
In this case I believe the Kura installer for Windows would redistribute some ``NSIS`` runtime for which we would need to file a separate CQ.
As said, the installer is optional and we can live without it for the moment.

About your PR:
* can you please double check that every file you added contains the Eclipse license header

* can you also double check that the license header reflects the latest format (with the copyright date updated to 2016), for example: https://github.com/eclipse/kura/blob/develop/kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/KuraException.java

* cfab4c2c3b629f96317078336f60bfce4d9fe087: I understand this was needed however it entirely eliminates the possibility of overriding the working directory via the cloud request. Unfortunately EC always sends the request with the working directory set to "/tmp" which would not work in your case.
For the time being let's live with this.

* 28c4873eb458f78da2ba6f1c7fdb6fb223214018: Please take care of indentation, spacing and replace calls to ``e.printStackTrace()`` with the logger. The worst part is that you had to add a dependency to an implementation package ``org.eclipse.kura.windows.system`` to get the system uptime and MAC address. This shows once more that org.eclipse.kura.core needs a refactor to remove OS dependent code.

* 1f9919b93236a53ad903abf82872ef237ae28691: I believe that 99% of the code is the same as ``org.eclipse.kura.linux.clock`` and that this OS independent code needs to be refactored elsewhere.
If possible, correct minor indentation issues.

* 497a243f5b2c1e0f67de8df0eb9fa6a055926f5e: Please remove the contents of kura/org.eclipse.kura.windows.net/src/main/resources since they don't apply to Windows.

* 34197fb49aabfa7e88831fd175508891d749531e: I don't see the need for this commit/bundle. Is there something Windows specific I'm missing?

* 20fb65b696fc8c2d414e467b9220699550094c45: same as above. Is there anything Windows specific?

* a9c224f170431e8b0b0e11096501b9f472b3262d: I believe you might want to use org.eclipse.kura.emulator.usb instead

* a9f21dc74d0a25270a61659c50c9775fa2c21441: I believe you might want to use  org.eclipse.kura.emulator.watchdog instead

malikm:
Thanks for the review Cristiano.
We will check the files required, meanwhile some comments.
Most of the Windows versions of bundles have only basic functionality, to make sure it works under Windows with the potential of extending the features later when needed.

https://github.com/eclipse/kura/commit/1f9919b93236a53ad903abf82872ef237ae28691 : Yes, most code is similar to Linux, but still contains a few pieces of Windows native code.

https://github.com/eclipse/kura/commit/34197fb49aabfa7e88831fd175508891d749531e : This one has removed the Linux-specific /dev requirements. It doesn't have dedicated USB support yet, which can be added later.

https://github.com/eclipse/kura/commit/20fb65b696fc8c2d414e467b9220699550094c45 : Needed to have the status service. Since there's no standard implementation of GPIO access in Windows, this currently relies on the emulator. Again, can be extended later with native code when needed.

https://github.com/eclipse/kura/commit/a9c224f170431e8b0b0e11096501b9f472b3262d : The org.usb4java / libusb4java seems to have native support for Windows as well. We haven't tested this yet, but we hope to utilize USB functions using that package. That will then most probably require a dedicated Kura USB package as well.

https://github.com/eclipse/kura/commit/a9f21dc74d0a25270a61659c50c9775fa2c21441 : This one is similar to GPIO. It was included for later extension if possible. Currently just a dummy one.

cdealti:
Hi Martin,

> 34197fb : This one has removed the Linux-specific /dev requirements. It doesn't have dedicated USB support yet, which can be added later.

Why don't you modify org.eclipse.kura.linux.position by adding a check like this:

			if(m_properties.get("port") != null){
				portName = (String) m_properties.get("port");

				if(portName != null) {
					if(portName.contains("/dev/") || portName.contains("COM")) {
						File f = new File(portName);
						if(f.exists()) {
							return true;
						}
					} else {
						List<UsbTtyDevice> utd = m_usbService.getUsbTtyDevices();	
						if(utd!=null){
							for (UsbTtyDevice u : utd) {
								if(portName.equals(u.getUsbPort())) {
									return true;
								}
							}
						}
					}
				}
			}

In this way you don't need an UsbService implementation for Windows and you can use the emulated one? I prefer to avoid having duplicated code if possible.

> 20fb65b : Needed to have the status service. Since there's no standard implementation of GPIO access in Windows, this currently relies on the emulator. Again, can be extended later with native code when needed.

Please remove this if you don't need it.


> a9f21dc : This one is similar to GPIO. It was included for later extension if possible. Currently just a dummy one.

If it's dummy, please remove and use the emulated one.

The above changes would reduce the number of lines of code of this contribution, making the CQ unneeded.

malikm:
https://github.com/eclipse/kura/commit/34197fb49aabfa7e88831fd175508891d749531e : Do we really want to include org.eclipse.kura.linux.position in the Windows package (with your proposed changes) ? Wouldn't it be confusing ?

https://github.com/eclipse/kura/commit/20fb65b696fc8c2d414e467b9220699550094c45 : If we remove the Windows status service, what should be used to replace it? Isn't this service mandatory ?

cdealti:
>34197fb : Do we really want to include org.eclipse.kura.linux.position in the Windows package (with your proposed changes) ? Wouldn't it be confusing ?

I don't have problems with this. I don't even find Linux specific code in the bundle.
In the future we could just rename it org.eclipse.kura.core.position.

>20fb65b : If we remove the Windows status service, what should be used to replace it? Isn't this service mandatory ?

Please replace with org.eclipse.kura.core.status (OS independent) and org.eclipse.kura.emulator.gpio

malikm:
Done all changes based on review, except refactoring.

Due to the unsigned commit https://github.com/eclipse/kura/pull/107/commits/6d65208a2a23b100ff72a6f6c4538e204f27ec48
I have rebased + squashed all commits after the PR into one with signed-offs. Hopefully it's OK now.

cdealti:
Hi Martin,
  The PR is now very difficult to read due to commits reverting changes contributed initially.
I'd try to cleanup the PR by attempting to squash related commits together.

For example a12383fede028076ba0da73c4dccc94b14f55c7a partially reverts 497a243f5b2c1e0f67de8df0eb9fa6a055926f5e. They should be logically squashed (via ``git rebase -i`` together. The resulting squashed commit should contain only the net additions, and will be easier to read.

You should try to squash together the other commits. You can try this incrementally until you are satisfied.

Another recommendation is to never merge a PR with the branch tracking the upstream develop branch. Use git rebase instead. I'm not sure c6a15c801f954d1eb3f9271a5136234e85cd9ec9 should appear as you last commit if you rebased. 

The final step should be to force a push (git push -f) from your local branch to malikm:KURA_WINDOWS. This will result in all the above commits being deleted and replaced with the squashed ones. At that point the PR should resemble the initial submissions and should be easy to read and review (like the original).

I know, it's a pain

malikm:
That merge was certainly not intentional, don't know how it happened. The only thing that worked was a hard reset to previous commit. After that a forced push did finalize the squash, so now there's just a single commit after PR.

cdealti:
OK, let's see if I can help directly looking at your fork

cdealti:
This PR is still strange and difficult to read or merge.
I expect to have a series of commits whose diffs **only** contain `+`.
Here we have initial commits of the PR followed by 290b676 which "undoes" (lots of ``-`` in diffs) most of the initial contribution.

I'd suggest to restart from the original commit series by making the requested changes one at a time and squashing appropriately to obtain commits with only "positive" diffs.

Some commits can be simply deleted via ``git rebase -i``, for example  34197fb.

I've also found the following issues:

* a4e14dc: buildkura_windows references a kura_eth local repository and Eurotech targets (e.g. Minigateway).
Please remove references to the kura_eth repository and targets.
In the same commit, what is the reason to add the ``if(getOsName().contains("Windows"))`` code to ``getPrimaryMacAddress()``?
Squash the resulting commit with this commit.


* 28c4873: you added a dependency to ``org.eclipse.kura.core`` from org.eclipse.kura.windows.system via ``Import-Package``.
Ideally org.eclipse.kura.core should be pure Java/OSGi (see #88).
  * What is the bundle exporting this package?
  * Does this mean that the Linux-based distributions needs this bundle too?
  * Wouldn't it fail to load the Windows native libraries on Linux?

malikm:
I agree that this PR has become difficult to read. That was mostly caused by the unsigned commit in the past...
It seems that the only way is to abandon this PR and start from scratch - create a new branch for PR, add all code changes with a single commit and do a new PR.
As for the other issues:
- buildkura_windows could/should be removed entirely
- I agree that org.eclipse.kura.core should not depend on org.eclipse.kura.windows.system. Since the native code is used on Windows systems only, it will not attempt to load the native library on Linux. What other solution do you suggest ?

cdealti:
> It seems that the only way is to abandon this PR and start from scratch
Can you adopt a strategy where both you and Mark appear as the authors of the contribution?
Maybe splitting the PR in two non-overlapping PRs or a single request with 2 non-overlapping commits?

> Since the native code is used on Windows systems only
Maybe, for the time being, just return 0 as the uptime for windows. We need to figure out how to remove the Linux and OSX dependencies from core in any case.

Alternatively, is there any way to get the uptime from Windows calling an external process?

malikm:
Should be possible using "net stats server". Will check this and replace if working.

markcullen:
I'm working with Martin on the Windows support for KURA. We would both like to be considered as commiters so that we can help to maintain and improve the Windows port. However we understand that it may be more difficult to get a pull request with multiple authors included so the question is should we provide one pull request with all the required changes or submit two pull requests one from each author that are applied together.

cdealti:
@malikm @markcullen Obsoleted by #313 and #319.

Issue #93, opened 2016-01-18
[Configuration] Empty Values In Service Properties


cdealti:
If I set the value of a service property to an empty string in the web UI and hit Apply the empty string is accepted and the corresponding service gets the empty string. When KURA gets restarted the value reverts to the default value from the metatype file. If I change the value to a non empty string the value is preserved across a KURA restart.
I'm changing command.working.directory but see the same behaviour on other properties.

cdealti:
As far as I know if the configuration with an "empty" value is validated by the Web UI, it means that the ```required``` attribute of the property in the service metatype is set to false.

* Does the service receives the ```null``` or the empty (```""```) String?

* How the value of the property is represented in the new xml snapshot that is created after applying the configuration change?

> When KURA gets restarted the value reverts to the default value from the metatype file

This surprises me. On startup, Kura should merge the configuration (from the latest snapshot file) with defaults (from the metatype) only for properties whose ```required``` attribute is set to true in the metatype.

cdealti:
See #109.

PR #109, opened 2016-03-18
Remove default values for username and password as they are optional

wnagele:
The parameters 'username' and 'password' have been made optional in #63. Because they were still having default values you could set them to empty values (for instance needed by AWS IoT) but once you restart Kura they would always be coming back with their default values. Setting them to empty defaults as many other optional parameters already use.

### DIFF
--- a/kura/org.eclipse.kura.core/OSGI-INF/metatype/org.eclipse.kura.core.data.transport.mqtt.MqttDataTransport.xml
+++ b/kura/org.eclipse.kura.core/OSGI-INF/metatype/org.eclipse.kura.core.data.transport.mqtt.MqttDataTransport.xml
@@ -41,7 +41,7 @@
             type="String"
             cardinality="0"
             required="false"
-            default="username" 
+            default=""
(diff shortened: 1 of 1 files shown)


MMaiero:
Hi Wolfgang,
thank you for your pull request. Currently we are evaluating if a more complete change to the configuration service is needed, in order to manage also empty values (refer also to bug #93). 
In my opinion, default values should be allowed also for non-required fields. We should be able to evaluate the configurations of running bundles and store in snapshots the empty properties of optional parameters. In this way we could be able to have defaults, but preserve the possibility to have an empty properties.

We will keep updated bug #93 and the feedback of the community is more than welcome for this issue.

Thanks,
Matteo

wnagele:
Default values of _username_ and _password_ for configuration options that are called _Username_ and _Password_ seem super pointless to me - no matter if the handling of default values changes. ;) Default values for other fields make good sense - but these properties should never use default values in a security context ...

cdealti:
@wnagele As of Kura 2.0.0 empty strings are supported in the configuration.

PR #115, opened 2016-04-12
Added ifconfig up and ifconfig down also for non yocto-based devices.…

MMaiero:
… Modified the logical behavior of the disableInterface method.


cdealti:
40960ba: remove comment ``Can we unify the below cases?`` in ``disableInterface``.

4579767: remove ``local`` prefix from localNewNetConfiguration.

955426a: ``localEnabledInterfaces`` and ``localDisabledInterfaces`` are not used in the scope they are declared. Move them in ``updateInterfacesLists()``. Also I don't quite like the prefix ``local``. Please omit it.

After changing use ``git rebase -i`` to squash related commits in order to only have "positive" diffs.
Also use to ``git push -f origin fix_ifdown_error`` to update the PR.

Eventually you should still 3 commits.

PR #180, opened 2016-05-03
Hotfix/#179 ble compatibility issue; close #179

### LINKED ISSUE #179: BluetoothLE: compatibility issue
There are at least 2 different types of gatttool response so far, however the existing code only works with the first scenario, 

Scenario 1: 
[   ][BC:6A:29:AE:CC:33][LE]> connect
[CON][BC:6A:29:AE:CC:33][LE]>

Scenario 2: 
[54:4A:16:5A:14:C6][LE]> connect
Attempting to connect to 54:4A:16:5A:14:C6
Connection successful
[54:4A:16:5A:14:C6][LE]> 

It took me quite a few hours to figure it out, nevertheless, I have fixed it and will share it later.

gavinying:


pierantoniomerlino:
Great! I'll test the fix asap. Thank you for the contribution!

cdealti:
@pierantoniomerlino please test this but don't merge this yet. Waiting for more feedback

pierantoniomerlino:
Tests:
- [ ] Bluez 4.101
- [ ] Bluez 5.23
- [ ] Bluez 5.33
- [ ] Bluez 5.39

pierantoniomerlino:
@gavinying 
Hi,
I did some test on the bluez-5.x, but I wasn't able to work with that.
It seems that the line returned to the Gatttool parser is like "[KConnection successful", so we have to check if the line contains the successful string.
Moreover, why 80e05126491c00b8a4f997b22febaf67c60ec5aa ?

Finally, it seems that the checkConnection method fails with bluez-5.x because a "\n" command doesn't trigger a new line.

gavinying:
@pierantoniomerlino 
You are right. In that case we would need regex to match the pattern as well. I will improve it. 
BTW, I thought only bluez 4.x is supported by kura, so now the latest 5.x version is also supported, right? 
for 80e0512 , I think once it receives REGEX_NOT_CONNECTED string, the `ready` flag shall be set to `true` so that the while loop in connect() won't keep waiting for timeout. 
Yes, for this type of gatttool response, it seems no "\n" response, have to think about the other way, however, I could prefer to do device connection checking in application layer.

pierantoniomerlino:
@gavinying 
IMHO, probably `String.contains()` is better than a regex for a static string.

Yes, Kura supports only Bluez-4.x. AFAIK, Bluez-5.x can be supported if we change the way Kura parses the output of Gatttool. I see that the problem is how to detect that a connection is successful (parsing "Connection successful" string as you suggested), that a connection is still alive (checkConnection() method) and that a device is not connected (we can't use the \n on gatttool).

For 80e05126491c00b8a4f997b22febaf67c60ec5aa, this makes sense, so we can remove the timeout check in the connect method.

gavinying:
Agree. We shall use String.contains(). 
Use `80e0512` could only avoid the waiting time when receives REGEX_NOT_CONNECTED string, like `[   ]`, but there's another situation that may never receive any response, so I think we should still keep the timeout checking in connect().

pierantoniomerlino:
@gavinying 
Thank you for the commit.
Regarding the m_ready variable, I think that we should set it to false in the REGEX_NOT_CONNECTED situation. When the method connect() is run, the processLine() method parse some lines before REGEX_CONNECTED. That lines trigger REGEX_NOT_CONNECTED, so the m_ready must be false, otherwise the connect() method returns with m_connected set to false.

gavinying:
@pierantoniomerlino 

To clarify this, I did some testing and also searched online, here are 4 scenarios I met so far,  
 
Case A: 
[   ][BC:6A:29:AB:DE:2B][LE]> connect
[CON][BC:6A:29:AB:DE:2B][LE]>

Case B: 
[   ][BC:6A:29:AB:DE:2B][LE]> connect
Connecting... connect error: xxxxxx
[   ][BC:6A:29:AB:DE:2B][LE]>

Case C: 
[54:4A:16:5A:14:C6][LE]> connect
Attempting to connect to 54:4A:16:5A:14:C6
Connection successful
[54:4A:16:5A:14:C6][LE]> 

Case D: 
[54:4A:16:5A:14:C6][LE]> connect
Attempting to connect to 54:4A:16:5A:14:C6
Error: connect error: xxxxxx
[54:4A:16:5A:14:C6][LE]> 

Did you find more cases that will unexpectedly trigger REGEX_NOT_CONNECTED condition during connect()? If it does, maybe we shall adjust the regex pattern to fix it. 

On second thought, if the m_ready cannot be set in condition that defined as REGEX_NOT_CONNECTED, then there's no point to use m_ready variable any more. 
What do you think?

pierantoniomerlino:
@gavinying 
When the bundle tries to connect, the processLine() parses

2016-05-11 14:21:22,163 [BluetoothProcess Input Stream Gobbler] DEBUG o.e.k.l.b.l.BluetoothGattImpl - Processing line : [   ][BC:6A:29:AE:CC:96][LE]>
2016-05-11 14:21:22,168 [BluetoothProcess Input Stream Gobbler] DEBUG o.e.k.l.b.l.BluetoothGattImpl - Processing line :  connect

2016-05-11 14:21:22,370 [BluetoothProcess Input Stream Gobbler] DEBUG o.e.k.l.b.l.BluetoothGattImpl - Processing line : Connecting... 
2016-05-11 14:21:22,371 [BluetoothProcess Input Stream Gobbler] DEBUG o.e.k.l.b.l.BluetoothGattImpl - Processing line : [CON][BC:6A:29:AE:CC:96][LE]>

So, the first line trigger a REGEX_NOT_CONNECTED. The lines are taken from kura.log.
However, the m_ready is used to avoid to wait until timeout when sending commands. But probably there is a better way to do this...

@gavinying @ccustine BTW, I tried to solve the issue regarding the checkConnection and parse connections/disconnections for the #209 and I create the new branch kura-bluez5. What do you think about this?

MMaiero:
Already fixed by PR #217. Closed and not merged.

PR #194, opened 2016-05-05
-Service Factory API and implementation

### LINKED ISSUE #182: Move Multitons API and implementation from kura2 to develop

LucaDazi:
-Multiton support in ConfigurationService and web UI

Signed-off-by: LucaDazi <luca.dazi@eurotech.com>

Fixes #182 

Edits to ConfigurationServiceImpl.
component.name no longer used to track Services in the WebUI. service.pid used instead.
Some test needed before merging into develop.


cdealti:
@LucaDazi wouldn't be better if the methods added to the ``ConfigurationService`` mapped more closely those of the ``ConfigurationAdmin``? For example:

* ``newComponent`` should be renamed ``createFactoryConfiguration`` which should return a ``ComponentConfiguration``.
* In turn ``ComponentConfiguration`` should provide a ``delete`` method (like the ``Configuration`` interface of the Configuration Admin spec)
* At that point there will be no need for ``deleteConfigurableComponent``

What do you think?

kartben:
FWIW, and I know this is late to the game, I think it is a bit confusing to
use the term "multiton", when in the end we're just talking about
"components", really. I guess my point is along the lines of Cristiano's
comments: it would be nice to align as much as possible with the classic
OSGi terminology rather than have additional terms / different APIs than
what OSGi already provides.
My 2 cents :)

LucaDazi:
@cdealti @kartben Totally agree.
Only thing, though: we have already "broken" the OSGi Configuration Admin contract by putting the `update` methods inside the `ConfigurationService`

At this point I think leaving the deletion API inside the `ConfigurationService` makes sense.
API signature however should be changed from `deleteConfigurableComponent` to `deleteConfiguration`.

Pull Request updated.

kartben:
FWIW my point was more regarding the "Multiton" term which I believe is
confusing. At least it seemed to be to the people I started to discuss this
upcoming feature with.

cdealti:
@LucaDazi 
  it looks good, however:

* The OSGi spec prefers the expression ``Factory Component`` to ``Component Factory``. Can we rename new methods accordingly?
* Maybe it will be more efficient if I come to your desk ;)

@kartben Multiton was an internal nick name that is not going to appear anywhere (API, Javadoc etc.).

kartben:
@cdealti Oops, I didn't realize that, sorry! Sorry for ranting then :) That's going to be a really cool new feature that I'm really looking forward to!

cdealti:
@LucaDazi accordingly to Semantic Versioning the changes in the MANIFEST of  ``org.eclipse.kura.configuration`` **implementors**  should be [1.1,1.2)

LucaDazi:
@cdealti that's true for strict Semantic Versioning.
However in this case bundles implementing APIs from `org.eclipse.kura.configuration` are mostly using the ConfigurableComponent interface.

Changing the versioning of all the bundles using `ConfigurableComponent` is a bit odd, to me.
I totally agree on separating APIs from marker interfaces, abstract classes and listeners, as per #226

Issue #181, opened 2016-05-03
[GPS] position service: code review


pierantoniomerlino:
`org.eclipse.kura.linux.position.PositionServiceImpl` has become messy.
Its configuration suggests an exclusive access to a serial port (UART, USB-to-serial) providing the NMEA sentences. This works well with standalone GPS receivers (for example an Ublox GPS receiver).

In the course of time we had to support NMEA sentences provided by Telit modems which combine a modem and a GPS receiver in the same module.

In Kura modems are managed by the net admin bundle and the modem configuration and initialization happens there. Moreover the net admin may reset the modem if it stops responding or hangs.
Also, on some modems the same serial port is used for sending AT commands (by the net admin) and receiving the NMEA sentences (by the position service).

To cope with this some workarounds have been added to the position service implementation.
These are based on various OSGi events resulting in a messy design and unreliable operation.

P1: check reliable operation when GPS is provided by the modem and:
- The modem is reset by net admin
- Kura is restarted

Postponed:
- When Kura is restarted and GPS service was enabled, the network service can't communicate with the modem and reset it
- How to manage USB and built-in modem GPS device
- Modify GPS enable AT command
- Filter USB events
- Check the configuration passed to the bundle

pierantoniomerlino:
P1: check reliable operation when GPS is provided by the modem and:

- [ ] The modem is reset by net admin
- [ ] Kura is restarted
Postponed:

- [ ] When Kura is restarted and GPS service was enabled, the network service can't communicate with the modem and reset it
- [ ] How to manage USB and built-in modem GPS device
- [ ] Modify GPS enable AT command
- [ ] Filter USB events
- [ ] Check the configuration passed to the bundle

pierantoniomerlino:
Partially solved in #224

pierantoniomerlino:
The configuration provided by the configurable component is relative to the USB GPS. When a modem GPS is enabled, the latter is used. When the modem GPS is disabled, the USB is used.

cdealti:
@ibinshtok can you please review?

ibinshtok:
@pierantoniomerlino , I am getting these exceptions when switching from modem GPS back to onboard GPS. This is on HE910 modem.
2016-06-08 14:31:24,336 [GpsSerialCommunicate] ERROR o.e.k.l.p.GpsDevice - Exception in gps read - {}
java.lang.NullPointerException
[...]

PR #224, opened 2016-05-17
Fixed modem reset and kura restart issues for gps

pierantoniomerlino:
This PR partially solves the issue #181. 
In particular, the modem reset and kura restart issues are addressed.


pierantoniomerlino:
Fixed the remaining bugs in #181

pierantoniomerlino:
@ibinshtok 
could you please review this pull request?

Thank you!

cdealti:
@ibinshtok Can you please review?

kura/org.eclipse.kura.linux.position/src/main/java/org/eclipse/kura/linux/position/GpsDevice.java
```diff
@@ -314,11 +325,7 @@ public boolean doPollWork() {
 							c = in.read();
 						} catch (Exception e) {
 							s_logger.error("Exception in gps read - {}", e);
-							try {
-								Thread.sleep(1000);
-							} catch (InterruptedException e1) {
-								s_logger.warn("Interrupted - {}", e1);
```
ibinshtok: @pierantoniomerlino would removing try/catch block around Thread.sleep(1000) here cause the doPollWork() method return true instead of false? In original code, if in.read() throws an exception, we would seep for 1 sec and return false which, in turn, will cause disconnect. Now, if in.read() throws an exception and Thread.sleep(1000), in the catch block throws InterruptedException, it would be caught in the outer catch block and the doPollWork() method returns true; no disconnect will follow. Is this desired behavior?

kura/org.eclipse.kura.linux.position/src/main/java/org/eclipse/kura/linux/position/PositionServiceImpl.java
```diff
@@ -318,12 +329,21 @@ else if(UsbDeviceRemovedEvent.USB_EVENT_DEVICE_REMOVED_TOPIC.contains(event.getT
...
+				m_properties.put(BITSPERWORD, event.getProperty(ModemGpsEnabledEvent.DataBits));
+				m_properties.put(STOPBITS, event.getProperty(ModemGpsEnabledEvent.StopBits));
+				m_properties.put(PARITY, event.getProperty(ModemGpsEnabledEvent.Parity));
+				m_properties.put(MODEM, "true");
+				
+				if (m_gpsDevice != null) {
+					Properties currentConfigProps = m_gpsDevice.getConnectConfig();
+					Properties serialProperties = getSerialConnectionProperties(m_properties);
+					if ((currentConfigProps != null) && (serialProperties != null)) {
+						checkProperties(currentConfigProps, serialProperties);
```
ibinshtok: @pierantoniomerlino , the checkProperties() is a void method that only compares current properties with serial properties and logs the "same configuration, no need to reconfigure GPS device" if properties are equal. Don't we want to call the updated() method (in the handleEvents()) based on result of this comparison?

kura/org.eclipse.kura.linux.position/src/main/java/org/eclipse/kura/linux/position/GpsDevice.java
pierantoniomerlino: You are right. My goal was to simplify and make more readable the code. I'll encapsulate the sleep and catch in a private method.

kura/org.eclipse.kura.linux.position/src/main/java/org/eclipse/kura/linux/position/PositionServiceImpl.java
pierantoniomerlino: @ibinshtok You are totally right! I forgot to change the method return type after a refactoring... However the logic is the same : when a GPS enabled will arrive, the method check if the properties are changed. If so, it triggers an update.

pierantoniomerlino:
@ibinshtok The last commit should fix the bugs. Please take a look on it. 
Thank you for the comments!

pierantoniomerlino:
@ibinshtok : did you take a look on the last modifications?

pierantoniomerlino:
@ibinshtok : I updated the PR adding a fix for the enable/disable procedure. However I wasn't able to replicate the issue you reported with the HE910. I tested the code both with a HE910 and GE910.
Could you please take a look on this PR?

dwoodard1:
@pierantoniomerlino @ibinshtok What is the status of this PR? Can we rebase and merge?

ibinshtok:
@dwoodard1 @pierantoniomerlino I think we need to rebase and retest before we merge.

pierantoniomerlino:
@dwoodard1 @ibinshtok : I'll rebase the branch asap.

cdealti:
Closing due to history rewrite.

PR #225, opened 2016-05-18
Fix wifi config hide passwords

ibinshtok:
@MMaiero, can you verify this, please? I have tested it on my RG 10-11 having wlan0 in AP mode and wlan1 in Station mode. Both hostapd and wpa-supplicant used respective config files from the /tmp folder which were removed after startup. Respective configuration files in the /etc/ directory have passwords encrypted.


kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/configuration/ConfigurationService.java
```diff
@@ -80,6 +80,13 @@
 	 */
 	public ComponentConfiguration getComponentConfiguration(String pid) throws KuraException;
 	
+	/**
+	 * Returns component configuration properties for specified PID
+	 * @param pid - the ID of the component whose configuration is requested
+	 * @return component configuration properties
+	 * @throws KuraException
+	 */
+	public Map<String, Object> getConfigurableComponentConfigurationProperties(String pid) throws KuraException;
```
MMaiero: @ibinshtok I know that @LucaDazi is working on similar methods in his multi ton branch. Maybe we should get in touch with him in order to know if there is a similar method in his implementation that can be used here

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/configuration/Password.java
```diff
@@ -13,24 +13,26 @@
 
 public class Password 
 {
-	private char[] password;
+	protected char[] m_password;
```
MMaiero: @ibinshtok why we need to widen the visibility? I've seen that in WifiPassword.java you are using m_password but, wouldn't it better to keep the password variable private and use the public getPassword to return the password in the subclass?

ibinshtok:
@MMaiero sure, I'll modify m_password beck to private and use getter.

kura/org.eclipse.kura.linux.net/src/main/java/org/eclipse/kura/linux/net/wifi/HostapdManager.java
```diff
@@ -26,51 +35,38 @@
...
 			Thread.sleep(1000);
-
 		} catch(Exception e) {
 			throw KuraException.internalError(e);
 		}
 		finally {
 			if (proc != null) {
 				ProcessUtil.destroy(proc);
 			}
+			if (!OS_VERSION.equals(KuraConstants.Intel_Edison.getImageName() + "_" + KuraConstants.Intel_Edison.getImageVersion() + "_" + KuraConstants.Intel_Edison.getTargetName())) {
```
MMaiero: @ibinshtok since we are using it many times, would have sense to have this string static and evaluated once?

kura/org.eclipse.kura.linux.net/src/main/java/org/eclipse/kura/linux/net/wifi/HostapdManager.java
```diff
@@ -156,4 +154,58 @@ public static String getHostapdConfigFileName(String ifaceName) {
...
+	private static File generateHostapdConfigFile(String ifaceName, Password passkey, WifiSecurity wifiSecurity) throws KuraException {
+		File retConfigFile = new File(privGetHostapdConfigFileName(ifaceName));
+		File configFile = new File(getHostapdConfigFileName(ifaceName));
+		if(!configFile.exists()) {
+			throw KuraException.internalError("Config file does not exist: " + configFile.getAbsolutePath());
+		}
+		BufferedReader br = null;
+		PrintWriter pw = null;	
+		try {
+			br = new BufferedReader(new FileReader(configFile));
```
MMaiero: @ibinshtok be careful here: you opened a FileReader but not closed. I think this can lead to a leak. Same also for the following FileWriter

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/net/wifi/WifiPassword.java
```diff
@@ -0,0 +1,112 @@
...
+					Long.parseLong(part3, 16);
+				} catch(Exception e) {
+					throw KuraException.internalError("the WEP key (passwd) must be all HEX characters (0, 1, 2, 3, 4, 5, 6, 7, 8, 9, a, b, c, d, e, and f");
+				}
+			} else if ((passKey.length() == 5)
+					|| (passKey.length() == 13)
+					|| (passKey.length() == 16)) {
+				
+				// 5, 13, or 16 ASCII characters
+				passKey = toHex(passKey);
```
MMaiero: @ibinshtok Since the method returns void, is it useful this assignment?

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/net/wifi/WifiPassword.java
```diff
@@ -0,0 +1,112 @@
...
+	/*
+	 * This method converts supplied string to hex
+	 */
+	private String toHex(String s) {
+		if (s == null) {
+			return null;
+		}
+		byte[] raw = s.getBytes();
+
+		StringBuffer hex = new StringBuffer(2 * raw.length);
```
MMaiero: @ibinshtok Do we need a thread-safe object? Or we could use String Buffer?

kura/org.eclipse.kura.linux.net/src/main/java/org/eclipse/kura/linux/net/wifi/HostapdManager.java
```diff
@@ -156,4 +154,58 @@ public static String getHostapdConfigFileName(String ifaceName) {
 		}
 		return sb.toString();
 	}
+	
+	private static String privGetHostapdConfigFileName(String ifaceName) {
```
MMaiero: @ibinshtok do you think we could merge the privGetHostapdConfigFileName with the already existing getHostapdConfigFileName? Maybe checking if a certain path exists and returning the correct path back. Is it feasible?

kura/org.eclipse.kura.linux.net/META-INF/MANIFEST.MF
```diff
@@ -12,6 +12,7 @@ Import-Package: javax.microedition.io,
  org.apache.commons.io;version="2.4.0",
  org.eclipse.kura;version="[1.1,2.0)",
  org.eclipse.kura.comm;version="[1.0,2.0)",
+ org.eclipse.kura.configuration;version="[1.0,1.1)",
```
MMaiero: @ibinshtok If I'm not wrong, this should be [1.0,2.0) since you are using Password and not implementing it. Or you are using configuration to implement a configurable component?

kura/org.eclipse.kura.linux.net/src/main/java/org/eclipse/kura/linux/net/wifi/WpaSupplicantManager.java
```diff
@@ -181,4 +198,60 @@ public static String getWpaSupplicantConfigFilename(String ifaceName) {
...
+		File retConfigFile = new File(privGetWpaSupplicantConfigFilename(ifaceName));
+		File configFile = new File(getWpaSupplicantConfigFilename(ifaceName));
+		if(!configFile.exists()) {
+			throw KuraException.internalError("Config file does not exist: " + configFile.getAbsolutePath());
+		}
+		BufferedReader br = null;
+		PrintWriter pw = null;	
+		try {
+			br = new BufferedReader(new FileReader(configFile));
+			pw = new PrintWriter(new FileWriter(retConfigFile));
```
MMaiero: @ibinshtok Same as before, I think that the FileReader and the FileWriter should be managed and closed.

kura/org.eclipse.kura.net.admin/src/main/java/org/eclipse/kura/net/admin/visitor/linux/util/WifiVisitorUtil.java
```diff
@@ -0,0 +1,120 @@
...
+			for (ComponentConfiguration cc : componentConfigurations) {
+				if (NETWORK_CONFIGURATION_SERVICE_PID.equals(cc.getPid())) {
+					netConfigServiceComponentConfig = cc;
+					break;
+				}
+			}
+			if (netConfigServiceComponentConfig != null) {
+				Map<String, Object> props = netConfigServiceComponentConfig.getConfigurationProperties();
+				if (props.containsKey(key)) {
+					passphrase = ((Password)props.get(key)).toString();
```
MMaiero: @ibinshtok We cannot suppose to have a Password object here: we must be compliant with old snapshot 0s that have the wifi password with type String.

Issue #215, opened 2016-05-13
[Bluetooth] Expose buzzer and leds on SensorTag Example


pierantoniomerlino:

ccustine:
@pierantoniomerlino Are you working on this currently?  If not, I may be able to do this for you in the next 1-2 weeks. Let me know!

pierantoniomerlino:
@ccustine 
Hi,
thank you for the proposal!
I'm not working on it, so feel free to work on it.

BTW, time ago I developed a Android app based on the official TI SensorTag app and I used the buzzer and leds. Here the code: https://github.com/eurotech/edc-examples/tree/master/SensorTag

pierantoniomerlino:
Hi @ccustine!

Did you have the possibility to work on this issue? Do you need some testing?

pierantoniomerlino:
Hi @ccustine,

did you work on this issue? Since we need it for a demo, could you share something? Otherwise it is not a problem for me to work on it.

Thank you!

pierantoniomerlino:
Solved by #246.

ccustine:
I tested this thoroughly yesterday and everything works well.

cdealti:
@ccustine Thanks! 
By the way how is the buzzer on your SensorTag 2? On mine is very weak and unstable.
It has a firmware dated July 2015. I've downloaded the latest BLE stack from TI to see if there is a later version or if we can improve the situation by compiling a new one.

ccustine:
@cdealti Yeah it seems like the frequency is very low so the sound is quite bad. Maybe that is for reduced power requirements :-)

cdealti:
@ccustine you know the alarm on cheap quarz watches... They are loud and run off a coin battery. Tomorrow i'll take a look to the firmware. I'm pretty sure it can work better. I'll let you know my findings
________________________________
Da: Chris Custine <notifications@github.com>
Inviato: mercoledì 8 giugno 2016 21.47.50
A: eclipse/kura
Cc: De Alti, Cristiano; Mention
Oggetto: Re: [eclipse/kura] [Bluetooth] Expose buzzer and leds on SensorTag Example (#215)


@cdealti<https://github.com/cdealti> Yeah it seems like the frequency is very low so the sound is quite bad. Maybe that is for reduced power requirements :-)

-

cdealti:
@ccustine I have a new firmware with improved buzzer support [here](https://github.com/cdealti/sensortag). Sounds good!

PR #246, opened 2016-05-30
Added support for leds and buzzer on SensorTag example

pierantoniomerlino:


pierantoniomerlino:
The PR solves #215.

ccustine:
@pierantoniomerlino I was just thinking about starting this so I am glad I checked here first.  I will test this with my use case and let you know if I find any issues.  Thanks for doing this.

Issue #250, opened 2016-05-31
KuraCloudComponent should allow more than 2 topic levels


ccustine:
#The code in KuraCloudComponent only allows 2 topic levels (appId/topic) which is fine when using the bridged mode, but this hampers portability if you change the camel route to use native cloud client.  Code in question is [here](https://github.com/eclipse/kura/blob/9701210c393e818c13cc2d4f198b2495139304f7/kura/org.eclipse.kura.camel/src/main/java/org/eclipse/kura/camel/cloud/KuraCloudComponent.java#L44-L44). Discussed this with @hekonsek .

hekonsek:
Many thanks, Chris. I will pick it up from here.

hekonsek:
Fixed.

PR #261, opened 2016-06-06
Increase topic length (for #250)

ccustine:
This allows topic length greater than 2 while still checking for a complete appId/topic. This should be a valid fix for #250.


MMaiero:
Hi @ccustine, could you please sign your commit?

ccustine:
@MMaiero Done, thanks for pointing that out.

hekonsek:
Merged. Many thanks for the contribution, Chris!

PR #264, opened 2016-06-06
Modified emulator code to check if the snapshot folder is available. …

### LINKED ISSUE #252: Seed initial configuration when running the Emulator
With the introduction of Multiple Connection the default communication stack is created starting from the snapshot_0.

When running the Emulator Kura starts without a snapshot_0, thus leading to the lack of a communication stack, which can be created later by the gogo shell commands.

A snapshot_0 for the emulator should be created as part of the Emulator bundle resources and used to seed initial configuration when needed.

MMaiero:
…If not, creates it and copies snapshot_0.xml file inside.

Closes #252


cdealti:
@MMaiero I don't see the ``snapshot_0.xml`` file in the bundle's resources. But maybe I'm blind.

MMaiero:
@cdealti Absolutely right! Pushed now

PR #268, opened 2016-06-07
Modified service locator to get multiple services and to return servi…

### LINKED ISSUE #256: [Web UI] Show connection status for each Communication Stack
The Status tab of the WebUI should include the information for all the available Communication Stacks.
Right now only the default communication stack information is represented.

MMaiero:
…ce references. Added code in gwt status to support the displaying of information for zero or more connections.

This PR closes #256


cdealti:
@MMaiero it looks like we put everything in a single table. Should we use a separate table for every connection (Cloud and Data Service, Cloud and Data Service 2, etc)?

Then we could add Connect/Disconnect button for every connection table (not sure if buttons can be added to tables).

Should we check for null services before looping? Alternatively return empty lists or arrays.

MMaiero:
@dwoodard1 Could you give a look to this PR?

PR #273, opened 2016-06-08
Dots in mqtt client id are now implicitly substituted by dashes.

### LINKED ISSUE #272: [WebUI] Change message on client-id under MqttDataTransport tab
The . (dot) is not allowed in the client-id. Add it to the message.

MMaiero:

Closes #272


MMaiero:
@pierrepitiot Could you check it?

PR #274, opened 2016-06-08
Removed net and gps info from device - profile tab as redundant with …

### LINKED ISSUE #270: [WebUI] Remove Network Information from Device tab
In the Device tab the Network Information shows the first active network interface. Often it is the loopback interface. Do it make sense?

MMaiero:
…what in status.

Closes #270


MMaiero:
@pierrepitiot Could you check it?

PR #276, opened 2016-06-09
Fixed exception in ssl service related to bugs in configuration service.

MMaiero:

@dwoodard1 Could you check if this solves your issue?


MMaiero:
To be fully tested:
- if it correctly works with ssl
- if changes to values in web ui are correctly done
- if configuration changes from cloud are correctly applied
and don't result in corrupted configurations/snapshots

PR #277, opened 2016-06-10
Fix factory metatype

MMaiero:
Performed, also, general refactoring in ConfigurationServiceImpl


MMaiero:
@dwoodard1 @pierrepitiot Could you check if this PR solves the issues with metatype and the multiple connections work as well?

PR #282, opened 2016-06-13
Status icon will display connection status only for the default conne…

### LINKED ISSUE #279: [Web UI] Cloud connection icon should show the connection status of the default connection
For the time being the red/green cloud connection icon will only show the status of the default connection.

MMaiero:
…ction stack.

@cdealti Could you check it? It's based on a trick. But I think that for a temporary fix before the refactoring for 2.1 is ok

Closes #279


MMaiero:
@pierrepitiot Could you check this PR? Thanks!

Issue #284, opened 2016-06-13
[WebUI] Empty strings in arrays are not managed


pierantoniomerlino:
When I try to save changes in the publisher demo, the kura.log reports an exception due to empty strings in short array.

pierantoniomerlino:
Solved by #285.

PR #285, opened 2016-06-13
Fixed empty strings bug

pierantoniomerlino:

Solved issue #284.

### DIFF
--- a/kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/server/GwtComponentServiceImpl.java
+++ b/kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/server/GwtComponentServiceImpl.java
@@ -425,7 +425,7 @@ public class GwtComponentServiceImpl extends OsgiRemoteServiceServlet implements
 				if (!value.trim().isEmpty())
 					values.add(value);
 			}
-			return defaultValues;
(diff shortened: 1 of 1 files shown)


pierantoniomerlino:
@MMaiero : could you please review it?

kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/server/GwtComponentServiceImpl.java
```diff
@@ -360,59 +359,72 @@ private Object getObjectValue(GwtConfigParameter gwtConfigParam, String strValue
...
+					values.add( new Password(value));		
 			}
 			return values.toArray( new Password[]{});
 
 		case STRING:
+			for (String value : defaultValues) {
+				if (!value.trim().isEmpty())
+					values.add(value);
+			}
 			return defaultValues;
```
MMaiero: @pierantoniomerlino why, in the string case, we are stepping through all the strings in defaultValues, adding them to values and then we return defaultValues as in the old code?

pierantoniomerlino: Copy-and-paste error. My fault.

Issue #271, opened 2016-06-08
[Bluetooth] BluetoothUtil always use the default adapter (e.g. hci0)


cdealti:
The adapter specified in the configuration of the BluetoothLE Sensor Tag app is not respected.
BluetoothGattImpl and BluetoothUtil always assume the default (hci0) adapter.

pierantoniomerlino:
When the gatttool command is run, no adapter is specified. In BluetoothUtil.java the startSession method must be modified adding -i hciX.

PR #286, opened 2016-06-14
Fixed bug on default adapter with gatttool

pierantoniomerlino:
The PR solves #271

### DIFF
--- a/kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/bluetooth/BluetoothGatt.java
+++ b/kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/bluetooth/BluetoothGatt.java
@@ -29,6 +29,15 @@ public interface BluetoothGatt {
 	 * 
 	 * @return If connection was successful
 	 */
+	public boolean connect() throws KuraException;
+
+	/**
+	 * Connect to devices GATT server with a given adapter.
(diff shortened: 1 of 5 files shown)


kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/bluetooth/BluetoothGatt.java
```diff
@@ -29,7 +29,7 @@
 	 * 
 	 * @return If connection was successful
 	 */
-	public boolean connect() throws KuraException;
+	public boolean connect(String adapterName) throws KuraException;
```
MMaiero: @pierantoniomerlino This is API code: we cannot change it in this way. I suggest you to overload the connect method adding one with the adapterName parameter, but we cannot remove the old method. You need, also, to add a `@since` in the comment to specify that this API is available since org.eclipse.kura.bluetooth;version="1.2.0" like @cdealti did here: https://github.com/eclipse/kura/blob/develop/kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/configuration/ConfigurationService.java#L85

pierantoniomerlino:
@MMaiero : Ok, I'll add the method instead of modify the old one. Moreover, I'll add the since directive.

PR #287, opened 2016-06-14
Reverted wifi password changes

MMaiero:
@cdealti @ibinshtok Could you check if everything is ok for you?


MMaiero:
Merging: @ibinshtok checked and the PR seems to work

PR #291, opened 2016-06-15
Fix multiple connection buttons

dwoodard1:
This pull request adds multiple connect/disconnect buttons to support multiple connection channels in Kura.

@MMaiero can you review?


MMaiero:
@dwoodard1 As now we have the modal to display the connect/disconnect buttons for all the connections, would it be a good idea to unify the connect and disconnect buttons in status in order to only open the modal window? Because, for what I see, if you click either the connect or the disconnect button, you get access to the same modal and you are allowed to connect or disconnect a connection independently from the button you clicked before.

MMaiero:
@dwoodard1 However this is only a stylistic item. The PR works really good!

MMaiero:
@dwoodard1 @cdealti I'll merge this PR. Eventually the single button for connect/disconnect in status will be added in the release branch

PR #298, opened 2016-06-19
Platform support for PC Engines APU1

unverbraucht:
Dear Kura team,

I have added initial support for the embedded x86_64 APU1 platform from PC Engines running Debian. For now there's no reason why the resulting DEB shouldn't run on any x86_64 Debian system since I haven't added the GPIOs yet. This is a quick "port" in the sense that's its a mixture between the Intel Edison and the Raspbian profiles. I'll flesh it out if I find the time, right now it's running nicely and it's not a primary target for us.

If you see the point in adding platforms that probably only a few people use (compared to RasPis at least ;) then please feel free to merge. Feedback is also welcome of course :)

### DIFF
--- a/kura/distrib/src/main/sh/extract_upgrade.sh
+++ b/kura/distrib/src/main/sh/extract_upgrade.sh
@@ -22,6 +22,9 @@ TIMESTAMP=`date +%Y%m%d%H%M%S`
 LOG=/tmp/kura_upgrade_${TIMESTAMP}.log
 ABSOLUTE_PATH=`readlink -m $0`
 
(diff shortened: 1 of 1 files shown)


cdealti:
@unverbrauch thanks for your contribution. There is an IP validation check failed that you should fix.
Please fill in the [Eclipse CLA](https://eclipse.org/legal/CLA.php).

unverbraucht:
@cdealti I have signed the CLA, but my Signed-Off-By address is wrong. I'll try to amend the commit later on.

Issue #302, opened 2016-06-24
OSGi console "help" command fails


kartben:
```
osgi> help 
---Extension Registry Commands---
	ns [-v] [name] - display extension points in the namespace; add -v to display extensions
	pt [-v] uniqueExtensionPointId - display the extension point and extensions; add -v to display config elements

---Service Component Runtime---
	list/ls [-c] [bundle id] - Lists all components; add -c to display the complete info for each component;
			use [bundle id] to list the components of the specified bundle
	component/comp <component id> - Prints all available information about the specified component;
			<component id> - The ID of the component as displayed by the list command
	enable/en <component id> - Enables the specified component;
			<component id> - The ID of the component as displayed by the list command
	disable/dis <component id> - Disables the specified component;
			<component id> - The ID of the component as displayed by the list command
	enableAll/enAll [bundle id] - Enables all components; use [bundle id] to enable all components of the specified bundle
	disableAll/disAll [bundle id] - Disables all components; use [bundle id] to disable all components of the specified bundle

---Kura Logger Commands---
	getkuraloglevel [logger_name] - shows the Kura log level for all loggers, or for a specific logger if specified
	setkuraloglevel (ALL | <logger_name>) <log_level> - sets the Kura log level for a given logger, or all loggers if 'ALL'
		The valid log_level options are: ALL, TRACE, DEBUG, INFO, WARN, ERROR, FATAL, OFF

gogo: ClassCastException: java.lang.String cannot be cast to [Ljava.lang.String;
```

This is because the "createConfiguration" command is not registered properly.
Would be nice to see this fixed for 2.0.0, as it's quite annoying to lose access to the OSGi console help.

I will submit a PR.

kartben:
Well actually it seems to be something that's been fixed in Gogo command 0.12.0, or at least they implemented a more flexible way to parse the "osgi.command.function" property, and it can now be either a single String, or a String array.
Since Kura still relies on Gogo command 0.8.0, the fix for now should be to just make sure the "osgi.command.function" property is a String array, as it's what Gogo 0.8 only supports.

cdealti:
Fixed by PR #304.

PR #303, opened 2016-06-24
Fix issue #302. The "osgi.command.function" property better be an arr…

kartben:
…ay so as the command can properly be handled with old-ish versions of Apache Gogo shell/command.

Signed-off-by: Benjamin Cabé <benjamin@eclipse.org>


kartben:
closing this PR as PR #304 will eventually bring the fix to develop.

PR #319, opened 2016-07-12
Windows support, part 2/2

markcullen:
Signed-off-by: Mark Cullen <mark.cullen@eurotech.com>


cdealti:
Well done!

cdealti:
@markcullen this PR contains files with an IBM Copyright (though licensed under the EPL).
Can you provide links or the origin of these files?
Since you are not the author of these files and they are licensed under the EPL I assume it is enough to retain the original header and copyright and add a line like this:

```
Contributors:
     IBM - initial API and implementation
     Eurotech - this file has been modified to...
```
@ctron would this work?

ctron:
I don't think so. Basically it should be possible. But I do think that you will run into troubles with the overall IP log of Kura. Which you need to manually alter then.

You can however add a second author using the "signed off" mechanism, which should be the original author then.

But I would say you should forward this question to emo-ip-team@eclipse.org

cdealti:
@markcullen while testing this PR I've noticed that the org.eclipse.kura.windows.system bundle is installed in every distribution, for example the one for the raspberry PI which does not make much sense.

PR #320, opened 2016-07-13
Provide Oomph setup and clean up build issues

ctron:
This changes adds an Oomph based setup which follows the
"Getting Started" tutorial. It will allow the user to choose between
a minimal set as described in the tutorial or a full Kura developer
setup.

For this it was necessary to also fix a few project issues in order to
create a, mostly, error free workspace.
 
The steps to be taken for the end user are:

 * Install Java 8
 * Download and run the Eclipse installer
  * Select Eclipse IDE edition
  * Select "Eclipse Kura"
  * Enter location for installation
  * Wait a few minutes

In addition this change fixes a few project settings issues:

 * org.eclipse.soda.dk.comm package path (actually is
src/main/java/org/eclipse/soda/dk instead of
src/main/java/org/eclipse/soda/dk/comm)
 * Don't ignore Tycho and WTP maven plugins but install m2eclipse
adapters instead


ctron:
I did amend the pull request so that the user now has the choice to select whether to include all Kura project (developer) or only the ones from the getting started tutorial (user).

Also does the setup now import all projects as Eclipse PDE projects (thanks to the Tycho Connector) and removed the unnecessary m2e settings from all child modules (see PR #323).

dwoodard1:
@ctron I recently merged a PR with very similar POM changes. Would it be possible for you to update this PR to include only the changes to support Ommph? It would be nice to have those changes isolated in one PR.

ctron:
Yes, I will do that.

ctron:
So I cleaned up the PR. It is also rebased and the changes on the POM files got merged.

Regarding the POM files, which I changed as well. I did what the other PR did, but for the m2eclipse settings. First of all they can be at the parent project, as the other PR did it, and second several of them are not needed if you have the proper m2eclipse adapters installed, which the Oomph setup does.

In order for the Oomph setup pick up the projects types and create proper Eclipse projects, the m2eclipse "ignore" configuration must be removed.

In the case that you don't use the Oomph setup, a m2eclipse quick fix will allow you to install those connectors at a later time.

dwoodard1:
@ctron Thanks!

PR #322, opened 2016-07-14
Clean up static lock initialization

ctron:
The static lock was re-initialized in the classes constructor. The
static lock variable is now statically initialized send the static field
set to final in order to prevent further updates in this lock instance.


cdealti:
@ctron is this needed. It's a singleton. And s_lock should not be static at all.

ctron:
If you ask me the 's_lock' is not needed at all. It could be done by simple 'synchronized' methods.

hekonsek:
Jens, maybe you could refactor the existing code to use synchronized methods instead of this lock object? And create new pull request? And obviously close this one then :) .

What do you think?

ctron:
Closing in favor of PR #392

PR #323, opened 2016-07-14
Clean up POM files and move m2e declarations to top level project

ctron:
This change moves the m2e declarations required to silence the build
errors triggered by unknown maven target to the top level project in
"manifest_pom.xml"

This cleans up most child module POM files from the repeating POM
declarations.


ctron:
I would like to revoke this PR in favor or PR #320.

PR #325, opened 2016-07-15
Fix a possible Context Class Loader leak

ctron:
In the case of an exception in the Web UI the context class loader
will not be cleaned up correctly. This may lead to all kinds of issues
later on.


cdealti:
@ctron thanks!

PR #327, opened 2016-07-15
Fix possible resource leaks in web2

ctron:
This change fixes a few possible resource leaks in the web2 project.


cdealti:
@ctron Kura source still aims to be compatible to Java 6

ctron:
Is there a reason for that?

ctron:
Switched the discussion to the other thread. Seems more focused on the topic.

Issue #330, opened 2016-07-15
Setting Maven compiler settings to Java 7


ctron:
As described in the Eclipse Wiki [1] Kura has a dependency on Java 7.

However the maven compiler settings are still at Java 6.

I would like to raise this to Java 7 for Kura 2.0.

[1] https://wiki.eclipse.org/Kura/Getting_Started#Java

cdealti:
@ctron see my comment to #327

ctron:
Maybe I do continue here ...

So what is the reason for that sticking with source level compatibility on 1.6?

Since this would mean that I can switch to EE 7, but cannot use "try-with-resources"?

cdealti:
Exactly

ctron:
I don't see any reason at all. The other way round, maybe.

So let's do it of nobody objects with a real reason.

hekonsek:
Actually even Camel requires Java 1.7 to work. :) It will be nice to upgrade to Java 7. 

Anybody has something against upgrading JDK to 1.7 for Kura 2.0? @kartben @mcarrer @dwoodard1 ?

cdealti:
For the build we use Tycho. If I change the plugin configuration in ``manifest_pom.xml`` to compile Java 7 code:

```
                        <plugin>
                                <groupId>org.eclipse.tycho</groupId>
                                <artifactId>tycho-compiler-plugin</artifactId>
                                <version>${tycho-version}</version>
                                <configuration>
                                        <source>1.7</source>
                                        <target>1.6</target>
                                        <compilerArgument>-warn:[+|-]warning_tokens_separated_by_comma</compilerArgument>
                                        <!-- <compilerArgument>-err:-forbidden</compilerArgument> -->
                                        <!-- <useJDK>BREE</useJDK> -->
                                </configuration>
                        </plugin>
```

I get:

```
[ERROR] Failed to execute goal org.eclipse.tycho:tycho-compiler-plugin:0.20.0:compile (default-compile) on project org.eclipse.kura.api: Fatal error compiling: Target level '1.6' is incompatible with source level '1.7'. A target level '1.7' or better is required -> [Help 1]
```

Our industrial grade gateways all ship Oracle Java SE Embedded 7 and 8 so we could set both ``source`` and ``target`` to 1.7.

I cannot exclude however that there aren't gateways in the field with a Java 6 VM and/or if Kura can still run on a Java 6 VM.

ctron:
I agree both source and target level should be the same.

[...]

PR #331, opened 2016-07-15
Fix concurrency issue in command service

ctron:
The command service was implemented in a way that it did execute all
command written out to a single shell script. This can cause the issue
that two requests issued at the same time would fail.

This change does write out seperate script files using the
Files.createTempFile method. It also does not fork "chmod" and
"dos2unix" since those operations can be performed inside Java which
saves two additional process forks.

Also is the "cd" statement not necessary when the ProcessBuilder
is instructed to change the directory first. For this the process
started can now accept an optional working directory.

This change also includes a unit test for the command service.


hekonsek:
Hi Jens, 

I'd like to merge the PR, but you should downgrade Java to 1.6 before I do that. We should stick for Java 1.6 for a while. Can you do this?

Thanks!

ctron:
In theory: yes

But rather than rewriting those "try-with-resources" sections and bloating the source code again, I would suggest we make a decision on #330.

Kura itself needs Java 7 already. And I don't see any advantage in sticking with Java 6 for the source code. But maybe I just don't see it.

If we come to conclusion that Java 6 is still a requirement I will re-work the patch for Java 6.

dwoodard1:
@ctron Why does Kura need Java 7?

dwoodard1:
@ctron Apologies, I just saw the note in #330 about Apache Camel, I will respond there.

hekonsek:
Jens, I propose to downgrade this pull request to use Java 6 for a while. When are are done with Java 7 upgrade, then we could create an another PR to make this code more Java 7 friendly.

Is that OK?

ctron:
Well the changes make use of the Java 7 NIO2 methods like for example `Files.setPosixFilePermissions`.

I would like to spend a bit more time on making a decision/plan about Java 7. I think this PR can wait little while longer. If Java 6 is a hard requirement, then we can still rework it. But of possible I would like to save the time.

hekonsek:
I guess that @cdealti is concerned (#330) about customers using Java 6 in production. I would say that we should migrate this PR to 1.6 then, if you ask me :) .

ctron:
Superseded by #868

PR #344, opened 2016-07-18
Fix issue with ClassCastException in OSGi console

ctron:
This change fixes an issue which breaks the OSGi console. All
registered commands must register functions as String array. However
using the "value" attribute in OSGi DS does register a String instead
of String[]. In order to register a String[] the element data has to
be used.

----

I do think this issue is rather serious since is breaks the whole `help` functionality of the OSGi console.


kartben:
I think this is already fixed with issue #302 / PR #304 :-)

ctron:
So it seems this was not ported to the devel branch:

https://github.com/eclipse/kura/blob/develop/kura/org.eclipse.kura.core.cloud/OSGI-INF/cloudServiceFactory.xml

MMaiero:
It will be ported to develop when Kura 2 will be released.

hekonsek:
Hi Jens,

Can you confirm that we can close this PR then?

Thanks!

ctron:
:+1:

Issue #343, opened 2016-07-18
[Web2]: Setting the "name" property to anything different than the "id" sets an empty value


ctron:
Setting the "name" attribute to anything different than the "id" attribute in the metatype model will cause the "name" to rendered as label, but fail to retrieve the data from the generated form since the label is used as ID:

```java
String id = ((FormLabel) fg.getWidget(i)).getText();
param = m_configurableComponent.getParameter(id.trim().replaceAll("\\*$", ""));
```

A possible workaround would be to set the ID into the title an also retrieve it from there, making the label just a label:

```java
FormLabel formLabel = new FormLabel();
if (param.isRequired()) {
  formLabel.setText(param.getName() + "*");
} else {
  formLabel.setText(param.getName());
}
formLabel.setTitle(param.getId());
```

and:

```java
param = m_configurableComponent.getParameter(fg.getWidget(i).getTitle());
```

cdealti:
Closed by #345.

PR #345, opened 2016-07-18
Clean up and fix the service field generation a bit

ctron:
This change includes:

 * Allow String type fields with more than 255 characters
 * Handles the case no minimum and maximum are set for String type
   fields
 * Use the ID for internal field lookup in all places, #343
 * fix up log/error output
 * Use a TextArea instead of a plain TextBox when the label contains
   the magic suffix or has a cardinality of -1 set
   
### DIFF
--- a/kura/examples/org.eclipse.kura.example.camel.quickstart/OSGI-INF/metatype/org.eclipse.kura.example.camel.quickstart.GatewayRouter.xml
+++ b/kura/examples/org.eclipse.kura.example.camel.quickstart/OSGI-INF/metatype/org.eclipse.kura.example.camel.quickstart.GatewayRouter.xml
@@ -6,7 +6,7 @@
         <Icon resource="OSGI-INF/logo.png" size="32" />
 
         <AD id="camel.route.xml" name="camel.route.xml"
-            description="Camel XML route definition &#x200b;&#x200b;&#x200b;&#x200b;&#x200b;&#x200b;&#x200b;&#x200b;&#x200b;&#x200b;"
+            description="Camel XML route definition|TextArea"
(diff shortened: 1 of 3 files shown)


cdealti:
@ctron I believe 1 and 2 have already been fixed in another PR, right?
I would be interested to 5. Please note that accordingly to the Meta Type spec a cardinality < 0 is valid implying a ``Vector`` type

ctron:
I think only 1 is fixed.

I do think we need a strategy for switching between text box and text area. -1 may not be an optimal solution. Maybe we can use the label itself, but in a prettier way, like a suffix "|TextArea" and "|TextBox". With the rule that if the suffix of the label is "|(.*)" then it will be removed from the actual label. And if it is unknown it will fall back to default.

ctron:
Maybe let's discuss that tomorrow.

dwoodard1:
@ctron I think it looks okay, but we are currently in the middle of releasing v2.0.1. Let's get the release out, then review this in more detail later this week. It seems that for item 3 we were using name instead of ID for a specific reason, but I will need to check that.

dwoodard1:
I was thinking of cases where we use <Option> within the <AD>. We had special handling of cases where the label was not equal to the value. This metatype file is the best to reference [1]. For most cases, we have label == value, but some (ex: protocol.version) do not. 

I don't think the proposed changes should affect this in any way. I will test to verify.

[1] https://github.com/eclipse/kura/blob/develop/kura/org.eclipse.kura.core/OSGI-INF/metatype/org.eclipse.kura.core.data.transport.mqtt.MqttDataTransport.xml

PR #347, opened 2016-07-19
Reset the original state of the accessible flag

ctron:
This change does reset the original state of the accessible attribute
of the field "languages".


hekonsek:
+1 Many thanks for the contribution, Jens!

Issue #296, opened 2016-06-16
[Web UI] Status | Connection: if the connection username is not set, "null" is shown


cdealti:
If the username is not set (left blank) in the MqttDataTransport, it is shown as "null" in the Connection Status page of the Web UI.

This was found in an 2.0.0 SNAPSHOT so it might have already been fixed.

pierrepitiot:
The bug is still present. Null is returned by org.eclipse.paho.client.mqttv3.MqttConnectOptions.getUserName().

PR #349, opened 2016-07-20
Fix username displayed null in Device Status.

pierrepitiot:
When the username is empty in MqttDataTransport configuration, Device Status in the web ui displays username null. 
This PR solves issue #296


MMaiero:
@pierrepitiot I would suggest you to first rebase the branch with develop, in order to have all the latest changes.

Other than that, I think that the best solution is to catch the null value at the level of the dataTransportService implementation: in this way the get part will be untouched and easier to read, while the mqttDataTransport class can be changed in this way to catch the null values:

`public String getUsername() {
		if (m_clientConf != null) {
		    String username = m_clientConf.getConnectOptions().getUserName();
		    if (username != null) {
		        return username;
		    }
		}
		return "";
	}`

I would also do the same check for the other getters we have there, in order to be sure that we return a meaningful string or an empty string and not a null value.

What do you think?

pierrepitiot:
I think you are right, rather than rebase this branch, I will close this PR and open a new one on updated branch and modify MqttDataTransport.getUsername. 
PR replaced by PR #394

PR #361, opened 2016-07-22
Use Import-Package in preference of Require-Bundle

ctron:
This change converts to Require-Bundle statements to Import-Package in
order to remove hard dependencies on bundles.


cdealti:
+1 thanks.

hekonsek:
Merged. Many thanks, Jens!

Issue #221, opened 2016-05-13
[Camel] Do we need a Camel API in the org.eclipse.kura.api bundle?


cdealti:
@hekonsek in Kura we package APIs in the ``org.eclipse.kura.api`` bundle. Typically, implementation bundles do not export API packages and where they do that it's only for internal use by other Kura bundles.

For Camel we don't have anything in the API bundle.

It looks to me that the Camel "API" is [CamelRouter](https://github.com/eclipse/kura/blob/develop/kura/org.eclipse.kura.camel/src/main/java/org/eclipse/kura/camel/router/CamelRouter.java). Well, it's not actually a service interface but it's looks to me that ``CamelRouter`` plays a role similar to that of [Cloudlet](https://github.com/eclipse/kura/blob/develop/kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/cloud/Cloudlet.java) which is in the Kura API bundle but it's not a service interface but a base class that needs to be extended.

I'm not sure what to do for Camel

cdealti:
This issue is postponed and will be addressed in Kura-2.1.0

cdealti:
@hekonsek  I believe we need to extract a Java interface from the ``CamelRouter`` and add it to the ``org.eclipse.kura.api`` bundle. Let's rename the current ``CamelRouter`` to ``CamelRouterImpl`` and reserve ``CamelRouter`` for the interface.

 Consumers will not extend ``CamelRouterImpl``. Instead they will request a new ``CamelRouter`` to a ``CamelService`` via a ``newCamelRouter`` factory method.

It's more in the OSGi spirit for consumers to use an interface instead of exposing the details of an implementation.

Also the name of the ``org.eclipse.kura.camel`` bundle and the names of the packages need to be refactored.

For implementation bundles we are starting following the convention of adding a **.provider** suffix.

For internal packages we are following the [Eclipse naming convention](https://wiki.eclipse.org/Naming_Conventions)
so the packages in the ``org.eclipse.kura.camel.provider`` bundle would start with ``org.eclipse.kura.internal.camel``.

ctron:
I would strongly recommend against adding another technology to the "org.eclipse.kura.api" bundle. It already seems overloaded with functionality to me, and the idea of OSGi is to keep thing modular. Adding every technology in Kura to the "api" bundle will simply create a "super bundle".

[...]

Issue #205, opened 2016-05-11
All languages from Camel Core should be loaded by CamelRouter


hekonsek:

ctron:
This one should be closed by  #387

Issue #204, opened 2016-05-11
All data formats from Camel Core should be loaded by CamelRouter


hekonsek:

ctron:
This was done by PR #365

Issue #200, opened 2016-05-09
[Camel] Do we need to add sun.*,com.sun.* bootdelegation to Kura?


cdealti:
As explained in the [quickstart](https://rhiot.gitbooks.io/rhiotdocumentation/content/quickstarts/kura_camel_quickstart.html) bootdelegation must be added to config.ini:

```
org.osgi.framework.bootdelegation=sun.*,com.sun.*
```

Are there alternatives to bootdelegation?

cdealti:
@hekonsek as you reported this is necessary (for now). Please **submit this as a pull-request**.
You can create a PR branch directly in the upstream Kura repository

ctron:
I do think that this one is resolved with PR #365.

hekonsek:
Resolved by Jens and Camel upgrade.

PR #365, opened 2016-07-25
Refactoring Apache Camel XML router

ctron:
This change upgrades to Apache Camel 2.17.2 and does remove the
unnecessary workarounds for older Camel versions. Languages and
components get auto-detected by OSGi now.

Also are the Kura related CamelRouter classes being split up in:
* A basic Kura based Camel component
* A basic XML routes based component
* A basic XML routes component which gets configured as OSGi service
component
* A generic Kura component based on a single XML snippet with UI support

This change also fixes the "sun.misc" import issue in a way that the
bootdelegation is not required.


cdealti:
@ctron thanks for the effort. It would be great if you could also address (and comment) the following github issues related to the Camel integration:

#221, #125, #200 

If you install the zebhub plugin for Chrome, filter by the ``kura-camel`` label and looking at the ``Backlog`` and ``Todo`` pipelines you should have a better picture of the status.

ctron:
@cdealti I will have a look at those issues.

I am using Firefox. Is there an addon for it?

ctron:
So checking all three issues, I think that #200 is fixed by this change, I would strongly recommend against #221 and #125 is an issue, not sure what causes it, I will dig into this. But then again it would be a bug fix on top of this change.

ctron:
I just saw that #204 and #205 should be solved by this upgrade of camel as well.

ctron:
I am closing this PR in favor of splitting it up into smaller ones.

PR #369, opened 2016-07-27
ConfigurationListener removed from ConfigurationService implementation.

### LINKED ISSUE #358: createFactoryConfiguration seems to always save snapshots
The new `createFactoryConfiguration` method in the CloudService generates a snapshot even if the `takeSnapshot` flag is set to false.

e.g. when creating a new communication stack, the result is three new snapshots in the snapshots folder, instead of just one, as expected.

On another note, the `deleteFactoryConfiguration` seems to correctly handle the `takesnapshot` flag.

LucaDazi:
fixes #358 

Note: ConfigurationListener has been removed from ConfigurationServiceImpl.
Full QA on snapshot management should be performed.
ConfigurationService no more relies on adding the Component pid to a temporary list in order to trigger a snapshot on configuration change.

Patch has been tested for local and remote configuration, and for Factory Components creation and removal.


kura/org.eclipse.kura.core.configuration/src/main/java/org/eclipse/kura/core/configuration/ConfigurationServiceImpl.java
```diff
@@ -75,7 +75,7 @@
 /**
```
MMaiero: @LucaDazi You can clean two unused imports:
import org.osgi.service.cm.ConfigurationEvent;
import org.osgi.service.cm.ConfigurationListener;

PR #379, opened 2016-08-01
#200: remove the need for bootdelegation to sun.misc

ctron:
This change does remove the necessary to add the bootdelegation to
sun.misc by creating two fragments which do the proper wiring.

One fragment is extending the "system.bundle" to export "sun.misc" and
the other one is extending "camel-core" to import "sun.misc" as it
should be done by camel.

Still, this will only work with a JRE which actually provides "sun.misc"
and it will probably stop working with Java 9.

AFAIK the core issue will be fixed by Camel 2.18 in a future release.

---

The reason why the two fragments are in the target-platform project and are not build with tycho
is simply because tycho and PDE both try to resolve "sun.misc" based on the assumption that these are actual implementation packages.


hekonsek:
Merged. Many thanks for the contribution, Jens!

Issue #125, opened 2016-04-21
[Camel] "BundleContext is no longer valid" error after restarting/reinstalling Camel+Kura quickstart bundle


hekonsek:
```
2016-04-21 11:55:30,034 [Camel (camel-4) thread #16 - timer://xmltopic] WARN  o.a.c.c.t.TimerConsumer - Error processing exchange. Exchange[ID-raspberrypi-46944-1461238784533-3-136][Message: org.eclipse.kura.message.KuraPayload@7812fb]. Caused by: [org.apache.camel.ResolveEndpointFailedException - Failed to resolve endpoint: seda://myapp:xmltopic due to: Cannot auto create component: seda]
org.apache.camel.ResolveEndpointFailedException: Failed to resolve endpoint: seda://myapp:xmltopic due to: Cannot auto create component: seda
	at org.apache.camel.impl.DefaultCamelContext.getEndpoint(DefaultCamelContext.java:587)
	at org.apache.camel.impl.DefaultProducerTemplate.resolveMandatoryEndpoint(DefaultProducerTemplate.java:453)
	at org.apache.camel.impl.DefaultProducerTemplate.sendBodyAndHeaders(DefaultProducerTemplate.java:229)
	at org.eclipse.kura.camel.camelcloud.CamelCloudClient.doPublish(CamelCloudClient.java:181)
	at org.eclipse.kura.camel.camelcloud.CamelCloudClient.publish(CamelCloudClient.java:90)
	at org.eclipse.kura.camel.cloud.KuraCloudProducer.process(KuraCloudProducer.java:75)
	at org.apache.camel.util.AsyncProcessorConverterHelper$ProcessorToAsyncProcessorBridge.process(AsyncProcessorConverterHelper.java:61)
	at org.apache.camel.processor.SendProcessor.process(SendProcessor.java:141)
	at org.apache.camel.management.InstrumentationProcessor.process(InstrumentationProcessor.java:77)
	at org.apache.camel.processor.RedeliveryErrorHandler.process(RedeliveryErrorHandler.java:460)
	at org.apache.camel.processor.CamelInternalProcessor.process(CamelInternalProcessor.java:190)
	at org.apache.camel.processor.Pipeline.process(Pipeline.java:121)
	at org.apache.camel.processor.Pipeline.process(Pipeline.java:83)
	at org.apache.camel.processor.CamelInternalProcessor.process(CamelInternalProcessor.java:190)
[...]

PR #381, opened 2016-08-01
Fix issue #125

ctron:
This PR contains two fixes for the Camel integration which fix issue #125

1. Fix a possible deadlock when shutting down the camel context
2. Fix a cloud client leak

It is followed by a third patch selecting the newer version of the bundle in the "distrib" project.


hekonsek:
Many thanks, Jens. Merged!

PR #382, opened 2016-08-01
Use a copy-on-write list in order to support concurrent modifications

ctron:
This change uses a CopyOnWriteArrayList instead of a simple LinkedList
for managing the listeners. This is done in order to support lock-free
add/remove operations while not running into a
ConcurrentModificationException


hekonsek:
+1 Merged. Many thanks!

PR #394, opened 2016-08-02
Fix username displayed null in WebUI Device Status.

pierrepitiot:
This PR solves issue #296 and replace PR #349 .


MMaiero:
@pierrepitiot Why we are not doing the same check also for getBrokerUrl()?

pierrepitiot:
Because the other fields can be empty but there is a check on broker URL that prohibits an empty value (you cannot apply it). Do you think we should check it also?

MMaiero:
I would do it, just to keep consistency: the empty value check is performed by the web ui, but a third party app can rely on mqttdatatransport and get null value, if someone programmatically set it

pierrepitiot:
Ok I will fix it.

PR #395, opened 2016-08-02
Remove dependency on slf4j-log4j12

ctron:
The project should only depend on slf4j-api and not on slf4j-log4j12.

slf4j-log4j12 is only required during the runtime of the module and
so the selection of the slf4j implemenation should be left to the
OSGi container/distribution.


cdealti:
@MMaiero should we explain how to also update kura.build.properties?

MMaiero:
@cdealti Yes, we need to document that. Maybe in the contributing.md file? Or do we have to create a specific file in the wiki?

Issue #340, opened 2016-07-18
[Configuration] Metatype not updated in the Web UI


cdealti:
How to recreate:

1. Install a Deployment Package containing a bundle with a ConfigurableComponent
2. Update the metatype of the ConfigurableComponent, for example by changing the description in an ``AD`` element
3. Uninstall/Install the Deployment Package again
4. The metatype is not updated in the Web UI until Kura is restarted

I guess it's a bug of the ConfigurationService.

PR #397, opened 2016-08-02
Modified validation logic, added better check for string size.

### LINKED ISSUE #338: [Web2]: String input forced to a maximum length of 255
Input fields of type "String" are forced to a maximum limit of 255 characters. Although components may specify more in their metatype description, the following code will not allow text input with more than 255 characters:

```java
if ((String.valueOf(box.getText().trim()).length()) > Math.min(configMaxValue, 255)) {
```

MMaiero:

Closes #338 
@pierrepitiot


pierrepitiot:
The min/max values in metatype are still affected by issue #340 . They cannot be updated by a next release of the bundle.

PR #400, opened 2016-08-03
Fix build warnings

ctron:
This change fixes a few build warnings based on wrong paths.


hekonsek:
Agreed. :)

Many thanks for the fix Jens!

Issue #364, opened 2016-07-25
Kura main process does not automatically restart


ctron:
If anything happens to the main Kura process (JVM) then the process is not being restarted.

MMaiero:
The only possibility we see here is to install monit.
We think that if the Kura user wants to have the framework restarted, probably he will install monit by himself.

ctron:
I can think of a lot way to resolve this. A "forever loop script". A proper systemd configuration.

MMaiero:
This has been addressed years ago. Closing.

PR #410, opened 2016-08-04
Remove call to System.exit()

ctron:
System.exit() should never be called in an OSGi environment!

### DIFF
--- a/kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/system/SystemServiceImpl.java
+++ b/kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/system/SystemServiceImpl.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2011, 2016 Eurotech and/or its affiliates
+ * Copyright (c) 2011, 2016 Eurotech and others
  *
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
@@ -8,6 +8,7 @@
  *
  * Contributors:
  *     Eurotech
+ *     Red Hat Inc - allow opting-out of System.exit()
  *******************************************************************************/
 package org.eclipse.kura.core.system;
 
@@ -47,9 +48,6 @@ import org.osgi.service.component.ComponentException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-
-/**
- */
 public class SystemServiceImpl implements SystemService 
 {
 	private static final Logger s_logger = LoggerFactory.getLogger(SystemServiceImpl.class);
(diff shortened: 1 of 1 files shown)


kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/system/SystemServiceImpl.java
```diff
@@ -1138,8 +1136,7 @@ private static void createDirIfNotExists(String fileName)
 		// Make sure the configuration directory exists - create it if not		
 		File file = new File(fileName);
 		if(!file.exists() && !file.mkdirs()) {
-			s_logger.error("Failed to create the temporary configuration directory: " + fileName);
-			System.exit(-1);
+			s_logger.error("Failed to create the temporary configuration directory: {}", fileName);
```
hekonsek: @ctron Can you just throw an exception here instead? We should not continue the flow of the execution here.

kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/system/SystemServiceImpl.java
```diff
@@ -1138,8 +1136,8 @@ private static void createDirIfNotExists(String fileName)
...
+			throw new RuntimeException(String.format ("Failed to create the temporary configuration directory: %s", fileName));
```
ctron: Done!

cdealti:
I just wonder what was the intent of the original code. I guess it was added deliberately.
Note that in production we use monit to monitor and restart Kura.

ctron:
So it would continuously restart then. And this is also not the case for Kura due to #364

cdealti:
Maybe. However, failing to create directories looks like an unrecoverable error. Is throwing an exception here enough?

ctron:
Many different services can run inside an OSGi container. Shutting it down because one service fails seems completely wrong to me.

cdealti:
Everything in Kura depends on the SystemService. If it does not initialize properly Kura will not start or will behave erratically. By exiting you have a chance that the next restart will work or monit, noticing too many restarts might reboot the system and at the next reboot you will have a change that Kura will start properly. As you know these embedded systems must run unattended and must be able to recover without intervention.

I just wonder if this was the intent of the original code. It looks to me that it would be safer to keep it until we have not understood the implications of such a change.

ctron:
So, who can shed a light on this?

ctron:
@mcarrer @dwoodard1 maybe you can shed some light on this?

I would assume that a stable system, even if it is not working as expected, is better than a box being in a constant reboot loop.

cdealti:
Well, I think you can simulate this occurrence with your patch and look at the result.

________________________________
Da: Jens Reimann <notifications@github.com>
Inviato: giovedì 4 agosto 2016 19.08.42
A: eclipse/kura
Cc: De Alti, Cristiano; Comment
Oggetto: Re: [eclipse/kura] Remove call to System.exit() (#410)


@mcarrer<https://github.com/mcarrer> @dwoodard1<https://github.com/dwoodard1> maybe you can shed some light on this?

I would assume that a stable system, even if it is not working as expected, is better than a box being in a constant reboot loop.

-

ctron:
As agreed on our call I will update this PR in order to keep the default behaviour, but allowing an opt-out. And then I will start a document on how we can resolve this situation in a better way with a future version of Kura.

ctron:
I did update the PR accordingly.

ctron:
As discussed I drafted up a solution beyond the opt-out [1]

[1] https://docs.google.com/document/d/1Db2uo_WUXKww-uWQcYdi5I_qbvqF69W3q4eMvQSsj8o

PR #416, opened 2016-08-04
Fix possible NPEs

ctron:
This change fixes two more possible NPEs.


kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/server/GwtNetworkServiceImpl.java
```diff
@@ -105,6 +106,10 @@
     public ArrayList<GwtNetInterfaceConfig> findNetInterfaceConfigurations() throws GwtKuraException {
         ArrayList<GwtNetInterfaceConfig> result= privateFindNetInterfaceConfigurations();
 
+        if(result==null) {
+        	return null;
+        }
+
```
cdealti: @ctron can you just add blanks before and after ``==``?

ctron:
Changed and rebased

PR #424, opened 2016-08-05
Fix missing "org.eclipse.soda.dk.comm" dependency

ctron:
Commit 3675b9bde0c5994e14a963cd9963ac2c1bef0fe2 did change the version to 1.2.2-SNAPSHOT where it in fact it has to be "1.2.2.SNAPSHOT", so ".SNAPSHOT" instead of "-SNAPSHOT".

This was necessery due to the commit 7a18e616c97859664dd44c2d4cf781f27e5b69b2 which did change the version from 1.2.1 to 1.2.2-SNAPSHOT.

---
It seems there are some issues with the combination of tools involved here which make it tricky to use -SNAPSHOT versions. I incremented the "soda comm" dependency to simply used 1.2.2 instead.


cdealti:
@MMaiero You also took a look at this issue, didn't you?

ctron:
Closing this one out in favor of an upcoming PR which re-structures the whole target platform setup.

PR #425, opened 2016-08-08
Fix modem reset rg10 11

ibinshtok:
Fixed modem reset for RG 10-11 by setting pin direction once in a 'static' block. Fixed the 'NoSuchPortException' in a SerialModemDriver by removing HE910 from supported serial modems for RG 10-11.


ibinshtok:
@MMaiero can you check and merge this pull request, please?

kura/org.eclipse.kura.linux.net/src/main/java/org/eclipse/kura/linux/net/modem/ModemDriver.java
```diff
@@ -256,17 +289,13 @@ private void toggleGpio(String gpio, String gpioPath, String directionPath, Stri
...
-        
+	private void invertGpioValue(String gpio, String gpioPath, String directionPath, String valuePath) throws IOException {
         FileReader fGpioOldValueReader= null;  //read current value
         BufferedReader fGpioOldValue= null;
         int oldValue= 0;
         try {
             fGpioOldValueReader= new FileReader(valuePath);
-            fGpioOldValue= new BufferedReader(fGpioOldValueReader);
-            oldValue= Integer.parseInt(fGpioOldValue.readLine());
+            oldValue = fGpioOldValueReader.read() - 48;
```
MMaiero: @ibinshtok Honestly I prefer the old buffered reader instead of having this line and its magic number

MMaiero: @ibinshtok otherwise, if we want to keep this new line as is, we must clean the code and remove fGpioOldValue as it's not used

MMaiero:
@ibinshtok So, we are not using also GPIO 30 for internal modem reset?

MMaiero:
@ibinshtok as discussed when we met, I would also suggest to use invertGpioValue instead of toggleGpio: it worked better in both my devices

ibinshtok:
@MMaiero I made changes you suggested:
* Used invertGpioValue() call instead of toggleGpio() in the turnModemOn() and turnModemOff() methods.
* Used BufferedReader() in the invertGpioValue() method.
I did not use gpio30 for internal modem for two reasons:
1. invertGpioValue() for gpio60 seems to work fine; I tested it many times.
2. gpio30 seems to depend on gpio60 in this way
a. If gpio60 is set to 1: setting gpio30 to 1 turns modem off, setting gpio30 to 0 turns modem back on.
b. If gpio60 is set to 0: setting gpio30 to 1 turns modem off, but setting gpio30 to 0 does not turn modem back on.
So I think it would be safer to leave things as is and don't use gpio30. What do you think?

PR #428, opened 2016-08-08
Fix missing camel bundle in dev workspace

ctron:
The bundle "org.eclipse.kura.camel" is missing from the developer workspace. This change adds it.


ctron:
This is an issue which affects Kura 2.0.0

PR #446, opened 2016-08-11
Target platform camel bundles are now in -SNAPSHOT version. Updated v…

MMaiero:
…ersion for camel examples. Added steps for camel bundles dp generation when in release version.


MMaiero:
@cdealti Could you check this?

Issue #290, opened 2016-06-15
[Web UI] Refresh button in network configuration deselects the currently selected interface


cdealti:
The refresh button deselects the currently selected interface, for example wlan0,  and sets it to the loopback interface.

dwoodard1:
Should be fixed in #448 .

dwoodard1:
@cdealti @ibinshtok Can one of you review this?

PR #450, opened 2016-08-11
Fix to retain selected network interface on refresh.

dwoodard1:
This PR is for #290

### DIFF
--- a/kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/client/ui/Network/NetworkInterfacesTableUi.java
+++ b/kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/client/ui/Network/NetworkInterfacesTableUi.java
@@ -53,6 +53,7 @@ public class NetworkInterfacesTableUi extends Composite {
 	}
 
 	private static final Messages MSGS = GWT.create(Messages.class);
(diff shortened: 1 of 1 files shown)


kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/client/ui/Network/NetworkInterfacesTableUi.java
```diff
@@ -54,6 +54,7 @@
 
 	private static final Messages MSGS = GWT.create(Messages.class);
 	private final GwtNetworkServiceAsync gwtNetworkService = GWT.create(GwtNetworkService.class);
+	private final String SELECTED_INTERFACE = "ui.selected.interface";
```
MMaiero: @dwoodard1 Maybe this variable could be static final

cdealti:
@dwoodard1 It works for me. Ready to merge if you accept @MMaiero suggestion

dwoodard1:
@MMaiero I am not sure about the static variable. I think we want a separate instance for each instantiation. In case you are working with the Kura UI with separate tabs/windows/browsers.

dwoodard1:
@MMaiero Made your suggested change, have a look.

PR #452, opened 2016-08-12
Modified target-platform build to have correct versions when in -SNAP…

MMaiero:
…SHOTS. Modified ant, distrib/pom.xml and .gitignore accordingly.

@pierantoniomerlino or @cdealti Could you check this?


dwoodard1:
@MMaiero @ctron Has been working on a substantial rework of the target-platform. Let's wait until he submits the PR and see if we agree it is a better approach. You can take a look at the upcoming PR here: https://github.com/ctron/kura/tree/feature/fix_build_3

ctron:
The way this is done in the new way is to copy the actual artifacts from the local build, extract the OSGi information from those artifacts, create the versions property file from that and use this in the final distrib step to copy exactly those versions. This also works with -SNAPSHOT versions.

The change in this branch can now also assemble DPs fir modbus and can and provides a P2 repository which can be consumed by others.

Issue #242, opened 2016-05-25
[Web UI] Help is not always visible in the new web ui


pierrepitiot:
Help says "Mouse over enabled items on the left to see help" but the vertical spacing is too large and at a certain point help is not more visible

dwoodard1:
@pierrepitiot Can you give a specific example of this? All of the help text I see looks okay.

dwoodard1:
@cdealti @MMaiero @pierantoniomerlino @ibinshtok @LucaDazi I believe Pierre is on Holiday. Has anyone else seen this issue? Can you provide instructions for repeating?

pierantoniomerlino:
@dwoodard1 I tried to replicate the issue, but I have never seen it.

MMaiero:
I have no ways to test here but, if I remember well, if you go in the modem tab and scroll down, the help box does not scroll down. So, you cannot get help.

cdealti:
Yes. Select Network | <modem> | Cellular and scroll down the configuration. The help box does not scroll down accordingly. I believe we can postpone it.

pierantoniomerlino:
Ah! Now I understand.

dwoodard1:
@cdealti I have a fix, it is pretty simple. As long as we don't want to release today it can stay.

dwoodard1:
@MMaiero @pierantoniomerlino Should be better now, please confirm.

PR #454, opened 2016-08-12
Fix for dialog on Modem help text.

dwoodard1:
Fix for #242


MMaiero:
@dwoodard1 I would do the same also for the wireless tab

MMaiero:
@dwoodard1 However, I see that if you reduce the browser window width, the help text box disappears

dwoodard1:
@MMaiero I will check the wireless tab, thought I had checked them all. Not sure what to do about the small screens. Where would the dialog go? I think this would be a bigger design change than should be covered in this PR.

cdealti:
@MMaiero Shall we merge?

Issue #396, opened 2016-08-02
iwScanTool fails to parse scan results


lmarshall-eth:
There's a network in my area which causes iwScanTool to fail when parsing the results of `iw dev wlan0 scan`.

The following error is produced:
```
2016-08-02T09:07:46.384Z [qtp27081630-40] WARN org.eclipse.kura.web.server.util.KuraExceptionHandler - RPC service non-application error
  org.eclipse.kura.KuraException: An internal error occurred. scan operation has failed
...
...
  Caused by: org.eclipse.kura.KuraException: An internal error occurred. error parsing scan results
  	at org.eclipse.kura.linux.net.util.iwScanTool.scan(iwScanTool.java:162)
  	at org.eclipse.kura.net.admin.NetworkAdminServiceImpl.getWifiHotspots(NetworkAdminServiceImpl.java:1161)
  	... 38 more
  Caused by: java.lang.NumberFormatException: For input string: "IBSS"
  	at sun.misc.FloatingDecimal.readJavaFormatString(Unknown Source)
  	at sun.misc.FloatingDecimal.parseFloat(Unknown Source)
  	at java.lang.Float.parseFloat(Unknown Source)
  	at org.eclipse.kura.linux.net.util.iwScanTool.parse(iwScanTool.java:372)
  	at org.eclipse.kura.linux.net.util.iwScanTool.scan(iwScanTool.java:160)
  	... 39 more
```

The failure is caused by the following item in the list:
```
BSS 00:11:22:33:44:55(on wlan0)
	TSF: 10275279415182 usec (118d, 22:14:39)
	freq: 2412
	beacon interval: 100 TUs
	capability: IBSS Privacy (0x0012)
	signal: -84.00 dBm
	last seen: 8070 ms ago
	Information elements from Probe Response frame:
	SSID: TheSSID
	Supported rates: 1.0* 2.0 5.5 11.0 6.0 9.0 12.0 18.0 
	DS Parameter set: channel 1
	IBSS ATIM window: 0 TUs	Extended supported rates: 24.0 36.0 48.0 54.0 
	WMM: information: 01 00
```

Specifically, the line:
`IBSS ATIM window: 0 TUs	Extended supported rates: 24.0 36.0 48.0 54.0`

This seems like a bug in iw (version 3.11), and it is really supposed to be two separate lines.
[...]

PR #477, opened 2016-08-26
refs #396: Fix parsing and refactor

lmarshall-eth:
Signed-off-by: Lee Marshall <lee.marshall@eurotech.com>

Here, we fix the problem with parsing Extended Supported Rates and split the parsing up into separate modules.


ibinshtok:
@cdealti Yes, will do

PR #484, opened 2016-08-29
Modified wrong references in some snapshots, modified sslManager serv…

MMaiero:
…ice to perform a different check for the default password and the password change.


MMaiero:
@dwoodard1 Could you check if this PR is ok for you?

PR #499, opened 2016-09-02
[Windows] Add 32-bit/x86 (Win32-nn) build & fix build failure

malikm:
Signed-off-by: Martin Malik <martin.malik@eurotech.com>


malikm:
Travis build fails because NSIS compiler is not installed:

> [exec] WARNING: The NSIS tools are not installed the Windows installer program will not be built.
> : Could not find file /home/travis/build/eclipse/kura/kura/distrib/target/kura_2.1.0-SNAPSHOT_Win32-nn/kura_2.1.0-SNAPSHOT_Win32-nn_x86.exe to copy.

ctron:
I would recommend to make that an optional maven profile which can be activated and deactivated when needed.

ctron:
Still NSIS is supported by travis ci, so you just need to request the package in the travis ci file.

malikm:
I'll fix this by not failing when the file is not present, so those without NSIS won't have troubles.

cdealti:
@malikm I'd follow @ctron suggestion. @ctron can you take care of reviewing and merging this?

cdealti:
@ctron please review #499

ctron:
I will do so.

Btw, is there a CQ for NSIS?

ctron:
@malikm could you also point to me some README or wiki page what the expectation on NSIS are, regarding versions, OS, …

kura/distrib/src/main/resources/Win32-nn/start_kura_debug.bat
```diff
@@ -0,0 +1,32 @@
+@echo off
+
+set KURA_ROOT=c:\opt\eclipse
+set KURA_HOME=%KURA_ROOT%\kura\kura
+set KURA_HOME_URI=/%KURA_HOME:\=/%
+set KURA_DATA=%KURA_ROOT%\kura\data
+set KURA_TEMP=\tmp\.kura
```
ctron: Isn't there something like %TEMP% on windows?

target-platform/org.eclipse.kura.windows.service/src/main/c/main.c
```diff
@@ -34,6 +35,25 @@ char KuraJavaArgs[] = "java \
...
+-Dtarget.device=windows \
+-Declipse.ignoreApp=true \
+-Dkura.home=%KURA_HOME% \
+-Dkura.data.dir=%KURA_DATA% \
+-Dkura.configuration=file:%KURA_HOME_URI%/kura.properties \
+-Dkura.custom.configuration=file:%KURA_HOME_URI%/kura_custom.properties \
+-Ddpa.configuration=%KURA_HOME%\\dpa.properties \
+-Dlog4j.configuration=file:%KURA_HOME_URI%/log4j.properties \
+-Djdk.tls.trustNameService=true \
+-jar \"%~dp0\\plugins\\org.eclipse.osgi_3.8.1.v20120830-144521.jar\" \
```
ctron: Is it a wise idea to hardcode that in a C file? Looking forward to the Equinox 3.11.0 upgrade, this will cause some major issues because the final target can no longer be built by just switching a maven profile.

ctron:
Several files contain a copyright header starting 2011. Where they really written in 2011? Otherwise please update them reflecting the correct year.

kura/distrib/src/main/resources/Win32-nn/Kura.nsi
```diff
@@ -0,0 +1,435 @@
+!include "Library.nsh"
```
ctron: Beside not having a copyright header, this file is copied from somewhere else [1]. For external content you will need to file a CQ.

[1] http://nsis.sourceforge.net/mediawiki/index.php?title=ReplaceInFile&oldid=20490

ctron:
In general, Eclipse copyright headers should state "Copyright (c) 2011, 2016 Eurotech and others"

malikm:
See here: http://nsis.sourceforge.net/Features

kura/distrib/src/main/resources/Win32-nn/start_kura_debug.bat
malikm: Yes and that's actually done by the setup.
Our idea was to keep hard-coded paths in the first instance, so that a user could run Kura without installing - by just copying the files to fixed directory locations (similar to Linux paths).
The NSIS setup replaces all these hard-coded paths with final destinations.

target-platform/org.eclipse.kura.windows.service/src/main/c/main.c
malikm: I agree, it's not the best idea. We will handle this later when required.

kura/distrib/src/main/resources/Win32-nn/Kura.nsi
malikm: Yes, the ReplaceInFile function was copied from that public library.
The rest is using standard NSIS functions and libraries, so you might consider a whole NSIS CQ if needed.

kura/distrib/src/main/resources/Win32-nn/start_kura_debug.bat
ctron: But wouldn't it be much easier for a user to rely on OS specific behavior, like %TEMP% instead of requiring him to create a "/tmp" directory first. Which might not even be possible on some machines due to user access restrictions?

malikm: See: "if not exist c:\tmp mkdir c:\tmp"
This batch file is also intended for debug purposes only.

ctron: So that script will definitely fail in companies which don't allow access to "C:\". For home users that may be fine, but for company users?

ctron:
So to wrap it all up:
* I am missing a CQ for NSIS and its dependencies
* There is copied code from some external source without a CQ
* The Equinox version is hard-coded, that should be fixed
* The license header has to be fixed (at least the copyright year and the type of contribution)
* File `main.c` [1] contains copied code from [2], no copyright information about that is is provided in the license header. Maybe the source is even a different one.

[1] https://github.com/malikm/kura_windows_2/blob/37166c2a517572523db029b228cce4a3b4ed33bb/target-platform/org.eclipse.kura.windows.service/src/main/c/main.c#L60
[2] http://stackoverflow.com/questions/3659694/how-to-replace-substring-in-c/32503536

kura/distrib/src/main/resources/Win32-nn/start_kura_debug.bat
malikm: If UAC is active (high) and the batch file is not run as Admin, it will fail.
We expect those going to debug to have advanced knowledge.

malikm:
If [1] would be based on the function published on [2] (which seems is heavy modified), what would cure this? Mentioning [2] in the copyright header, or replace the entire function with new code ?
BTW, all this code has already made it into the repo during the initial PR. This PR is just an extension for 32-bit support.

malikm:
Can we postpone resolving the hard-coded Equinox version until a bit later?
This will require more changes, also to files beyond this PR.

cdealti:
> Can we postpone resolving the hard-coded Equinox version until a bit later?
> This will require more changes, also to files beyond this PR.

From a milestone perspective I think so. The new Equinox target platform is not supposed to go to Kura 2.1.0.

However, what changes would be needed to externalize these in batch or configuration file?

malikm:
> However, what changes would be needed to externalize these in batch or configuration file?

Not sure yet what would be the ideal solution. One proposal might be a single batch file modified accordingly by build_equinox_distrib.xml (to replace Equinox version) and then read by setup/launcher to be modified further (for final paths) and executed.

cdealti:
> Not sure yet what would be the ideal solution. One proposal might be a single batch file modified accordingly by build_equinox_distrib.xml (to replace Equinox version) and then read by setup/launcher to be modified further (for final paths) and executed.

I believe this is acceptable. We use the same approach for the Linux builds.

ctron:
So let's open a new issue right away tracking this!

ctron:
> If [1] would be based on the function published on [2], which was heavy modified, what would cure this? Mentioning [2] in the copyright header, or replace the entire function with new code ?
> BTW, all this code has already made it into the repo during the initial PR. This PR is just an extension for 32-bit support.

You can a) ask for permission to use this code from the original author and file a CQ for that and mention it in the copyright header, b) replace it with a version you authored.

Just that it happened in the past, doesn't make it right!

malikm:
> Just that it happened in the past, doesn't make it right!

Sure, I just wanted to point out, that we already have code inside that will need attention.
I'll go for option b).

ctron:
Fine with me. So that means that we have to ask EMO if the Eclipse Webmaster needs to purge those files from the history of this repository.

malikm:
So except the CQ for NSIS are we OK now with all the rest ?

ctron:
I think we need a general build-and-test CQ, currently there exists none. For this PR this seems to cover NSIS and MsBuild. Do you know of anything else?

A build and test dependency CQ would only cover the use of this tool. Resources which are not authored by you need a dedicated CQ if they are distributed in any way (including the wiki or GitHub). So for example copying a snippet from some webpage (like the RIF method) would require a CQ, including the source, the author, the license, … Since that function (probably) is not part of the standard NSIS distribution this would not be covered by a dedicated NSIS CQ.

But I guess you need a full CQ for NSIS anyway. If you want to distribute an NSIS installer, then it will (probably) attach some "EXE" code or some resources, not authored by you, to your payload data. If this "setup.exe" should be distributed from Eclipse download servers, then the Kura distribution does contain code from NSIS. So a full CQ for NSIS is required before making a release of Kura.

Maybe it would be a good idea to read through the Eclipse Project Handbook [1] and its references.

[1] https://www.eclipse.org/projects/handbook/

ctron:
I am currently in the process of setting up a docker build for testing NSIS. Since I had no information it took a bit longer. I hope to have this tested by tomorrow.

ctron:
So I manage to build all that stuff in a docker container, using Ubuntu 16.04. Since RHEL 7 does not seem to have some "nsis" package ready, at least for me, I had to make use of docker. I am not sure if the HIPP instance can provide docker. I do know they support WIX though, which btw is approved.

However I cannot make an installation test. If you like I can provide you with the Dockerfile or binaries it produces.

Please be aware of the fact though that you cannot distribute those binaries as of now, since there is now CQ for NSIS and NSIS does embed code in the resulting EXE file.

malikm:
Yes, please provide the binaries and I will run a test.

ctron:
https://tresor.it/s#W75az8iZEGdyBzciS3r1Cg
https://tresor.it/s#CHkKtUQlaRdMTxc1yig4oQ

malikm:
Thanks!
Looks like some recent changes in the repo have broken a few things in the Windows version. I need to fix this now.

malikm:
Changes in 6cbddeb1845c892011f736f3d2889feac2849270 were not merged into the windows target yet.
This again brings the idea of sharing common build code for both windows and linux targets.

cdealti:
@ctron @malikm can you confirm the ONLY legal issue we have already introduced in the **develop** with commit:

https://github.com/eclipse/kura/commit/741220741bf613a43040afaa890f37f96ccf0c6d

 is for https://github.com/eclipse/kura/blob/develop/kura/distrib/src/main/resources/Win64-nn/Kura_x64.nsi ?

and that the problem is the ``_ReplaceInFile`` macro copied from elsewhere?

If this is the case can we create a separate issue for this one?

malikm:
Not just the _ReplaceInFile function in the nsi, but also str_replace in:
https://github.com/eclipse/kura/blob/develop/target-platform/org.eclipse.kura.windows.launcher/src/main/c/main.c
and
https://github.com/eclipse/kura/blob/develop/target-platform/org.eclipse.kura.windows.service/src/main/c/main.c
The latter function has been replaced by this PR.

ctron:
@cdealti I cannot confirm this without doing a full check. But I also would recommend to the through the rest of the code and see if there are additional cases like that. Since this only popped up in the review of this single PR.

malikm:
I'm working on replacing the `_ReplaceInFile` function with own code, will attach it to this PR once ready.

cdealti:
@malikm @ctron Thanks. In the meantime we need to create a legal issue so that @dwoodard1 can take appropriate actions with EMO.

malikm:
All known problematic functions are now replaced with own code.

cdealti:
@malikm @ctron About #545, EMO (Eclipse Management Organization) has decided that we need to delete from the repository history the problematic commit. @malikm this **does not** mean pushing another commit with a re-implementation of the two functions.
Deleting a commit from the history is a serious thing that breaks existing forks, pull requests etc.

Since we are forced to do this, to minimize the resulting mess and idea would be to delete the commit and insert another commit  in the same point of the history containing the original but sanitized contribution (via git rebase).

I also believe that the changes in **this** PR should be moved in the sanitized commit replacing the problematic one so I think this should be closed. 

**We can discuss the details later on but it's important to update all the existing IP issues in #545.**

malikm:
I'm sorry for the troubles this has caused.
I'm not experienced in this area, but wouldn't it be simpler to file CQs for those 2 functions, rather than breaking the repo? Not sure if that would work afterwards, or if it would make things simpler at all...
I assume the NSIS function requires the same CQ as the rest of NSIS, which we need anyway.

ctron:
A CQ for GPL code won't be approved. [1]

And I doubt that you find clear licensing terms for the NSIS code snippet on the wiki. You may of course get in contact with the author and ask for clarification.

So since re-writing history seems to be necessary anyway, just clean up both.

I really suggest that you read through the Eclipse project handbook [2]!

[1] https://www.eclipse.org/legal/EclipseLegalProcessPoster.pdf
[2] https://www.eclipse.org/projects/handbook/

cdealti:
@ctron @malikm Is it OK to close this one until we a clear picture of #545?

malikm:
This fixes also other problems like 37166c2a517572523db029b228cce4a3b4ed33bb
Would it be possible to squash this with the original Windows PRs ? That would also replace all offending functions (if possible).

cdealti:
@malikm @markcullen @ctron I'd like to close this pull request and create an issue wrapping up the contents of this discussion about MinGW vs Visual Studio and NSIS vs Wix.
@malikm can you please do that?

malikm:
Yes

cdealti:
OK, then please close it when done. Thanks!

malikm:
See: https://github.com/eclipse/kura/issues/637

Issue #326, opened 2016-07-15
org.eclipse.kura.web2 is missing about.html


ctron:
The project `org.eclipse.kura.web2`is missing the required `about.html` file.

Since there seem to be some icon resources in the project, this has to be explicitly mentioned in the "Third Party Content" section.

dwoodard1:
Which icon resources are you referring to? I believe all the icons/images (except Kura logo) come from the bootstrap js/css. Are you seeing something different?

ctron:
The files under org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/client/ui/resources/icons

ctron:
That is fine if they come from bootstrap. But still external dependencies have to be named in the `about.html`

ctron:
Also see https://www.eclipse.org/legal/epl/about.php

dwoodard1:
@ctron I forgot about the icons directory. Those do not come from bootstrap and I am not sure they are needed.

@MMaiero Can we get rid of these icons? My quick search shows we are only using two (for connection status). Seems like a lot of icons to only be using two.

MMaiero:
@dwoodard1 Yes, I think we can remove them

PR #504, opened 2016-09-05
Added about* files in web2 bundle.

MMaiero:

@dwoodard1 Could you check this?


kura/org.eclipse.kura.web2/about.html
```diff
@@ -0,0 +1,28 @@
...
+at <a href="http://www.eclipse.org/legal/epl-v10.html">http://www.eclipse.org/legal/epl-v10.html</a>.
+For purposes of the EPL, &quot;Program&quot; will mean the Content.</p>
+
+<p>If you did not receive this Content directly from the Eclipse Foundation, the Content is 
+being redistributed by another party (&quot;Redistributor&quot;) and different terms and conditions may
+apply to your use of any object code in the Content.  Check the Redistributor's license that was 
+provided with the Content.  If no such license exists, contact the Redistributor.  Unless otherwise
+indicated below, the terms and conditions of the EPL still apply to any source code in the Content
+and such source code may be obtained at <a href="http://www.eclipse.org/">http://www.eclipse.org</a>.</p>
+
```
ctron: This file is missing the section regarding 3rd party content in this bundle!

MMaiero:
@ctron That is known. @dwoodard1 is checking that and will eventually push a commit to this PR.

dwoodard1:
This fix addresses Issues #326

dwoodard1:
@MMaiero Can you review? The icon doesn't look quite as nice, but I don't think it is noticeable. And it allows us to get rid of the resources folder entirely.

@ctron With that folder gone, is there a need for the 3rd party content in the about file? I believe everything is now covered by CQs.

ctron:
There was no CQ for that?

Also if you have a CQ, if it is part of this bundle, you need that section. e.g. if you put a JAR or JS, CSS file in a bundle, which is 3rd party content, you do need a CQ, but you also need to add that section.

ctron:
Wasn't that [CQ 8207](https://dev.eclipse.org/ipzilla/show_bug.cgi?id=8207)?

Browsing through the module that should be fine for the bundle. But we should check the resulting binary. It may be that the GWT compile process does add some static resources for GWT.

dwoodard1:
There was a CQ for the icons that were just removed. They were used in the origin web UI.

Thanks for the explanation, I think I finally understand what you mean. Some times I am a bit slow. I will get the about file amended accordingly.

ctron:
I just checked the final JAR. There is "fontawesome" in the bundle. So we do need that section anyway.

ctron:
And bootstrap

ctron:
I think we should add a task for checking all bundles regarding this point!

dwoodard1:
So, for Font Awesome, do we need a different section in the about file for each file? For example:

www/denali/css/font-awesome-4.3.0.min.cache.css
www/denali/fonts/FontAwesome.otf 
www/denali/fonts/fontawesome-webfont.eot
www/denali/fonts/fontawesome-webfont.svg 
etc.

Or can we simply have one section to cover a directory: www/denali/css/fonts ?

ctron:
No, this would be for each CQ.

dwoodard1:
@ctron I have updated the about.html to include 3rd party content. Can you have a look and see if it is what's expected?

@MMaiero @cdealti I have looked and don't see any reason not to remove the resources folder containing the icons. 3rd party bundles can't access these resources, they are only available to the web bundle itself through [this class](https://github.com/eclipse/kura/blob/develop/kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/client/ui/resources/Resources.java). Let me know if you disagree. Again, we have an approved CQ for these, but I think cutting any size from the web bundle would be a good thing.

ctron:
So I just double checked the target JAR file. I see that we now have a section for Fontawesome and GwtBootstrap. What is still unclear to me are the following entries:

```
    45404  07-28-2016 16:27   www/denali/fonts/glyphicons-halflings-regular.ttf
    23424  07-28-2016 16:27   www/denali/fonts/glyphicons-halflings-regular.woff
   108738  07-28-2016 16:27   www/denali/fonts/glyphicons-halflings-regular.svg
    20127  07-28-2016 16:27   www/denali/fonts/glyphicons-halflings-regular.eot
    18028  07-28-2016 16:27   www/denali/fonts/glyphicons-halflings-regular.woff2

   117150  07-28-2016 16:27   www/denali/css/bootstrap-3.3.2.min.cache.css
    19963  07-28-2016 16:27   www/denali/css/bootstrap-theme-3.3.2.min.cache.css
    35452  07-28-2016 16:27   www/denali/js/bootstrap-3.3.2.min.cache.js

    95931  07-28-2016 16:27   www/denali/js/jquery-1.11.2.min.cache.js
```

dwoodard1:
Yes, it took me a while to figure out exactly how things were approved. That is why I was asking if each folder needed an entry in the about.html. Here is what I found:

[CQ10136](https://dev.eclipse.org/ipzilla/show_bug.cgi?id=10136): This covers gwtbootstrap3. The source code for this CQ included bootstrap JS and CSS files, Font Awesome (icon files and CSS), Glyphicons icon files, and JQuery. As you can see in the CQ, the IP team requested a separate CQ for Font Awesome and JQuery.

[CQ10506](https://dev.eclipse.org/ipzilla/show_bug.cgi?id=10506): CQ for Font Awesome. The source code for this CQ only included the icon files.

[CQ10504](https://dev.eclipse.org/ipzilla/show_bug.cgi?id=10504): CQ for JQuery. I agree, this should be added to the about.html file.

ctron:
Ok I see. Then I misunderstood your question earlier, sorry for that.

So you do need an entry for each dependency which is distributed in that bundle. Including transitive dependencies if they show up in the final bundle.

dwoodard1:
That is what I was going for, so wouldn't the below be acceptable? It should cover all the 3rd party content. The only one I am not sure of is Glyphicons. [Their license page](https://glyphicons.com/license/) seems to have a special handling for bootstrap, which I assume is why a separate CQ was not required. 

<h3>Third Party Content</h3>
<p>
The Content includes items that have been sourced from third parties as set out below. If you did not receive this Content directly from the Eclipse Foundation, the following is provided for informational purposes only, and you should look to the Redistributor's license for terms and conditions of use.
</br></br>
<b>font-awesome-4.3.0</b>
This package provides scalable graphics and CSS files for rendering certain graphics within the Eclipse Kura UI. This package can be obtained from http://fontawesome.io/#modal-download and is licensed under MIT and SIL OFL 1.1. 
</p>
<b>gwtbootstrap3-0.9.1</b>
This is a wrapper library for Bootsrap and provides the framework for UI elements within the Eclipse Kura UI. This package can be obtained from https://github.com/gwtbootstrap3/gwtbootstrap3/tree/0.9.1 and is licensed under Apache License v2.0.
<b>jquery-1.11.2</b>
This package provides the JQuery library required by gwtbootstrap3. The package can be obtained from https://github.com/jquery/jquery and is licensed under the MIT license.

ctron:
I was just digging through the list of Kura CQs regarding that issue. I am currently writing on my mobile, so please forgive me for not adding links.

In general yes. Only bootstrap itself seems to be missing. Sharon wanted to have an additional CQ for that so I would say it has to be added?

dwoodard1:
@MMaiero Can you please review? The about file should be good to go, so I just need confirmation that removing the icons is okay with everyone.

cdealti:
Can we rebase and merge this ASAP?

dwoodard1:
@cdealti Yes, I will work on this now, but please review. I don't see any reason to not remove all the unused Resource icons, but I may have missed something.

cdealti:
@MMaiero any objections to removing the icons?

MMaiero:
For me is ok. I'll merge as soon as I complete some tests to verify that everything works as expected.

PR #524, opened 2016-09-07
Initial support for different repositories. Selection performed via p…

MMaiero:
…roperty or profile.

@cdealti Could you check it?

### DIFF
--- a/kura/manifest_pom.xml
+++ b/kura/manifest_pom.xml
@@ -82,19 +82,6 @@
 		<sonar.jacoco.reportPath>target/jacoco.exec</sonar.jacoco.reportPath>
 	</properties>
 
-	<distributionManagement>
-			<repository>
-				<id>repo.eclipse.org</id>
-				<name>Eclipse Kura Repository - Releases</name>
-				<url>https://repo.eclipse.org/content/repositories/kura-releases/</url>
-			</repository>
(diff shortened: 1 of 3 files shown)


kura/manifest_pom.xml
```diff
@@ -137,6 +124,55 @@
...
+				</property>
+				<activeByDefault>false</activeByDefault>
+			</activation>
+			<properties>
+			</properties>
+			<distributionManagement>
+				<repository>
+					<id>releases</id>
+					<name>releases</name>
+					<url>http://172.17.0.2:8081/content/repositories/releases</url>
```
ctron: I don't think it is a good idea to use internal addressing in an open source project. I would recommend to use something like maven settings for that.

kura/pom_pom.xml
```diff
@@ -113,6 +100,55 @@
...
+					<name>!useTestRepo</name>
+				</property>
+				<activeByDefault>true</activeByDefault>
+			</activation>
+			<properties>
+			</properties>
+			<distributionManagement>
+				<repository>
+					<id>repo.eclipse.org</id>
+					<name>Kura Repository - Releases</name>
```
ctron: Is there something wrong with using the proper name "Eclipse Kura"?

Issue #523, opened 2016-09-07
[Speed up build] Checkums are generated multiple times


ctron:
To me it looks the process of generating checksums (which is rather time consuming and CPU intensive) is generating checksums of the same artifacts over and over again during one build process:

The following snippet from a standard Kura build shows that both executions (raspberry-pi-2-deb-checksums and raspberry-pi-2-nn-deb-checksums) generate the checksums for the same files.

So instead of generating N checksums for N device profiles, it actually generated N² checksums.

```
[INFO] --- checksum-maven-plugin:1.3:files (raspberry-pi-2-deb-checksums) @ distrib ---
[INFO] kura_2.1.0-SNAPSHOT_raspberry-pi-2-nn_installer.deb - MD5 : bae6a0fbdabc1684884f193238619295
[INFO] kura_2.1.0-SNAPSHOT_raspberry-pi-2-nn_installer.deb - SHA-1 : 38ebdc386952efc00d1561e6e95630d480ee0d91
[INFO] kura_2.1.0-SNAPSHOT_raspberry-pi_installer.deb - MD5 : b2b21eb4e8007d56a984a1ba8542c2ba
[INFO] kura_2.1.0-SNAPSHOT_raspberry-pi_installer.deb - SHA-1 : cae98a14696750d4381338a14aff58f0521c7a04
[INFO] kura_2.1.0-SNAPSHOT_raspberry-pi-bplus-nn_installer.deb - MD5 : 290f5c962a6ef692c9ad4d2aec09131b
[INFO] kura_2.1.0-SNAPSHOT_raspberry-pi-bplus-nn_installer.deb - SHA-1 : 472e55def127bf1f253fa7f76106771d67e6f02d
[INFO] kura_2.1.0-SNAPSHOT_raspberry-pi-nn_installer.deb - MD5 : 5b0a6a79e5b738cdf04434e83da03834
[INFO] kura_2.1.0-SNAPSHOT_raspberry-pi-nn_installer.deb - SHA-1 : 6dbf91462aac859f5eb5291bcc9c64bd327fc5df
[INFO] kura_2.1.0-SNAPSHOT_beaglebone_debian_installer.deb - MD5 : f3191b0fa3254a1dd5d04093ca05ecfa
[INFO] kura_2.1.0-SNAPSHOT_beaglebone_debian_installer.deb - SHA-1 : a8ede9162e604a72dc2b8ea6ce50e35535408204
[...]

PR #526, opened 2016-09-07
Modified checksum calculation.

MMaiero:
Signed-off-by: MMaiero <matteo.maiero@eurotech.com>

Closes #523 

@ctron

PR #527, opened 2016-09-07
#523 only calculate checksums once

ctron:
This changed does restructure the build so that they get only calculated
once. Before that each device profile did generate the checksums for
all other device profiles in addition.

The main change is that the calls to "build_equinix_distrib.xml" get
shifted to the "package" phase and that there is only one call to the
checksum plugin in the "install" phase.

In order to not let the checksum plugin fail when do "deb" profile is
active, a dummy file is being generated early in the build.


ctron:
Superseded by #526

PR #528, opened 2016-09-07
Fix the lookup of the default value, now done by id.

ctron:
This change does fix another location where values of the properties map
got addressed by name instead of id.


cdealti:
@ctron what are the backward compatibility implications of this change?

ctron:
There should be none. Currently the name and ID have to be equal. But with one of the latest changes in the Web UI, that was changed. So that the name actually is a name.

But if you currently do use this feature, having a different value in name than in ID, then components don't get started because the are not configured because there is no default by looking it up "by name" instead of "by id".

So if you follow the path of the past (name equals ID) then there is no difference. But if you use the name as label now, then it will fail.

cdealti:
@ctron I confess I'm confused.
Can you clarify how you interpret ``id`` and ``name`` of the Meta Type? Is ``name`` meant as the "display name" (the label) of the attribute in the Kura Web UI while the **name** of the property in the Config Admin matches ``id``?

Also note that the [Configuration XML serializer](https://github.com/eclipse/kura/blob/develop/kura/org.eclipse.kura.core.configuration/src/main/java/org/eclipse/kura/core/configuration/util/serializers/XmlJavaComponentConfigurationsMapper.java) has an ``name`` attribute which which represent the name of the property in the Config Admin

ctron:
Yes, that's how I (an I guess OSGi) interpreted it. See also [1], [2].

[1] http://felix.apache.org/documentation/subprojects/apache-felix-metatype-service.html
[2] http://blog.osgi.org/2011/03/metatypes.html

cdealti:
The ``ConfigurationService`` has a ``getComponentConfiguration`` methods that returns a ``ComponentConfiguration`` from which you can access the ``AD``s of the Meta Type.
A consumer of this API must now be aware of the distinction between ``name`` and ``id``.
We don't know which one it is using right now.

ctron:
So this works in combination with PR #345, if you plan on not merging this one, please also be sure to revert the "ID" logic in PR #345. Maybe coordinate with @dwoodard1 who wanted to check into this topic.

cdealti:
I've just tested the original behaviour with Kura 2.0.1 (pre #345 and this PR of course) with the example publisher bundle.
I've modified the following AD like this
```
        <AD id="publish.rate"  
            name="Publish Rate"
            type="Integer"
            cardinality="0" 
            required="true"
            default="1000"
            min="1"
            description="Default message publishing rate in milliseconds (min 1)."/>
```

What happens is that the Kura Web UI shows the the ``name`` attribute ``Publish Rate`` but as we can see **without** the default value (BAD)

image

Also the ExamplePublisher component does not receive the default value of the property from the ConfigurationService (BAD)

```
2016-09-15 17:26:34,145 [Component Resolve Thread (Bundle 7)] INFO  o.e.k.e.p.ExamplePublisher - Activating ExamplePublisher...
2016-09-15 17:26:34,146 [Component Resolve Thread (Bundle 7)] INFO  o.e.k.e.p.ExamplePublisher - Activate - metric.char: a
2016-09-15 17:26:34,147 [Component Resolve Thread (Bundle 7)] INFO  o.e.k.e.p.ExamplePublisher - Activate - metric.byte: 119
2016-09-15 17:26:34,147 [Component Resolve Thread (Bundle 7)] INFO  o.e.k.e.p.ExamplePublisher - Activate - metric.short: 32759
2016-09-15 17:26:34,147 [Component Resolve Thread (Bundle 7)] INFO  o.e.k.e.p.ExamplePublisher - Activate - publish.appTopic: data/metrics
2016-09-15 17:26:34,147 [Component Resolve Thread (Bundle 7)] INFO  o.e.k.e.p.ExamplePublisher - Activate - metric.integer: 2147483599
2016-09-15 17:26:34,148 [Component Resolve Thread (Bundle 7)] INFO  o.e.k.e.p.ExamplePublisher - Activate - component.id: 46
2016-09-15 17:26:34,148 [Component Resolve Thread (Bundle 7)] INFO  o.e.k.e.p.ExamplePublisher - Activate - metric.string.array: [Ljava.lang.String;@1780029
2016-09-15 17:26:34,148 [Component Resolve Thread (Bundle 7)] INFO  o.e.k.e.p.ExamplePublisher - Activate - metric.string.oneof: string.value.option.1
2016-09-15 17:26:34,149 [Component Resolve Thread (Bundle 7)] INFO  o.e.k.e.p.ExamplePublisher - Activate - component.name: org.eclipse.kura.example.publisher.ExamplePublisher
2016-09-15 17:26:34,149 [Component Resolve Thread (Bundle 7)] INFO  o.e.k.e.p.ExamplePublisher - Activate - metric.temperature.initial: 10.0
2016-09-15 17:26:34,150 [Component Resolve Thread (Bundle 7)] INFO  o.e.k.e.p.ExamplePublisher - Activate - metric.short.array: [Ljava.lang.Short;@7bde9c
2016-09-15 17:26:34,150 [Component Resolve Thread (Bundle 7)] INFO  o.e.k.e.p.ExamplePublisher - Activate - objectClass: [Ljava.lang.String;@13056fa
2016-09-15 17:26:34,150 [Component Resolve Thread (Bundle 7)] INFO  o.e.k.e.p.ExamplePublisher - Activate - Publish Rate: 1000
2016-09-15 17:26:34,151 [Component Resolve Thread (Bundle 7)] INFO  o.e.k.e.p.ExamplePublisher - Activate - metric.double: 4.29496729599998E9
2016-09-15 17:26:34,151 [Component Resolve Thread (Bundle 7)] INFO  o.e.k.e.p.ExamplePublisher - Activate - publish.retain: false
2016-09-15 17:26:34,151 [Component Resolve Thread (Bundle 7)] INFO  o.e.k.e.p.ExamplePublisher - Activate - metric.temperature.increment: 0.1
2016-09-15 17:26:34,151 [Component Resolve Thread (Bundle 7)] INFO  o.e.k.e.p.ExamplePublisher - Activate - metric.string: string.value
2016-09-15 17:26:34,151 [Component Resolve Thread (Bundle 7)] INFO  o.e.k.e.p.ExamplePublisher - Activate - metric.password: xea2sebrvKJQEW1YRDEEGg==
2016-09-15 17:26:34,152 [Component Resolve Thread (Bundle 7)] INFO  o.e.k.e.p.ExamplePublisher - Activate - publish.qos: 0
2016-09-15 17:26:34,152 [Component Resolve Thread (Bundle 7)] INFO  o.e.k.e.p.ExamplePublisher - Activate - service.pid: org.eclipse.kura.example.publisher.ExamplePublisher
2016-09-15 17:26:34,152 [Component Resolve Thread (Bundle 7)] INFO  o.e.k.e.p.ExamplePublisher - Activate - metric.boolean: false
2016-09-15 17:26:34,153 [Component Resolve Thread (Bundle 7)] INFO  o.e.k.e.p.ExamplePublisher - Activate - kura.service.pid: org.eclipse.kura.example.publisher.ExamplePublisher
2016-09-15 17:26:34,153 [Component Resolve Thread (Bundle 7)] INFO  o.e.k.e.p.ExamplePublisher - Activate - metric.float: 32766.98
2016-09-15 17:26:34,153 [Component Resolve Thread (Bundle 7)] INFO  o.e.k.e.p.ExamplePublisher - Activate - metric.integer.fixed: 101
2016-09-15 17:26:34,153 [Component Resolve Thread (Bundle 7)] INFO  o.e.k.e.p.ExamplePublisher - Activate - metric.long: 4294967295
2016-09-15 17:26:34,154 [Component Resolve Thread (Bundle 7)] INFO  o.e.k.e.p.ExamplePublisher - Getting CloudApplicationClient for EXAMPLE_PUBLISHER...
2016-09-15 17:26:34,155 [Component Resolve Thread (Bundle 7)] INFO  o.e.k.e.p.ExamplePublisher - Update ExamplePublisher - Ignore as properties do not contain TEMP_INITIAL_PROP_NAME and PUBLISH_RATE_PROP_NAME.
2016-09-15 17:26:34,155 [Component Resolve Thread (Bundle 7)] INFO  o.e.k.e.p.ExamplePublisher - Activating ExamplePublisher... Done.
```

So having ``name`` != ``id`` was not supported by Kura 2.0.1 and so #345 and this PR cannot break backward compatibility:

1. ``name`` is used as the property label in the Web UI (Nice)
2. ``id`` represent the key of the property in the component configuration. Unfortunately all our examples have constants like ``PUBLISH_RATE_PROP_NAME`` for the key which is misleading. They should be changed to ``PUBLISH_RATE_PROP_ID``.
3. In the serialized XML configurations the attribute **name** of the ``<esf:property>`` tag will match ``id`` of the Meta Type and not ``name``. Misleading and it cannot be changed
4. If we merge this PR we enable the distinction between id and name but existing cloud platforms will not be able to support this immediately (Everyware Cloud).

However, if you rebase this it will be merged. @dwoodard1 do you agree?

dwoodard1:
Agreed. I think this is a good change. It can make the configurable components more readable by not forcing name == id. 

@ctron Can you rebase so we can merge this and #345 ?

ctron:
I will do that.

kura/org.eclipse.kura.core.configuration/src/main/java/org/eclipse/kura/core/configuration/util/ComponentUtil.java
```diff
@@ -331,7 +332,7 @@ public static Tocd readObjectClassDefinition(Bundle bundle, String pid)
 
 				for (AD attrDef : attrDefs) {
 
-					String name = attrDef.getName();
+					String name = attrDef.getId();
```
cdealti: @ctron just to avoid confusion can you please rename the ``name`` local variable to ``id``?

ctron: +1

ctron:
Rebased and made the requested change of the variable name.

cdealti:
Thanks!

PR #529, opened 2016-09-07
Support for custom distribution repositories

MMaiero:
Custom repos ids and urls can be specified via maven properties.

@ctron Is it ok for you?

@cdealti

### DIFF
--- a/kura/manifest_pom.xml
+++ b/kura/manifest_pom.xml
@@ -81,15 +81,19 @@
 		<sonar.dynamicAnalysis>reuseReports</sonar.dynamicAnalysis>
 		<sonar.jacoco.reportPath>target/jacoco.exec</sonar.jacoco.reportPath>
 		
+		<release.repo.id>repo.eclipse.org</release.repo.id>
 		<release.repo.url>https://repo.eclipse.org/content/repositories/kura-releases/</release.repo.url>
+		<snapshot.repo.id>repo.eclipse.org</snapshot.repo.id>
 		<snapshot.repo.url>https://repo.eclipse.org/content/repositories/kura-snapshots/</snapshot.repo.url>
 	</properties>
 	
 	<distributionManagement>
 		<repository>
+			<id>${release.repo.id}</id>
 			<url>${release.repo.url}</url>
 		</repository>
 		<snapshotRepository>
+			<id>${snapshot.repo.id}</id>
 			<url>${snapshot.repo.url}</url>
 		</snapshotRepository>
 	</distributionManagement>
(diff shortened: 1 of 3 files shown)


ctron:
Why don't you just use `mvn -DaltDeploymentRepository` for doing that? [1]

[1] http://maven.apache.org/plugins/maven-deploy-plugin/deploy-mojo.html#altDeploymentRepository

MMaiero:
We used that in our vagrant test machine, but it requires the maven-deploy-plugin in version 2.8.1 to work. This approach implies a small change, but it should be working also with other plugin versions. Our approach is to have a test build for releases as close as the effective release environment, in order to prevent issues like for kura 2.0.1.

ctron:
I just did check the SVN repo ... this option is at least there since version 2.4, which is about 8 years old.

Using that property achieves that in an even closer manner. And it uses standard maven options instead of Kura specifics once again.

MMaiero:
We had an issue with that property with version 2.7.

ctron:
What kind of issue?

MMaiero:
We were unable to deploy the artifacts

ctron:
Could you be a bit more specific.

dwoodard1:
For some reason, when trying to use this option on the Eclipse Hudson server, we received ‘Unauthrozied’ errors. You can see the build output here [1]. It didn’t make much sense to me, but we were trying to push out the release so didn’t spend too much time investigating.

[1] https://hudson.eclipse.org/kura/job/kura-release/43/console

Thanks,
--Dave

On Sep 7, 2016, at 11:54, Jens Reimann <notifications@github.com<mailto:notifications@github.com>> wrote:


Could you be a bit more specific.

ctron:
Well I guess you are not authorized on the HIPP instance since there the Eclipse webmaster sets up a profile for you to do so. And you are using proxies.

But on the Eclipse Servers you already have a deployment profile. So there should be no need for that.

I guess the best thing here is to resolve the actual issue instead of creating another Kura specific way of doing things.

dwoodard1:
We should see the same behavior on the develop build as well. I will set that up and see if we get the same error. If so, I will contact the Eclipse webmaster.


On Sep 7, 2016, at 16:22, Jens Reimann <notifications@github.com<mailto:notifications@github.com>> wrote:


Well I guess you are not authorized on the HIPP instance since there the Eclipse webmaster sets up a profile for you to do so. And you are using proxies.

But on the Eclipse Servers you already have a deployment profile. So there should be no need for that.

I guess the best thing here is to resolve the actual issue instead of creating another Kura specific way of doing things.

ctron:
Maybe you should have a look at https://wiki.eclipse.org/Services/Nexus#Deploying_a_jar_to_repo.eclipse.org

Issue #492, opened 2016-08-31
[Windows] org.eclipse.kura.windows.system bundle is installed in every distrib


cdealti:
@markcullen @malikm 
...while it should only be installed on Windows related distributions.

malikm:
Would a custom profile for target-platform to build Windows components be the right solution ?

cdealti:
@malikm I don't think this is necessary for now. Just make sure that during the build of ``distrib`` Windows specific bundles are not copied to Linux based distribs.

malikm:
Should be fixed by #532

PR #532, opened 2016-09-09
Don't copy Win components to Linux distribs

malikm:
Fix for #492

Signed-off-by: Martin Malik <martin.malik@eurotech.com>


cdealti:
@malikm Thanks!

Issue #262, opened 2016-06-06
Maven failure when dev-env isn't built


lmarshall-eth:
When building without dev-env, for example with:
mvn -f pom_pom.xml -Pweb,raspberry-pi-2
the staging/target-definition directory isn't created.
This causes a build failure for the new maven archetype module.

dwoodard1:
@cdealti This is not a trivial fix. The archetype portion of the build relies on the target platform generated by the dev-env profile. I suggest postponing this until after the 2.0.0 release as all CI builds use the dev-env profile.

@lmarshall-eth your RPI2 installers were still generated even though the build failed. The archetype creation is the last step in the build.

cdealti:
@dwoodard1 any updates?

dwoodard1:
I need to look into this again. The archetype module can't build unless the target-platform has been created, hence the dependency on dev-env. I will likely just move the archetype module into a profile. I will try and give it a look today.

MMaiero:
@dwoodard1 Do we need to keep this issue open? Or we close it and, eventually, open another for the tools and dev-env correlated profiles?

dwoodard1:
@MMaiero I think close this and we can review later. This would only be an issue if someone is specifically trying to build the archetypes and doesn't use -Ptools,dev-env. I would also like to investigate using PackageDrone to collect the target-platform instead of getting it from the build. This may be the best solution, but probably a low priority item at the moment.

PR #534, opened 2016-09-09
Fix archetype build

dwoodard1:
This fix addresses issue #262 

@lmarshall-eth Can you confirm this works for you?


lmarshall-eth:
Can confirm the main build now succeeds (my bug is fixed), though the maven archetype still fails if dev-env hasn't been built.

dwoodard1:
@lmarshall-eth Thanks Lee, you are correct, if you include the -Ptools profile you will also have to include dev-env. There is not an easy work around for this, but we can take care of building this module in our CI builds. 

@MMaiero Can you merge this PR?

PR #541, opened 2016-09-13
Fix javadoc for get kura version

ibinshtok:
@cdealti please take a look at modified comments ...


cdealti:
It's OK Ilya. And I don't think we need to uptick the API for this Javadoc change only.

PR #542, opened 2016-09-14
Re-use existing code to parse kura.properties

ctron:
This change replaces two code sections which parse kura.properties on
their own instead of re-use the data from SystemService. That will lead
to the defect that the SystemService does allow a few more locations
for kura.properties, while those two code fragments only perform the
single lookup.


cdealti:
@ctron we discussed about reworking the commits by removing the formatting to better understand what has really changed

cdealti:
Can you please squash them together?

ctron:
Can't you squash them when merging? I would hate to loose the history unless it is being merges as is.

cdealti:
The PR is difficult to review to me

ctron:
I understand, and I did fix the most that I can do. Looking at the "Files changed" section you should only get the summarized (squashed) changes anyway.

However the file "LinuxProcessUtil" is just messed up with DOS formatting. And although I did revert it back to "DOS" encoding, it seems that on the way to GitHub this gets fixed (properly) to Unix encoding.

ctron:
I would suggest that you check it out locally, run "unix2dos" locally on the file and then view the diff.

dwoodard1:
@LucaDazi I have merged this PR. For CryptoServiceImpl, Jens made a change to use SystemService to get properties instead of using System.getProperty. This made sense to me since we already have methods in SystemService to return what we need.

LucaDazi:
@dwoodard1 ok, got it.
My concern is that previously we didn't use Declarative Services so to avoid a dependency on the SystemService.
In this way the CryptoService would start first, as it has no dependencies.

With this change, the CryptoService will have to wait the SystemService to start, before starting itself.
In other words, generally speaking, services related to security should be the first to start, in order to perform security checks before starting anything else.

But if you guys already discussed it, no problem. Sorry to have brought this up... just curiosity.

dwoodard1:
@LucaDazi That may be a valid concern. Maybe run it by Matteo and see if this is a problem. Jens made the same change in other bundles, but we can revert the CryptoService if we need to.

ctron:
Could you please write down requirements like that? Also, I don't see any checks in the CryptoService, so which checks do you mean? Any why can't they be started after the SystemService?

PR #546, opened 2016-09-14
Minor fixes for the startup scripts

ctron:
This change does use CDATA sections since one of script does use >> as
text data.

Additionally all scripts now show a message when Kura is not started
due to it is already running.


MMaiero:
@ctron If I start Kura in a raspberry pi with `sudo /etc/init.d/kura start` I get `Starting Kura/opt/eclipse/kura/bin/start_kura_background.sh: 36: /opt/eclipse/kura/bin/start_kura_background.sh: amp: not found
.`

Could you check?
Thanks

ctron:
@MMaiero should be fixed

PR #552, opened 2016-09-20
Support for RPi running Fedora via Fedberry, Pidora, etc.

ccustine:
Signed-off-by: Chris Custine <ccustine@apache.org>


ccustine:
Tested on Fedora 20 via Pidora and Fedora 24 via Fedberry. Tested with eth0 as WAN interface. Limited testing using hostapd, dhcpd, and bind.

Feel free to review and let me know of any requested changes.

ccustine:
Updated and added copyright headers in several files. Added README directions for getting started and installing Fedberry.
@ctron Let me know if there copyright additions are correct or if I need more modifications.

cdealti:
@ccustine it looks good to me but I haven't checked the copyright stuff.

ccustine:
Fixed one issue with the systemd service and pushed changes.

ctron:
Regarding the copyright headers. Please also change existing headers from:

```
    Copyright (c) 2011, 2016 Eurotech and/or its affiliates

    Contributors:
      Eurotech
```

To:

```
    Copyright (c) 2011, 2016 Eurotech and others

    Contributors:
      Eurotech
      Red Hat Inc - This is what I did
```

ctron:
"existing" meaning .. files you touched!

cdealti:
@ccustine If you make the copyright changes this is ready to be merged

kartben:
I just tested this PR on a Samsung ARTIK board and it is way cool, if you ask me.

I have two quick questions/comments:

- Is there anything that's really Raspberry Pi specific in this? Maybe the Device I/O mappings? But I have to admit I am not very familiar with DIO myself so I don't know for sure. My point is that maybe this is in fact a generic "Fedora on ARM" packaging?
- the huge condition block in kura/org.eclipse.kura.linux.net/src/main/java/org/eclipse/kura/linux/net/dns/LinuxNamed.java is having me a bit worried, as this really doesn't look like the right way to easily allow people to add their own Kura portings without touching/recompiling the Kura core code. Could this kind of stuff be moved to e.g a properties file?

dwoodard1:
There are three main reasons listed below for the separate installers for each platform. There may be a better way of doing this, but this methodology was around pre Kura. 

1. Networking. If you are using Kura to control your network interfaces, the available interfaces may be different depending on the platform (number of eth ports, WiFi or no, firewall settings, etc.). Kura is smart enough to "learn" of existing interfaces once it starts, but the installer is useful for getting the interfaces in a known state on install. 
2. Device I/O. You already mentioned this. There is a jdk.dio.properties file that maps the pins available on the target.
3. JVM settings. You can tweak some JVM settings (-Xms and -Xmx) per target. This is probably less important, but may be relevant on more constrained devices.

Regarding the condition block, yes, that is not an ideal approach. If you want to create another issue around that we can take a look at improving how that is done.

cdealti:
@ccustine ping

ccustine:
@cdealti Looks like the code formatting update recently has blown up this PR. I will rebase and fix conflicts shortly.

cdealti:
@ccustine thanks! Make sure to update the copyright headers as suggested above by @ctron

cdealti:
@ccustine sorry to bother you. Can you please fix the copyright headers and rebase so we can merge this before rewriting the history?

cdealti:
@ccustine we would like to merge this tomorrow because we really need to rewrite the repository history. If you cannot make it to rebase it today we will do that on your behalf tomorrow. Please forgive the hurry.

ccustine:
@cdealti This should be ready to go now. Take a look and let me know if I need to fix anything else.

cdealti:
+1 for me. If we need to fix something we will do that after this has been merged.
Thanks!

PR #560, opened 2016-09-22
Fixed bug on ConfigurationService about kura.service.pid

pierantoniomerlino:

### DIFF
--- a/kura/org.eclipse.kura.core.configuration/src/main/java/org/eclipse/kura/core/configuration/ConfigurationServiceImpl.java
+++ b/kura/org.eclipse.kura.core.configuration/src/main/java/org/eclipse/kura/core/configuration/ConfigurationServiceImpl.java
@@ -501,6 +501,7 @@ public class ConfigurationServiceImpl implements ConfigurationService {
             s_logger.warn("Either kura.service.pid {} or service.pid {} are null", pid, servicePid);
             return;
         }
+        
         if (!m_allActivatedPids.contains(pid)) {
             // register the component instance
             s_logger.info("Registration of ConfigurableComponent {} by {}...", pid, this);
@@ -1073,7 +1074,7 @@ public class ConfigurationServiceImpl implements ConfigurationService {
                             // set kura.service.pid if missing
                             Map<String, Object> newProperties = new HashMap<String, Object>(props);
                             if (!newProperties.containsKey(ConfigurationService.KURA_SERVICE_PID)) {
-                                newProperties.put(ConfigurationService.KURA_SERVICE_PID, newProperties.get(Constants.SERVICE_PID));
+                                newProperties.put(ConfigurationService.KURA_SERVICE_PID, config.getPid());
                             }
 
                             cfg.update(CollectionsUtil.mapToDictionary(newProperties));


kura/org.eclipse.kura.core.configuration/src/main/java/org/eclipse/kura/core/configuration/ConfigurationServiceImpl.java
```diff
@@ -497,10 +497,16 @@ synchronized void registerComponentOCD(String metatypePid, Tocd ocd, boolean isF
     }
 
     synchronized void registerComponentConfiguration(String pid, String servicePid, String factoryPid) {
+
```
cdealti: Why is this needed? Can you explain the root cause?

ctron:
I think it would be good if we have a bit more information in PRs. Not only this one, but in general. Each PR should have some information about what has been done and why. If it is a bug, maybe also add a link to the original bug report.

kura/org.eclipse.kura.core.configuration/src/main/java/org/eclipse/kura/core/configuration/ConfigurationServiceImpl.java
pierantoniomerlino: If a component has not a servicePid or a factoryPid, the method returns and the component is not added to m_allActivatedPids. The method that assigns the pid to the servicePid is updateWithDefaultConfiguration, but this is never executed because of the initial check on the pid/servicePid.
We found this bug when a component configuration, without kura.service.pid, was loaded from the snapshot_0.

pierantoniomerlino:
I agree on this. Usually the PRs are related to an issue or bug and it is linked to the PR.
In this specific case, I found the bug working on a different one and I forget to open the issue.

Best,
Pier

On 22 Sep 2016, at 12:20, Jens Reimann <notifications@github.com<mailto:notifications@github.com>> wrote:


I think it would be good if we have a bit more information in PRs. Not only this one, but in general. Each PR should have some information about what has been done and why. If it is a bug, maybe also add a link to the original bug report.

cdealti:
Thanks @pierantoniomerlino the property ``kura.service.pid`` has been introduced in Kura 2.0.0 to support creating component configurations (for "factory" components) whose identity is persisted across ``clean`` restarts of the OSGi framework where the ``service.pid`` would change every time (at least with Equinox 3.8.1).
The ``kura.service.pid`` property is set in the component configuration by the Kura Configuration Service, either by the ``createFactoryConfiguration`` method or when configurations are loaded from the snapshot.

In theory the ``kura.service.pid`` property should only be set in the configurations of a "factory" component but at that point you will have components with and without that property.

Also, the various ``pid`` parameters in the Configuration Service API would represent either the ``kura.service.pid`` or the ``service.pid`` respectively for "factory" and "regular" components.
Messy, isn't it?

Eventually the decision was to set the property for every component.

Here something is not working as expected but I believe the root cause has not been understood

pierantoniomerlino:
@cdealti : the problem, indeed, is in the loadLatestSnapshotInConfigAdmin method. If the properties don't contain a kura.service.pid entry, it is added by the method, but the value is null. The fix will set the property to service.pid value.
This PR will fix #564.

cdealti:
@pierantoniomerlino please update the PR by squashing the two commits into one and adding a meanigful message describing the issue.

PR #568, opened 2016-09-26
Test Refactor and Cleanup

cdealti:
The Kura test framework needs to be redesigned.
In the meantime let's cleanup and improve the existing tests.


cdealti:
@ctron can you please review?
You can run the tests in multiple ways:
1. From Eclipse using the Cura_Core_Test.launch. However this is likely to be a mess as with all the launch configurations
2. From Maven (do not skip the tests)
3. Starting the ``org.eclipse.kura.test`` bundle (also from one of the emulator launch configurations)

The ``org.eclipse.kura.test`` is a ConfigurableComponent. If you run it you should see this in the Web UI:

screenshot from 2016-09-27 12 11 02


We are going to refactor the test framework and this is a WIP and will change again soon. In the meantime can you take a look and see if this is worth to merge?

ctron:
I did wanted to use the test runner. But it seems that the configuration service is not picking it up, since the configuration policy is unset.

I will try the launcher and maven next.

ctron:
The launcher throws some errors. First the Java version was not right, then it was complaining of "no selection", now it simply fails to start.

Running the test from Maven simply fails with:

```
Tests run: 1, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 1.175 sec <<< FAILURE!

Results :

Tests in error: 
  org.eclipse.kura.core.test.AllCoreTests: Invalid client configuration

Tests run: 1, Failures: 0, Errors: 1, Skipped: 0
```

I would suggest to focus on the maven part of this. Since that was at least easy to run.

I do think that getting the integration build up and running should have priority. If that means dropping broken unit tests for the moment, then so be it. But right now that complete build is broken. If we drop what is broken, then at least the remaining unit tests can be used.

Merging it is fine, if you can work through the rest of the issues. Since we still don't have any timeline for a 2.1 release, nothing is being broken by merging this. But it will make it easier for you to work on this in the develop branch since you don't need to care about re-basing and merging when other things change.

cdealti:
Thanks for the review

> I did wanted to use the test runner. But it seems that the configuration service is not picking it up, since the configuration policy is unset.

I'll fix this ASAP

> The launcher throws some errors. First the Java version was not right, then it was complaining of "no selection", now it simply fails to start.

I'll take a look but we always need to fix these launchers

> Running the test from Maven simply fails with

Can you make sure your **Ethernet cable is pugged** This usually happens on Linux where there is no link on the Ethernet port or the primary network interface is neither "eth0" nor "en0". What OS are you running?

ctron:
> Can you make sure your Ethernet cable is pugged This usually happens on Linux where there is no link on the Ethernet port or the primary network interface is neither "eth0" nor "en0". What OS are you running?

I am running RHEL 7. The ethernet is up and running. However the interface name "enp0s31f6". And I can't change that.

cdealti:
Closing this. I will create another PR after applying the formatting

Issue #573, opened 2016-09-27
Commit 41c3180 breaks data entry


ctron:
One of the recent changes seems to have broken the input of integers. It fails now with the following error:

```
Tue Sep 27 14:52:28 GMT+200 2016 ErrorLogger INFO: : (TypeError) : a is nullcom.google.gwt.core.client.JavaScriptException: (TypeError) : a is null
   at Unknown.RLb(http://puddlejumper/kura)
   at Unknown.OLb(http://puddlejumper/kura)
   at Unknown.QLb(http://puddlejumper/kura)
   at Unknown.sMb(http://puddlejumper/kura)
   at Unknown.jm(http://puddlejumper/kura)
   at Unknown.Pn(http://puddlejumper/kura)
   at Unknown.Hn(http://puddlejumper/kura)
   at Unknown.gU(http://puddlejumper/kura)
   at Unknown.Wl(http://puddlejumper/kura)
   at Unknown.iU(http://puddlejumper/kura)
   at Unknown.sU(http://puddlejumper/kura)
   at Unknown.lZ(http://puddlejumper/kura)
   at Unknown.K$(http://puddlejumper/kura)
   at Unknown.je(http://puddlejumper/kura)
   at Unknown.me(http://puddlejumper/kura)
   at Unknown.le/<(http://puddlejumper/kura)
   at Unknown.anonymous(Unknown)
```

MMaiero:
@ctron Is this happening for all the services? I've tried with a build done this morning and I have no issues

ctron:
Not with all. I just dropped my DP int a fresh build of Kura and now run into this issue. But I guess I have to dig deeper.

ctron:
I do have the same error now when running Kura from the IDE. So this has to come from one of the recent commits.

MMaiero:
I'll try to do some tests. Could you share the dp you are using, in order to test in the same conditions?

ctron:
I am not sure it is the integer field. I do have three fields: boolean, integer and string.

ctron:
https://tresor.it/s#EdRf67lEQoxJkW3Ol7uJVg

ctron:
Setting the generation to "DETAIL" I do get:

```
[...]

PR #574, opened 2016-09-28
#573: Fix client side NPE on field entries

ctron:
This change does revert the regression introduced by commits 41c3180c9054ac1871687cff00bf36df8c22f5a1 and 06ce1f387ee17cf6c0c1c5b890b1c860d389c656 which does shift the information of the "id" to the "title" attribute rather then the field's label.

The original commit 06ce1f387ee17cf6c0c1c5b890b1c860d389c656 did not fill the title though for fields other than the text field and commit 41c3180c9054ac1871687cff00bf36df8c22f5a1 only reverted this partly, removing the null check, letting the client run into a NPE.


ctron:
@MMaiero Please review

Issue #567, opened 2016-09-24
build-all script needs to incorporate the user selection of specific profile


amitjoy:
Currently build-all script builds Kura distribution for all platforms. I believe, the script must include a choice based on whether user needs to build the distribution at all or the user needs the distribution for all platforms or for any specific platform.

ctron:
You already can push any arguments to the build-all script. Those arguments will be passed on to all mvn calls.

So calling buil-all -Pfoo will activate profile "foo" on all builds.

amitjoy:
That also constrains users to know all the profiles beforehand. This could be easily solved if we provide a script which actually prompts users with proper information about whether the user wants to build the distribution or not and if yes, user would be able to choose a profile from a list of available ones.

I have been confronting this problem since months and finally I had modified the build-all script which prompts me before it starts building Kura.

ctron:
Which is bad for automated processes ;-)

Of course you could add a second script for that.

I would rather like to see more documentation around the build, than adding just another layer of complexity which has to be maintained. Just think what happens when you add another target platform. This adds another location which you have to think of.

amitjoy:
For automated processes, you are right but for user-centric needs, I believe, we should provide an easy-to-use utility **build-selective.sh** script.

ctron:
Yes, that would work. And I agree that there should be a way to speed up things.

But IMHO it should be done in a way which makes the build less complicated.

amitjoy:
Yeah you are right. The script must not complicate the build process. That is the reason, it would be really helpful for developers to use such **build-selective.sh** script to build the selective distribution build.

The problems which I have been confronting since months is that I am changing something on WebUI bundle and I want a distribution build for RPi. And if I mistakenly forget to include the specific profile while running the script or through the usual maven commands, I have to wait for the all builds to complete or I have to explicitly cancel the process as soon the build process gets complete for my specific build profile.

This gets more complex for new developers as they would never know that the different profiles do exist in Kura. They would run usual build-all script and it will run all the profiles. At least I confronted this when I started to work initially.

That's why I want it to be simplified so that anyone will have the opportunity to choose whether he/she wants to build for all the distribution profiles or selective one or no distribution at all.

ctron:
I agree, the build process has to be more simple. But not only from an end user perspective. Also from a maintainer perspective. Right now, everything is just built. A process taking more time than necessary, but adding a new target is covered by that.

[...]

PR #576, opened 2016-09-28
Added functionality to install Dialog based on running platform

amitjoy:
Signed-off-by: Amit Kumar Mondal <admin@amitinside.com>

### DIFF
--- a/build-menu.sh
+++ b/build-menu.sh
@@ -9,6 +9,7 @@
 # 
 # Contributors:
 #     Red Hat Inc - initial API and implementation
+#     Amit Kumar Mondal (admin@amitinside.com)
 #
 
 set -e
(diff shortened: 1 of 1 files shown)


amitjoy:
#567

build-menu.sh
```diff
@@ -18,15 +19,26 @@ set -e
...
+  # check if Mac OSX
+  if [[ "$OSTYPE" == "darwin"* ]]; then
+        echo "Installing Dialog for Mac OSX..."
+        brew install dialog
+  #check if Linux
+  elif [[ "$OSTYPE" == "linux-gnu" ]]; then
+      #check if redhat
+      if [ -f /etc/redhat-release ]; then
+        echo "Installing Dialog for Redhat Platform..."
+        sudo yum install dialog
```
kartben: how about people who use macports?

amitjoy: I believe it's the very first version of the script. It would be extended with more commands as soon as we know how to execute different commands in different platforms using other installers. The current script is just a head-start.

ctron:
I really, really wouldn't execute that immediately. I personally hate it when scripts do that.

Printing out a command which needs to be executed is helpful though.

So I would suggest to re-work that patch to print out instead of execute those commands. This will also prevent errors in case the script did not get it right.

build-menu.sh
ctron: @kartben Contributions are welcome :wink

amitjoy:
@ctron You are right! It would make much sense to show the commands instead. I have corrected it. So, based on the current running platform, the script would display the respective command.

build-menu.sh
amitjoy: @ctron As I mentioned earlier, it's just a first version. I will research more to incorporate several other platforms. Till then, we can provide this as a base script to use for building selective distributions.

amitjoy: @ctron Added RedHat Family distros and SuSE check to display respective command to install dialog

amitjoy:
@ctron Can you have a look at it?

ctron:
I tried to work with you on this PR, but it seems that since you created this PR based on "develop" I cannot amend the PR. Not sure what is the case for that. So I simply took over your changes, refined it a bit and added you as "Also-by".

I did refine the detection a bit so that we now have a final "else" branch which simply shows that there is not help available. I also did not use the "lsb-release" check since that file is no longer present on Ubuntu of Debian. At least when I tested it.

I did test the current code with Docker images for rhel:latest, centos:latest, ubuntu:latest, debian:latest opensuse:latest and gentoo/stage3-amd64. Gentoo triggered the "else" branch saying that no help is available.

It would be cool if someone could verify the Mac OS X branch, since setting up a Mac OS instance for testing costs me quite a while.

PR #585, opened 2016-10-01
CRLF fix

MMaiero:
This PR fixes the CRLF problem by superseding commit 03a06aac4a8df079de893371ba46074b1adbf630, adding a build step in manifest_pom and pom_pom verifying for wrong files (uses the maven-checkstyle-plugin in version 2.14 that was already approved by Eclipse with this CQ https://dev.eclipse.org/ipzilla/show_bug.cgi?id=9584) and, finally, adding an initial version of the .gitattributes file.

This branch is taken from the other branch named enh_code-refactor


ctron:
You don't need a full CQ for that. Just did it to the build&test CQ, which still needs to be created.

MMaiero:
@pierantoniomerlino Could you check if it works for you?

ctron:
I would suggest spitting this off into two commits. One can be done in a branch, adding the new checkstyle stuff. The other one should be done directly on the "develop" branch. Since you simply can't tell if you are overwriting any change by merging this in!

MMaiero:
I've rebased the branch to be even with develop. I've also checked the modified files to be sure that they are at the same status as they were when formatted or, if changed, the changes are still there. So, it should be ok now. I've also performed a full build and the check style step does not fail

ctron:
+1 for merging right now!

PR #409, opened 2016-08-03
#408 Let service tracker only track relevant services

ctron:
This change re-implements the service tracking of the default 
configuration service implementation to use standard service trackers
and only track relevant services from the beginning.

In addition all service registrations have been fixed to actually export
the ConfigurableService and SelfConfiguringService.

Beside improving the performance this also prevents services to be
activated which would not have been activated otherwise e.g. in Karaf.

Signed-off-by: Jens Reimann <jreimann@redhat.com>

PR #591, opened 2016-10-04
Fix recursive loop exception when starting with Apache Felix

ctron:
This change will detect when the ConfigurationServiceImpl is being
checked by the tracker and ignore that service.

This can cause a recursive exception on Apache Felix since the service
manager will start the ConfigurationServiceImpl, which will then
start the service tracker, which will then try to get the
ConfigurationServiceImpl.


ctron:
Will be fixed with #409

Issue #372, opened 2016-07-28
[BLE] BluetoothGattImpl.getCharacteristics(String startHandle, String endHandle) and getServices() methods return empty list


Erwan-S:
It seems that ANSI Escape sequences send by gatttool (according to https://git.kernel.org/cgit/bluetooth/bluez.git/tree/attrib/interactive.c) remain in strings treated by BluetoothGattImpl.

Response data filter corresponding to `getCharacteristics` is `startsWith("handle:")`  but received string begin by `[Khandle:` (with [K from the erase line escape sequence). So the method returns an empty list.

Likewise for `getServices`.

Kura version : 2.0.0
Bluez version : 5.23
Plateform : Raspberry Pi 3 with Linux raspberrypi 4.4.11-v7+

PR #593, opened 2016-10-05
remove erase and graphics rendition ANSI sequences

Erwan-S:
cf. https://github.com/eclipse/kura/issues/372


dwoodard1:
Hello, thanks for the contribution. Please note, before we can review/consider the PR, you must sign the Eclipse CLA and also sign your commit. I would recommend you read the [contributing guidelines](https://github.com/eclipse/kura/blob/develop/CONTRIBUTING.md) .

Erwan-S:
It's done.

ctron:
I don't fully understand the change TBH. But wouldn't it be simpler to set the TERM variable to something which doesn't use ANSI?

Erwan-S:
@ctron : I, first, tried your solution and it didn't work. Furthermore, why adding a system configuration dependency ?

ctron:
What kind of dependency do you mean? You can set environment variables when starting an application.

Erwan-S:
Do you think that is better to change kura launch script than remove ansi character in gatttool response handler ? 

Anyhow, as I previously said, xterm solution doesn't work. When it creates response, gatttool doesn't consider xterm value (You can see that in interactive.c from gatttool source code).

cdealti:
This PR has been screwed up by the Git history rewrite and needs to be resubmitted.

Issue #596, opened 2016-10-05
Replace all calls to ConfigurationAdmin.getConfiguration(String)


ctron:
All calls to `org.osgi.service.cm.ConfigurationAdmin.getConfiguration(String)` should be replaced to calls of `org.osgi.service.cm.ConfigurationAdmin.getConfiguration(String, String)` providing `null` as the location parameter.

Most calls in Kura already are already migrated to that call, but not all. Using Apache Felix ConfigAdmin actually enforces the default binding which is generated otherwise and will not configure some components. Also see OSGi spec for `getConfiguration(String)`:

 > Bind its location to the calling bundle's location.

That effectively binds the configuration to the Kura configuration bundle which is wrong. It should instead be bound to the bundle containing the service. Leaving the argument `null` would do that automatically once the service is registered.

PR #597, opened 2016-10-05
Fix issue #596 by providing `null` as location

ctron:
This change does fix the issue of wrongly bound configurations by
calling getConfiguration(String,String) with `null` as location
argument.

### DIFF
--- a/kura/org.eclipse.kura.test/src/org/eclipse/kura/test/RemoteTargetTest.java
+++ b/kura/org.eclipse.kura.test/src/org/eclipse/kura/test/RemoteTargetTest.java
@@ -198,7 +198,7 @@ public class RemoteTargetTest {
         // hijack the settings
         try {
             Configuration mqttConfig = this.m_configAdmin
-                    .getConfiguration("org.eclipse.kura.core.data.transport.mqtt.MqttDataTransport", null);
+                    .getConfiguration("org.eclipse.kura.core.data.transport.mqtt.MqttDataTransport",  "?");
             Dictionary<String, Object> mqttProps = mqttConfig.getProperties();
             mqttProps.put("broker-url", "mqtt://broker-sandbox.everyware-cloud.com:1883/");
             mqttProps.put("topic.context.account-name", "EDC-KURA-CI");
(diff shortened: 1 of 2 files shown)


amitjoy:
According to the OSGi specification, the location needs to be set to `?` as it bypasses the location check with all the security measures.

ctron:
I don't think you need to bypass anything. Just bind the configuration to the correct location.

amitjoy:
It's mentioned in OSGi specification to have it set to `?`. That is advised if the system runs with security.

ctron:
Can you point to me that advisory?

amitjoy:
You can check OSGi Compendium Specification (104.4.1 - Location Binding)

amitjoy:
You can check at the OSGi enRoute's website as well.

http://enroute.osgi.org/services/org.osgi.service.cm.html

ctron:
I just read through the specs. That feature is present at least since OSGi 4.3. So I agree, it is the better way to go.

I will amend the PR tomorrow.

ctron:
PR is amended. All locations will use `?` now. Thanks for explaining that!

amitjoy:
👍

kura/org.eclipse.kura.test/src/org/eclipse/kura/test/RemoteTargetTest.java
```diff
@@ -198,15 +198,15 @@ private void startingTests() {
...
+                    .getConfiguration("org.eclipse.kura.core.data.transport.mqtt.MqttDataTransport",  "?");
             Dictionary<String, Object> mqttProps = mqttConfig.getProperties();
             mqttProps.put("broker-url", "mqtt://broker-sandbox.everyware-cloud.com:1883/");
             mqttProps.put("topic.context.account-name", "EDC-KURA-CI");
             mqttProps.put("username", "EDC-KURA-CI");
             mqttProps.put("password", "PYtv3?s@");
             mqttConfig.update(mqttProps);
 
-            Configuration dataConfig = this.m_configAdmin.getConfiguration("org.eclipse.kura.data.DataService");
+            Configuration dataConfig = this.m_configAdmin.getConfiguration("org.eclipse.kura.data.DataService", null);
```
MMaiero: Why here the getConfiguration parameter is null?

ctron: No reason at all. Fixed.

dwoodard1: FYI, Cristiano has cleaned up the test code in #568 . Probably okay to merge, but this will cause merge conflicts with that PR.

ctron: I guess there are conflicts anyway due to the code formatting

cdealti:
As far as I'm concerned this can be merged

Issue #595, opened 2016-10-05
ComponentMetaTypeBundleTracker is processing every bundle twice on startup


ctron:
The implementation of `ComponentMetaTypeBundleTracker` is overriding the `open` method an manually processing all bundles a second time during the `open` call. In the log file you can see that all bundles get processed twice. Once by the "adding" method (since the call to `super.open` comes first) and second by the open call itself.

```java
    @Override
    public void open() {
        s_logger.info("Opening ComponentMetaTypeBundleTracker...");
        super.open();
        s_logger.debug("open(): getting bundles...");
        Bundle[] bundles = this.m_context.getBundles();
        if (bundles != null) {
            for (Bundle bundle : bundles) {
                if (bundle.getState() == Bundle.ACTIVE) {
                    s_logger.debug("open(): processing MetaType for bundle: {}...", bundle.getSymbolicName());
                    processBundleMetaType(bundle);
                    s_logger.debug("open(): processed MetaType for bundle: {}", bundle.getSymbolicName());
                } else {
                    s_logger.debug("open(): bundle: {} is in state: {}. MetaType will be processed by addingBundle()",
                            bundle.getSymbolicName(), bundle.getState());
                }
            }
        }
        s_logger.debug("open(): done");
    }
```

PR #598, opened 2016-10-05
#595: Don't override ComponentMetaTypeBundleTracker#open()

ctron:
This change does remove the override of method open() since already
existing bundles will be automatically announced by the call to open().


dwoodard1:
@ctron I don't believe we can remove the override. There is a private method processBundleMetaType(Bundle b) that is called in the overridden method. This method updates bundles based on the Kura metatype definitions.

LucaDazi:
My understanding is that the BundleTracker will trak components in the addingBundle method even if they have been added to the runtime before opening the Service Tracker, so in fact there should be no need of overriding the open method.

dwoodard1:
@LucaDazi You are correct, forgot that we were overriding the addingBundle method which also calls the needed private method. Sorry about that, I think this PR is okay.

hekonsek:
Merging this one then. Thanks, Jens!

Issue #599, opened 2016-10-05
WebUI fails completely is `PositionService` is not registered


ctron:
If there is no instance of `PositionService` registered with OSGi, then the whole Web UI fails with an "internal error".

Although the code checks for a `null` reference:

```java
PositionService positionService = ServiceLocator.getInstance().getService(PositionService.class);

if (positionService != null) {
```

The actual call to `org.eclipse.kura.web.server.util.ServiceLocator.getService(ServiceReference<T>)` will throw an exception.

cdealti:
@ctron what is the state of this issue? If you are using Zenhub please update the fields on the left column (Pipeline, Milestone, Assignee etc) so that we can properly track the progress.

ctron:
This one was merged already.

I don't use Zenhub.

PR #600, opened 2016-10-05
#599: Fix "internal error" exception when PositionService is missing

ctron:
This change will handle the case when the PositionService is missing.
It will also not leak service references.


amitjoy:
👍

cdealti:
@MMaiero can you please take a look?

hekonsek:
+1 This optional method is a great way to handle service absence. Merging.

Issue #603, opened 2016-10-05
Can "json 20090211" be removed?


ctron:
A few bundles declare a dependency on "json 20090211":

```xml
<groupId>org.json</groupId>
<artifactId>json</artifactId>
<version>20090211</version>
```

But to me it seems it is not used. Can it be removed?

cdealti:
If it still builds and runs after removing it, yes!

ctron:
Ok, then I will work on that.

ctron:
Looks like it.

ctron:
This affects the deployment on Karaf.

Issue #226, opened 2016-05-19
[API] We should review how we use OSGi Semantic Versioning


cdealti:
For the Kura API we use OSGi Semantic Versioning of packages.
Every exported package is qualified with a version that we upgrade if something changes.

Since we aim to backward compatibility we only make changes to the API that require a change in the minor number of the exported package.

For example in #194 we are adding methods to `org.eclipse.kura.configuration.ConfigurationService`.
Due to this we also have to change the minor version to the  `org.eclipse.kura.configuration` package exported in the MANIFEST of the `org.eclipse.kura.api` bundle, from version 1.0.0 to 1.1.0.

This will not cause problems to **consumers** of the `org.eclipse.kura.configuration` API if they are properly importing the `org.eclipse.kura.configuration` package:

```
Import-Package: org.eclipse.kura.configuration;version="[1.0,2.0)",
```

However, consumers of `org.eclipse.kura.configuration` are rare.

Virtually all Kura apps are **implementors** of the  `org.eclipse.kura.configuration` API for example because they implement `org.eclipse.kura.configuration.ConfigurableComponent` and if they properly use semantic versioning:

```
Import-Package: org.eclipse.kura.configuration;version="[1.0,1.1)",
```

They will have their imports not satisfied which is good for Semantic Versioning but it's not good for customers wanting to upgrade Kura without upgrading the apps (and in this example there would be no need to upgrade the apps).

There is really nothing we can do in this case, only hope that customer apps are not following Semantic Versioning in this case.

The situation is the same for the various listeners we have in the API.

For the future:
We should have a finer granularity for packages to have better control on versioning: `ConfigurableComponent` should belong to an `org.eclipse.kura.configuration.component`

LucaDazi:
Agree on putting marker interfaces, abstract classes and listeners on dedicated packages.

We could start duplicating classes where needed and deprecating the old ones.
We'll have to take that step, sooner or later.

cdealti:
Something has been done for the Data Service and Data Transport Service listeners.
New ``org.eclipse.kura.data.listener`` and `` org.eclipse.kura.data.transport.listener`` have been created to properly support the multiple connections feature.
The old [listeners](https://github.com/eclipse/kura/blob/develop/kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/data/DataServiceListener.java) in ``org.eclipse.kura.data`` have been deprecated.

cdealti:
To clearly state the requirements:

[...]

PR #604, opened 2016-10-06
#603: remove dependency on JSON 20090211

ctron:


amitjoy:
I believe the `Bundle-Vendor` manifest header must be `Eclipse Foundation`, instead of `Eclipse Kura`

ctron:
I can be either "Eclipse.org" or the full project name, You can have a look at the Eclipse software installation dialog on the tab "Plugins".

ctron:
At least the Bundle Manifest has been updated, that should be reflected in a version uptick.

amitjoy:
@ctron The `Provider` in the plugins sections actually shows it in different way and I agree with it. It is actually taken from the `feature.xml`. It would be nice to provide the real name of the vendor of the product. My personal opinion would be to go with `Eclipse Foundation` instead of `Eclipse.org` or `Eclipse Kura` as this doesn't logically belong to the `Bundle-Vendor`

amitjoy:
Apart from it, I agree with @MMaiero as the Bundle version is also a part of semantic versioning scheme and if there is no functional change in the code, I believe we should not uptick the versions.

ctron:
@amitjoy Why not follow the way other Eclipse projects do it? Why create a special way?

The problem is with the current build system then in this case changes will not get pulled in into the next build.

MMaiero:
@ctron Could you explain me what you mean with "The problem is with the current build system then in this case changes will not get pulled in into the next build."?

amitjoy:
@ctron That was just my opinion. If everyone agrees with `Eclipse.org`, I am also okay with it.

ctron:
I just looked into the Eclipse guide for versions [1]

> For example, a bug has been fixed in the code, the plug-in manifest has changed, documentation has changed, compiler settings have changed.

I think we touched two of these points for this change.

[1] https://wiki.eclipse.org/Version_Numbering#When_to_change_the_service_segment

ctron:
@amitjoy I don't say there is anything wrong with your opinion. But I also think that this an Eclipse project and it should follow the path other Eclipse projects follow for years. Putting in `Eclipse Foundation` does not add any value to the project IMHO, so I don't see the necessity for a special way.

amitjoy:
@ctron Then we would go for it. I would suggest you to make a single commit with all the `Bundle-Vendor` changes in all the existing bundles. It would be nice to do it this way instead of doing it in different commits.

ctron:
I would let GitHub to the job and issue "Squash and merge" for this PR.

MMaiero:
For what I know, till now we used [this](https://www.osgi.org/wp-content/uploads/SemanticVersioning.pdf) as reference for semantic versioning as reported also [here](https://github.com/eclipse/kura/issues/226). 
If we want to change to [this](https://wiki.eclipse.org/Version_Numbering#When_to_change_the_service_segment) we need to be fully adherent and not using only parts.

However, I don't see a big problem if we continue to follow the approach used till now: the changes to bundle manifests will be in the code base and will go into production when the bundle code will be updated. It will take few Kura versions, but I don't think that people will complain if the bundle-vendor is not updated...

ctron:
I don't really see how this conflicts with the upticking of the micro-version:

> A difference in the micro part does not signal any backward compatibility issues. The micro 
number is used to fix bugs that do not affect either consumers or providers of the API.

So something in the build got fixed (dependency + manifest). The project's source code advanced and no longer matches the build tag.

I will contact EMO and ask their interpretation on [1]. If there is a need to change then bundle vendor, then we should do that. If not we can wait another version or two.

[1] https://www.eclipse.org/legal/guidetolegaldoc2.php

cdealti:
@ctron the point here is that you don't need to change the versions ``kura/distrib/config/kura.build.properties`` **right now** only due to this change. The next Kura release will not have new bundle versions unless in the meantime there is a real bugfix.

ctron:
@cdealti I understand that. That is why I did ask EMO of the Bundle-Vendor should be updated or not. If it should be, then that would be a bug fix. Otherwise I can downtick the version in the properties file.

ctron:
Beside that, the POM files of this version have to be released with the Nexus repository. Since they do contain the dependency and trigger a wrapper bundle on Karaf:

```
Feature contains followed bundles:
  mvn:org.eclipse.kura/org.eclipse.kura.linux.bluetooth/1.0.6-SNAPSHOT start-level=80
  mvn:org.eclipse.kura/org.eclipse.kura.linux.clock/1.0.7-SNAPSHOT start-level=80
  wrap:mvn:org.json/json/20090211 start-level=80
```

ctron:
So I did get back a pointer from EMO [1]. It clearly states:
> Bundle manifests contains a "Bundle-Vendor" entry set to "Eclipse &lt;project name&gt;"

So we are required to set the field to "Eclipse Kura". Not "Eclipse Foundation", "Eclipse.org" or anything else.

Since there where no objections to other parts of this PR, I would to go on and merge this.

[1] https://wiki.eclipse.org/Development_Resources/HOWTO/Release_Reviews#Checklist

PR #605, opened 2016-10-06
Fix a few resource leaks

ctron:
This change does properly close input streams after loading of 
properties.

As the Java documentation for `load` states:

 > The specified stream remains open after this method returns.

It also wraps some close calls into a finally section in order to ensure
that the close is called in case of an error.


amitjoy:
👍

kura/org.eclipse.kura.core/META-INF/MANIFEST.MF
```diff
@@ -3,7 +3,7 @@ Bundle-ManifestVersion: 2
 Bundle-Name: org.eclipse.kura.core
 Bundle-SymbolicName: org.eclipse.kura.core;singleton:=true
 Bundle-Version: 1.0.10.qualifier
-Bundle-Vendor: EUROTECH
+Bundle-Vendor: Eclipse Kura
```
amitjoy: I believe the `Bundle-Vendor` must be `Eclipse Foundation`

PR #607, opened 2016-10-06
Start the Web UI component immediately

ctron:
The Web UI component should be started immediately on startup. If the
immediate attribute is not set, the component will become a "delayed"
component and only be activated once there is a reference to it (also
see 112.5.4 of the OSGi compendium specification). However no one will
actively reference this component.


amitjoy:
👍

cdealti:
@MMaiero can you please take a look?

hekonsek:
+1 Merging.

Issue #592, opened 2016-10-04
Again Java 7 code in BREE 6 bundle


ctron:
Please fix or switch to Java 7:

KuraJDBCPooledDatasource:

```
    @Override
    public Logger getParentLogger() throws SQLFeatureNotSupportedException {
        throw new SQLFeatureNotSupportedException("Not supported!");
    }
```

`getParentLogger` in not defined in Java 6

PR #610, opened 2016-10-10
Enable BREE validation on Travis CI build

ctron:
This change adds a BREE validation profile to the existing tycho build.
A profile can be used to enable this validation and by default this
profile is not active.

This profile is turned on during the Travis CI build, so that code can
be validated against the actual configured BREE.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,8 +3,12 @@ language: java
 sudo: false
 
 jdk:
-  - oraclejdk7
+  - oraclejdk8
 
-script: ./build-all.sh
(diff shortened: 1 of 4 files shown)


ctron:
The Travis CI validation currently is expected to fail since the current state of Kura does not compile properly. See issue #592.

cdealti:
+1 This must be merged asap. And enabled on Hudson as well

ctron:
I will keep issue #592 alive until I validated it on HIPP.

PR #621, opened 2016-10-11
Another set of changes to remove CRLF line endings.

MMaiero:
.catalog files are left untouched

@dwoodard1 Could you check it builds in your environment?


ctron:
Why are the .catalog files untouched. They will also generate changes directly after checkout.

MMaiero:
Catalog files are generated by the build. Them must not be touched by committers apart for exceptional cases. In the past we had some bad issues with those files, resulting in broken updates. For this reason I want to investigate carefully the effects of changing those files: it could affect the scripts used for upgraders.

ctron:
Then at least the gitattributes should be changed for those files. Checking out a git repo should not mark those files as changed. Especially if the are so fragile.

MMaiero:
I've updated the .gitattributes file and now cloning the branch does not show modified files. So it should be fine. Can someone check and eventually merge? Thanks

MMaiero:
Is there anyone that can give a try to this branch so we can merge to develop and close this issue?
Thanks

ctron:
After checkout I now get:

```
# On branch fix_crlf-part2
nothing to commit, working directory clean
```

PR #623, opened 2016-10-12
Fix NPE when JDK DIO properties file is not configured

ctron:
I cases where the configuration is missing the code currently runs into
a NPE. This change does fix this by throwing an IllegalStateException
with a proper message instead.


hekonsek:
Merged. Many thanks Jens!

PR #625, opened 2016-10-12
Kura should not publish modified artifacts under "org.eclipse" namespace

ctron:
This change does change the group id of the modified soda comm bundle to
"org.eclipse.kura".


hekonsek:
+1 Many thanks, Jens.

Issue #539, opened 2016-09-13
[Cloud] Required changes in CloudServiceFactory


MMaiero:
While working on #515 I've noticed the impossibility to relate a CloudServiceFactory instance with its cloud stacks.
In my opinion we should change that API moving its package to version 1.1.0, including a new variable named `kura.cloud.service.factory.pid` and a new method `getComponentConfigurations(String pid)`.

The `kura.cloud.service.factory.pid` will be used by the compliant cloud services instances to relate the specific instance with the constructor factory that generated it. In this way it will be possible, by searching in the cloud component properties, to map the instance with the factory.
The new UI developed to solve #515 will be able to filter the services that provide the `kura.cloud.service.factory.pid` giving them the possibility to be grouped and managed in the new "Cloud Services" section. To keep backward compatibility, the old cloud services implementations will continue to be available in the "Services" section of the web ui.

The `getComponentConfigurations(String pid)` will be useful to get the component configurations for all the components generated by the factory: the method will return a `List<ComponentConfiguration>` that, for the specified cloud service instance, will fetch the configuration of all the components of the identified cloud stack.

ctron:
Why not simply make the requirement that every service belonging to that group has to register an additional OSGi property, like `kura.cloud.service.pid`. All services sharing the same value are grouped together.

That way you would avoid adding another method.

MMaiero:
I've created a new branch named enh_cloudServices-separate-ui where I've implemented a first version of the cloud services Ui for web2.

It required to move the api and core packages to version 1.0.10 (snapshot) and upgrading org.eclipse.kura.cloud.factory to version 1.1.0.

At the end, the CloudServiceFactory api required the definition of a new property, KURA_CLOUD_SERVICE_FACTORY_PID and the definition of a new method getStackComponentsPids.

The getStackComponentsPids takes the pid of the cloud service and gets, from the factory, the list of pids that compose the cloud stack. The choice to add this method is related to the fact that only the factory knows how to create the stack and the final naming. This also guarantees to get the same order for every call, in a defined cloud stack (we cannot think to always have the schema cloudService-DataService-DataTransport).

Furthermore I've added few properties in some component definition files:

- **kura.ui.service.hide**: specified in a service pertaining to a cloud stack, allows the web ui to hide this service when rendering the services section;

- **kura.ui.csf.pid.default**: optional cloud factory component parameter that allows to specify a default suggested format for the pid that the user has to put in the web ui;

[...]

Issue #515, opened 2016-09-06
[Web UI] Implement Cloud Services UI


cdealti:
The Web UI lacks the ability to create, configure and manage Cloud Services.
See #444 and #189 for a discussion on the requirements.

![image](https://cloud.githubusercontent.com/assets/1225835/18272119/667012b6-7436-11e6-9cbc-b06159f50c99.png)

cdealti:
@ctron Matteo will take care of this. Please coordinate with him if you plan to work on the affected parts of the Web UI.

pgufAtIBM:
its true, if connection to cloud (in multiple cloud support) is there, is not visible in web ui. you can see it in log file kura.log. 
currently you also cannot select which cloud service is currently active in system -> status of web ui.
I have used DataService -> autoconnect = true. Then your cloud instance connects to desired cloud (if connection is successful you can see only in your cloud platform (f. ex. Watson IoT) or in Kura log

cdealti:
@pgufAtIBM thanks for the feedback. Do you agree on the plan to modify the Web UI as shown in the above mockup? As you can see in the mockup your WatsonIoT Cloud Service Factory is listed.

pgufAtIBM:
I like this approach, so a user can select the cloud component, and not x services on the left side are listed.
What would be good: able to provide logging parts to web ui. So the user can see that diagnostic logs for example are published to cloud. or why a device cannot be instanciated on cloud

MMaiero:
Closing due to #626 merging

Issue #189, opened 2016-05-05
[Camel] Camel integration in the Kura CloudService


cdealti:
@hekonsek this feature leverages the Kura multiple connections feature recently added on the develop branch.

You need to implement a new [CloudServiceFactory](https://github.com/eclipse/kura/blob/develop/kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/cloud/factory/CloudServiceFactory.java) and a component registering a `CloudService` in the OSGi framework.

The Camel implementations will be `CamelCloudServiceFactory` and `CamelCloudService`.

The Kura [default](https://github.com/eclipse/kura/blob/develop/kura/org.eclipse.kura.core.cloud/src/main/java/org/eclipse/kura/core/cloud/factory/DefaultCloudServiceFactory.java) implementation of the `CloudServiceFactory` can be used as a reference implementation.

Please note that the CloudService implementation must also be a Kura `ConfigurableComponent` so, in the case of the `CamelCloudService`, its configuration could be the Camel XML routes for that CloudService instance (the same configuration of the Camel quickstart example).

ctron:
@cdealti Maybe you can help a little bit here. I am trying to look for code referencing the `CloudServiceFactory` interface. And to me it seems as if it is not really used in Kura. There is the "default" implementation. But not more. No one seems to call or use the interface. Unless I have missed a spot.

cdealti:
@ctron it is mostly intended for the Web UI so that it can get registered ``CloudServiceFactory`` OSGi services and present a consistent UI to create connection instances and show their configurations.

Can you create an issue for this? See also #444

ctron:
I have to admit I fail to see the basic idea behind the way this is done. On the one hand there is a common interface, on the other hand it is not used but every component has to create it own way of adding services.

Maybe you have some description, beside the Javadoc, what the intended use cases of this construct is. As far as I understood, once the method is called in order to create that service, it still has to be configured. So instead of making a custom console command, as described in issue #444, shouldn't there be a common console command providing this functionality? Or did I misunderstand that?

cdealti:
> I have to admit I fail to see the basic idea behind the way this is done. On the one hand there is a common interface, on the other hand it is not used but every component has to create it own way of adding services.

What do you exactly mean?

> Maybe you have some description, beside the Javadoc, what the intended use cases of this construct is.

[...]

PR #626, opened 2016-10-12
Changes to support multiple cloud services from local WebUi

MMaiero:
This PR will add support, in the local WebUi, to multiple cloud services. Currently the implementation is fully working for the default stack. For Camel I've only updated versions and added some placeholders that eventually will need a full implementation to be used with the new setup.
Eventually, I can do a rollback of those changes if Camel cloud support will be not managed by this new feature.


cdealti:
@ctron Can you kindly review? See also #515 for the requirements and #539 for the proposed implementation.
If you agree this PR would be a good chance to finalize #189.

cdealti:
@ctron it's important to merge this before rewriting the history. Can you please review?

ctron:
I already started and did not get any response from @MMaiero 

Or maybe I did something wrong the new workflow for reviewing.

ctron:
But I don't completely understand the rush. Why can't we re-base afterwards?

cdealti:
I'd like to merge what we can before rewriting the history.

ctron:
Ok then I would suggest to fix the comments I made and proceed with merging. Anything which goes wrong can be fixed at a later time.

MMaiero:
I had no notifications till today when comments were made.
However I've updated the pull request with the suggested changes and rebased on develop.

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/cloud/factory/CloudServiceFactory.java
```diff
@@ -115,6 +124,8 @@
      */
     void createConfiguration(String pid) throws KuraException;
```
ctron: Add documentation

ctron: And request from callers to handle a "null" result

ctron:
Ok, I see. Sorry fur that. Seems like you need to "submit" the comments on the new GitHub review process in order to make them visible.

PR #630, opened 2016-10-13
Minor cleanups to the Linux ClockServiceImpl

ctron:
This change:

 * Uses a static instance of the ClockEvent in order to not generate
   empty instances
 * Shows the return code when the hwlock command fails
 * Cleans up a few minor things

### DIFF
--- a/kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/clock/ClockEvent.java
+++ b/kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/clock/ClockEvent.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2011, 2016 Eurotech and/or its affiliates
+ * Copyright (c) 2011, 2016 Eurotech and others
  *
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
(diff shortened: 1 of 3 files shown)


kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/clock/ClockEvent.java
```diff
@@ -23,6 +25,8 @@
     /** Topic of the ClockEvent */
     public static final String CLOCK_EVENT_TOPIC = "org/eclipse/kura/clock";
 
+    public static final ClockEvent EMPTY = new ClockEvent(Collections.<String, Object> emptyMap());
```
MMaiero: @ctron Analyzing the api changes I went through this.
Why we need this EMPTY element in the APIs? Being public, this implies a change on the clock package version in manifest file. 
Furthermore, I don't see a big advantage in having it here. Maybe it's better to preserve the api and move this definition in the impl class. What do you think?

MMaiero:
@ctron probably you have not received the message. I'll report it following:

> @ctron Analyzing the api changes I went through this.
Why we need this EMPTY element in the APIs? Being public, this implies a change on the clock package version in manifest file. 
Furthermore, I don't see a big advantage in having it here. Maybe it's better to preserve the api and move this definition in the impl class. What do you think?

ctron:
Didn't see that, sorry.

Yes, we could do that!

ctron:
Can you do that and just push it?

Issue #640, opened 2016-10-17
[Modem] NoClassDefFoundError: Could not initialize class org.eclipse.kura.linux.net.modem.SupportedUsbModems


MMaiero:
This issue is caused by commit 9f5870537e36afa6641dc6d3d237686e82407f02 
Everything works right if the device has no internal modem, while, if the modem is available, the execution will stop at boot due to a `NoClassDefFoundError` on class `org.eclipse.kura.linux.net.modem.SupportedUsbModems`.

The issue is triggered by `SupportedUsbModemsInfo.isSupported` that tries to call `SupportedUsbModems.isSupported`. Before executing the `isSupported` method, the `SupportedUsbModems` static block will invoke `getLsusbInfo` that will call `LinuxProcessUtil.startWithStats`. 
The static block in `LinuxProcessUtil` modified by 9f5870537e36afa6641dc6d3d237686e82407f02 tries to get a service reference of the `SystemService`, but the result of this invocation is a null value. This is due to the fact that the `SystemService` component definition has a dependency on the `NetworkService`, but the network service needs the `SystemService` to execute: an indirect dependency loop is generated in this way, preventing the code to execute properly.

@ctron Could you take a look at this?

MMaiero:
This is the stack trace:

`!ENTRY org.eclipse.equinox.ds 4 0 2016-10-17 14:28:09.260
!MESSAGE [SCR] Exception while activating instance org.eclipse.kura.linux.net.NetworkServiceImpl@13b2f29 of component org.eclipse.kura.net.NetworkService  
!STACK 0
java.lang.reflect.InvocationTargetException
	at sun.reflect.GeneratedMethodAccessor4.invoke(Unknown Source)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:606)
	at org.eclipse.equinox.internal.ds.model.ServiceComponent.activate(ServiceComponent.java:235)
	at org.eclipse.equinox.internal.ds.model.ServiceComponentProp.activate(ServiceComponentProp.java:146)
	at org.eclipse.equinox.internal.ds.model.ServiceComponentProp.build(ServiceComponentProp.java:347)
	at org.eclipse.equinox.internal.ds.InstanceProcess.buildComponent(InstanceProcess.java:620)
	at org.eclipse.equinox.internal.ds.InstanceProcess.buildComponents(InstanceProcess.java:197)
[...]

PR #644, opened 2016-10-18
#640: Fix issue initializing modem

ctron:
This is workaround in order to fix the initialization issue around SystemService.

Currently the SystemService depends on the NetworkService for providing network information (which seems not optimal). The network manager calls back to the core bundle, which is hosting the "linux utils" (which is not optimal either).

This change does a lazy initialization of the LinuxProcessUtil since only two methods require the initialization information of this class.

Second the SupportedUsbModems class is implemented now using commons-exec for calling "lsusb" which allows to drop the dependency on the "linux.utils" package or "kura.core".

This change depends on CQ12140 [1]

[1] https://dev.eclipse.org/ipzilla/show_bug.cgi?id=12140


ctron:
Looks like dropping the dependency right now doesn't work out.

cdealti:
@ctron we agreed on not creating new PRs before rewriting the history. I propose to postpone (close) it.

ctron:
You can just keep it open. I will rebase it.

dwoodard1:
I think we can leave this open for now, until we have a better understanding of how to recover from the Git history rewrite. However, as discussed, lets please hold off on new PRs until we have made the history rewrite.

ctron:
Rebased after history rewrite.

cdealti:
If you managed to rebase this after the history rewrite it would be helpful to share this on the mailing list.
Thanks

ctron:
@dwoodard1 Has this been merged into the release branch?

kura/org.eclipse.kura.linux.net/META-INF/MANIFEST.MF
```diff
@@ -9,10 +9,12 @@ Service-Component: OSGI-INF/*.xml
...
 Bundle-ActivationPolicy: lazy
 Import-Package: javax.microedition.io,
- org.apache.commons.io;version="2.4.0",
+ org.apache.commons.exec;version="[1.3.0,2.0.0)",
+ org.apache.commons.io;version="[2.4.0,3.0.0)",
+ org.apache.commons.io.output;version="[2.4.0,3.0.0)",
  org.eclipse.kura;version="[1.1,2.0)",
  org.eclipse.kura.comm;version="[1.0,2.0)",
- org.eclipse.kura.core.linux.util;version="[1.0,2.0)",
+ org.eclipse.kura.core.linux.util;version="1.0.0",
```
MMaiero: I think this should be a range, as it was before.

MMaiero: The changes seem to be ok, but I had no time to test that effectively those work. I'll do that on Monday when our offices will be re-opened.

ctron: True.

MMaiero: *sorry, I meant Wednesday.

ctron: I read it as "next working day" 😉

ctron:
The CQ has been approved

ctron:
@MMaiero The version import is fixed

target-platform/config/kura.target-platform.build.properties
```diff
@@ -28,6 +28,8 @@ jdk.dio.version=1.0.6-SNAPSHOT
 org.junit.version=4.10.0.v4_10_0_v20120426-0900
 org.apache.commons.fileupload.version=1.2.2.v20111214-1400
 org.apache.commons.io.version=2.4.0
+commons-exec.version=1.3
```
MMaiero: I would collapse commons-exec.version and org.apache.commons.exec.version together referring as org.apache.commons.exec.version=1.3
I think that that duplication can be potentially misleading

target-platform/p2-repo-common/pom.xml
```diff
@@ -172,6 +172,11 @@
 									<version>${org.apache.commons.fileupload.version}</version>
 								</artifactItem>
 								<artifactItem>
+								    <groupId>org.apache.commons</groupId>
+								    <artifactId>commons-exec</artifactId>
+								    <version>${commons-exec.version}</version>
```
MMaiero: I would have org.apache.commons.exec.version instead or commons-exec.version. In the configuration file I would keep org.apache.commons.exec.version=1.3

kura/distrib/src/main/ant/build_equinox_distrib.xml
```diff
@@ -995,6 +996,8 @@ Resource-Processor: org.eclipse.kura.deployment.customizer.upgrade.rp.UpgradeScr
 				value=", reference:file:${kura.install.dir}/kura/plugins/jdk.dio_${jdk.dio.version}.jar@3" /> 
 			<entry key="osgi.bundles" operation="+"
 				value=", reference:file:${kura.install.dir}/kura/plugins/org.apache.commons.net_${org.apache.commons.net.version}.jar@3:start" />
+			<entry key="osgi.bundles" operation="+"
+                value=", reference:file:${kura.install.dir}/kura/plugins/org.apache.commons.exec_${org.apache.commons.exec.version}.jar@2:start" />
```
MMaiero: out of curiosity, is there a particular reason why commons.exec starts with a starting level of 2?

ctron: If should be available early. As the modem lookup is early as well. However I guess the "start" keyword could be omitted altogether since there is (AFAIK) no activator present in the bundle.

Then again most bundles in Kura are "started" anyway.

target-platform/p2-repo-common/pom.xml
ctron: The build script can't handle this right now. The final name will be converted by P2 from 1.3 to 1.3.0, so the file name will be first 1.3 and then 1.3.0. I can hardcode this, but this seems even more wrong to me.

ctron:
That change was there before the release branch was created. So to couldn't point to the release branch back then. It would have been good to first merge all required changes and then create the release branch.

I would suggest to test and squash & merge this change first into develop and then cherry pick the squashed commit.

MMaiero:
I'm sorry, probably I'm missing the point, but if you simply close this PR and do it again pointing from github to release-2.1.0 branch would not be enough?

ctron:
On the other hand you could just is as is. Squash & rebase and then cherry pick ... 

If this is any problem and the PR is fine, then I can of course do that for you.

MMaiero:
As discussed today, for me is fine to merge this PR. @dwoodard1 could you do that and cherry peak to 2.1.0?

dwoodard1:
Yes, I will take care of it today.

ctron:
As agreed I did remove the "start" tag from the bundle.

dwoodard1:
Merged into release-2.1.0:
a5e92bac53c49c14e833019195766028e5cb28fa
795cd8cb225d4446caecb24d96b9c380d7538cd1
cf776ae14515cfe3a2dd54f21e87c519a4b2eaf4
3448c33067f8d66a0ddbcfa72075f8886a42b529

Issue #641, opened 2016-10-17
[Camel] KuraCloudConsumer - Failed to subscribe exception


MMaiero:
At boot the following exception is thrown

`2016-10-17 16:02:41,581 [Component Resolve Thread (Bundle 7)] WARN  o.e.k.c.c.KuraCloudConsumer - Failed to subscribe
org.eclipse.kura.KuraNotConnectedException: Not connected.
    at org.eclipse.kura.core.data.transport.mqtt.MqttDataTransport.subscribe(MqttDataTransport.java:445)
    at org.eclipse.kura.core.data.DataServiceImpl.subscribe(DataServiceImpl.java:430)
    at org.eclipse.kura.core.cloud.CloudClientImpl.subscribe(CloudClientImpl.java:155)
    at org.eclipse.kura.camel.cloud.KuraCloudConsumer.performSubscribe(KuraCloudConsumer.java:90)
    at org.eclipse.kura.camel.cloud.KuraCloudConsumer.doStart(KuraCloudConsumer.java:49)
    at org.apache.camel.support.ServiceSupport.start(ServiceSupport.java:61)
    at org.apache.camel.impl.DefaultCamelContext.startService(DefaultCamelContext.java:3275)
    at org.apache.camel.impl.DefaultCamelContext.doStartOrResumeRouteConsumers(DefaultCamelContext.java:3569)
    at org.apache.camel.impl.DefaultCamelContext.doStartRouteConsumers(DefaultCamelContext.java:3505)
    at org.apache.camel.impl.DefaultCamelContext.safelyStartRouteServices(DefaultCamelContext.java:3435)
    at org.apache.camel.impl.DefaultCamelContext.safelyStartRouteServices(DefaultCamelContext.java:3448)
    at org.apache.camel.impl.DefaultCamelContext.startRouteService(DefaultCamelContext.java:3325)
    at org.apache.camel.impl.DefaultCamelContext.startRoute(DefaultCamelContext.java:976)
    at org.apache.camel.impl.DefaultCamelContext.startRouteDefinitions(DefaultCamelContext.java:3301)
    at org.apache.camel.impl.DefaultCamelContext.addRouteDefinitions(DefaultCamelContext.java:908)
    at org.eclipse.kura.camel.runner.AbstractRoutesProvider.applyRoutes(AbstractRoutesProvider.java:24)
    at org.eclipse.kura.camel.runner.CamelRunner.setRoutes(CamelRunner.java:558)
[...]

PR #648, opened 2016-10-18
#641: Check connection state before trying to subscribe

ctron:
If the subscription fails due to a "not connected" state, then the
consumer will re-subscribe at a later time anyway.

### DIFF
--- a/kura/org.eclipse.kura.camel/src/main/java/org/eclipse/kura/camel/cloud/KuraCloudConsumer.java
+++ b/kura/org.eclipse.kura.camel/src/main/java/org/eclipse/kura/camel/cloud/KuraCloudConsumer.java
@@ -32,7 +32,7 @@ public class KuraCloudConsumer extends DefaultConsumer implements CloudClientLis
 
     private final CloudClient cloudClient;
 
-    public KuraCloudConsumer(Endpoint endpoint, Processor processor, CloudClient cloudClient) {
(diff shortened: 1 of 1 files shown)


kura/org.eclipse.kura.camel/src/main/java/org/eclipse/kura/camel/cloud/KuraCloudConsumer.java
```diff
@@ -88,6 +89,8 @@ private void performSubscribe() {
         try {
             this.log.debug("Perform subscribe: {} / {}", this.cloudClient, getEndpoint().getTopic());
             this.cloudClient.subscribe(getEndpoint().getTopic(), 0);
+        } catch (KuraNotConnectedException e) {
+            this.log.info("Failed to subscribe", e);
```
amitjoy: I believe the log level should be error or warning.

ctron: I believe so too .. but @MMaiero  thinks otherwise .. see #641

cdealti:
@ctron we agreed on not creating new PRs before rewriting the history. I propose to postpone (close) it.

ctron:
I will amend this PR as it indeed should be fine to test and subscribe then only of connected. If the connection tests fails the listener will fire the subscription since it is registered before test and set operation. If the connection check is true and the connection gets lost in the middle, there will be an exception.

The log level will go back to warning.

ctron:
Amended and renamed PR.

PR #653, opened 2016-10-18
Fix wifi issues 2.0.1

dwoodard1:
Hotfixes for WiFi for the 2.0.2 service release.

@cdealti @ibinshtok Who should review this? It needs to be done tomorrow so we can proceed with the 2.0.2 release.


cdealti:
@dwoodard1 @ibinshtok I assume this has been extensively tested.
+1 for me.

Issue #637, opened 2016-10-17
[Windows] Use of NSIS and MinGW


malikm:
The use of **NSIS** is currently not covered by a CQ. Thus it's currently not possible to distribute an NSIS-based installer from Eclipse servers.
As an alternative, WiX has been suggested, which is approved. However a setup based on WiX is not possible to build on Linux-based machines (only Windows is supported). This doesn't fulfill the initial requirement to build a complete Windows installation on Linux systems too.

Native libraries and executables for Windows are currently build using **MinGW**. This allows building those files on both Windows- and Linux-based machines. However there's no full CQ for MinGW yet, and the code produced is linked against MinGW libraries.
As an alternative, a build using Microsoft Visual Studio has been suggested. This however doesn't allow building the required binaries on Linux systems and the restriction to dynamically link against Microsoft libraries would require shipping of Microsoft Visual Studio Redistributable package with Kura.

Further discussion details can be found here: https://github.com/eclipse/kura/issues/545

dwoodard1:
@malikm I had a chance to discuss the legal issues around NSIS at a recent Eclipse event. The question that came up, is NSIS necessary? Instead of distributing a *.exe, could we just distribute a Windows shell program? I understand some of the advantages of having the *.exe, but a quicker path to getting Kura Windows out may be the shell program.

ctron:
A shell program for setting up Kura is quite ugly compared to the standard way of using an MSI file. So what speaks against using MSI with WIX?

diegovisentin:
Have you already evaluated the use of GNOME msitools ? [https://wiki.gnome.org/msitools](https://wiki.gnome.org/msitools)

cdealti:
Due to all these legal aspects and the lack of interest on the Windows port I propose to close this issue and remove from Kura all the Windows specific code.
@dwoodard1 @malikm @markcullen what do you think?

cdealti:
Closing. Windows code will be removed

PR #667, opened 2016-10-21
[Windows] Fix Windows build and add 32-bit support

malikm:
- Fixes Windows files after history rewrite
- Add x86/32-bit build
- Additional updates to build process

Signed-off-by: Martin Malik martin.malik@eurotech.com


cdealti:
Didn't we agree on removing MinGW and NSIS [1]?

[1] https://github.com/eclipse/kura/issues/637

malikm:
AFAIK this was about removing the binaries produced by those tools, but source code should be OK.
Am I wrong ?

cdealti:
My understanding is that we want the native libraries built with Visual Studio and the Wix installer.
In this way the Windows installer can be downloaded from eclipse.org.

malikm:
Or keep NSIS and MinGW without binary files and let them build locally by users until a CQ for those is done ?
@ctron Is a CQ for NSIS and MinGW not feasible ? If Windows tools are approved, why something like NSIS and MinGW would not be possible ?

cdealti:
I'll leave the final word to @ctron. However we have time to file **full** CQs for NSIS and MinGW if you insist using those.

ctron:
So I we would go the way @malikm proposes we would still need a CQ, but a "works-with" would suffice IMHO. So Kura can be used without that, even on windows. And people could build it themselves. I don't even think we do need a CQ for MinGW, since there should be nothing MinGW specific those parts.

But then you would be able to distribute those binaries (installer, EXE) from Eclipse download servers.

Going with a complete CQ would allow exactly that. But it would take more time and has the probability to fail completely.

The Eclipse Foundation sometimes grants "exempt" pre-requisite CQs. Those are not always formalized, which is a point for improvement. Nobody is perfect :wink: The idea behind that is simply some sort of risk management. So for example Java can be used without a CQ, but also it cannot be distributed. Looking at Visual Studio, and here comes my personal opinion, nothing official, it is probably the fact that Microsoft has a clean IP record. So the chance to get sues for using the Visual Studio compiler is pretty unlikely. However the situation with MinGW may be different. Must not, but could be. Now when you want to bet on it, which horse would you pick?

ctron:
To sum it up. If you want to distribute it, you need a full CQ before merging. If you don't want to distribute the binaries, a works-with request would suffice.

If you go with Visual Studio and WiX you simple need to add this to the build and test CQ and are done as long as you don't use any WiX addons which actually add external content to the final MSI (like DLLs or images).

malikm:
Due to the disadvantages of using WiX and MSVC mentioned in #637, I'd prefer to stay with NSIS and MinGW.
If everybody agrees, can we start the CQ process ?

cdealti:
As @ctron already mentioned it's unlikely MinGW would be approved [1], [2].

[1] https://dev.eclipse.org/ipzilla/show_bug.cgi?id=8773
[2] https://wiki.eclipse.org/Tools_PMC_Meeting_2014-11-12

malikm:
My understanding wasn't "unlikely", but less strict. If it's really "unlikely" we might go ahead without pre-compiled and shipped binaries. Later compile using MSVC, but need to add the MSVC redistributable package are a pre-requisite.
Do we need to distribute the whole Windows package ? If not, is there anything else preventing accepting this PR ?

cdealti:
Closing as no decision has been made for what compiler and installer to use.

PR #676, opened 2016-10-27
Kura Wires Web Integration

amitjoy:


kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/EventHandlerServlet.java
```diff
@@ -0,0 +1,99 @@
+/**
+ * Copyright (c) 2011, 2016 Eurotech and/or its affiliates
```
ctron: Thus file seems to be new. Please use the correct year for copyright.

kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/client/ui/wires/WiresPanelUi.ui.xml
```diff
@@ -11,36 +9,173 @@
 
     Contributors:
       Eurotech
-
+      Amit Kumar Mondal (admin@amitinside.com)
```
ctron: Please use "Eurotech" as legal entity

kura/org.eclipse.kura.web2/src/main/webapp/joint.shapes.devs.js
```diff
@@ -0,0 +1,169 @@
+joint.shapes.devs = {};
```
ctron: Please add a copyright header

target-platform/p2-repo-common/pom.xml
```diff
@@ -82,6 +82,11 @@
 									<version>${org.eclipse.kura.sun.misc.version}</version>
 								</artifactItem>
 								<artifactItem>
+								    <groupId>org.json</groupId>
```
ctron: As already mentioned earlier, for a new component please use some JSON library which is OSGi ready on maven central and does not have legal issues.

cdealti:
@ctron what are the conventions for the date?

1. A new file will only have the year of the initial contribution
2. Every time the file is modified we add a comma followed by the year of the change

Is that all?

amitjoy:
@cdealti I agree with @ctron. I also use only 2016 for the copyright headers but mistakenly it stayed in the aforementioned file. I would amend it.

kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/EventHandlerServlet.java
amitjoy: @ctron 👍

kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/client/ui/wires/WiresPanelUi.ui.xml
amitjoy: @ctron Would you kindly be more specific?

kura/org.eclipse.kura.web2/src/main/webapp/joint.shapes.devs.js
amitjoy: @ctron This is part of JointJS library which is already been approved previously if I am not wrong. I just made some changes to it to cope up with wires.

target-platform/p2-repo-common/pom.xml
amitjoy: @ctron As the code base is really big for the Wires Web Integration, I would first concentrate on integrating it and then changing or modifying it. As this would be in different release, I believe we would have no problem in using JSON for now and if all the committers agree on having GSON for more flexibilities, I would have no problem in changing that later on.

So, for now, we can keep it and let the interested users play with Wires and then we can discuss on the JSON or GSON library integration.

cdealti:
Since the header stuff is asked every time it should be clarified in the CONTRIBUTING.md

MMaiero:
I agree, I'll open an issue to track it

amitjoy:
@ctron As we discussed, I have updated the headers as well as the about.html.

amitjoy:
@cdealti @dwoodard1 @ctron I believe we can go ahead with the CQ. What do you think?

target-platform/p2-repo-common/pom.xml
ctron: So, as agreed on EclipseCon, go ahead with org.json as long as we replace this dependency before 2.2 will be released.

kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/client/ui/wires/WiresPanelUi.ui.xml
ctron: As discussed at EclipseCon, just use your name. If someone at Eurotech did co-author on that file add them in the contributors list and add "and others".

kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/EventHandlerServlet.java
ctron: Also, use your name instead of "Eurotech".

kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/EventHandlerServlet.java
```diff
@@ -0,0 +1,99 @@
+/**
+ * Copyright (c) 2016 Eurotech and/or its affiliates
```
ctron: Please use your name instead of "Eurotech".

kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/client/ui/wires/PropertiesUi.java
```diff
@@ -0,0 +1,1339 @@
+/*******************************************************************************
+ * Copyright (c) 2016 Eurotech and/or its affiliates
```
ctron: Same here.... And for the others.

ctron:
The whole header stuff is already explained by Eclipse. We could link to it, but I don't see any benefit in replicating it.

ctron:
So I am not sure if we should continue merging or fix issues before that? What is the strategy with PRs in general?

I am fine with merging right away and amending later. But we should take the same approach then for all PRs.

amitjoy:
@ctron #678 and #688 are created to track progress of amending copyright headers for Wires Core, Asset Model and Wires Web.

mcarrer:
It is correct to leave Eurotech in the copyright header.

Amit is currently an intern in Eurotech and large contributions to the files in this PR (including architecture, design and coding) have been done by Eurotech employees.

ctron:
So then it would not be "Amit" in the contributors list, but "Eurotech".

amitjoy:
@mcarrer Thanks for the clarification.

@ctron According to [1], the contributors list would contain the names of the persons who actually wrote the source code. So, my proposition would be to have the names of the contributors in the Contributors list in header and that's why I have included my name. And as @mcarrer denoted, Eurotech would still be the copyright owner as I am an intern in Eurotech. Some source codes have been written as a joint work with Eurotech Employee @LucaDazi and those source codes have **Eurotech** in the contributors list in the headers also.

So, I believe the headers in the sources would still be the same as it is now.

[1] - http://www.eclipse.org/legal/copyrightandlicensenotice.php

ctron:
I think you are mixing up things here.

Currently Kura doesn't follow that approach. Currently contributors are the legal entities, namely "Eurotech". I am happy if we change this, but then we should do this for all contributions, not just Kura Wires.

And if we would follow the recommendation it would be "Amit … (Eurotech)":

>  It is good practice to list the name of the author with their employer if the latter owns the copyright. For example: "John Smith (XYZ Corp.)".

On the other hand the same document your are referencing strongly suggest to take into account the contributions made, including their description. So I agree, let's follow the way Eclipse proposes, but just do it all the way and not pick a few things here and there. As people get confused over time which rules to follow and which not.

amitjoy:
@ctron I understand completely that it would be good for group of fixes to be mentioned in the header but still I am skeptical about every minor change to be mentioned in the header too as the header would be too bloated and eventually would create blobs.

ctron:
As the document you are referencing to explains:

> Subsequent authors are listed on proceding lines with a short description of the contribution. If there is an associated Bugzilla report number, then it is good practice to include it here. Over time, this section actually becomes a log of the contributions. Although it could become quite lengthy in some cases, it should be maintained for two important reasons - to acknowledge the contributions of individuals (especially when their contribution is owned by their employer), and to assist with tracking the pedigree of the code (who wrote what).

kura/org.eclipse.kura.web2/pom.xml
```diff
@@ -1,19 +1,13 @@
...
      All rights reserved. This program and the accompanying materials
      are made available under the terms of the Eclipse Public License v1.0
      which accompanies this distribution, and is available at
      http://www.eclipse.org/legal/epl-v10.html
 
-    Contributors:
-      Eurotech
-      Red Hat Inc - remove dependency on slf4j:logj12
-         - Fix build warning
-	  Amit Kumar Mondal (admin@amitinside.com)
```
ctron: a) this seems to be aligned using tabs, not spaces
b) If we go for the Eclipse proposal this would be "Amit Kumar Mondal (Eurotech)"

ctron:
I general I think we should wait until the discussion around the copyright headers is completed. Not do a special way for Kura wires.

If we do decide to use authors instead of legal entities, then we need to do this for all PRs not just for Wires.

If we decide against that then the contributors would only be "Eurotech".

amitjoy:
@ctron Actually it's a general consensus here in Eurotech to have the copyright header the way it is done in my last commit. As Kura Wires is a pretty new feature, we thought of applying the copyright headers first to Wires and then to the rest of the source code base for maintaining a common copyright header for all.

As you have referred the document [1], we believe it would be nice to not include the contributors list in the header at all so that the headers remain clean.

[1] - https://www.eclipse.org/projects/handbook/#ip-copyright-headers

amitjoy:
@dwoodard1 @ctron @cdealti @MMaiero Any update on merging it?

MMaiero:
I'll merge it as soon as possible. Don't worry 😄

PR #685, opened 2016-10-28
Enable Kura API analysis

cdealti:
- Annotated org.eclipse.kura.api with @ProviderType OSGi annotations
- Annotated org.eclipse.kura.api with @noimplement Javadoc tags
- Enabled CI API analysis through bnd-baseline-maven-plugin
- Updated MANIFEST as suggested by bnd-baseline-maven-plugin
- TODO: add bnd-baseline-maven-plugin 3.4.0-SNAPSHOT to build/test CQ
- TODO: upgrade bnd-baseline-maven-plugin to 3.4.0 when released

Signed-off-by: Cristiano De Alti cristiano.dealti@eurotech.com


kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/bluetooth/BluetoothConnector.java
```diff
@@ -20,8 +20,13 @@
 import javax.microedition.io.Connection;
 import javax.microedition.io.ConnectionNotFoundException;
 
+import org.osgi.annotation.versioning.ProviderType;
 import org.osgi.service.io.ConnectorService;
 
+/**
+ * @noimplement
+ */
```
amitjoy: It would be better if you can use the full version of @noimplement with full sentence.

cdealti:
API interfaces defaults to consumer type ``@ConsumerType``.

API interfaces annotated with ``@ProviderType`` (and ``@noimplement`` Javadoc tag) are not meant to be implemented by clients (API consumers).

Clients (API consumers, application bundles) will import both consumer and provider type interfaces with ``[X.Y, X+1.0)`` disregarding if they are implementors.

API Providers (for example the Kura bundles) will import provider type interfaces with ``[X.Y,X.Y+1)`` and consumer type interfaces with ``[X.Y,X+1.0)``.

Adding a method to a consumer type interface (for example a listener interface) requires incrementing the MAJOR number because it is considered a breaking change for clients.

Adding a method to a provider type interface does not break compatibility for clients and only requires incrementing the MINOR number.

To test this:

* Remove ``~/.m2/repository/``
* Run the build. Make sure the following lines are logged building ``org.eclipse.kura:org.eclipse.kura.api``
```
[INFO] The baseline version was found to be 1.0.9
[INFO] Baselining check succeeded checking org.eclipse.kura:org.eclipse.kura.api:jar:1.1.0-SNAPSHOT against org.eclipse.kura:org.eclipse.kura.api:jar:1.0.9
```
* Add a method to a consumer type API interface, for example ``org.eclipse.kura.data.listener.DataServiceListener``.
  This is a breaking change for clients and the build will fail suggesting to update the MAJOR number of the ``org.eclipse.kura.data.listener`` package
* Add a method to a provider type API interface, for example ``org.eclipse.kura.data.DataService``.
  This is not a breaking change for clients (but it is for providers) and the build will fail suggesting to update the MINOR number of the ``org.eclipse.kura.data`` package

@dwoodard1 can you please review?

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/cloud/CloudClient.java
```diff
@@ -74,7 +75,10 @@
  * Note that the default subscription of a CloudClient allows remote servers
  * or applications running on other devices to publish messages addressed
  * to specific applications running on specific devices.
+ * 
+ * @noimplement
```
amitjoy: The same applies here

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/cloud/CloudPayloadProtoBufDecoder.java
```diff
@@ -23,6 +25,8 @@
      * @param payload
      * @return
      * @throws KuraException
+     * 
+     * @noimplement This interface is not intended to be implemented by clients.
```
amitjoy: @noimplement needs to be at the class level Javadoc

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/cloud/CloudPayloadProtoBufEncoder.java
```diff
@@ -24,6 +26,8 @@
      * @param gzipped
      * @return
      * @throws KuraException
+     * 
+     * @noimplement This interface is not intended to be implemented by clients.
```
amitjoy: Same here.

cdealti:
@dwoodard1 the build is failing because it attempts to use org.eclipse.kura.api 1.0.10-SNAPSHOT as a baseline which is not correct. Let me check.

@amitjoy thanks for reviewing it

cdealti:
@amitjoy Done.
@dwoodard1 Can you please review and test (instructions are above)?

amitjoy:
👍

dwoodard1:
Sorry for the delay, looks great @cdealti!

dwoodard1:
@cdealti Had to revert this PR. It is failing on Hudson, checking into it now.

PR #695, opened 2016-10-31
Clean up base64 processing

ctron:
This change forwards base 64 encoding to the already existing
dependency commons-net.

By re-using code from commons-net the reflection based method invocation
can be replaced by a proper method call.

In addition a unit test was added to verify the same output as the old
implementation.


kura/org.eclipse.kura.core.crypto/src/main/java/org/eclipse/kura/core/crypto/CryptoServiceImpl.java
```diff
@@ -92,73 +93,19 @@ protected void activate() {
...
-            convertedData = base64EncodeJava8(encryptedBytes);
-        } catch (Exception e) {
-
+            /* This try-catch would not be necessary if we would have a
+             * more recent version of commons.net of Java 7.
+             * But UTF-8 is always supported with Java.
+             */
+            return new String(Base64.encodeBase64(encryptedBytes, false), "UTF-8");
+        } catch (UnsupportedEncodingException e) {
+            throw new RuntimeException(e);
```
amitjoy: Instead of throwing generic Runtime Exception, it would be better off with KuraRuntimeException, in which you can specify the KuraErrorCode.

ctron: I don't see a big difference, the exception will never ever be thrown and the whole try-catch will be gone with java 7 anyway.

MMaiero:
Since Kura 3.0.0 will only target Java 8, why not thinking of using only [base64 java 8](https://docs.oracle.com/javase/8/docs/api/java/util/Base64.html) primitives? Indirectly it was already used. And we don't need the dependency on commons.net.

ctron:
Updated to use Java 8 only. Rebased.

cdealti:
@MMaiero please vote

PR #698, opened 2016-11-01
Fix ble process leak

dwoodard1:
This fix was applied to the Kura v2.0.2 hotfix release, this merges the change into the v2.1.0 release as well. 

@pierantoniomerlino Can you double check that everything looks ok?


pierantoniomerlino:
@dwoodard1 In kura.build.properties the first line is "<<<<<<< HEAD": could you check it? The rest is fine.

dwoodard1:
@pierantoniomerlino Thanks, I have addressed this in the last commit.

PR #699, opened 2016-11-01
Add DependencyManagement to Camel Examples

dwoodard1:
This fix adds the DependencyManagement section to the pom so artifacts can
be distributed to the Nexus repo.

@ctron I assume we want to distribute these. If not we will need to skip the deploy step during a release build.

### DIFF
--- a/kura/examples/org.eclipse.kura.example.camel.aggregation/pom.xml
+++ b/kura/examples/org.eclipse.kura.example.camel.aggregation/pom.xml
@@ -24,7 +24,20 @@
 		</repository>
 	</repositories>
 
-	<build>
+    <distributionManagement>
+        <repository>
+            <id>repo.eclipse.org</id>
+            <name>Kura Repository - Releases</name>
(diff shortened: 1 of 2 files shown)


ctron:
The question is if we do want to distribute those from the nexus instance. It doesn't hurt. But since it is a manual install I think the download.eclipse.org location would be more appropriate.

dwoodard1:
We are pushing the other examples, so it made since to me to push the Camel examples as well. I agree, they should be made available for direct download. I think we should create deployment packages and add them to the Eclipse Marketplace.

ctron:
Agreed, push all of them. But make them available on download.eclipse.org ... and maybe link them to the marketplace.

So can I merge and cherry-pick this one?

dwoodard1:
You can merge. No need to cherry pick, it is based on release-2.1.0. We will merge it back into develop after the release.

ctron:
Well, but when will that be? I would like to see both branches staying together in order to avoid rebase issues. Especially it we already know that we want to take over the change into develop.

Issue #691, opened 2016-10-31
NPE when CloudService isn't configured


ctron:
I can't fully grasp the situation at the moment. But it somehow seems that the new multi-cloud-service-ui cannot handle the situation where the actual CloudService instance is not configured by the ConfigurationService. The UI runs into a NPE.

Since most methods in the Web UI are undocumented, it is hard to figure out what is expected from the different services.

ctron:
So to me it looks a bit like as if the UI does not take into account the actual configured components. It takes the "id" of the CloudInstance instance and assumes that it has some configuration in the ConfigurationService. However as far as I understood, the new method `getStackComponentsPids` should be in charge to return the actual PIDs of what makes up the stack.

ctron:
So when the Web UI is being built, I would expect the `getStackComponentsPids` to be called at least once. However a breakpoint being set at that methods does not get triggered.

ctron:
I think the issue is located in the method `void selectConnection(GwtCloudConnectionEntry selection)`, where, instead of the factory id is used, it is fishing for factory ID by reading out the configuration.

I would assume that a direct get with the method `selection.getCloudFactoryPid()` should already provide the factory ID.

MMaiero:
@ctron could you point me to an example that causes this issue? Why that component has not to be tracked by the configuration service? Thanks

ctron:
Just use the camel cloud service factory.

MMaiero:
ok, tomorrow I'll do some tests

ctron:
Wednesday 😉

MMaiero:
yes, it's time for me to go to bed :-D

ctron:
Ok, I think this can be fixed by refactoring the Web UI 2 in way that the factory ID is discovered and not provided. `org.eclipse.kura.web.server.GwtCloudServiceImpl.findStackPidsByFactory(String, String)` could use the actual property from the service referenced instead.

I will give it a try.

ctron:
So with the bugfix it works for me both ways. I did create a PR on the release branch.

PR #704, opened 2016-11-02
Fix issue #691, #732 & Implement CloudServiceFactory for Camel

ctron:
This change does fix issue #691 by re-implementing the way the UI
backend does locate the components the communication stack is made up.

Instead of doing another round-trip to the server the backend extracts
the factory id from the service, locates the factory instance and
fills the result in one step.

This change also reduces a few assumptions made on the services, like
the requirement on a CloudService that is has to be configurable, the
requirement on the factory to be implemented using OSGi DS.


ctron:
I did amend the PR as discussed with a fix for #732

ctron:
@MMaiero @cdealti Ok, I think it looks good.

ctron:
@MMaiero @cdealti  Can this be merged?

MMaiero:
I'm reviewing it. I'll post my comments soon 😄

ctron:
Thanks @MMaiero , I just wanted to be sure that we are both waiting for something to happen :wink:

kura/org.eclipse.kura.core.cloud/META-INF/MANIFEST.MF
```diff
@@ -15,6 +15,7 @@ Import-Package: org.eclipse.kura;version="[1.0,2.0)",
  org.eclipse.kura.cloud.factory;version="[1.1,1.2)",
  org.eclipse.kura.command;version="[1.1,1.2)",
  org.eclipse.kura.configuration;version="[1.0,2.0)",
+ org.eclipse.kura.configuration.metatype;version="[1.0.0,2.0.0)",
```
MMaiero: We usually omit micros for version ranges.

kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/server/util/ServiceLocator.java
```diff
@@ -89,20 +95,92 @@ public static ServiceLocator getInstance() {
...
+     * Lookup services and iterate over their instances
+     *
+     * @param serviceClass
+     *            the services to look for
+     * @param consumer
+     *            the consumer which will be called for each service instance
+     * @throws GwtKuraException
+     *             if any service consumer throws an exception
+     */
+    public static <T> void withAllServices(final Class<T> serviceClass, final ServiceConsumer<T> consumer)
```
MMaiero: same for withOptionalService

kura/org.eclipse.kura.core.cloud/META-INF/MANIFEST.MF
ctron: So that means that you can't use the Eclipse IDE anymore for editing. And it doesn't provide any technical benefit.

kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/server/GwtCloudServiceImpl.java
```diff
@@ -34,34 +38,99 @@
...
+        public int compare(GwtCloudConnectionEntry o1, GwtCloudConnectionEntry o2) {
+            int rc;
+
+            rc = o1.getCloudFactoryPid().compareTo(o2.getCloudFactoryPid());
+            if (rc != 0) {
+                return rc;
+            }
+
+            rc = o1.getCloudServicePid().compareTo(o2.getCloudServicePid());
+            if (rc != 0) {
```
ctron: True, it just keeps the pattern that way. So adding another check follows the pattern. But if this is a problem for you I can change it.

kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/server/util/ServiceLocator.java
ctron: At least there is JavaDoc :wink: 

Well `withOptionalService` was already there. I again, I kept the pattern. It could also be "runWith" or "consumeOptionalService" ...

MMaiero: well, withOptionalService was added by commit 8a5c94415035b8b5256fd2969eda38a2e019f383 and, for what I know, is not released. So, we are eventually still in time to rename it 😉

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/cloud/factory/CloudServiceFactory.java
```diff
@@ -150,4 +152,21 @@
...
+     * <br>
+     * The IDs returned by this method must not necessarily point to registered OSGi services. But if they do, they must
+     * point only to instances of the {@link CloudService}.
+     * 
+     * @return the set of services, never returns {@code null}
+     * @throws KuraException
+     *
+     * @since {@link org.eclipse.kura.cloud.factory} 1.1.0
+     */
+    Set<String> getConfigurations() throws KuraException;
```
ctron: `getManagedCloudServicePids`?! .. double plural sounds strange.

ctron: done

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/cloud/factory/CloudServiceFactory.java
```diff
@@ -150,4 +152,21 @@
      * @throws KuraException
      */
     void deleteConfiguration(String pid) throws KuraException;
+
+    /**
+     * Return a set of services managed by this factory
+     * <br>
+     * It is up to the factory how it does assembles this. The IDs returned by this list must be the PIDs assigned
```
ctron: Done.

kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/client/ui/CloudServices/CloudServiceConfigurationsUi.java
```diff
@@ -170,42 +139,42 @@ public void onFailure(Throwable ex) {
...
-                                new AsyncCallback<List<GwtConfigComponent>>() {
-
                             @Override
                             public void onFailure(Throwable ex) {
-                                logger.log(Level.SEVERE, ex.getMessage(), ex);
                                 FailureHandler.handle(ex, EntryClassUi.class.getName());
                             }
 
                             @Override
-                            public void onSuccess(List<GwtConfigComponent> result) {
```
ctron: Sorry, but I don't understand that comment.

kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/server/util/ServiceLocator.java
ctron: True ... so make a decision.

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/cloud/factory/CloudServiceFactory.java
```diff
@@ -8,10 +8,12 @@
  *
  * Contributors:
  *     Eurotech
+ *     Red Hat Inc - API enhancements
```
ctron: Done

kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/client/ui/CloudServices/CloudServiceConfigurationsUi.java
MMaiero: Please have a look at https://github.com/eclipse/kura/issues/539#issuecomment-256005042 point 1

ctron: I still don't understand it. The order of stack PIDs it not changed. Maybe you can point me the code where you think the issue is.

kura/org.eclipse.kura.core.cloud/META-INF/MANIFEST.MF
ctron: Done

ctron:
So all comments should be addressed.

kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/server/util/ServiceLocator.java
dwoodard1: +1 for @MMaiero naming scheme.

ctron: Done. So I hope this doesn't spark another discussion, but I did change the name from `withOptionalService` -> `applyToServiceOptionally`, since this considers services optionally. At some point there may be a method `applyToService` which actually requires a service and would fail if the service is not present.

kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/client/ui/CloudServices/CloudServiceConfigurationsUi.java
ctron: So as discussed, please re-test and maybe provide a screenshot of what is wrong. Otherwise I think we are good to merge.

kura/org.eclipse.kura.camel.cloud.factory/src/org/eclipse/kura/camel/cloud/factory/internal/CamelManager.java
```diff
@@ -180,7 +189,8 @@ public Tocd getDefinition() {
         final Set<String> ids = new TreeSet<>();
         try {
 
-            final Collection<ServiceReference<CamelFactory>> refs = FrameworkUtil.getBundle(CamelManager.class).getBundleContext().getServiceReferences(CamelFactory.class, null);
+            final Collection<ServiceReference<CamelFactory>> refs = FrameworkUtil.getBundle(CamelManager.class)
```
amitjoy: You can reuse the `getServiceReferences(BundleContext, Class<T>, String)` method from `ServiceUtil`

ctron: We discussed yesterday about the required changes for this PR, we probably will find all kinds of other things we could to in addition. But let's finish this up, it is open for two weeks already.

Beside that, I can't that class in the Kura setup. Maybe you can point me to it.

amitjoy: 👍

ctron: Could you still point me towards that class. It sounds promising.

amitjoy: The class `ServiceUtil` resides in `org.eclipse.kura.util` bundle.

ctron: Which is not present in the release-2.1.0 branch :wink: 

So I do think that for 3.0 we should move the static methods from the `ServiceLocator` class to the new `ServiceUtils` class.

MMaiero:
Merging. Eventually we will address later the last issue about ordering. So you can work in parallel

MMaiero:
This is what I meant.
If you comment the alphabetical ordering in GwtComponentServiceImpl you will end up with:
ordering

that is not the ordering given by the factory.

ctron:
I will try to test this.

ctron:
So according to the code this should be:

```java
componentPids.add(pid);
componentPids.add(dataServicePid);
componentPids.add(dataTransportServicePid);
```

And the result is:
mcc1

Which seems correct to me.

I will do another full run, cleaning all local maven caches and browser caches. Maybe you can do the same and then we compare this again.

Following the code path, I can't see how that is possible. But maybe you can point me to the code which is wrong in your opinion.

MMaiero:
Could you try out commenting sortConfigurationsByName in GwtComponentServiceImpl?
I was able to have that different ordering in that way.

ctron:
Ok I see what you mean. But this method was there before. In the same call stack. So the sort order was the same as before this PR.

ctron:
So do you want to remove the call to the sort method?

MMaiero:
> Ok I see what you mean. But this method was there before. In the same call stack. So the sort order was the same as before this PR.

Yes, the ordering returned by that method was there before, but my previous PR tried to force the factory ordering [1]. Did you have the chance to try it?

> So do you want to remove the call to the sort method?

No, I don't want to. It was just a way to show you what @cdealti meant. That method can continue to return alphabetically ordered services. 

[1] https://github.com/ctron/kura/blob/382ea9b2d6a627ebe96523884ffbe271d777759c/kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/client/ui/CloudServices/CloudServiceConfigurationsUi.java#L173-L211

ctron:
I did try and it provided the screenshot above. Is there something wrong with that?

MMaiero:
Strange. I've tried with a clean build and the following changes:

```
diff --git a/kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/client/ui/CloudServices/CloudServiceConfigurationsUi.java b/kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/client/ui/CloudServices/CloudServiceConfigurationsUi.java
index 2c0df50..86cad18 100644
--- a/kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/client/ui/CloudServices/CloudServiceConfigurationsUi.java
+++ b/kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/client/ui/CloudServices/CloudServiceConfigurationsUi.java
@@ -129,7 +129,6 @@ public class CloudServiceConfigurationsUi extends Composite {
     }
 
     private void getCloudStackConfigurations(String factoryPid, String cloudServicePid) {
-
         this.gwtCloudService.findStackPidsByFactory(factoryPid, cloudServicePid, new AsyncCallback<List<String>>() {
 
             @Override
@@ -139,42 +138,42 @@ public class CloudServiceConfigurationsUi extends Composite {
             }
 
             @Override
-            public void onSuccess(List<String> result) {
-                final List<String> pidsResult = result;
+            public void onSuccess(final List<String> pids) {
 
                 CloudServiceConfigurationsUi.this.gwtXSRFService
                         .generateSecurityToken(new AsyncCallback<GwtXSRFToken>() {
 
+                    @Override
+                    public void onFailure(Throwable ex) {
+                        FailureHandler.handle(ex, EntryClassUi.class.getName());
+                    }
+
+                    @Override
+                    public void onSuccess(GwtXSRFToken token) {
+                        CloudServiceConfigurationsUi.this.gwtComponentService.findFilteredComponentConfigurations(token,
+                                new AsyncCallback<List<GwtConfigComponent>>() {
+
                             @Override
                             public void onFailure(Throwable ex) {
+                                logger.log(Level.SEVERE, ex.getMessage(), ex);
                                 FailureHandler.handle(ex, EntryClassUi.class.getName());
                             }
 
                             @Override
-                            public void onSuccess(GwtXSRFToken token) {
-                                CloudServiceConfigurationsUi.this.gwtComponentService
-                                        .findFilteredComponentConfigurations(token,
-                                                new AsyncCallback<List<GwtConfigComponent>>() {
-
-                                                    @Override
-                                                    public void onFailure(Throwable ex) {
-                                                        logger.log(Level.SEVERE, ex.getMessage(), ex);
-                                                        FailureHandler.handle(ex, EntryClassUi.class.getName());
-                                                    }
-
-                                                    @Override
-                                                    public void onSuccess(List<GwtConfigComponent> result) {
-                                                        boolean isFirstEntry = true;
-                                                        for (GwtConfigComponent pair : result) {
-                                                            if (pidsResult.contains(pair.getComponentId())) {
-                                                                renderTabs(pair, isFirstEntry);
-                                                                isFirstEntry = false;
-                                                            }
-                                                        }
-                                                    }
-                                                });
+                            public void onSuccess(List<GwtConfigComponent> result) {
+                                boolean isFirstEntry = true;
+                                for (String cloudStackPid : pids) {
+                                    for (GwtConfigComponent pair : result) {
+                                        if (cloudStackPid.equals(pair.getComponentId())) {
+                                            renderTabs(pair, isFirstEntry);
+                                            isFirstEntry = false;
+                                        }
+                                    }
+                                }
                             }
                         });
+                    }
+                });
             }
         });
     }
diff --git a/kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/server/GwtComponentServiceImpl.java b/kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/server/GwtComponentServiceImpl.java
index 6552353..ec4d565 100644
--- a/kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/server/GwtComponentServiceImpl.java
+++ b/kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/server/GwtComponentServiceImpl.java
@@ -355,7 +355,7 @@ public class GwtComponentServiceImpl extends OsgiRemoteServiceServlet implements
 
             List<ComponentConfiguration> configs = cs.getComponentConfigurations();
             // sort the list alphabetically by service name
-            sortConfigurationsByName(configs);
+            // sortConfigurationsByName(configs);
 
             for (ComponentConfiguration config : configs) {
                 GwtConfigComponent gwtConfigComponent = createGwtComponentConfiguration(config);
@@ -434,7 +434,7 @@ public class GwtComponentServiceImpl extends OsgiRemoteServiceServlet implements
 
             List<ComponentConfiguration> configs = cs.getComponentConfigurations();
             // sort the list alphabetically by service name
-            sortConfigurationsByName(configs);
+            // sortConfigurationsByName(configs);
 
             for (ComponentConfiguration config : configs) {
                 GwtConfigComponent gwtConfigComponent = createGwtComponentConfiguration(config);
```

The result obtained from the web UI is the following:
schermata 2016-11-20 alle 11 21 01

in which the order should be given by the factory and not by the alphabetical ordering given by GwtComponentServiceImpl.

I've uploaded the compiled code here: https://s3-us-west-2.amazonaws.com/kura-folder/ordering/kura_2.1.0_raspberry-pi-2-3-nn_installer.deb
Could you try it and tell me if it works also for you? Thanks

dwoodard1:
@MMaiero I get the same results. What exactly is the problem?

MMaiero:
@dwoodard1 It depends on what you mean with "the same results" 😄: If, with the last code, you see the tabs ordered as CloudService, DataService, MqttDataTransport, great! It means that even if the GwtComponentServiceImpl methods do not apply an ordering, the ordering of the tabs is the one imposed by the factory. And this was my intent, also following what suggested by @cdealti in https://github.com/eclipse/kura/issues/539#issuecomment-256005042 point 1.

> What exactly is the problem?

The current code orders the tabs only due to an alphabetical ordering imposed by methods in GwtComponentServiceImpl.

ctron:
But what does this have to do with this PR?

MMaiero:
Simply this PR modified this code: https://github.com/eclipse/kura/pull/704/files#diff-0c6798e271027ef5dad10bf301d4f7c5L195

ctron:
But still I don't see a changed call path with changes a behavior here. Maybe you can point that out. That would definitely help.

cdealti:
@ctron @dwoodard1 @MMaiero let's try to sort this out.
With the Kura default stack you will not see any difference between the alphabetical order and the hierarchical order returned by ``CloudServiceFactory.getStackComponentsPids`` because in this case it happens to be same order.

Trying other stacks (see https://github.com/eclipse/kura/issues/539#issuecomment-256005042) I've found that the hierarchical order is preferable (personal taste).

In the end, @MMaiero is just saying that to obtain the hierarchical order we should apply to the develop the above diff.

@ctron is it OK for you to submit @MMaiero in a new PR and stop this discussion?

ctron:
I think I do understand what is required in the end. What I still don't understand what it has to do with the changes of this PR. If so if this isn't related to this PR, yes, please open a new issue for tracking this. If this is a defect introduced by this PR, then let's stay here.

cdealti:
No, there is no defect in this one. The requested change will be addressed in a new PR.
Thanks

ctron:
@cdealti thanks for that clarification.

Maybe we can open a new issue in the future if the topic is out of the scope of the PR. I think that would help.

PR #705, opened 2016-11-02
Clean up and deprecate QuickSort

ctron:
* Forward all calls to plain Java methods
* Deprecate class and methods, comment
* Prevent subclassing


kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/util/QuickSort.java
```diff
@@ -8,160 +8,35 @@
  *
  * Contributors:
  *     Eurotech
+ *     Red Hat Inc - Re-implement using pure Java methods
```
cdealti: Can you please remove "Re-implement using pure Java methods"?

ctron:
Updated and rebased.

PR #711, opened 2016-11-02
Cleanup and bug fixes on org.eclipse.kura.core.comm

ctron:
This change cleans up a bunch of dead code and oddities. It also fixes
a bug on Windows, therefore the version uptick.

Changes are:

* Remove unnecessary code
* Fix path separator (Windows bug)
* Fix hex output
* Only generate hex output when necessary
* Let the connection implement Closable
* Cleanup code style
* Use same timeout calculation (long instead of Date) for both methods


kura/org.eclipse.kura.core.comm/src/main/java/org/eclipse/kura/core/comm/CommConnectionImpl.java
```diff
@@ -121,110 +118,115 @@ public DataOutputStream openDataOutputStream() throws IOException {
...
+        this.outputStream.write(message, 0, message.length);
+        this.outputStream.flush();
     }
 
     @Override
     public synchronized byte[] sendCommand(byte[] command, int timeout) throws KuraException, IOException {
-        if (command != null) {
-            s_logger.debug("sendMessage() - {}", getBytesAsString(command));
+        if (command == null) {
+            throw new NullPointerException("Serial command must not be null");
```
amitjoy: You can use Kura's Preconditions static factories

kura/org.eclipse.kura.core.comm/src/main/java/org/eclipse/kura/core/comm/CommConnectionImpl.java
```diff
@@ -121,110 +118,115 @@ public DataOutputStream openDataOutputStream() throws IOException {
...
-            s_logger.debug("sendMessage() - {}", getBytesAsString(command));
 
-            if (this.m_outputStream == null) {
-                openOutputStream();
-            }
-            if (this.m_inputStream == null) {
-                openInputStream();
-            }
+        if (command == null) {
+            throw new NullPointerException("Serial command must not be null");
```
amitjoy: same here

kura/org.eclipse.kura.core.comm/src/main/java/org/eclipse/kura/core/comm/CommConnectionImpl.java
```diff
@@ -54,42 +56,37 @@
...
-        this.m_commUri = commUri;
-        this.m_serialPort = null;
-        this.m_port = this.m_commUri.getPort();
-        this.m_baudRate = this.m_commUri.getBaudRate();
-        this.m_dataBits = this.m_commUri.getDataBits();
-        this.m_stopBits = this.m_commUri.getStopBits();
-        this.m_parity = this.m_commUri.getParity();
-        this.m_flowControl = this.m_commUri.getFlowControl();
-        this.m_timeout = this.m_commUri.getTimeout();
+        this.commUri = commUri;
```
amitjoy: if commUri is null, then rest of the initialisations result in NPE. It's better to check for null or you can use Preconditions.

kura/org.eclipse.kura.core.comm/src/main/java/org/eclipse/kura/core/comm/CommConnectionImpl.java
```diff
@@ -54,42 +56,37 @@
...
-                this.m_serialPort.setFlowControlMode(this.m_flowControl);
+                this.serialPort.setSerialPortParams(baudRate, dataBits, stopBits, parity);
+                this.serialPort.setFlowControlMode(flowControl);
             } catch (UnsupportedCommOperationException e) {
-                e.printStackTrace();
+                logger.error("Failed to configure COM port", e);
+                // TODO shouldn't we throw an IOException here
             }
         } else {
             throw new IOException("Unsupported Port Type");
```
amitjoy: My personal preference is to use only the KuraException as a checked exception and KuraRuntimeException as an unchecked one. Whenever I need to throw any checked exception, I wrap it in KuraException and throw it. The same for unchecked exception (wrap in KuraRuntimeException).

kura/org.eclipse.kura.core.comm/src/main/java/org/eclipse/kura/core/comm/CommConnectionImpl.java
```diff
@@ -121,110 +118,115 @@ public DataOutputStream openDataOutputStream() throws IOException {
...
+            this.serialPort = null;
         }
     }
 
     @Override
     public synchronized void sendMessage(byte[] message) throws KuraException, IOException {
-        if (message != null) {
-            s_logger.debug("sendMessage() - {}", getBytesAsString(message));
+        if (message == null) {
+            throw new NullPointerException("Messahe must not be null");
```
amitjoy: You can use Preconditions here as well.

kura/org.eclipse.kura.core.comm/OSGI-INF/comm.xml
```diff
@@ -10,9 +10,10 @@
 
     Contributors:
       Eurotech
+      Red Hat Inc - clean up
 
 -->
-<scr:component xmlns:scr="http://www.osgi.org/xmlns/scr/v1.1.0" activate="activate" deactivate="deactivate" immediate="true" name="org.eclipse.kura.core.comm.CommConnectionFactory">
+<scr:component xmlns:scr="http://www.osgi.org/xmlns/scr/v1.1.0" immediate="true" name="org.eclipse.kura.core.comm.CommConnectionFactory">
```
amitjoy: It's better to have this service component delayed as no other service reference is actually consumed by this component. It can be activated on demand.

kura/org.eclipse.kura.core.comm/src/main/java/org/eclipse/kura/core/comm/CommConnectionFactory.java
```diff
@@ -8,49 +8,26 @@
...
-
     @Override
     public Connection createConnection(String name, int mode, boolean timeouts) throws IOException {
         try {
-            CommURI uri = CommURI.parseString(name);
-            return new CommConnectionImpl(uri, mode, timeouts);
+            return new CommConnectionImpl(parseString(name), mode, timeouts);
+        } catch (IOException e) {
+            throw e; // re-throw
         } catch (Throwable t) {
```
amitjoy: You can catch only Exception and then throw it suppressing in KuraException. This way we would not be needing multi-catch blocks.

kura/org.eclipse.kura.core.comm/src/main/java/org/eclipse/kura/core/comm/CommConnectionImpl.java
```diff
@@ -121,110 +118,115 @@ public DataOutputStream openDataOutputStream() throws IOException {
 
     @Override
     public synchronized OutputStream openOutputStream() throws IOException {
-        if (this.m_outputStream == null) {
-            this.m_outputStream = this.m_serialPort.getOutputStream();
+        if (this.outputStream == null) {
```
amitjoy: serialPort null check would be good

kura/org.eclipse.kura.core.comm/src/main/java/org/eclipse/kura/core/comm/CommConnectionImpl.java
```diff
@@ -108,10 +105,10 @@ public DataInputStream openDataInputStream() throws IOException {
 
     @Override
     public synchronized InputStream openInputStream() throws IOException {
-        if (this.m_inputStream == null) {
-            this.m_inputStream = this.m_serialPort.getInputStream();
+        if (this.inputStream == null) {
+            this.inputStream = this.serialPort.getInputStream();
```
amitjoy: serialPort null check would be good

kura/org.eclipse.kura.core.comm/src/main/java/org/eclipse/kura/core/comm/CommConnectionImpl.java
```diff
@@ -121,110 +118,115 @@ public DataOutputStream openDataOutputStream() throws IOException {
...
-                this.m_outputStream.close();
-                this.m_outputStream = null;
+            if (this.outputStream != null) {
+                this.outputStream.close();
+                this.outputStream = null;
             }
 
-            this.m_serialPort.close();
-            this.m_serialPort = null;
+            this.serialPort.close();
```
amitjoy: serialPort null check would be good

kura/org.eclipse.kura.core.comm/src/main/java/org/eclipse/kura/core/comm/CommConnectionImpl.java
```diff
@@ -287,16 +289,16 @@ private synchronized ByteBuffer getResponse(int timeout, int demark) throws IOEx
...
             return null;
-        } else {
-            StringBuffer sb = new StringBuffer();
-            for (byte b : bytes) {
-                sb.append("0x").append(Integer.toHexString(b)).append(" ");
-            }
+        }
 
-            return sb.toString();
+        StringBuffer sb = new StringBuffer(bytes.length * 3);
```
amitjoy: Out of curiosity, is there any specific reason of preferring StringBuffer to StringBuilder?

kura/org.eclipse.kura.core.comm/src/main/java/org/eclipse/kura/core/comm/CommConnectionImpl.java
ctron: Yes, there could be a bunch more null checks overall.

kura/org.eclipse.kura.core.comm/src/main/java/org/eclipse/kura/core/comm/CommConnectionImpl.java
ctron: From what I have seen the constructor should take care if this.

kura/org.eclipse.kura.core.comm/src/main/java/org/eclipse/kura/core/comm/CommConnectionImpl.java
ctron: But I would wait for Java 7 and use some more common Objects.* checks instead. I really hope we can at least advance to Java 7 in this release.

kura/org.eclipse.kura.core.comm/src/main/java/org/eclipse/kura/core/comm/CommConnectionImpl.java
ctron: No. Should be a StringBuilder.

kura/org.eclipse.kura.core.comm/src/main/java/org/eclipse/kura/core/comm/CommConnectionImpl.java
amitjoy: I have created a Helper Class - `Preconditions` which does the similar thing in Kura-specifc way to throw Unchecked Exception if a condition is never met.

kura/org.eclipse.kura.core.comm/src/main/java/org/eclipse/kura/core/comm/CommConnectionFactory.java
amitjoy: Actually the usual way to throw a checked exception in kura is by wrapping it in KuraException. And in case of unchecked one, the usual way is to wrap in KuraRuntimeException.

kura/org.eclipse.kura.core.comm/src/main/java/org/eclipse/kura/core/comm/CommConnectionImpl.java
ctron: I know. But I would like to stick with standard Java functionality in order to make it easier for people to understand. More java standard, less Kura specific.

kura/org.eclipse.kura.core.comm/OSGI-INF/comm.xml
ctron: I agree. If @cdealti agrees to this as well, I am happy to change it.

kura/org.eclipse.kura.core.comm/src/main/java/org/eclipse/kura/core/comm/CommConnectionFactory.java
ctron: So the method already declares a thrown IOException, which is common Java behavior in such cases. I really don't see any benefit in introducing yet another Kura specific wrapper here. Adding another level in the Exception stack, being unable to distinguish this outside of the method call.

amitjoy: The idea is to never implementation specific checked or unchecked exceptions escalate to the higher layers. `IOException` occurs at the communication layer and the business layer do not need to know about such exceptions.

In case of such scenario, we could either - 

1. Convert `IOExcepton` to `KuraException`, if the client of this bundle is expected to recover from such exceptional cause
2. Convert `IOException` to `KuraRuntimeException`, if the client of this bundle `cannot` do anything

This is what Kura follows now to separate lower level exceptions from the higher level.

kura/org.eclipse.kura.core.comm/src/main/java/org/eclipse/kura/core/comm/CommConnectionImpl.java
```diff
@@ -44,7 +46,7 @@
             String existingDirs = System.getProperty("java.ext.dirs");
```
amitjoy: Would be good if you declare `java.ext.dirs` as a constant so that it can be reused in different places without duplicating.

kura/org.eclipse.kura.core.comm/OSGI-INF/comm.xml
MMaiero: @ctron please, do it

kura/org.eclipse.kura.core.comm/src/main/java/org/eclipse/kura/core/comm/CommConnectionImpl.java
```diff
@@ -121,115 +118,135 @@ public DataOutputStream openDataOutputStream() throws IOException {
...
+    }
+    
     @Override
     public synchronized void sendMessage(byte[] message) throws KuraException, IOException {
-        if (message != null) {
-            s_logger.debug("sendMessage() - {}", getBytesAsString(message));
+        checkIfClosed();
+        
+        if (message == null) {
+            throw new NullPointerException("Messahe must not be null");
```
ctron: Done

kura/org.eclipse.kura.core.comm/src/main/java/org/eclipse/kura/core/comm/CommConnectionImpl.java
ctron: Done

kura/org.eclipse.kura.core.comm/src/main/java/org/eclipse/kura/core/comm/CommConnectionImpl.java
ctron: Fixed.

cdealti:
@amitjoy OK to merge?

kura/org.eclipse.kura.core.comm/src/main/java/org/eclipse/kura/core/comm/CommConnectionImpl.java
amitjoy: Could you kindly add the required preconditions in the constructor?

kura/org.eclipse.kura.core.comm/src/main/java/org/eclipse/kura/core/comm/CommConnectionFactory.java
amitjoy: Could you kindly catch only `IOException` here instead of throwing `IOException` in both the catch blocks? or If you believe it can also encounter any other exception apart from `IOException`, you can use multi-catch construct.

kura/org.eclipse.kura.core.comm/src/main/java/org/eclipse/kura/core/comm/CommConnectionImpl.java
amitjoy: Please add a null check precondition here as well.

kura/org.eclipse.kura.core.comm/src/main/java/org/eclipse/kura/core/comm/CommConnectionImpl.java
ctron: Same what?

ctron:
Updated and rebased

cdealti:
@amitjoy please vote

kura/org.eclipse.kura.core.comm/src/main/java/org/eclipse/kura/core/comm/CommConnectionFactory.java
```diff
@@ -8,49 +8,26 @@
...
```
amitjoy: Could you kindly catch only IOException here instead of throwing IOException in both the catch blocks? or If you believe it can also encounter any other exception apart from IOException, you can use multi-catch construct.

kura/org.eclipse.kura.core.comm/src/main/java/org/eclipse/kura/core/comm/CommConnectionImpl.java
amitjoy: This was old comment about using NPE. I have commented it before we started using exceptions in a different way. Just ignore it 😄.

kura/org.eclipse.kura.core.comm/src/main/java/org/eclipse/kura/core/comm/CommConnectionImpl.java
amitjoy: 👍

Issue #706, opened 2016-11-02
Run unit tests on Travis


ctron:
The Travis CI build should be configured to run unit tests for validation.

ctron:
This is merged in 3.0.0

PR #712, opened 2016-11-03
#706: Enable unit tests when running on Travis CI

ctron:
This PR allows to run the unit tests with the `build-all.sh` script on request. The default is still not to run those tests.

On Travis CI the flag is set and unit tests get executed. The intention is to fail the build if unit tests fail.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,6 @@ before_script:
   - ls /usr/lib/jvm
   - cp kura/setups/toolchains/toolchains-travis.xml ~/.m2/toolchains.xml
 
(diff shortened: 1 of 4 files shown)


ctron:
I forgot to mention that this PR also disables the unit tests of `org.eclipse.kura.core.test` since they don't seem to work and that would cause the build to fail. Disabling this module from the test run allows at least the other unit tests to be used.

AFAIK @cdealti is working on a replacement for this. So we can re-enable this as soon as it is fixed.

hekonsek:
Merged. Thanks!

Issue #683, opened 2016-10-27
Control Prefix in Request/Response


dwoodard1:
The control prefix in CloudService was made configurable as many MQTT brokers didn't like the '$' prefix. However, this prefix is still coded in parts of Kura. This is impacting Request/Response communication in the CloudConfigurationHandler. The CloudService uses [KuraTopic](https://github.com/eclipse/kura/blob/develop/kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/message/KuraTopic.java) to parse topics, but this class has the below logic. While a bug has been reported on this specific issue, control topics are not being process correctly if the prefix is changed from "$EDC".

``` java
        // prefix   

        int index = 0;
        int offset = 0; // skip a slash
        if (this.m_topicParts[0].startsWith("$")) {
            this.m_prefix = this.m_topicParts[index];
            offset += this.m_prefix.length() + 1;
            index++;
        }
```

The other related issue is in CloudCallService. This implementation has this:

``` java
StringBuilder sbReqTopic = new StringBuilder("$EDC").append("/").append(ACCOUNT_NAME_VAR_NAME).append("/")
                .append(deviceId).append("/").append(appId).append("/").append(appTopic);

StringBuilder sbRespTopic = new StringBuilder("$EDC").append("/").append(ACCOUNT_NAME_VAR_NAME).append("/")
              .append(CLIENT_ID_VAR_NAME).append("/").append(appId).append("/").append("REPLY").append("/")
                .append(requestId);
```

The CloudCallService has no knowledge of the CloudService control prefix, so I am not sure of the best solution here other than making CloudService a dependency.

It appears to me we will need API changes to address these issues, so it would be nice if we can get this fixed in the v2.1.0 release.

cdealti:
@dwoodard1 The ``CloudCallService`` should not even be an API. It was only meant for testing purposes.
There are also some other things to consider given that CloudService is not a singleton anymore (MutlipleCloudConnections)

dwoodard1:
@cdealti I agree regarding the CloudCallService, so the change here is less important. For the CloudService, I was thinking if we changed the constructor (or added an overload) of KuraTopic to to accept the control prefix from the CloudService, that should take care of Factories as well. This would also prevent API breakages if we overload.

cdealti:
+1

PR #719, opened 2016-11-04
Fix for Issue #683

dwoodard1:
This commit properly handles customization of the control
topic prefix when decoded Kura topics. See issue #683 

@MMaiero @pierantoniomerlino Can one of you verify/merge? To test, change the default control topic by removing the "$". Then, attempt to use the CONF-V1 application to request a configuration.


kura/org.eclipse.kura.api/META-INF/MANIFEST.MF
```diff
@@ -22,7 +22,7 @@ Export-Package: org.eclipse.kura;version="1.2.0",
  org.eclipse.kura.data.transport.listener;version="1.0.0",
  org.eclipse.kura.db;version="1.0.0",
  org.eclipse.kura.linux.udev;version="1.0.0",
- org.eclipse.kura.message;version="1.0.0",
+ org.eclipse.kura.message;version="1.0.1",
```
MMaiero: @dwoodard1 The change in api determined an overloading of KuraTopic method. In my opinion, this implies that the api version for the message package should be increased to 1.1.0.

dwoodard1:
@MMaiero Updated, please review.

PR #720, opened 2016-11-04
Fix wifi issues 2.1.0

ibinshtok:
@MMaiero can you take a look at this PR, please?


kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/linux/util/LinuxProcessUtil.java
```diff
@@ -351,4 +393,22 @@ public static String getInputStreamAsString(InputStream stream) throws IOExcepti
         }
         return sb.toString();
     }
+    
+    public static boolean waitProcess(int pid, long poll, long timeout) {
```
MMaiero: @ibinshtok What is the difference between this waitProcess and the one in PppLinux?

kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/linux/util/LinuxProcessUtil.java
```diff
@@ -268,6 +268,48 @@ public static int getPid(String command, String[] tokens) throws Exception {
             }
         }
     }
+    
+    public static int getPid(int pid) throws Exception {
```
MMaiero: wouldn't be better to have this method named as isProcessAlive(int pid) or something related?

kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/linux/util/LinuxProcessUtil.java
```diff
@@ -351,4 +393,22 @@ public static String getInputStreamAsString(InputStream stream) throws IOExcepti
...
     }
+    
+    public static boolean waitProcess(int pid, long poll, long timeout) {
+        boolean exists = false;
+        try {
+            final long startTime = System.currentTimeMillis();
+            long now;
+            do {
+                Thread.sleep(poll);
+                exists = (getPid(pid) >= 0) ? true : false;
```
MMaiero: maybe we could use the code in PppLinux and have `exists = LinuxProcessUtil.stop(pid);`

kura/org.eclipse.kura.linux.net/src/main/java/org/eclipse/kura/linux/net/wifi/WpaSupplicantManager.java
```diff
@@ -161,28 +142,53 @@ public static boolean isTempRunning() throws KuraException {
...
-            if (cmd != null && !cmd.isEmpty()) {
-                LinuxProcessUtil.start(cmd);
-                if (ifaceName != null) {
-                    LinuxNetworkUtil.disableInterface(ifaceName);
+            if (!s_isIntelEdison) {
+                int pid = getPid(ifaceName);
+                if (pid >= 0) {
+                    s_logger.info("stopping wpa_suplicant for the {} interface, pid={}", ifaceName, pid);
+
+                    boolean exists = LinuxProcessUtil.stop(pid);
```
MMaiero: this part of code is the same as in HostapdManager.java, but also in PppLinux.disconnect
I think that maybe is better to have all this code centralized in LinuxProcessUtil or other utility class.

dwoodard1:
@MMaiero This was WiFi hotfix that was included in the v2.0.2 release. @ibinshtok back ported it for the upcoming v2.1.0 release.

kura/org.eclipse.kura.net.admin/src/main/java/org/eclipse/kura/net/admin/NetworkAdminServiceImpl.java
```diff
@@ -945,13 +946,48 @@ public void enableInterface(String interfaceName, boolean dhcp) throws KuraExcep
         try {
             NetInterfaceType type = LinuxNetworkUtil.getType(interfaceName);
 
-            if (!LinuxNetworkUtil.hasAddress(interfaceName)
-                    || type == NetInterfaceType.WIFI && !LinuxNetworkUtil.isLinkUp(interfaceName)) {
+            NetInterfaceStatus status = NetInterfaceStatus.netIPv4StatusUnknown;
+            WifiMode wifiMode = WifiMode.UNKNOWN;
+            WifiConfig wifiConfig = null;
+            WifiInterfaceState wifiInterfaceState = null;
+            if (type == NetInterfaceType.WIFI) {
```
MMaiero: @ibinshtok I think this section of code should be simplified/split into different methods: it is hard to read/debug.

kura/org.eclipse.kura.net.admin/src/main/java/org/eclipse/kura/net/admin/NetworkAdminServiceImpl.java
```diff
@@ -1539,7 +1575,7 @@ private void wifiModeWait(String ifaceName, WifiMode mode, int tout) {
     // Private Methods
     //
     // ----------------------------------------------------------------
-    private void enableWifiInterface(String ifaceName) throws KuraException {
+    private void enableWifiInterface(String ifaceName, NetInterfaceStatus status, WifiMode wifiMode, WifiConfig wifiConfig) throws KuraException {
```
MMaiero: @ibinshtok Could this signature be simplified? Maybe it's possible to use the wifiMode inside the wifiConfig. In this way we could promote, a bit, the readability.

MMaiero:
@dwoodard1 Ok, so it arises another question: do we have to increase the bundles versions, since those changes were applied in 2.0.2? I think we shouldn't if we want to keep the code in 2.0.2 and this is only a code update. Otherwise, if we plan to apply further changes, @ibinshtok version update is absolutely ok.

dwoodard1:
I think the bundle versions have to be updated. The v2.0.2 bundles didn't have the manifest changes introduced in the release-2.1.0 branch. Also, the release build will fail if we don't update the bundle versions.

MMaiero:
@pierantoniomerlino Could you perform some checks on gateways to double check that the changes made are correct?
Thanks

kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/linux/util/LinuxProcessUtil.java
MMaiero: since this method is back ported from a previous release, and is public, I would leave it as is to not break something

kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/linux/util/LinuxProcessUtil.java
MMaiero: From the refactoring, the result is that the waitProcess in PppLinux is not needed

ibinshtok:
@MMaiero I've tested changes you made it it works. I think you can merge this PR.

pierantoniomerlino:
@ibinshtok @MMaiero For me is OK.

MMaiero:
Ok, thanks. Merging

PR #735, opened 2016-11-09
Structure for test projects & dependency updates

markoer:
added new dependencies: mockito, objenesis
updated junit and hamcrest versions

Signed-off-by: markoer <marko.erker@comtrade.com>

### DIFF
--- a/kura/manifest_pom.xml
+++ b/kura/manifest_pom.xml
@@ -73,6 +73,7 @@
         <module>org.eclipse.kura.windows.command</module>
         <module>org.eclipse.kura.windows.clock</module>
         <module>examples</module>
+        <module>test</module>
     </modules>
 
     <properties>
@@ -117,7 +118,7 @@
                 <repository>
                     <id>p2-repo-equinox_3.8.1</id>
                     <layout>p2</layout>
-                    <url>file:///${kura.basedir}/target-definition/equinox_3.8.1/repository/</url>
+                    <url>file:////${kura.basedir}/target-definition/equinox_3.8.1/repository/</url>
                 </repository>
             </repositories>
         </profile>
@@ -133,7 +134,7 @@
                 <repository>
                     <id>p2-repo-equinox_3.8.1</id>
                     <layout>p2</layout>
-                    <url>file:///${kura.basedir}/target-definition/equinox_3.8.1/repository/</url>
+                    <url>file:////${kura.basedir}/target-definition/equinox_3.8.1/repository/</url>
                 </repository>
             </repositories>
         </profile>
@@ -188,7 +189,7 @@
         <repository>
             <id>p2-repo-common</id>
             <layout>p2</layout>
-            <url>file:///${kura.basedir}/target-definition/common/repository/</url>
+            <url>file:////${kura.basedir}/target-definition/common/repository/</url>
         </repository>
         <repository>
             <id>kura_addons</id>
--- a/kura/test/org.eclipse.kura.core.testutil/META-INF/MANIFEST.MF
+++ b/kura/test/org.eclipse.kura.core.testutil/META-INF/MANIFEST.MF
@@ -1,0 +1,11 @@
+Manifest-Version: 1.0
+Bundle-ManifestVersion: 2
+Bundle-Name: org.eclipse.kura.core.testutil
+Bundle-SymbolicName: org.eclipse.kura.core.testutil;singleton:=true
+Bundle-Version: 1.0.0.qualifier
+Bundle-Vendor: Eclipse Kura
+Bundle-RequiredExecutionEnvironment: JavaSE-1.6
+Import-Package: org.slf4j;version="1.6.4"
+Bundle-ActivationPolicy: lazy
+Comment: Unit testing utilities
+Export-Package: org.eclipse.kura.core.testutil
--- a/kura/test/org.eclipse.kura.core.testutil/pom.xml
+++ b/kura/test/org.eclipse.kura.core.testutil/pom.xml
@@ -1,0 +1,29 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+
+Copyright (c) 2016 Eurotech and/or its affiliates and others
+
+  All rights reserved. This program and the accompanying materials
+  are made available under the terms of the Eclipse Public License v1.0
+  which accompanies this distribution, and is available at
+  http://www.eclipse.org/legal/epl-v10.html
+
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
+        <groupId>org.eclipse.kura</groupId>
+        <artifactId>test</artifactId>
+        <version>2.1.0-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>org.eclipse.kura.core.testutil</artifactId>
+    <packaging>eclipse-plugin</packaging>
+    <version>1.0.0-SNAPSHOT</version>
+
+    <properties>
+        <kura.basedir>${project.basedir}/../..</kura.basedir>
+    </properties>
+</project>
--- a/kura/test/org.eclipse.kura.core.testutil/src/main/java/org/eclipse/kura/core/testutil/TestUtil.java
+++ b/kura/test/org.eclipse.kura.core.testutil/src/main/java/org/eclipse/kura/core/testutil/TestUtil.java
@@ -1,0 +1,113 @@
+/*******************************************************************************
+ * Copyright (c) 2016 Eurotech and/or its affiliates and others
+ *
+ *   All rights reserved. This program and the accompanying materials
+ *   are made available under the terms of the Eclipse Public License v1.0
+ *   which accompanies this distribution, and is available at
+ *   http://www.eclipse.org/legal/epl-v10.html
+ ******************************************************************************/
+package org.eclipse.kura.core.testutil;
+
+import java.lang.reflect.Field;
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class TestUtil {
+
+    private static final Logger logger = LoggerFactory.getLogger(TestUtil.class);
+
+    private static Field getField(Object svc, String fieldName) throws NoSuchFieldException {
+        Field field = null;
+        Class clazz = svc.getClass();
+        while (!(clazz == Object.class || field != null)) {
+            try {
+                field = clazz.getDeclaredField(fieldName);
+                break;
+            } catch (NoSuchFieldException e) {
+                // don't worry about it, here
+            }
+            clazz = clazz.getSuperclass();
+        }
+
+        if (field == null) {
+            throw new NoSuchFieldException(String.format("Field not found: %s", fieldName));
+        }
+
+        return field;
+    }
+
+    public static Object getFieldValue(Object svc, String fieldName) throws NoSuchFieldException {
+        Object result = null;
+
+        Field field = getField(svc, fieldName);
+
+        try {
+            field.setAccessible(true);
+            result = field.get(svc);
+        } catch (IllegalArgumentException e) {
+            logger.warn(e.getMessage(), e);
+        } catch (IllegalAccessException e) {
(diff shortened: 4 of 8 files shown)


cdealti:
@markoer you don't need to delete and recreate the PR due to my review.
Just fix and rebase your pull request local branch and update the PR with ``git push -f origin test_structure``

kura/test/org.eclipse.kura.core.testutil/META-INF/MANIFEST.MF
```diff
@@ -0,0 +1,16 @@
+Manifest-Version: 1.0
+Bundle-ManifestVersion: 2
+Bundle-Name: org.eclipse.kura.core.testutil
+Bundle-SymbolicName: org.eclipse.kura.core.testutil;singleton:=true
+Bundle-Version: 1.0.0.qualifier
+Bundle-Vendor: Eclipse Kura
+Bundle-RequiredExecutionEnvironment: JavaSE-1.6
+Import-Package: javax.comm,
+ junit.framework,
```
ctron: Shouldn't we got for `org.junit` instead of this one?

kura/test/org.eclipse.kura.core.testutil/META-INF/MANIFEST.MF
```diff
@@ -0,0 +1,16 @@
...
+ org.junit,
+ org.slf4j;version="1.6.4"
+Bundle-ActivationPolicy: lazy
+Comment: Unit testing utilities
+Require-Bundle: org.junit;bundle-version="4.12.0",
```
ctron: There is only one case where a bundle import may be necessary. Sometimes Tycho Surefire gets confused when the test suite is actually a fragment. Then it **may** be necessary to resort to bundle imports. But in general package imports should be preferred!

kura/test/org.eclipse.kura.core.testutil/about_files/epl-v10.html
```diff
@@ -0,0 +1,261 @@
+<?xml version="1.0" encoding="ISO-8859-1" ?>
```
ctron: That file is only needed for products, not for bundles.

kura/test/org.eclipse.kura.core.testutil/pom.xml
```diff
@@ -0,0 +1,32 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+
+    Copyright (c) 2011, 2016 Eurotech and/or its affiliates
```
ctron: Please correct the year. Unless this file was created in 2011.

kura/test/org.eclipse.kura.core.testutil/src/main/java/org/eclipse/kura/core/testutil/TestUtil.java
```diff
@@ -0,0 +1,114 @@
...
+            logger.warn(e.getMessage(), e);
+        }
+
+        return result;
+    }
+
+    private static Method getMethod(Object svc, String methodName, Class... paramTypes) throws NoSuchMethodException {
+        Method method = null;
+        Class<?> clazz = svc.getClass();
+        hwile: while (!(clazz == Object.class || method != null)) {
```
ctron: typo: "hwile"

kura/test/org.eclipse.kura.core.testutil/src/main/java/org/eclipse/kura/core/testutil/TestUtil.java
```diff
@@ -0,0 +1,114 @@
...
+        } catch (IllegalAccessException e) {
+            logger.warn(e.getMessage(), e);
+        } catch (NoSuchFieldException e) {
+            logger.warn(e.getMessage(), e);
+        }
+
+        return result;
+    }
+
+    private static Method getMethod(Object svc, String methodName, Class... paramTypes) throws NoSuchMethodException {
```
ctron: Why not simply use Class#getMethod? [1]

[1] https://docs.oracle.com/javase/7/docs/api/java/lang/Class.html#getMethod%28java.lang.String,%20java.lang.Class...%29

markoer:
I so hoped this wouldn't ruin my peaceful weekend :(

kura/manifest_pom.xml
```diff
@@ -117,7 +118,7 @@
                 <repository>
                     <id>p2-repo-equinox_3.8.1</id>
                     <layout>p2</layout>
-                    <url>file:///${kura.basedir}/target-definition/equinox_3.8.1/repository/</url>
+                    <url>file:////${kura.basedir}/target-definition/equinox_3.8.1/repository/</url>
```
markoer: It doesn't work with only 3, so, yes, it looks like it...

kura/test/org.eclipse.kura.core.testutil/META-INF/MANIFEST.MF
markoer: OK

kura/test/org.eclipse.kura.core.testutil/pom.xml
```diff
@@ -0,0 +1,32 @@
...
+
+     All rights reserved. This program and the accompanying materials
+     are made available under the terms of the Eclipse Public License v1.0
+     which accompanies this distribution, and is available at
+     http://www.eclipse.org/legal/epl-v10.html
+
+    Contributors:
+      Eurotech
+
```
markoer: OK

kura/test/org.eclipse.kura.core.testutil/src/main/java/org/eclipse/kura/core/testutil/TestUtil.java
```diff
@@ -0,0 +1,114 @@
...
+            try {
+                field = clazz.getDeclaredField(fieldName);
+                break;
+            } catch (NoSuchFieldException e) {
+            }
+            clazz = clazz.getSuperclass();
+        }
+
+        if (field == null) {
+            throw new NoSuchFieldException("Field not found");
```
markoer: OK...

kura/test/org.eclipse.kura.core.testutil/src/main/java/org/eclipse/kura/core/testutil/TestUtil.java
```diff
@@ -0,0 +1,114 @@
...
+        }
+
+        return result;
+    }
+
+    private static Method getMethod(Object svc, String methodName, Class... paramTypes) throws NoSuchMethodException {
+        Method method = null;
+        Class<?> clazz = svc.getClass();
+        hwile: while (!(clazz == Object.class || method != null)) {
+            // method = clazz.getDeclaredMethod(methodName);
```
markoer: why not...

kura/test/org.eclipse.kura.core.testutil/src/main/java/org/eclipse/kura/core/testutil/TestUtil.java
```diff
@@ -0,0 +1,114 @@
...
+
+    public static Object invokePrivate(Object svc, String methodName, Object... params) throws Throwable {
+        Method method = getMethod(svc, methodName);
+
+        method.setAccessible(true);
+
+        try {
+            Object result = method.invoke(svc, params);
+            return result;
+        } catch (IllegalAccessException e) {
```
markoer: yes
some are harmless and some actually report a problem

kura/test/pom.xml
```diff
@@ -0,0 +1,99 @@
...
+            </plugin>
+        </plugins>
+    </build>
+
+    <properties>
+        <kura.basedir>${project.basedir}/..</kura.basedir>
+    </properties>
+
+    <modules>
+<!--
```
markoer: remove what? modules or properties?

target-platform/p2-repo-common/pom.xml
```diff
@@ -221,6 +221,23 @@
...
+									<groupId>org.mockito</groupId>
+									<artifactId>mockito-core</artifactId>
+									<version>${mockito-core.version}</version>
+								</artifactItem>
+								<artifactItem>
+									<groupId>org.objenesis</groupId>
+									<artifactId>objenesis</artifactId>
+									<version>${objenesis.version}</version>
+								</artifactItem>
+<!--
```
markoer: not really sure why, but OK

kura/test/org.eclipse.kura.core.testutil/META-INF/MANIFEST.MF
markoer: will clean up...

kura/test/org.eclipse.kura.core.testutil/about_files/epl-v10.html
markoer: good to know, will remove it

kura/test/org.eclipse.kura.core.testutil/pom.xml
markoer: it's a copy of a file created then :)
will replace this header

kura/test/org.eclipse.kura.core.testutil/src/main/java/org/eclipse/kura/core/testutil/TestUtil.java
markoer: getMethod only returns public methods which this class has no interest whatsoever in. getDeclaredMethod is the closest that might do, but it doesn't since it requires parameter types to be specified, if method takes them. And the plan is not to complicate invocation where only one method of that name is available.

kura/test/org.eclipse.kura.core.testutil/src/main/java/org/eclipse/kura/core/testutil/TestUtil.java
markoer: Intentional, as is e.g. clazz from one line above: 'while' is a reserved word and cannot be used as a label.

kura/test/org.eclipse.kura.core.testutil/src/main/java/org/eclipse/kura/core/testutil/TestUtil.java
ctron: Understood. And why do you need to call private methods?

kura/test/org.eclipse.kura.core.testutil/src/main/java/org/eclipse/kura/core/testutil/TestUtil.java
ctron: I am just trying to understand what you are doing here. To meed this all seems totally overcomplicated.

Why not use a simple `equals` comparison and simple return the method:

```java
Class<?> clazz = svc.getClass();

while (clazz != Object.class) {
    Method[] methods = clazz.getDeclaredMethods();
    for (Method m : methods) {
        if (m.getName().equals(methodName)) {
            return m;
        }
    }
    clazz = clazz.getSuperclass();
}

throw new NoSuchMethodException(String.format("Method not found: %s", methodName));
```

With Java 8 this may still be broken since interfaces can have methods as well, so you would need to iterate over all interfaces in addition. Checking which ones you already traversed.

So do you really need this. Or can't you at least take some library which already implements that?

kura/test/org.eclipse.kura.core.testutil/src/main/java/org/eclipse/kura/core/testutil/TestUtil.java
markoer: To test them...

kura/test/org.eclipse.kura.core.testutil/src/main/java/org/eclipse/kura/core/testutil/TestUtil.java
markoer: You will understand my reasons with the next pull request when this method is extended. It is not enough to simply return the method there. Parameters will be checked after a potential method is found.
And as far as I know, Kura is Java 6.

ctron: So maybe we should wait reviewing this PR until it is complete?

Kura is Java 8 -> https://github.com/eclipse/kura/blob/develop/README.md#system-requirements

cdealti:
@ctron can you please dismiss your review so that we can merge this PR?
I know that @markoer will push more stuff in the testing area and it's likely this first contribution can be later improved. I'd like to have this merged today.

@markoer at the end of the review please consider squashing your commits into a single commit and rebase your branch against upstream/develop. There must not be merge commits in your pull request.
About your commits messages, please follow these [guidelines](https://dev.eclipse.org/mhonarc/lists/kura-dev/msg00689.html)

cdealti:
@markoer the last two commits shouldn't be there (I don't believe you touched the files of a12665b).
Can you please try rebasing again?

PR #745, opened 2016-11-10
Solved possible IllegalArgumentException due to null pid

MMaiero:
In slow devices it could happen that the pid is tracked but the map that maps the
pid with a servicePid has not the correct entry. This could lead to an
IllegalArgumentException in org.eclipse.equinox.internal.cm.ConfigurationAdminImpl.checkPID


MMaiero:
@pierantoniomerlino Could you check also this PR?

PR #758, opened 2016-11-15
Small build fixes to include wires

MMaiero:
Implies changes in manifest-pom and web2 bundles to build and include wires as
dependencies.


MMaiero:
@ctron Since Kura 3.0.0 will go for Java 1.8 I think we need to relax the Travis bree check. Cloud you do it?

ctron:
There is no need to relax the BREE step. Just declare the correct BREE on the bundle.

ctron:
Let's discuss this during the project call today.

ctron:
I think setting the proper BREE should be enough.

PR #759, opened 2016-11-15
Changes to include platform-dependent jdk.dio native libs in installers 

MMaiero:
The change affects distrib.
Only the specific native library will be included for each platform and architecture.


ctron:
@MMaiero Could you explain what the change should do

PR #772, opened 2016-11-16
Fix a few simple build warnings

ctron:


cdealti:
Good example of what should not require a PR IMHO.

PR #775, opened 2016-11-17
Fix the build & switch to Equinox 3.11.1

### LINKED ISSUE #741: Remove MaxPermSize argument
Every time I start up Kura I get a reminder that MaxPermSize is deprecated:

```
[me@my ~]# /opt/eclipse/kura/bin/start_kura.sh 
OpenJDK Zero VM warning: ignoring option MaxPermSize=64m; support was removed in 8.0
```

I think it time to remove this for Kura 3.0

### LINKED ISSUE #769: Version problems for org.eclipse.kura.api
Currently both `release-2.1.0` and `develop` declare version `1.0.10` for `org.eclipse.kura.api`.

However the content is actually different and the "release" branch is missing the wires additions.

This causes the problem locally when you are working with both versions that Maven will choose 1.0.10 over 1.0.10-SNAPSHOT, so the Web UI module doesn't compile anymore.

IMHO the version for `org.eclipse.kura.api` should be `1.1.0` now.

ctron:
This change:
 * **Fixes the build**
 * Switches to Equinox 3.11.1
 * Fixes #769 in order to compile again
 * Fixes #741 as Java 8 is now the minimum requirement

I am pretty sure there are many things which can be improved. But I vote for fixing the build first and improving after merging the PR.


target-platform/p2-repo-equinox_3.11.1/pom.xml
```diff
@@ -0,0 +1,168 @@
...
+				</executions>
+				<configuration>
+					<compress>false</compress>
+					<!-- additional arguments example <additionalArgs>-configs win32.win32.x86</additionalArgs> -->
+				</configuration>
+			</plugin>
+			<plugin>
+				<artifactId>maven-antrun-plugin</artifactId>
+                <version>1.8</version>
+				<executions>
```
amitjoy: I believe the XML Formatter Profile is not configured properly.

ctron: There is none. As this always messes up files. I wish those files would have a common ways of dealing with this. If you enable "special characters view" you can see the mess these files are in.

@MMaiero  @cdealti  So if this is the only problem I would like to go ahead and merge this

ctron:
@amitjoy whitespaces are fixed ... just in case you haven't looked at the actual whitespacing of those files:

ws

amitjoy:
👍

PR #777, opened 2016-11-17
Fixed serialization issue introduced by PR #704

MMaiero:
Closes #776 

Signed-off-by: MMaiero <matteo.maiero@eurotech.com>

Issue #776, opened 2016-11-17
Serialization issue in cloud window


MMaiero:
This issue was introduced by PR #704 and is due to the fact that the `GwtCloudConnectionState` enum cannot be serialized between GWT server and client part.
The web ui will end up with the following error:
<img width="1184" alt="schermata 2016-11-17 alle 11 29 12" src="https://cloud.githubusercontent.com/assets/10448126/20385992/ed9bfd8e-acb9-11e6-9136-9942b12cbe68.png">

while the kura.log file registers the following exception:

``` 
2016-11-17 10:28:40,797 [qtp8908375-23] WARN  / - Exception while dispatching incoming RPC call
com.google.gwt.user.client.rpc.SerializationException: Type 'org.eclipse.kura.web.shared.model.GwtCloudConnectionState' was not included in the set of types which can be serialized by this SerializationPolicy or its Class object could not be loaded. For security purposes, this type will not be serialized.: instance = DISCONNECTED
	at com.google.gwt.user.server.rpc.impl.ServerSerializationStreamWriter.serialize(ServerSerializationStreamWriter.java:667)
	at com.google.gwt.user.client.rpc.impl.AbstractSerializationStreamWriter.writeObject(AbstractSerializationStreamWriter.java:130)
	at com.google.gwt.user.client.rpc.core.java.util.Map_CustomFieldSerializerBase.serialize(Map_CustomFieldSerializerBase.java:53)
	at com.google.gwt.user.client.rpc.core.java.util.HashMap_CustomFieldSerializer.serialize(HashMap_CustomFieldSerializer.java:39)
	at com.google.gwt.user.client.rpc.core.java.util.HashMap_CustomFieldSerializer.serializeInstance(HashMap_CustomFieldSerializer.java:51)
	at com.google.gwt.user.client.rpc.core.java.util.HashMap_CustomFieldSerializer.serializeInstance(HashMap_CustomFieldSerializer.java:28)
	at com.google.gwt.user.server.rpc.impl.ServerSerializationStreamWriter.serializeImpl(ServerSerializationStreamWriter.java:789)
	at com.google.gwt.user.server.rpc.impl.ServerSerializationStreamWriter.serialize(ServerSerializationStreamWriter.java:669)
	at com.google.gwt.user.client.rpc.impl.AbstractSerializationStreamWriter.writeObject(AbstractSerializationStreamWriter.java:130)
[...]

PR #780, opened 2016-11-17
#776: Fix enum issue with GWT

ctron:
This change works around a bug in the GWT compiler which sometimes
does not pick up a type for serialization.

Also see: http://stackoverflow.com/a/992745/222044

This obsoletes #777

### DIFF
--- a/kura/org.eclipse.kura.web2/pom.xml
+++ b/kura/org.eclipse.kura.web2/pom.xml
@@ -252,7 +252,7 @@
 				<artifactId>maven-compiler-plugin</artifactId>
 				<version>3.3</version>
 				<configuration>
-					<source>1.6</source>
+					<source>1.7</source>
 					<target>1.7</target>
 				</configuration>
 			</plugin>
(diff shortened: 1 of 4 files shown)


kura/org.eclipse.kura.web2/pom.xml
```diff
@@ -252,7 +252,7 @@
 				<artifactId>maven-compiler-plugin</artifactId>
 				<version>3.3</version>
 				<configuration>
-					<source>1.6</source>
+					<source>1.7</source>
```
MMaiero: Why you need to have 1.7 here?

ctron: It is a leftover from a test. But binary code already is 7. So that setting doesn't change anything.

MMaiero: You mean due to the fact that the target is 1.7?

MMaiero: According to this [document](http://docs.oracle.com/javase/8/docs/technotes/tools/windows/javac.html) the -target is not so relevant if -source is specified

ctron: How do you come to this conclusion from this document? What I read is that the source is validated against Java 6 rules:

> Specifies the version of source code accepted. The following values for release are allowed:
> …
> No language changes were introduced in Java SE 6. However, encoding errors in source files are now reported as errors instead of warnings as in earlier releases of Java Platform, Standard Edition.

Where the resulting byte code would be Java 7:

> Generates class files that target a specified release of the virtual machine. Class files will run on the specified target and on later releases, but not on earlier releases of the JVM. 


Which, in that combination, doesn't make any sense to me.

Issue #778, opened 2016-11-17
[Net] Refactor needed in org.eclipse.kura.core.net


cdealti:
This includes:

* Parameter validation
* Refactor to remove code duplication
* Consistent byte order 

It will be submitted by @djurod

cdealti:
@markoer @djurod can we close it?

djurod:
This was resolved with PR #781

It can be closed

PR #781, opened 2016-11-17
issue #778: Refactored NetworkUtil class

djurod:
Changes:
* Removed code duplication.
* Added checking of inputs: a KuraException is thrown in case of invalid
  inputs.
* Fixed the byte order in conversion of IP addresses.
* Code cleanup.

### DIFF
--- a/kura/org.eclipse.kura.core.net/src/main/java/org/eclipse/kura/core/net/util/NetworkUtil.java
+++ b/kura/org.eclipse.kura.core.net/src/main/java/org/eclipse/kura/core/net/util/NetworkUtil.java
@@ -11,72 +11,47 @@
  *******************************************************************************/
 package org.eclipse.kura.core.net.util;
 
-import java.util.StringTokenizer;
-
-import org.eclipse.kura.KuraErrorCode;
-import org.eclipse.kura.KuraException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-public class NetworkUtil {
+public final class NetworkUtil {
 
     private static final Logger s_logger = LoggerFactory.getLogger(NetworkUtil.class);
+    
+    private NetworkUtil() {
+    }
 
     public static String calculateNetwork(String ipAddress, String netmask) {
-        int ipAddressValue = 0;
-        int netmaskValue = 0;
-
-        StringTokenizer st = new StringTokenizer(ipAddress, ".");
-        for (int i = 24; i >= 0; i -= 8) {
-            ipAddressValue = ipAddressValue | Integer.parseInt(st.nextToken()) << i;
-        }
-
-        st = new StringTokenizer(netmask, ".");
-        for (int i = 24; i >= 0; i -= 8) {
-            netmaskValue = netmaskValue | Integer.parseInt(st.nextToken()) << i;
-        }
+        int ipAddressValue = NetworkUtil.convertIp4Address(ipAddress);
+        int netmaskValue = NetworkUtil.convertIp4Address(netmask);
 
         int network = ipAddressValue & netmaskValue;
         return dottedQuad(network);
     }
 
     public static String calculateBroadcast(String ipAddress, String netmask) {
-        int ipAddressValue = 0;
-        int netmaskValue = 0;
-
-        StringTokenizer st = new StringTokenizer(ipAddress, ".");
-        for (int i = 24; i >= 0; i -= 8) {
-            ipAddressValue = ipAddressValue | Integer.parseInt(st.nextToken()) << i;
-        }
-
-        st = new StringTokenizer(netmask, ".");
-        for (int i = 24; i >= 0; i -= 8) {
-            netmaskValue = netmaskValue | Integer.parseInt(st.nextToken()) << i;
-        }
+        int ipAddressValue = NetworkUtil.convertIp4Address(ipAddress);
+        int netmaskValue = NetworkUtil.convertIp4Address(netmask);
 
         int network = ipAddressValue | ~netmaskValue;
         return dottedQuad(network);
     }
 
-    public static String getNetmaskStringForm(int prefix) throws KuraException {
-        if (prefix >= 0 && prefix <= 32) {
+    public static String getNetmaskStringForm(int prefix) {
+        if (prefix >= 1 && prefix <= 32) {
             int mask = ~((1 << 32 - prefix) - 1);
             return dottedQuad(mask);
         } else {
-            throw new KuraException(KuraErrorCode.INTERNAL_ERROR, "invalid prefix ");
+            throw new IllegalArgumentException("prefix is invalid: " + Integer.toString(prefix));
         }
     }
 
-    public static short getNetmaskShortForm(String netmask) throws KuraException {
+    public static short getNetmaskShortForm(String netmask) {
         if (netmask == null) {
-            throw new KuraException(KuraErrorCode.INTERNAL_ERROR, "netmask is null");
+            throw new IllegalArgumentException("netmask is null");
         }
 
-        int netmaskValue = 0;
-        StringTokenizer st = new StringTokenizer(netmask, ".");
-        for (int i = 24; i >= 0; i -= 8) {
-            netmaskValue = netmaskValue | Integer.parseInt(st.nextToken()) << i;
-        }
+        int netmaskValue = NetworkUtil.convertIp4Address(netmask);
 
         boolean hitZero = false;
         int displayMask = 1 << 31;
@@ -88,7 +63,7 @@ public class NetworkUtil {
             } else {
                 if (hitZero) {
                     s_logger.error("received invalid mask: " + netmask);
-                    throw new KuraException(KuraErrorCode.INTERNAL_ERROR, "received invalid mask: " + netmask);
+                    throw new IllegalArgumentException("netmask is invalid: " + netmask);
                 }
 
                 count++;
@@ -101,19 +76,27 @@ public class NetworkUtil {
     }
 
     public static String dottedQuad(int ip) {
-        StringBuffer sb = new StringBuffer(15);
-        for (int shift = 24; shift > 0; shift -= 8) {
-            // process 3 bytes, from high order byte down.
-            sb.append(Integer.toString(ip >>> shift & 0xff));
-            sb.append('.');
+        String[] items = new String[4];
+        for (int i = 3; i >= 0; i--) {
+            int value = ip & 0xFF;
+            items[i] = Integer.toString(value);
+            ip = ip >>> 8;
         }
 
-        sb.append(Integer.toString(ip & 0xff));
-        return sb.toString();
+        return String.join(".", items);
     }
 
     public static int convertIp4Address(String ipAddress) {
+        if (ipAddress == null) {
+            throw new IllegalArgumentException("ipAddress is null");
+        }
+
         String[] splitIpAddress = ipAddress.split("\\.");
+
+        if (splitIpAddress.length != 4) {
+            throw new IllegalArgumentException("ipAddress is invalid: " + ipAddress);
+        }
+
         short[] addressBytes = new short[4];
 
         for (int i = 0; i < 4; i++) {
@@ -125,8 +108,17 @@ public class NetworkUtil {
     }
 
     public static int packIp4AddressBytes(short[] bytes) {
+        if ((bytes == null) || (bytes.length != 4)) {
+            throw new IllegalArgumentException("bytes is null or invalid");
+        }
+
         int val = 0;
-        for (int i = 3; i >= 0; i--) {
+        for (int i = 0; i < 4; i++) {
+            if ((bytes[i] < 0) || (bytes[i] > 255)) {
+                throw new IllegalArgumentException(
+                        "bytes is invalid; value is out of range: " + Integer.toString(bytes[i]));
+            }
+
             val = val << 8;
             val |= bytes[i];
         }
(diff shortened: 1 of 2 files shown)


kura/org.eclipse.kura.core.net/src/main/java/org/eclipse/kura/core/net/util/NetworkUtil.java
```diff
@@ -101,19 +75,28 @@ public static short getNetmaskShortForm(String netmask) throws KuraException {
...
+        
+        String dottedValue = String.join(".", items);
+        return dottedValue;
     }
 
-    public static int convertIp4Address(String ipAddress) {
+    public static int convertIp4Address(String ipAddress) throws KuraException {
+        if (ipAddress == null) {
+            throw new KuraException(KuraErrorCode.INVALID_PARAMETER);
+        }
```
amitjoy: Null Parameters is a runtime problem and I believe it should be `KuraRuntimeException`. You can use `Preconditions` static factory methods container class for such conditions.

kura/org.eclipse.kura.core.net/src/main/java/org/eclipse/kura/core/net/util/NetworkUtil.java
```diff
@@ -101,19 +75,28 @@ public static short getNetmaskShortForm(String netmask) throws KuraException {
...
+        
         String[] splitIpAddress = ipAddress.split("\\.");
+        
+        if (splitIpAddress.length != 4) {
+            throw new KuraException(KuraErrorCode.INVALID_PARAMETER);
+        }
```
amitjoy: Same as before.

kura/org.eclipse.kura.core.net/src/main/java/org/eclipse/kura/core/net/util/NetworkUtil.java
```diff
@@ -124,72 +107,127 @@ public static int convertIp4Address(String ipAddress) {
         return NetworkUtil.packIp4AddressBytes(addressBytes);
     }
 
-    public static int packIp4AddressBytes(short[] bytes) {
+    public static int packIp4AddressBytes(short[] bytes) throws KuraException {
+        if ((bytes == null) || (bytes.length != 4)) {
+            throw new KuraException(KuraErrorCode.INVALID_PARAMETER);
+        }
```
amitjoy: Same here as well.

kura/org.eclipse.kura.core.net/src/main/java/org/eclipse/kura/core/net/util/NetworkUtil.java
```diff
@@ -124,72 +107,127 @@ public static int convertIp4Address(String ipAddress) {
...
+        
         int val = 0;
-        for (int i = 3; i >= 0; i--) {
+        for (int i = 0; i < 4; i++) {
+            if ((bytes[i] < 0) || (bytes[i] > 255)) {
+                throw new KuraException(KuraErrorCode.INVALID_PARAMETER);
+            }
```
amitjoy: --Same--

kura/org.eclipse.kura.core.net/src/main/java/org/eclipse/kura/core/net/util/NetworkUtil.java
```diff
@@ -124,72 +107,127 @@ public static int convertIp4Address(String ipAddress) {
...
-            retVal[i * 2 + 1] = (byte) Short.parseShort(sb.toString().substring(2, 4), 16);
+        for (int i = 0; i < 8; i++) {
+            try {
+                String octet = ip6Split[i];
+                int value = Integer.parseInt(octet, 16);
+                
+                if ((value < 0) || (value > 0xFFFF))
+                {
+                    throw new KuraException(KuraErrorCode.INVALID_PARAMETER);
+                }
```
amitjoy: It should be a runtime exception.

kura/org.eclipse.kura.core.net/src/main/java/org/eclipse/kura/core/net/util/NetworkUtil.java
```diff
@@ -124,72 +107,127 @@ public static int convertIp4Address(String ipAddress) {
...
+                
+                int k = i * 2;
+                retVal[k] = (byte) ((value >>> 8) & 0xFF);
+                retVal[k + 1] = (byte) (value & 0xFF);
+            } catch (NumberFormatException e) {
+                throw new KuraException(KuraErrorCode.INVALID_PARAMETER);
+            }
```
amitjoy: Same here. Runtime Exception.

kura/org.eclipse.kura.core.net/src/main/java/org/eclipse/kura/core/net/util/NetworkUtil.java
```diff
@@ -124,72 +107,127 @@ public static int convertIp4Address(String ipAddress) {
...
-        StringBuffer sb = new StringBuffer();
-        for (int i = 0; i < 16; i = i + 2) {
-            sb.append(Integer.toHexString(0xFF & bytes[i]));
-            sb.append(Integer.toHexString(0xFF & bytes[i + 1]));
-            if (i != 14) {
-                sb.append(":");
-            }
+    public static String convertIP6Address(byte[] bytes) throws KuraException {
+        if ((bytes == null) || (bytes.length != 16)) {
+            throw new KuraException(KuraErrorCode.INVALID_PARAMETER);
```
amitjoy: Runtime

kura/org.eclipse.kura.core.net/src/main/java/org/eclipse/kura/core/net/util/NetworkUtil.java
```diff
@@ -124,72 +107,127 @@ public static int convertIp4Address(String ipAddress) {
...
+        String ipAddress = String.join(":", items);
+        return ipAddress;
     }
 
-    public static String macToString(byte[] mac) {
-        StringBuilder sb = new StringBuilder();
+    public static String macToString(byte[] mac) throws KuraException {
+        if ((mac == null) || (mac.length != 6)) {
+            throw new KuraException(KuraErrorCode.INVALID_PARAMETER);
+        }
```
amitjoy: Runtime Exception as well

kura/org.eclipse.kura.core.net/src/main/java/org/eclipse/kura/core/net/util/NetworkUtil.java
```diff
@@ -124,72 +107,127 @@ public static int convertIp4Address(String ipAddress) {
...
+        
         byte[] bytes = new byte[6];
         for (int i = 0; i < 6; i++) {
-            bytes[i] = (byte) Integer.parseInt(st.nextToken(), 16);
+            String item = items[i];
+            if (item.isEmpty() || (item.length() > 2)) {
+                throw new KuraException(KuraErrorCode.INVALID_PARAMETER);
+            }
```
amitjoy: Runtime Exception

kura/org.eclipse.kura.core.net/src/main/java/org/eclipse/kura/core/net/util/NetworkUtil.java
```diff
@@ -124,72 +107,127 @@ public static int convertIp4Address(String ipAddress) {
...
+            
+            try {
+                bytes[i] = (byte) Integer.parseInt(items[i], 16);
+            } catch (NumberFormatException e) {
+                throw new KuraException(KuraErrorCode.INVALID_PARAMETER);
```
amitjoy: Same

kura/org.eclipse.kura.core.net/src/main/java/org/eclipse/kura/core/net/util/NetworkUtil.java
```diff
@@ -124,72 +107,127 @@ public static int convertIp4Address(String ipAddress) {
...
+        
+        String macString = String.join(":", items);
+        return macString;
     }
 
-    public static byte[] macToBytes(String mac) {
-        StringTokenizer st = new StringTokenizer(mac, ":");
+    public static byte[] macToBytes(String mac) throws KuraException {
+        if (mac == null) {
+            throw new KuraException(KuraErrorCode.INVALID_PARAMETER);
```
amitjoy: Same

kura/org.eclipse.kura.core.net/src/main/java/org/eclipse/kura/core/net/util/NetworkUtil.java
```diff
@@ -124,72 +107,127 @@ public static int convertIp4Address(String ipAddress) {
...
-            ;
-            String octet = ip6Split[i];
-            StringBuffer sb = new StringBuffer();
 
-            while (sb.length() + octet.length() < 4) {
-                sb.append("0");
-            }
-            sb.append(octet);
+        if (ip6Split.length != 8) {
+            throw new KuraException(KuraErrorCode.INVALID_PARAMETER);
```
amitjoy: Runtime

kura/org.eclipse.kura.core.net/src/main/java/org/eclipse/kura/core/net/util/NetworkUtil.java
```diff
@@ -72,11 +50,7 @@ public static short getNetmaskShortForm(String netmask) throws KuraException {
             throw new KuraException(KuraErrorCode.INTERNAL_ERROR, "netmask is null");
```
amitjoy: Needs to be `KuraRuntimeException`. There already exists an utility class `Preconditions` which performs all these necessary checks. You can reuse that.

kura/org.eclipse.kura.core.net/src/main/java/org/eclipse/kura/core/net/util/NetworkUtil.java
```diff
@@ -124,72 +107,127 @@ public static int convertIp4Address(String ipAddress) {
...
+            value = value >>> 8;
+        }
+        
+        return addressBytes;
     }
 
-    public static byte[] convertIP6Address(String fullFormIP6Address) {
+    public static byte[] convertIP6Address(String fullFormIP6Address) throws KuraException {
+        if (fullFormIP6Address == null) {
+            throw new KuraException(KuraErrorCode.INVALID_PARAMETER);
```
amitjoy: A runtime exceptional condition

kura/org.eclipse.kura.linux.net/src/main/java/org/eclipse/kura/linux/net/util/IWAPParser.java
```diff
@@ -34,7 +35,12 @@
     private IWSecuritySectionParser securityParser = null;
 
     public IWAPParser(String macAddressString) {
-        this.hardwareAddress = NetworkUtil.macToBytes(macAddressString);
+        try {
+            this.hardwareAddress = NetworkUtil.macToBytes(macAddressString);
+        } catch (KuraException e) {
+            // Hardware address is invalid
+            s_logger.info("Invalid hardware address");
```
amitjoy: error logger

kura/org.eclipse.kura.core.net/src/main/java/org/eclipse/kura/core/net/util/NetworkUtil.java
ctron: Or simply use standard java functionality Objects.requireNonNull()

kura/org.eclipse.kura.core.net/src/main/java/org/eclipse/kura/core/net/util/NetworkUtil.java
amitjoy: @cdealti Cool. No Problem. But I believe a null parameter check is a runtime problem which the client is not meant to handle to take precautions. And, that's why a null parameter check must incorporate an unchecked exception translation.

If you consider any null parameter check or any specific length check of an array or collection param, clients ain't meant to write precautionary codes to deal with such conditions. For a checked exception translation, we have to make sure that the exceptional conditions must necessitate clients to take precautions in case these conditions occur.

kura/org.eclipse.kura.core.net/src/main/java/org/eclipse/kura/core/net/util/NetworkUtil.java
amitjoy: `IllegalArgumentException` is also another possibility. This is also a perfect fit as an unchecked exception. But as a network operations is very low-level work, and the clients of such network operations reside at the high level, it would be good to throw High-level exceptions such as `KuraRuntimeException` or `KuraException` so that higher layers know about such exceptions.

If you believe that the clients of Network Operations are still at the low-level, we can use any relevant standard Java unchecked exceptions such as `IllegalArgumentException` or `NPE` and so on.

ctron: So IMHO it should be an unchecked exception in any case.

As for the rest I think it comes down to personal preference. As we agreed java.* exceptions are fine as well. If you use Kura*Exceptions then be specific and use proper error codes.

ctron:
+1 for:

> Add private constructors to the utility classes as it is not meant to be constructed. And also, you can make utility classes a non-overridable class with final modifier.

ctron:
And please use proper BREE validation. As this is cause for the build failure.

The best way to do this is to install all JDKs (1.6, 1.7 and 1.8) and register them with Eclipse. And you will get automatic error reporting on this.

kura/org.eclipse.kura.core.net/src/main/java/org/eclipse/kura/core/net/util/NetworkUtil.java
djurod: In a discussion with @cdealti it was decided to replace usage of KuraException in this class with IllegalArgumentException.

kura/org.eclipse.kura.core.net/src/main/java/org/eclipse/kura/core/net/util/NetworkUtil.java
djurod: As noted above, IllegalArgumentException will be used and NFE is also added as root cause (fixed in also other places).

kura/org.eclipse.kura.linux.net/src/main/java/org/eclipse/kura/linux/net/util/IWAPParser.java
djurod: With switching to use of IllegalArgumentException, changes in other classes are no longer needed. All changes in classes other than NetworkUtil will be reverted.

ctron:
+1

kura/org.eclipse.kura.core.net/src/main/java/org/eclipse/kura/core/net/util/NetworkUtil.java
```diff
@@ -125,73 +108,141 @@ public static int convertIp4Address(String ipAddress) {
...
+                bytes[i] = (byte) Integer.parseInt(items[i], 16);
+            } catch (NumberFormatException e) {
+                throw new IllegalArgumentException("mac is invalid: " + mac, e);
+            }
         }
 
         return bytes;
     }
+
+    private static String joinStrings(String separator, String[] items) {
```
amitjoy: You can reuse the `join(CharSequence, Iterable<?>)` method from `StringUtil`

ctron: Or just use Java 8

djurod: I did use Java 8, that is why the original build failed :)

@ctron I have seen a pull request (#775) that mentioned that Kura is switching to Java 8. Is it now allowed to use Java 8?

ctron: Yes it is!

djurod: Done

djurod:
I think all issues should now be resolved.

Issue #782, opened 2016-11-17
[Testing] Create unit tests for configuration service


markoer:
Add unit tests for configuration service implementation.

Test behavior of JUnit categories e.g. PublicTests (test API - or maybe all methods), PrivateTests (run only private implementation method tests), PublicOnlyTests (use only public API for setup with no private field modification).

cdealti:
@markoer can we close this?

markoer:
I suppose yes, once you decide what to do about #1075 (2 comments are open) and #1130 is merged.

PR #783, opened 2016-11-17
 #782: Added configuration service unit tests

markoer:
Service unit tests:
- new project
- updated test util with new method invocation
refactored configuration service for testability:
- added getDefaultProperties
added objenesis dependency


kura/test/org.eclipse.kura.core.configuration.test/src/test/java/org/eclipse/kura/core/configuration/ConfigurationServiceTest.java
```diff
@@ -0,0 +1,806 @@
...
+
+        ConfigurationServiceImpl cs = new ConfigurationServiceImpl();
+
+        Map<String, String> pids = (Map<String, String>) TestUtil.getFieldValue(cs, "m_servicePidByPid");
+        pids.put(pid, pid);
+
+        try {
+            cs.createFactoryConfiguration(factoryPid, pid, properties, takeSnapshot);
+
+            fail("Exception expected with existing pid value.");
```
ctron: Why aren't you declaring an expected Exception on the unit test itself my @Test(expected=?)?

markoer: more control

ctron: Which you need for what in this test case?

kura/test/org.eclipse.kura.core.configuration.test/META-INF/MANIFEST.MF
```diff
@@ -0,0 +1,24 @@
...
+ org.eclipse.kura.core.testutil,
+ org.eclipse.kura.crypto;version="1.2.0",
+ org.eclipse.kura.system;version="1.1.0",
+ org.junit,
+ org.mockito;version="1.10.19",
+ org.mockito.invocation;version="1.10.19",
+ org.mockito.stubbing;version="1.10.19",
+ org.osgi.framework,
+ org.osgi.service.cm,
+ org.osgi.service.component
```
markoer: will add

kura/test/org.eclipse.kura.core.configuration.test/META-INF/MANIFEST.MF
```diff
@@ -0,0 +1,24 @@
+Manifest-Version: 1.0
+Bundle-ManifestVersion: 2
+Bundle-Name: org.eclipse.kura.core.configuration.test
+Bundle-SymbolicName: org.eclipse.kura.core.configuration.test;singleton:=true
+Bundle-Version: 2.1.0.qualifier
+Bundle-Vendor: Eclipse Kura
+Bundle-RequiredExecutionEnvironment: JavaSE-1.6
+Import-Package: org.eclipse.kura;version="1.2.1",
+ org.eclipse.kura.configuration;version="1.1.0",
```
cdealti: Use semantic versioning for all org.eclipse.kura.* API packages

kura/test/org.eclipse.kura.core.configuration.test/META-INF/MANIFEST.MF
```diff
@@ -0,0 +1,24 @@
...
+ org.eclipse.kura.configuration.metatype;version="[1.0,2.0)",
+ org.eclipse.kura.core.configuration;version="[1.0,2.0)",
```
cdealti: We currently don't properly version Kura "internal" packages accordingly to Semantic Versioning so you should not rely on that. Maybe remove range for now (just set a minimum version for every import)

kura/test/org.eclipse.kura.core.configuration.test/pom.xml
```diff
@@ -0,0 +1,28 @@
...
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
+        <groupId>org.eclipse.kura</groupId>
+        <artifactId>test</artifactId>
+        <version>2.1.0-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>org.eclipse.kura.core.configuration.test</artifactId>
+    <packaging>eclipse-test-plugin</packaging>
```
cdealti: @markoer I believe we should explain that the initial idea was to provide both UT (under src/test/java) and IT (under src/main/java) and use both maven-surefire and tycho-surefire plugins.

This explains the ``eclipse-test-plugin`` packaging which is not yet needed by this first contribution.

kura/test/org.eclipse.kura.core.configuration.test/META-INF/MANIFEST.MF
cdealti: @markoer I'm confused if we really need a MANIFEST at this stage (see comment below for the pom file).
However, please take care of importing the org.eclipse.kura, org.eclipse.kura.configuration, org.eclipse.kura.crypto, org.eclipse.kura.system with proper ranges as described in the OSGi Semantic Versioning paper.

The  org.eclipse.kura.core.* packages are internal packages and we currently do not update their version properly so you should not rely on their version.
Do you really need all of them in the MANIFEST?

markoer: The dependencies on the Kura implementation packages are all necessary. But the package versions will be updated.

cdealti:
@markoer Can you please **rebase** this PR against the develop branch?

markoer:
No point in using this branch anymore due to too many conflicting changes. Will open a new one and hope it passes...

cdealti:
Pl

PR #799, opened 2016-11-18
Replace current process when starting Kura

ctron:
This change does replace the current process instead of forking another
instance and keeping the resources.

### DIFF
--- a/kura/distrib/src/main/ant/build_equinox_distrib.xml
+++ b/kura/distrib/src/main/ant/build_equinox_distrib.xml
@@ -1,6 +1,6 @@
 <!--
 
-    Copyright (c) 2011, 2016 Eurotech and others
+    Copyright (c) 2011, 2017 Eurotech and others
 
      All rights reserved. This program and the accompanying materials
      are made available under the terms of the Eclipse Public License v1.0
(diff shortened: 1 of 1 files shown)


dwoodard1:
Is this really necessary? My understanding is the potential saved resources is negligible. My concern is knowing which environment variables may be used when running with exec. We should catch this during our QA, but I am not sure it is worth the risk.

ctron:
What concerns with environment variables do you have? The process will have the same setup as the a child process would have.

dwoodard1:
Yes, my question is what environment gets used if called from exec. I am not an expert in this matter, just seems like potential problems for little benefit. Seems to be fixing something that isn't broken.

@cdealti You have more Linux experience, do you have any issues with this request?

ctron:
When a process is spawned it is forked and then exec'd. In this case it is just exec'd.

It makes things easier when handling the process as no new child process is being spawned.

ctron:
Maybe this [1] (section "Always 'exec' in Wrapper Scripts") helps a bit in understanding.

[1] http://www.projectatomic.io/docs/docker-image-author-guidance/

cdealti:
@dwoodard1 I'd give this a try. The chance is so minimal that it can be easily reverted in case of issues (unlikely). What do you think?

dwoodard1:
+1

Issue #779, opened 2016-11-17
[Testing] Unit Tests for org.eclipse.kura.core.net


cdealti:
It will be submitted by @djurod.
Requires #778.

cdealti:
@djurod can we close this?

djurod:
Not yet, there are still some pull requests pending. PR #1089 is still open and I at least one more PR is needed to include all written tests for core.net

cdealti:
@djurod can this be closed now?

djurod:
No, not yet. I still have to prepare a pull request (should be the last one for core.net) with the already written tests.

cdealti:
@markoer can you please check again with @djurod? Shall we postpone?

markoer:
The PR waits for @ibinshtok's decision regarding invocation of getByAddress(byte[]).

ibinshtok:
Closing this issue since associated PR#1336 was merged.

PR #800, opened 2016-11-18
Unit tests for NetworkUtil class

djurod:
issue #779: Prepared unit tests for NetworkUtil class from package org.eclipse.kura.core.net.util

### DIFF
--- a/kura/test/org.eclipse.kura.core.net.test/src/test/java/org/eclipse/kura/core/net/util/NetworkUtilTest.java
+++ b/kura/test/org.eclipse.kura.core.net.test/src/test/java/org/eclipse/kura/core/net/util/NetworkUtilTest.java
@@ -1,0 +1,489 @@
+/*******************************************************************************
+ * Copyright (c) 2016 Eurotech and/or its affiliates and others
+ *
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *******************************************************************************/
+package org.eclipse.kura.core.net.util;
+
+import static org.junit.Assert.assertArrayEquals;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.fail;
+
+import org.junit.FixMethodOrder;
+import org.junit.Test;
+import org.junit.runners.MethodSorters;
+
+@FixMethodOrder(MethodSorters.NAME_ASCENDING)
+public class NetworkUtilTest {
+
+    @Test
+    public void testCalculateNetwork() throws IllegalArgumentException {
+        String result = NetworkUtil.calculateNetwork("192.168.1.123", "255.255.255.0");
+        assertEquals("192.168.1.0", result);
+
+        result = NetworkUtil.calculateNetwork("10.100.250.1", "255.0.0.0");
+        assertEquals("10.0.0.0", result);
+    }
+
+    @Test(expected = IllegalArgumentException.class)
+    public void testCalculateNetworkNullAddress() {
+        NetworkUtil.calculateNetwork(null, "255.255.255.0");
+    }
+
+    @Test(expected = IllegalArgumentException.class)
+    public void testCalculateNetworkEmptyAddress() {
+        NetworkUtil.calculateNetwork("", "255.255.255.0");
+    }
+
+    @Test(expected = IllegalArgumentException.class)
+    public void testCalculateNetworkTooShortAddress() {
+        NetworkUtil.calculateNetwork("192.168.1", "255.255.255.0");
+    }
+
+    @Test(expected = IllegalArgumentException.class)
+    public void testCalculateNetworkTooLongAddress() {
+        NetworkUtil.calculateNetwork("192.168.1.123.1", "255.255.255.0");
+    }
+
+    @Test(expected = IllegalArgumentException.class)
+    public void testCalculateNetworkInvalidAddress() {
+        NetworkUtil.calculateNetwork("256.168.1.123", "255.255.255.0");
+    }
+
+    @Test(expected = IllegalArgumentException.class)
+    public void testCalculateNetworkNullSubnet() {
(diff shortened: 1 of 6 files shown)


kura/test/org.eclipse.kura.core.net.test/src/test/java/org/eclipse/kura/core/net/util/NetworkUtilTest.java
```diff
@@ -0,0 +1,799 @@
...
+        assertEquals(0xFF120000, result);
+
+        result = NetworkUtil.convertIp4Address("128.0.0.0");
+        assertEquals(0x80000000, result);
+
+        result = NetworkUtil.convertIp4Address("0.0.0.0");
+        assertEquals(0x00000000, result);
+    }
+
+    @Test
```
ctron: I still think it would be much clearer if it would be:

```java
@Test(expected=IllegalArgumentException.class)
public void testConvertIp4AddressNullAddress() {
     NetworkUtil.convertIp4Address(null);
}
```

djurodrljaca: I changed it locally and it worked as expected in Eclipse, but when I tried running the tests from command line (mvn clean test) all of the tests with @Test(expected=...) failed. I guess there is a problem in pom or in the manifest...

Any ideas how to solve this?

ctron: What is the error message?

djurod: There is no error message. The test build and runs, just that it reports that about 60 of the tests fail and all other pass. It looks like it just ignores the "(expected=...)" part of the annotation.

I will take a quick look at it tomorrow and if I don't manage to solve it quickly I will publish the build log here.

djurod: I found the problem. junit 4.12 was not added as a dependency in kura/test/pom.xml

djurod:
Any suggestions how to solve this:

> [ERROR] Failed to execute goal org.eclipse.tycho:tycho-surefire-plugin:0.26.0:test (default-test) on project org.eclipse.kura.core.net.test: No tests found. -> [Help 1]

I guess the problem here is that at least one test is expected in "src/main/java" by tyco surefire plugin (this is supposed to be reserved for integration tests), but I only have unit tests in "src/test/java" that are run by maven surefire plugin...

djurod:
I solved the problem by setting that option to false in manifest_pom.xml (see commit).

Is it OK to have this set in manifest_pom.xml or should it be set in kura/test/pom.xml or even in individual test projects?

ctron:
Well I would keep the flag at its default (true) because there is a reason for that. If you enable testing, but there are no test cases present, then something is wrong. Unless you do that on purpose, which sounds s bit weird, but then you can also set the flag manually.

I think it is more if a problem when, in the case something is wrong, if your test cases are not found, not executed, although you would expect that. So having an error here sounds like a good idea.

If you don't want to execute tests, the just don't define your project as a test project.

djurod:
I suspect you might not know what our plans were regarding testing, so here is a short explanation.

We plan to have all test in the "kura/test" folder (existing test should be moved there at some point). For unit tests we place the code in "src/test/java" and execute it with maven surefire plugin. For integration tests we place the code in "src/main/java" and execute it with tycho surefire plugin (existing tests were already done this way). With this unit tests are executed before the integration tests during the build.

So now we have a situation where we only have unit tests in a project and no integration tests and because of that the build fails.

I expected that changing this option "globally" is probably not good and probably the best solution would be to set this option for individual test projects (the ones that have only unit tests). Since I am new to java (and its "ecosystem") I don't know how to set this option in this test project. If you know what I need to do then please share :) ... otherwise I will ask my coworkers to help me with this on Monday.

ctron:
So for OSGi + Tycho you normally only do ".test" projects. The `src/test/java` construct is not being used. And you don't need any complex OSGi startup if you are simply doing unit tests. So also the tycho ".test" projects are more considered "unit test" than "integration tests".

You can just set the same configuration you would set globally in the local "pom" file. This will override this setting in the local file, but keep the default for all others.

djurod:
In case no one noticed, I updated the PR today and now it should be OK.

cdealti:
@ctron are you OK merging this. As @djurod already said, the idea was to put all the tests, unit tests and integration tests in .test bundles and have both tycho-surefire and maven-surefire plugins configured.
But we started from IT so maybe this PR is difficult to read now.

ctron:
As I said, traditionally this is not done that way when using Tycho. If you still want to go ahead that way. then you can do that of course. But I also don't see any benefit in doing so.

The way Tycho does that is to define all test cases in a "test fragment", which does have access to all internal packages anyway. There should be no need to define that dependency on "junit" using the POM file either.

In the end using two test plugins at the same time would probably make things more complex, and IHMO in this case, provide no benefit.

djurod:
@cdealti I think this should be ready to merge...

cdealti:
@ctron I think I agree with you. I also think that we can merge this to let @djurod continue his work.

@ctron @djurod @markoer: If the UT depended on the architecture, for example testing a method which spawns an OS process whose output may vary based on the OS, I might be interested to execute that test on the target too (I mean the Raspberry PI). In this case, the test should go in the test fragment because the plan is to eventually run these test fragments on the target.

ctron:
I think that all test could be run on all targets. For Jenkins this just another build slave where tests can be run. Yes it takes compile time, but if it is fully automated, you can simply add 5 Rasperry Pi build slaves and let it do this automatically.

cdealti:
It might work on an RPI which includes a toolchain but it's unlikely to work on production gateways and very low end (cost effective) hardware.

In the meantime I'll merge this and track this requirement elsewhere.

Thanks!

PR #802, opened 2016-11-19
Try a fix for the startup exception, missing "kura-cloud"

ctron:
This could be a possible fix for the start issue where the logs shows
an exception about the missing "kura-cloud" component.


ctron:
I just tested this one and it seems to work. However the one did work for me as well. So I can't say if it is an improvement. But at least it isn't going to make break things.

MMaiero:
Testing now. Give me few minutes

PR #808, opened 2016-11-21
Add missing dependencies after the Kura Wires merge

ctron:


amitjoy:
👍

dwoodard1:
@ctron This only addresses the Oomph setup, correct? Do we need the same change for the *.launch files and the system distributions (kura/distrib/src/main/osgi)?

ctron:
Correct. I don't know that. I never use them. Maybe Amit could make another PR for those.

dwoodard1:
Ok, we will create another PR. Out of curiosity, how do you update Eclipse using Oomph? I understand the initial setup, but I am unsure about how to update when changes are made like in this PR.

ctron:
Well the update comes with Git. Oomph only sets up git for you.

PR #813, opened 2016-11-22
Removed Cloud Services from Factory Instance Create Dialog Dropdown

### LINKED ISSUE #804: Remove DataService, CloudService and MqttDataTransport from the Factory Popup Dialog
Currently the `+` button to create factory instances also enlists the `DataService`, `CloudService` and `MqttDataTransport` services. For these aforementioned services, there already exists a dedicated UI. Hence, these services must not be enlisted in the popup dialog box.

amitjoy:
GWT 2.7 does not incorporate the use of Java 8 constructs and hence applied changes non-lambda way.

Fixes #804

### DIFF
--- a/kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/client/ui/EntryClassUi.java
+++ b/kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/client/ui/EntryClassUi.java
@@ -602,11 +602,11 @@ public class EntryClassUi extends Composite {
                             }
 
                             @Override
-                            public void onSuccess(List<String> result) {
+                            public void onSuccess(final List<String> result) {
                                 factoriesList.clear();
                                 factoriesList.addItem(SELECT_COMPONENT);
-                                for (String s : result) {
-                                    factoriesList.addItem(s);
+                                for (final String servicePid : result) {
+                                    factoriesList.addItem(servicePid);
                                 }
                             }
                         });
@@ -614,7 +614,6 @@ public class EntryClassUi extends Composite {
                 });
             }
         });
-
     }
 
     private class SelectValueChangeEvent extends ValueChangeEvent<String> {
--- a/kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/server/GwtComponentServiceImpl.java
+++ b/kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/server/GwtComponentServiceImpl.java
@@ -52,7 +52,8 @@ public class GwtComponentServiceImpl extends OsgiRemoteServiceServlet implements
 
     private static final String SERVICE_FACTORY_PID = "service.factoryPid";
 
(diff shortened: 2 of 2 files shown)


ctron:
Do you know if there is an update of GWT which would fix that?

ctron:
GWT 2.8 would allow support for Java 8 [1]

[1] http://www.gwtproject.org/release-notes.html#Release_Notes_2_8_0

amitjoy:
I am just curious to know if there a plan to move to GWT 2.8.

ctron:
Not yet. But ab upgrade from 2.7 to 2.8 sounds rather simple. And we would have a case for that right now.

amitjoy:
Shall we go ahead with this change? If it works, I would amend my changes to adopt to Java 8 way.

ctron:
I think we can merge this first and try an update. The PR already is an improvement.

amitjoy:
ok. cool. After this merge, we can create an issue to track the progress of changing to GWT 2.8 as well.

ctron:
I just tested this one and it seems that the approach doesn't work.

Having a hardcoded list won't work since there is e.g. the `org.eclipse.kura.camel.cloud.factory.CamelFactory` which now still gets shown.

amitjoy:
I believe only the Cloud Services must not be shown in this dropdown as there exists a dedicated UI for Cloud Services. The rest can stay as it is.

ctron:
`org.eclipse.kura.camel.cloud.factory.CamelFactory` is also a cloud service, as maybe others. The new Web UI covers all of them.

ctron:
Maybe it would be an idea to re-use the new `kura.ui.service.hide` property for filtering this.

amitjoy:
In Web Console, some services must be hidden in the service list (Web Console) and some in the Factory Dropdown. Cloud Services (DataService, CloudService and MqttDataTransport) are special services for which the factory creation is handled in a different UI. If there exists any other services which can be created as factory instances can be available in the dropdown list. `org.eclipse.kura.camel.cloud.factory.CamelFactory` can be listed in this dropdown if its factory creation is not handled in different UI. 

`kura.ui.service.hide` property can be a solution but we have to have two such properties to track the services to be hidden in two different places (in Web Console Service List and in Dropdown).

ctron:
So `org.eclipse.kura.camel.cloud.factory.CamelFactory `is covered by a different UI. And also other cloud service factories will be. I think a static list won't work here.

Since @MMaiero made the initial design, maybe he has some ideas

amitjoy:
@ctron Done.

MMaiero:
@amitjoy I've done a fresh build of the code few minutes ago (I don't know if in the meanwhile you rebased or made other changes and force pushed) but I get this:

schermata 2016-11-23 alle 13 36 02

Is it the expected behavior?

amitjoy:
This is because `org.eclipse.kura.camel.cloud.factory.CamelFactory` service component doesn't have any `kura.service.ui.hide` property set.

MMaiero:
ok, so it will be solved when we will merge 2.1.0 to develop. So, no problem at all. I think we can merge.

amitjoy:
👍

PR #815, opened 2016-11-22
Improved Kura installer script failure handling

### LINKED ISSUE #736: Kura installer continues despite errors occurred
Despite errors occur during the run of the Kura installer, the script continues to run:

```
Installing Kura...

Extracting Kura files
./kura_2.1.0_fedorapi_installer.sh: line 72: tar: command not found
sed: can't read /opt/eclipse/kura_*/install/kura_install.sh: No such file or directory

```

nicolatimeus:
The installer script now stops on failure.
All changed files are now backed up to a configurable directory.
The backed up files are restored on failure or on interrupt.
The installer now fails if Java 8 or greater is not installed.

Closes #736

### DIFF
--- a/kura/distrib/src/main/sh/extract.sh
+++ b/kura/distrib/src/main/sh/extract.sh
@@ -11,6 +11,10 @@
 #   Eurotech
 #
 
+set -e
+#Install failure trap
+trap 'do_restore; exit 1' SIGINT SIGTERM ERR
+
 INSTALL_DIR=/opt/eclipse
 TIMESTAMP=`date +%Y%m%d%H%M%S`
 LOG=/tmp/kura_install_${TIMESTAMP}.log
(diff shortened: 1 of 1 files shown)


ctron:
The script would still continue in the case e.g.`unzip` would be missing.

I think that
1) All checks should be done first, execute the main logic only after validation
2) The script should be run at least with `set -e`

nicolatimeus:
@ctron thanks for feedback
The script adds a trap for ERR that is used to restore the backup and stop the script if any command fails or if any binary is missing, i've not used `set -e` because this will probably interfere with this trap, anyway i could improve the script this way:

1) start the script with `set -e`
2) run the command `set +e` just before attaching the trap
3) check that the `unzip` and `tar` binaries are present before starting (let me now if there are other required binaries that could be missing at runtime)

PR #820, opened 2016-11-23
Upgrade to GWT 2.8.2

ctron:
This upgrades GWT to 2.8.0

**Note:** Don't merge until CQ12311 [1] has been approved

[1] https://dev.eclipse.org/ipzilla/show_bug.cgi?id=12311


ctron:
I did re-base it. Could you please have a look.

MMaiero:
Did you try it with Eclipse? It is giving me a lot of exceptions, especially in the bootstrap part, even if it is compiling in maven. Maybe there is something I've missed

ctron:
Yes I did and it worked fine. As well as with a `build-all.sh` build.

I can retest though. Maybe some later commit has messed things up.

MMaiero:
I've used the oomph setup. Don't know if this could be a possible cause.

amitjoy:
@ctron Any update?

MMaiero:
It is not in the plans for Kura 3.0.0. To discuss for Kura 4.0.0, but we need to consider the change (in terms of requirements of memory, upgrader, bandwidth) compared to the advantages

ctron:
As just discussed on the weekly Kura call, I will rebase this PR so that we can merge this for the next Kura release. We aim to go with GWT 2.8.2, which is covered by this CQ as 2.8.2 only is a service release to 2.8.0 and we do have an approved CQ for that.

ctron:
So from a first look that was easy. It compiles, but I haven't tested it after rebasing and switching to 2.8.2.

I will continue with this next week, because I think we can close this rather quickly.

ctron:
@MMaiero  @dwoodard1 The change works for me. I think this is ready to be merged.

MMaiero:
Testing it now

MMaiero:
The only strange thing I've seen so far moving across all the pages is that the list of snapshots is not available

ctron:
> The only strange thing I've seen so far moving across all the pages is that the list of snapshots is not available

I can reproduce the issue. There is an exception in the browser log. However GWT stack traces are not very helpful. But I will try to find the root cause of this.

ctron:
I just pushed a fix for this issue. I think this actually could already be an issue with previous GWT versions. As it is not really clear to which type to cast. However there already was an explicitly typed version of the same operation and so I simply used that.

The issue is now solved for me.

PR #821, opened 2016-11-23
Issue #782: Configuration service unit tests

markoer:


kura/test/org.eclipse.kura.core.configuration.test/META-INF/MANIFEST.MF
```diff
@@ -0,0 +1,24 @@
...
+Bundle-ManifestVersion: 2
+Bundle-Name: org.eclipse.kura.core.configuration.test
+Bundle-SymbolicName: org.eclipse.kura.core.configuration.test;singleton:=true
+Bundle-Version: 1.0.0.qualifier
+Bundle-Vendor: Eclipse Kura
+Bundle-RequiredExecutionEnvironment: JavaSE-1.6
+Import-Package: org.eclipse.kura;version="[1.2,2.0)",
+ org.eclipse.kura.configuration;version="[1.1,2.0)",
+ org.eclipse.kura.configuration.metatype;version="[1.0,2.0)",
+ org.eclipse.kura.core.configuration;version="1.0.2",
```
cdealti: I guess it would be better to make this test a fragment of the org.eclipse.kura.core.configuration plugin (see #832) avoiding to import org.eclipse.kura.core.configuration.*

kura/test/org.eclipse.kura.core.configuration.test/pom.xml
```diff
@@ -0,0 +1,42 @@
...
+
+    <parent>
+        <groupId>org.eclipse.kura</groupId>
+        <artifactId>test</artifactId>
+        <version>2.1.0-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>org.eclipse.kura.core.configuration.test</artifactId>
+    <version>1.0.0-SNAPSHOT</version>
+    <packaging>eclipse-plugin</packaging>
```
cdealti: packaging should be eclipse-test-plugin (I believe) if this project is going to use tycho-surefire-plugin.

kura/test/org.eclipse.kura.core.configuration.test/src/main/resources/kura_custom.properties
```diff
@@ -0,0 +1 @@
+kura.primary.network.interface=enp0s3
```
markoer: Well, I do... As we already discussed, many different names are now possible. If you want it changed, to what? If to eth0, I might as well remove the file.

kura/test/org.eclipse.kura.core.configuration.test/pom.xml
markoer: For now this does exactly this - prevent tycho from trying to run non-existing integration tests.

kura/test/org.eclipse.kura.core.configuration.test/src/main/resources/kura_custom.properties
ctron: I do think we should re-think the concept of a "primary" network interface.

PR #822, opened 2016-11-23
#779: Unit tests for EthernetInterfaceConfigImpl class

djurod:
- Prepared unit tests for EthernetInterfaceConfigImpl class from package
  org.eclipse.kura.core.net
- Fixed or implemented comparison of classes: AbstractNetInterface,
  EthernetInterfaceConfigImpl, and NetInterfaceAddressConfigImpl

### DIFF
--- a/kura/manifest_pom.xml
+++ b/kura/manifest_pom.xml
@@ -282,8 +282,8 @@
                 <artifactId>maven-compiler-plugin</artifactId>
                 <version>3.3</version>
                 <configuration>
-                    <source>1.6</source>
-                    <target>1.6</target>
+                    <source>1.8</source>
+                    <target>1.8</target>
                 </configuration>
             </plugin>
 
--- a/kura/org.eclipse.kura.core.net/src/main/java/org/eclipse/kura/core/net/AbstractNetInterface.java
+++ b/kura/org.eclipse.kura.core.net/src/main/java/org/eclipse/kura/core/net/AbstractNetInterface.java
@@ -12,6 +12,7 @@
 package org.eclipse.kura.core.net;
 
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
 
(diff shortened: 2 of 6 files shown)


djurod:
So there appears to be an issue with use of Java 1.8...

@cdealti @ctron any ideas how to solved this? According to @markoer one solution could be to change the compiler version in manifest_pom.xml, since apparently the next version of Kura will require minimal Java 1.8.

cdealti:
@djurod I'd agree with @markoer

ctron:
@djurod What problem is there?

djurod:
I now resolved the merge conflict and rebased to upstream again. Hopefully if will now build, but since it doesn't appear that anything changed in the build system I would expect that the build will fail again. If it does I increase the compiler version in manifest_pom.xml from 1.6 to 1.8 and try again.

ctron:
Can you explain what the problem is?

djurod:
Build is failing because of use of java features that are not present in java 1.6 (see build log on Travis CI):

> ERROR] /home/travis/build/eclipse/kura/kura/test/org.eclipse.kura.core.net.test/src/test/java/org/eclipse/kura/core/net/EthernetInterfaceConfigImplTest.java:[294,104] diamond operator is not supported in -source 1.6
>
>  (use -source 7 or higher to enable diamond operator)

Manifest for the project is configured to use java 1.8, but for some reason build still fails. I don't see any reason why it should except maybe because the java compiler version is set to 1.6 in manifest_pom.xml...

ctron:
I will look into this. For the moment just set source and target to 1.8.

ctron:
I just gave it a try. And this is exactly what I mean. To me it looks like the "src/test/java" code is compiled with settings from the "maven-compiler-plugin" while the "src/main/java" code is compiled with the "tycho-compiler-plugin".

If you use the diamond operator in "src/main/java" it works for me.

Please don't mix those two constructs. This will cause problems more problems than it will solve.

djurod:
With the latest commit the build no longer fails. Please review the changes.

ctron:
And yet the original problem still exists.

If you do that, at least set all BREE of all bundles  to Java 8 as well.

djurod:
I see these possibilities to solve this issue:
- Someone needs to create a new pull request that properly ports Kura to java 1.8 and I wait with this pull request until that is done.
- We merge this pull request (if there are no other issues with it) and require of all future pull requests to set the BREE of the changed bundles to java 1.8 (if it is not already). This could be checked during review phase.

Setting BREE to java 1.8 in all bundles inside this pull request does not make sense to me, as it is way out of context of what is being done here...

@cdealti could you also comment on this?

ctron:
There is a third option: Use the tools as they were intended to be used and only create `.test` fragment/bundles for testing. Using tycho-surefire only.

djurod:
@ctron your last comment probably fits into discussion regarding testing in issue #833. Should we first resolve that and then return to this?

djurod:
With the last commit I added a few more tests for the EthernetInterfaceConfigImpl class and I added unit tests for NetInterfaceAddressConfigImpl.

Issue #309, opened 2016-07-11
[Configuration Service] SelfConfiguringComponentS not tracked


cdealti:
The recent refactoring of the ConfigurationService broke tracking of SelfConfiguringComponents.
For example, they are not displayed in the list of services in the Kura Web UI and changes to their configuration are not saved to snapshots.

cdealti:
@ibinshtok can you please review?

cdealti:
Closed by #310.

PR #825, opened 2016-11-23
Fixed Bug to track SelfConfiguringComponents properly

### LINKED ISSUE #824: SelfConfiguringComponent are not registered properly
Due to continuous changes in Configuration Service, some codes got missing and wrongly replaced. This caused SelfConfiguringComponent to be registered with errors.

amitjoy:
The bug incurred a problem for which SelfConfiguringComponents are not
tracked properly by Configuration Service. Hence the dependent modules
such as Kura Wires encountered problems while creating Asset factory
instances.

Fixes #824

### DIFF
--- a/kura/org.eclipse.kura.core.configuration/src/main/java/org/eclipse/kura/core/configuration/ConfigurableComponentTracker.java
+++ b/kura/org.eclipse.kura.core.configuration/src/main/java/org/eclipse/kura/core/configuration/ConfigurableComponentTracker.java
@@ -90,7 +90,7 @@ public class ConfigurableComponentTracker extends ServiceTracker {
                             } else if (obj instanceof SelfConfiguringComponent) {
                                 s_logger.info("Adding SelfConfiguringComponent with pid {} and service pid {}", pid,
                                         servicePid);
-                                this.m_confService.registerSelfConfiguringComponent(servicePid);
+                                this.m_confService.registerSelfConfiguringComponent(pid, servicePid);
                             }
                         } finally {
                             this.context.ungetService(ref);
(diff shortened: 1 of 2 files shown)


cdealti:
mmh, let's see. I'm getting confused with all these PIDs around...

amitjoy:
@cdealti Yesterday I also got confused when I was swimming with PIDs 😆

amitjoy:
But trust me PIDs are very loyal. They would never betray. 😆

amitjoy:
@cdealti Actually this is a regression issue as well.

cdealti:
@amitjoy can you please find when the regression has been introduced? Maybe try with ``git bisect``

amitjoy:
I am not used to `git bisect` and would be able to find it later on. But just to update you quickly, it is actually introduced when a dedicated method for SelfConfiguringComponent registration has been written - `registerSelfConfiguringComponent(String)`. Previously this has been handled by the `registerComponentConfiguration(String, String, String)`. This method previously handled the registrations for the `ConfigurableComponent` and `SelfConfiguringComponent` both.

amitjoy:
The commit 8a14fc28e230c0c9da7b25ea6669a11f36816989
And the issue to solve #309 

This partly introduced the regression.

ctron:
So if we are talking about Kura 3.0.0, I would actually like to drop the `ConfigurableComponentTracker` in favor of proper service tracking.

ctron:
So only the code path which is effective when `org.eclipse.kura.core.configuration.trackOnlyRelevantServices` is `true` should remain.

amitjoy:
I agree with @ctron that currently `ConfigurableComponentTracker` does invoke the relevant registration methods as soon as it finds the proper services. This could be achieved using the inline trackers in `ConfigurationServiceImpl`. In this scenario, the services would be tracked based on a property as @ctron has mentioned in the comment.

PR #842, opened 2016-11-26
Re-enable Kura tests

cdealti:
This PR re-enables the old Kura tests.
To make the execution of org.eclipse.kura.core.test more reliable
the MQtt connection is established to a local broker.
Travis CI configuration has been modified to install Mosquitto.
To run the tests locally one has to install a broker.

The tests are crap and need to be heavily redesigned/refactored.
However, some tests are better than none.


cdealti:
@ctron Can you please check if they pass now?

ctron:
I try to repeat what I already wrote in my e-mail earlier. I do see several **severe** issues with this approach and strongly suggest against this way:

* This setup requires the "legacy" run mode of travis (with sudo). This is considered legacy (by Travis) for a reason. Those reasons are described here [1]. Making this changes will have effects on each and every validation build on Kura
* Having such a system dependency for running integration tests is a very bad idea. Having a manual step of installing a broker on the local system is burden, is something which some users may even not be able to do at all (Windows? No local admin rights? …) So the tests will probably just de-activated, like it was done in the past.
* Everyone will test with a different version of a broker. Now broker versions will have to be checked before an issue can be configured. Do you have the right version, the right dependencies, the correct configuration. Maybe it is your OS, maybe something else.

To me the issues described in [1] alone would be a reason not to follow this path.

However there is a very simple solution for this. We could use an embedded instance of Artemis [2] instead. This would instantly solve all of the issues above.

So I would like to learn about the advantages which Mosquitto provides over the Artemis approach.

[1] https://docs.travis-ci.com/user/migrating-from-legacy/
[2] https://activemq.apache.org/artemis/

cdealti:
@ctron thanks. We will look into this issues. However disabling tests is also not an option IMHO.

ctron:
Well I they don't work right now, then disabling those tests at least gives the other test a chance to run. So if those tests always fail, then it is better to disable them and have tests available at least for the rest.

I agree, those tests should be fixed re-enabled ASAP. But if this means that the whole Kura project massively suffers from that, then it is the wrong way to go.

If we merge this in, then every PR validation build will suffer from this. Every developer who wants to use unit tests will have to take care of installing an exact setup of a broker, which is probably not even documented. At least not in the wiki.

So of there are no advantages of using this setup of Mosquitto, then I would strongly vote for adapting Artemis, or to look for alternatives. If you need help, I can definitely do that.

I would suggest we discuss this is in the Kura project call tomorrow.

cdealti:
> This setup requires the "legacy" run mode of travis (with sudo). This is considered legacy (by Travis) for a reason. Those reasons are described here [1]. Making this changes will have effects on each and every validation build on Kura

fixed

> Having such a system dependency for running integration tests is a very bad idea. Having a manual step of installing a broker on the local system is burden, is something which some users may even not be able to do at all (Windows? No local admin rights? …) So the tests will probably just de-activated, like it was done in the past.

maybe. On Linux Mosquitto can run without root privileges from any directory.

> Everyone will test with a different version of a broker. Now broker versions will have to be checked before an issue can be configured. Do you have the right version, the right dependencies, the correct configuration. Maybe it is your OS, maybe something else.

Kura is meant to work with every MQTT 3.1.1 broker and for the tests we don't yet require a configuration

> If we merge this in, then every PR validation build will suffer from this. Every developer who wants to use unit tests will have to take care of installing an exact setup of a broker, which is probably not even documented. At least not in the wiki.

There is not setup. You need an MQTT 3.1.1 broker listening on 1883.

> So of there are no advantages of using this setup of Mosquitto, then I would strongly vote for adapting Artemis, or to look for alternatives. If you need help, I can definitely do that.

For tests, the advantage is size, startup time and it's an Eclipse project

I agree this is not perfect but we can run tests right now (whose compilation has been broken in the last five days, as far I can see - they build and pass in my Travis CI build).

ctron:
So unit tests are only supported on 64bit Linux running as Travis? When they could be running on all platforms, without the requirement of any manual setup?

cdealti:
I've embedded the Eclipse Moquette MQTT broker in the tycho surefilre build.
Tests pass locally but on Travis we are hitting the 4MB log limit which cause the build to fail.
When you have a chance can you take a look?

PR #843, opened 2016-11-27
Added Proper Javadoc to INTERNAL_ERROR for discouraged access

### LINKED ISSUE #841: Add Discouraged Access in Javadoc for INTERNAL_ERROR error code
We need to deprecate the use of `KuraErrorCode#INTERNAL_ERROR`

amitjoy:
Fixes #841

### DIFF
--- a/kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/KuraErrorCode.java
+++ b/kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/KuraErrorCode.java
@@ -12,47 +12,53 @@
 package org.eclipse.kura;
 
 /**
- * KuraErrorCode holds the enumeration of valid error codes for the exception message. For each defined enum value, a corresponding message should be defined in the properties bundle named:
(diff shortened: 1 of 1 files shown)


ctron:
We agreed upon not technically deprecating internal error. And I would be against it as long as we don't clear up all references to it at the same time.

Marking it as deprecated will cause numerous warnings in the IDE, which would then hide other warnings.

So as long as we stick to the original idea, please do not deprecate it.

Also I think there is something wrong with the code formatter. I had the feeling that something went wrong in the big code formatter commit, but please do double check if you are using the correct one.

amitjoy:
👍 Yeah I missed that conversation. No problem. There will be some amount of refactoring here and there. I will start with the refactoring in Kura Wires first which I can do with less effort then would like to cover the other areas.

I don't know if something has happened to the formatter in my Eclipse. I can still see in Eclipse Configurations that I have the Kura Code Formatter and CleanUp Formatter set.

ctron:
Do you have the version checked in by @MMaiero in the "setups/formatting" folder?

amitjoy:
@ctron All the formatters are in place.

PR #845, opened 2016-11-28
Fix warnings

ctron:
Fix a few warnings by importing org.junit instead of junit

### DIFF
--- a/kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/net/dns/DnsServerConfigIP.java
+++ b/kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/net/dns/DnsServerConfigIP.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2011, 2016 Eurotech and/or its affiliates
+ * Copyright (c) 2011, 2016 Eurotech and/or its affiliates and others
  *
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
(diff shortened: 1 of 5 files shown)


kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/net/dns/DnsServerConfigIP.java
```diff
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2011, 2016 Eurotech and/or its affiliates
+ * Copyright (c) 2011, 2016 Eurotech and/or its affiliates and others
```
dwoodard1: @ctron To be consistent with other headers, can you drop the "and others"?

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/net/firewall/FirewallOpenPortConfigIP.java
```diff
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2011, 2016 Eurotech and/or its affiliates
+ * Copyright (c) 2011, 2016 Eurotech and/or its affiliates and others
```
dwoodard1: Same as above

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/net/dns/DnsServerConfigIP.java
ctron: No, because this is a mixed copyright. Between Eurotech and Red Hat. And thus the copyright is "Eurotech (...) and others". But please check back with EMO if in doubt.

PR #847, opened 2016-11-28
Drop Web "classic" bundle

ctron:
A while ago we agreed upon dropping the classic web bundle


dwoodard1:
+1 For me. @cdealti @MMaiero Do you see any reason to not drop this bundle?

ctron:
So can we merge this?

PR #849, opened 2016-11-28
Show warnings and validate bootdelegation issues

ctron:


kura/org.eclipse.kura.core.configuration/META-INF/MANIFEST.MF
```diff
@@ -9,6 +9,7 @@ Service-Component: OSGI-INF/*.xml
 Bundle-ClassPath: .
 Bundle-ActivationPolicy: lazy
 Import-Package: javax.crypto,
+ javax.xml.bind,
```
dwoodard1: Is this change needed as part of this PR?

dwoodard1: That's a good reason.

PR #852, opened 2016-11-28
Fixed handling of configuration ADs with options

### LINKED ISSUE #851: [Web-UI] Label is sent instead of value for metatype parameters with options
If a metatype AD is declared with predefined options having a label and a value defined, the web ui will send to the server the text of the selected label instead of the actual value.
This also breaks the device display name configuration in the Cloud Services section.

nicolatimeus:
The actual value is now sent instead of the label in metatype parameters with options.

Closes #851


ctron:
Is there a reason for not using: `box.getSelectedValue()` ?

nicolatimeus:
No reason actually, I'll switch to that method and amend the commit.

PR #858, opened 2016-11-30
Updated release notes.

MMaiero:
Added missing release notes for 2.0.2.
First draft of the release notes for v. 2.1.0.

### DIFF
--- a/kura/distrib/RELEASE_NOTES.txt
+++ b/kura/distrib/RELEASE_NOTES.txt
@@ -8,9 +8,21 @@ Added Features :
 
 Bug Fixes :
   * Fixed github issues:
+    * #909: [Kura 2.1.0 QA] SSID selection dialog too small
+    * #903: [Kura 2.1.0 QA] Impossible to enter multiple DNS servers
+    * #901: [Kura 2.1.0 QA] changing username/password locks out admin access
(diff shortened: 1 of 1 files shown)


ctron:
I think there should also be a note on the AES encryption [1] and a note on Java 8 support.

Other than that I would suggest to have some "Cool new stuff" section in the beginning. Could also be the list from [1].

[1] https://projects.eclipse.org/projects/iot.kura/releases/2.1.0/review

ctron:
I would also add a marker on Java 8, since this was already discussed on the mailing list.

dwoodard1:
+1 for the "What's New" section. The references to issues are important, but most people aren't going to read those and decipher what has actually changed. We had this in previous release notes in a "Added Features" section.

@ctron Not sure what you mean by the "add a marker on Java 8" comment. Can you explain?

ctron:
I had a few run ins already people asking me why Java 8 is a problem with Kura. Also see the mailing list of Kura. So in order to clear any misunderstandings I would add, under "Known Issues" a note which says something like "Java 8 can be used as target platform but not for developing Kura components. This will be fixed in the next Kura release".

So people know what to expect.

ctron:
+1

MMaiero:
I think we are done with the release notes for Kura

PR #868, opened 2016-11-30
Fix concurrency issue in command service

ctron:
The command service was implemented in a way that it did execute all
command written out to a single shell script. This can cause the issue
that two requests issued at the same time would fail.

This change does write out separate script files using the
Files.createTempFile method. It also does not fork "chmod" and
`dos2unix` since those operations can be performed inside Java which
saves two additional process forks.

Also is the `cd` statement not necessary when the ProcessBuilder
is instructed to change the directory first.

This change also includes a unit test for the command service.


ctron:
This PR supersedes PR #331

kura/org.eclipse.kura.linux.command.test/META-INF/MANIFEST.MF
```diff
@@ -0,0 +1,8 @@
+Manifest-Version: 1.0
+Bundle-ManifestVersion: 2
+Bundle-Name: Unit tests for 'org.eclipse.kura.linux.command'
+Bundle-SymbolicName: org.eclipse.kura.linux.command.test
+Bundle-Version: 1.0.0.qualifier
```
amitjoy, line 5: Bundle-Vendor would be good

kura/org.eclipse.kura.linux.command.test/src/org/eclipse/kura/linux/command/CommandServiceTest.java
```diff
@@ -0,0 +1,54 @@
...
+
+public class CommandServiceTest {
+
+    private CommandService service;
+
+    @Before
+    public void setup() {
+        this.service = new CommandServiceImpl();
+    }
+
```
amitjoy, line 29: Similarly you can also provide an after method to dereference service. It is just a sugar coating for faster garbage collection.

kura/org.eclipse.kura.linux.command.test/META-INF/MANIFEST.MF
ctron: True

kura/org.eclipse.kura.linux.command.test/src/org/eclipse/kura/linux/command/CommandServiceTest.java
ctron: Nothing is referenced here. Setting a null value doesn't really help in modern java implementations, so leaving this as is would be fine unless we target Java 1.0 again.

amitjoy: 😆 Yeah. exactly and that's why I considered it as a sugar coating.

ctron:
So can this be merged now?

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/KuraErrorCode.java
```diff
@@ -54,5 +55,9 @@
 	/**
 	 * @since {@link org.eclipse.kura} 1.2.0
 	 */
-	INVALID_PARAMETER
+	INVALID_PARAMETER,
+	/**
+	 * Error executing process
+	 */
+	PROCESS_EXECUTION_ERROR
```
MMaiero: I think that a @since should be placed here.

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/KuraErrorCode.java
```diff
@@ -1,13 +1,14 @@
...
+ * Copyright (c) 2011, 2016 Eurotech and/or its affiliates and others
  *
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * which accompanies this distribution, and is available at
  * http://www.eclipse.org/legal/epl-v10.html
  *
  * Contributors:
- *     Eurotech
+ *     Eurotech and/or its affiliates
```
MMaiero, line 10: I don't think that we need the "and/or its affiliates" here. @dwoodard1

kura/org.eclipse.kura.linux.command.test/pom.xml
```diff
@@ -0,0 +1,30 @@
...
+      Red Hat Inc
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+	xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+	<modelVersion>4.0.0</modelVersion>
+
+	<parent>
+		<groupId>org.eclipse.kura</groupId>
+		<artifactId>kura</artifactId>
+		<version>2.1.0-SNAPSHOT</version>
```
MMaiero: This will cause a problem since the parent has version 3.0.0-SNAPSHOT

kura/org.eclipse.kura.linux.command/src/main/java/org/eclipse/kura/linux/command/CommandServiceImpl.java
```diff
@@ -8,170 +8,82 @@
...
+        
+        if (cmd == null || cmd.isEmpty()) {
+            return "<empty command>";
         }
-
-        // Delete script file if it exists
-        this.m_scriptFile = new File(SCRIPT_FILE);
-        if (this.m_scriptFile.exists()) {
+        
+        try {
```
MMaiero, line 48: what is the advantage of having these two nested try-catch blocks?

ctron: The finally section can throw an IOException as well, which is this way covered by the outer try-catch

ctron:
I added the requested changes.

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/KuraErrorCode.java
```diff
@@ -54,5 +55,10 @@
 	/**
 	 * @since {@link org.eclipse.kura} 1.2.0
 	 */
-	INVALID_PARAMETER
+	INVALID_PARAMETER,
+	/**
+	 * Error executing process
+	 * @since {@link org.eclipse.kura} 3.0.0
```
MMaiero: Maybe I'm wrong but I think that @cdealti for INVALID_PARAMETER and OS_COMMAND_ERROR with {@link org.eclipse.kura} referred the version for the org.eclipse.kura package inside the org.eclipse.kura.api project.

ctron: Updated

amitjoy:
@ctron Could you kindly resolve the conflicts?

ctron:
I am currently traveling. I can do this next week.

MMaiero:
Merged.
I've moved back api bundle version to 1.2.0-SNAPSHOT

PR #869, opened 2016-11-30
Clean up and document IOUtil

ctron:


kura/org.eclipse.kura.core/META-INF/MANIFEST.MF
```diff
@@ -12,6 +12,7 @@ Import-Package: javax.crypto,
  javax.net,
  javax.net.ssl,
  javax.sql,
+ org.apache.commons.io;version="[2.4,3)",
```
amitjoy: In Kura, the .0 is also mentioned everywhere as it is more visible at first glance. I suggest you to add 3.0.

kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/util/IOUtil.java
```diff
@@ -8,114 +8,84 @@
...
-        return resource;
+        return readResource(ctx.getBundle().getResource(resourceName));
     }
 
     /**
      * Reads a resource fully and returns it as a string.
      * 
-     * @param ctx
+     * @param bundle
+     *            the bundle to use for getting the bundle context
```
amitjoy: In the implementation the `BundleContext` is never used. Broadly speaking, in the implementation, there is no such call (`bundle.getContext()`).

kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/util/IOUtil.java
```diff
@@ -8,114 +8,84 @@
...
 
     /**
      * Reads a resource fully and returns it as a string.
      * 
      * @param ctx
+     *            the bundle context to use for locating the resource
      * @param resourceName
-     * @return
+     *            the name of the resource
+     * @return the content as string, or {@code null if the resource coude not be found}
```
amitjoy: could

kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/util/IOUtil.java
```diff
@@ -8,114 +8,84 @@
...
+    }
 
     /**
      * Reads a resource fully and returns it as a string.
      * 
-     * @param ctx
      * @param resourceName
-     * @return
+     *            the name of the resource
+     * @return the content as string, or {@code null if the resource coude not be found}
```
amitjoy: could

kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/util/IOUtil.java
```diff
@@ -8,114 +8,84 @@
...
     /**
      * Reads a resource fully and returns it as a string.
      * 
-     * @param ctx
+     * @param bundle
+     *            the bundle to use for getting the bundle context
      * @param resourceName
-     * @return
+     *            the name of the resource
+     * @return the content as string, or {@code null if the resource coude not be found}
```
amitjoy: could

kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/util/IOUtil.java
ctron: Then again this would break the interface and we would need to increment the major version number. Is that worth it? I don't think so.

ctron:
Updated and rebased. Good to be merged?

ctron:
@MMaiero @cdealti  can this be merged?

PR #870, opened 2016-11-30
Fix fedora issues

### LINKED ISSUE #742: Unable to establish stable network connection on Fedberry
This may be related to #721 but I am not sure.

While testing with Fedberry 24 I constantly run into issues where the network connection to the Raspberry PI is terminated and the PI is not evening pingable anymore.

I dug a bit through the Kura code and found a few locations where there are special cases for "Intel Edison", although they are more related to "systemd" instead.

Maybe there is an issue with "Network Manager", "systemd" or some other component. I am trying to track it down in this issue.

MMaiero:
This PR closes #742.


ctron:
Just to understand it, are we using NetworkManager and OpenVPN?
Also it seems that there are some hickups with whitespaces.

I will give it a test run now.

ctron:
Looks good to me.

However I am not sure what exactly was wrong. So I can't really test whatever got fixed.

MMaiero:
> Just to understand it, are we using NetworkManager and OpenVPN?

NetworkManager is a generic type that wraps all the programs related with networking like, for example, `wpa_supplicant`.
You're right we don't need OpenVPN. I've updated the pull request accordingly.

> Looks good to me.

Great

> However I am not sure what exactly was wrong. So I can't really test whatever got fixed.

This PR takes the following changes, as described in the descriptions of the commits:

- Disabled the default network manager. Kura is, by itself, a network manager: we cannot have both running.

- Added firewall init script to actually enable the feature.

- Added ifcfg files to specify the behavior of the network interfaces.

- Added selinux module to enable the network management features of Kura that, otherwise, would be blocked by the system.

ctron:
So if I can install this and manage the network interface and the firewall, then the changes should be ok?

MMaiero:
Yes. Please proceed with the QA.

ctron:
So I can set modify the firewall and change network configuration.

PR #872, opened 2016-11-30
Use Tycho for building Web 2

ctron:
This set of changes converts the build of the Web2 bundle to an actual tycho build. This allows to build all Kura bundles in one build reactor thus allowing us to build a final P2 repository in the end.


MMaiero:
I had a look, but I have few questions. Could we discuss about this tomorrow?

ctron:
Sure we can do that. If you already have some feedback, just let me know.

ctron:
So, as just discussed, the bundle is now named `.web2`

kura/distrib/pom.xml
```diff
@@ -350,13 +350,13 @@
 						</configuration>
 					</execution>
 					<execution>
-						<id>get-web2-bundle</id>
+						<id>get-web-bundle</id>
```
MMaiero: Just to keep consistency with what we agreed: why not keeping it as web2?

kura/distrib/pom.xml
```diff
@@ -350,13 +350,13 @@
...
 						<phase>generate-resources</phase>
 						<goals>
 							<goal>copy</goal>
 						</goals>
 						<configuration>
-							<skip>${org.eclipse.kura.web2.skip}</skip>
+							<skip>${org.eclipse.kura.web.skip}</skip>
```
MMaiero: Same as before

kura/distrib/src/main/ant/build_equinox_distrib.xml
```diff
@@ -121,8 +121,8 @@
 		<antcall target="wires-config" />
 
 		<!-- Conditional entries. Check for jar presence, and invoke relevant ant target-->
-		<available file="${project.build.directory}/plugins/org.eclipse.kura.web2_${org.eclipse.kura.web2.version}.jar" property="web2.jar.present" />
-		<antcall target="web2-config" />
+		<available file="${project.build.directory}/plugins/org.eclipse.kura.web2_${org.eclipse.kura.web2.version}.jar" property="web.jar.present" />
```
MMaiero: Same as in kura/distrib/pom.xml

ctron:
I thought we were only talking about the bundle itself.

Better that way?

PR #884, opened 2016-12-02
Resetted warnings when changing status in tcpip tab

### LINKED ISSUE #883: Invalid field validation state when switching from WAN to LAN
In the network UI:

Pre-conditions:
* eth0 is configured for WAN with DHCP

Steps:
* Select "manual" for eth0
* Clear the "gateway" field, may already be empty
* Press save -> error, OK
* Select "LAN" for eth0
* Press save -> error -> NOT OK

The field is no longer required when the mode is switched from WAN to LAN, but the field validation is not reset.

MMaiero:
Closes #883 


MMaiero:
@ctron Could you check if this solves your issue? Thanks

ctron:
I will do that this week.

PR #885, opened 2016-12-02
Fix for bind/named service

pierantoniomerlino:
Bind/named service and its configuration files are in different
locations depending on the OS. The modifications on this PR
add the support for different file locations.


pierantoniomerlino:
@MMaiero Fixed named for fedorapi. Could you please review the PR?

PR #886, opened 2016-12-02
Provide initial support for Karaf

ctron:
This set of patches provides initial build support for Apache Karaf.

It does generate a few Karaf feature XML files and a few example targets (docker and fedora). It also has an emulator instance for launching locally.


hekonsek:
Hi Jens

Awesome work! I am really glad to see Karaf support coming to Kura finally. All looks perfect for me. Merging then.

Thanks again!

PR #936, opened 2016-12-12
Fix the implementation of the Kura wires commands

ctron:
Contrary to the javadoc of the Wires commands this implementation does
not use the Gogo shell but the old, legacy, deprecated Equinox console.

However this does break the compatibility with Karaf.


hekonsek:
+1 we should be using gogo. Merged. Many thanks, Jens.

PR #953, opened 2016-12-15
Convert P2 repository to Maven 2 repository

ctron:
These changes convert the P2 repository generated during the build of Kura to a plain Maven 2 repository, suitable for uploading to Maven Central.


kura/maven-central/p2tom2.properties
```diff
@@ -0,0 +1,76 @@
...
+default.project.url=http://eclipse.org/kura
+
+developers=ctron
+
+developer.ctron.name=Jens Reimann
+developer.ctron.email=jreimann@redhat.com
+developer.ctron.url=https://dentrassi.de
+developer.ctron.organization=Red Hat Inc
+developer.ctron.organizationUrl=https://redhat.com
+
```
dwoodard1: I would like to see Eurotech added as a developer organization. I believe the other fields could be left blank if we don't won't to name specific committers. The 'name' attribute could be set to "Various" with the other fields empty.

ctron: Absolutely! That has to be updated. Let's check what other Eclipse projects do on maven central. Having "Eclipse Kura" as name would hopefully cut a long discussion around that short.

dwoodard1: A quick search looks like most projects are not filling in this information. It looks like Jetty and JGit are referencing each committer.

https://mvnrepository.com/artifact/org.eclipse.jetty/jetty-server/9.4.0.v20161208
https://mvnrepository.com/artifact/org.eclipse.jgit/org.eclipse.jgit/4.6.0.201612231935-r

dwoodard1: Also, I agree with keeping the discussion short. So, we can just add my information for the quickest fix. If you want, I can make a direct commit after the merge.

ctron: Well adding just your information is as wrong as adding just mine. So either we add all contributors, which is fine as well, or we stick to something more generic.

ctron: As discussed today we will go with the simple approach of listing the contributors page like here [1]

[1] https://search.maven.org/#artifactdetails|org.eclipse.platform|org.eclipse.swt|3.105.2|jar

ctron:
Rebased and updated.

Beside the URL tag there is also an ID named "eclipse.kura". This should not hurt and is simply due to the fact that the "p2tom2" tool currently requires in ID for the developer.

PR #980, opened 2017-01-04
#779: Unit tests for Wifi classes from core.net

djurod:
- Prepared unit tests for classes WifiAccessPointImpl,
  WifiInterfaceAddressConfigImpl and WifiInterfaceConfigImpl from
  package org.eclipse.kura.core.net
- Fixed or implemented comparison of classes: WifiAccessPointImpl,
  WifiInterfaceAddressConfigImpl

### DIFF
--- a/kura/org.eclipse.kura.core.net/src/main/java/org/eclipse/kura/core/net/WifiAccessPointImpl.java
+++ b/kura/org.eclipse.kura.core.net/src/main/java/org/eclipse/kura/core/net/WifiAccessPointImpl.java
@@ -11,6 +11,7 @@
  *******************************************************************************/
 package org.eclipse.kura.core.net;
 
+import java.util.Arrays;
 import java.util.EnumSet;
 import java.util.List;
 
@@ -116,7 +117,7 @@ public class WifiAccessPointImpl implements WifiAccessPoint {
     public String toString() {
         StringBuilder sb = new StringBuilder();
         sb.append("ssid=").append(this.ssid);
-        if (this.hardwareAddress != null && this.hardwareAddress.length > 0) {
+        if (this.hardwareAddress != null && this.hardwareAddress.length == 6) {
             sb.append(" :: hardwareAddress=").append(NetworkUtil.macToString(this.hardwareAddress));
         }
         sb.append(" :: frequency=").append(this.frequency).append(" :: mode=").append(this.mode);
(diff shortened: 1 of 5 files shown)


kura/org.eclipse.kura.core.net/src/main/java/org/eclipse/kura/core/net/WifiAccessPointImpl.java
```diff
@@ -141,4 +142,82 @@ public String toString() {
         }
         return sb.toString();
     }
+    
+    @Override
+	public int hashCode() {
```
amitjoy, line 147: To save verbosity, you could use `Objects.hash(Object...)` and `Objects.equals(Object, Object)`

kura/test/org.eclipse.kura.core.net.test/src/test/java/org/eclipse/kura/core/net/WifiAccessPointImplTest.java
```diff
@@ -0,0 +1,497 @@
+/*******************************************************************************
+ * Copyright (c) 2016 Eurotech and/or its affiliates and others
```
amitjoy: 2017

kura/test/org.eclipse.kura.core.net.test/src/test/java/org/eclipse/kura/core/net/WifiInterfaceAddressConfigImplTest.java
```diff
@@ -0,0 +1,311 @@
+/*******************************************************************************
+ * Copyright (c) 2016 Eurotech and/or its affiliates and others
```
amitjoy: 2017

kura/test/org.eclipse.kura.core.net.test/src/test/java/org/eclipse/kura/core/net/WifiInterfaceConfigImplTest.java
```diff
@@ -0,0 +1,156 @@
+/*******************************************************************************
+ * Copyright (c) 2016 Eurotech and/or its affiliates and others
```
amitjoy: 2017

amitjoy (changes requested):
Minor Changes

kura/org.eclipse.kura.core.net/src/main/java/org/eclipse/kura/core/net/WifiAccessPointImpl.java
djurod: Could you explain what you mean exactly?

I generated this with the Eclipse IDE, and as far as I understand the equal method needs to be explicitly overridden, otherwise the comparison of the contents is not done (only references are compared)...

kura/test/org.eclipse.kura.core.net.test/src/test/java/org/eclipse/kura/core/net/WifiAccessPointImplTest.java
djurod: The code was prepared in 2016, but I can fix it :)

kura/org.eclipse.kura.core.net/src/main/java/org/eclipse/kura/core/net/WifiAccessPointImpl.java
amitjoy: Yeah, you are right. Equals must be consistent with hashcode to have it work with hash-based data structures. The IDE generated implementations of these `equals` and `hashcode` methods are verbose and expose low-level complexities. That's why in Java 7, few useful static factory methods are provided to write your own `equals` and `hashcode` implementation in a cleaner way. Have a look at the `hash(Object...)` and `equals(Object, Object)` methods in `Objects`.

kura/test/org.eclipse.kura.core.net.test/src/test/java/org/eclipse/kura/core/net/WifiAccessPointImplTest.java
amitjoy: So it would be then - `Copyright (c) 2016, 2017 Eurotech and/or its affiliates and others`

kura/org.eclipse.kura.core.net/src/main/java/org/eclipse/kura/core/net/WifiAccessPointImpl.java
djurod: I checked this and also discussed it a little with @markoer, but in the end I came to a conclusion that the safest thing is to just generate this from Eclipse. It is less error prone than to manually create or update the methods. If any data member is changed you would just regenerate the methods using Eclipse...

amitjoy: I don't have any problem with the implementation that Eclipse IDE generates for `equals` and `hashcode`. I personally prefer using`Objects` to write less verbose implementations. It's just a syntactic sugar to the code. But if you feel that it would be easier for you to manage this way, I am definitely okay with that 👍 .

PR #998, opened 2017-01-11
Fix wires nulls

MMaiero:
The PR contains some bug fixes and NPE protection code, 
aligning the wires part to the latest changes in web2 bundle.


kura/org.eclipse.kura.wire.component.provider/src/main/java/org/eclipse/kura/internal/wire/publisher/CloudPublisher.java
```diff
@@ -377,7 +377,7 @@ private void publish(final List<WireRecord> wireRecords) {
 
         final AutoConnectMode autoConnectMode = this.options.getAutoConnectMode();
         final boolean autoConnectEnabled = this.dataService.isAutoConnectEnabled();
-        if ((AUTOCONNECT_MODE_OFF != autoConnectMode) && !autoConnectEnabled) {
+        if ((AUTOCONNECT_MODE_OFF != autoConnectMode) && autoConnectEnabled) {
```
amitjoy, line 380: Good catch 👍

kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/client/ui/wires/PropertiesUi.java
```diff
@@ -1106,6 +1131,48 @@ public void onValueChange(final ValueChangeEvent<String> event) {
         this.fields.add(formGroup);
     }
 
+    private TextBoxBase createTextBox(final GwtConfigParameter param) {
+        if (param.getDescription() != null && param.getDescription().contains("\u200B\u200B\u200B\u200B\u200B")) {
```
amitjoy, line 1135: This can be extracted as constant

amitjoy (commented):
Good refactoring 👍

kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/client/ui/wires/PropertiesUi.java
MMaiero: Yes, you're right.
Since we need to refactor this code to extend the AbstractServicesUI, I think it's better to track it in Issue #999 and do it once at that time.

amitjoy: 👍

MMaiero:
@amitjoy @cdealti if it's ok for you, I would merge it in order to have Kura ready for tomorrow's presentation.

amitjoy:
I would then suggest you to merge it. Go ahead with that.

PR #1001, opened 2017-01-12
Added Support For Blocking Queue to prevent out of memory issue

### LINKED ISSUE #1000: Memory leak and concurrency issue in 'EventHandlerServlet'
IMHO there is a huge memory leak and beside that a concurrency issue in the newly added `EventHandlerServlet`.

There is only a single queue for buffering events. This queue is unbounded and if no web session reads those events, then the queue will fill up until all memory is used up.

Beside that events which have been sent to one user won't be sent to others.

amitjoy:
Fixes #1000

### DIFF
--- a/kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/EventHandlerServlet.java
+++ b/kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/EventHandlerServlet.java
@@ -1,68 +1,95 @@
 /*******************************************************************************
- * Copyright (c) 2016 Eurotech and/or its affiliates and others
+ * Copyright (c) 2016, 2017 Eurotech and/or its affiliates and others
  *
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * which accompanies this distribution, and is available at
  * http://www.eclipse.org/legal/epl-v10.html
- * 
+ *
+ * Contributors:
+ *  Eurotech
+ *  Amit Kumar Mondal
+ *
  *******************************************************************************/
 package org.eclipse.kura.web;
 
+import static java.util.Objects.isNull;
+import static java.util.Objects.nonNull;
+import static java.util.Objects.requireNonNull;
+import static org.eclipse.kura.util.base.StringUtil.isNullOrEmpty;
+import static org.eclipse.kura.util.base.ThrowableUtil.stackTraceAsString;
+import static org.eclipse.kura.wire.WireSupport.EMIT_EVENT_TOPIC;
+import static org.osgi.service.event.EventConstants.EVENT_TOPIC;
+
+import java.io.Closeable;
 import java.io.IOException;
-import java.io.PrintWriter;
+import java.io.PrintStream;
 import java.util.Dictionary;
 import java.util.Hashtable;
-import java.util.Queue;
-import java.util.concurrent.ConcurrentLinkedQueue;
+import java.util.Map;
+import java.util.concurrent.BlockingQueue;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.LinkedBlockingQueue;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicReference;
 
 import javax.servlet.ServletException;
+import javax.servlet.ServletOutputStream;
 import javax.servlet.http.HttpServlet;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
+import javax.servlet.http.HttpSession;
 
-import org.eclipse.kura.wire.WireSupport;
 import org.osgi.framework.BundleContext;
 import org.osgi.framework.FrameworkUtil;
-import org.osgi.service.event.Event;
-import org.osgi.service.event.EventConstants;
+import org.osgi.framework.ServiceRegistration;
 import org.osgi.service.event.EventHandler;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
  * The Class EventHandlerServlet is responsible for interacting between Event
- * Admin and JS through Server Sent Events. This is mainly required for Kura
- * Wires to delegate the emit events.
+ * Admin and Javascript through Server Sent Events (SSE). This is mainly required
+ * for Kura Wires to delegate the emit events.
  */
 public final class EventHandlerServlet extends HttpServlet {
+
+    /** The Logger Instance. */
+    private static final Logger logger = LoggerFactory.getLogger(EventHandlerServlet.class);
+
+    /** Session Timeout in Seconds - 5 minutes */
+    private static final int MAX_INACTIVE_INTERVAL = 5 * 60;
+
+    /**
+     * Maximum size of the queue to maintain the events. The queue does not care if any
+     * event maintained by the queue is rejected. As long as a consumer thread can consume
+     * events from the queue, the system behavior is considered as expected.
+     */
+    private static final int MAX_SIZE_OF_QUEUE = 10;
 
     /** Serial Version */
     private static final long serialVersionUID = -8962416452919656283L;
 
-    /**
-     * Tracks all the changes associating a flag which signifies if the item has
-     * been read by JavaScript
-     */
-    private final Queue<String> m_sharedCache;
+    /** Bundle Context */
+    private BundleContext bundleContext;
 
-    /**
-     * Instantiates a new event handler servlet.
-     */
-    public EventHandlerServlet() {
-        this.m_sharedCache = new ConcurrentLinkedQueue<String>();
-        final Dictionary<String, Object> map = new Hashtable<String, Object>();
-        map.put(EventConstants.EVENT_TOPIC, WireSupport.EMIT_EVENT_TOPIC);
-        final BundleContext bundleContext = FrameworkUtil.getBundle(this.getClass()).getBundleContext();
-        bundleContext.registerService(EventHandler.class, new EventHandler() {
+    /** Used to track the new sessions */
+    private Map<String, HttpServletRequest> requests;
+    
+    /** {@inheritDoc} */
+    @Override
+    public void init() throws ServletException {
+        super.init();
+        this.bundleContext = FrameworkUtil.getBundle(this.getClass()).getBundleContext();
+        this.requests = new ConcurrentHashMap<>();
+    }
 
-            /** {@inheritDoc} */
-            @Override
-            public void handleEvent(final Event event) {
-                synchronized (bundleContext) {
-                    final String eventData = String.valueOf(event.getProperty("emitter"));
-                    EventHandlerServlet.this.m_sharedCache.add(eventData);
-                }
-            }
-        }, map);
+    /** {@inheritDoc} */
+    @Override
+    public void destroy() {
+        super.destroy();
+        this.requests.clear();
     }
 
     /**
@@ -80,18 +107,148 @@ public final class EventHandlerServlet extends HttpServlet {
     @Override
     public void doGet(final HttpServletRequest request, final HttpServletResponse response)
             throws ServletException, IOException {
+        // track the new session
+        final String sessionId = request.getParameter("session");
+        if (!isNullOrEmpty(sessionId)) {
+            this.requests.put(sessionId, request);
+        }
 
+        // set the response headers for SSE
         response.setContentType("text/event-stream");
         response.setCharacterEncoding("UTF-8");
         response.setHeader("Cache-control", "no-cache");
         response.setHeader("Connection", "keep-alive");
+        response.setHeader("Access-Control-Allow-Origin", "*"); // required for IE9
 
-        final PrintWriter writer = response.getWriter();
-        while (!this.m_sharedCache.isEmpty()) {
-            writer.write("data: " + this.m_sharedCache.poll() + "\n\n");
+        final HttpSession session = request.getSession(false);
+        // timeout is required if the browser beforeunload event is not fired in exceptional circumstances.
+        // If there exists a session in which a browser is opened, the beforeunload event is expected to
+        // be fired as soon as the browser or the tab is closed. But if by any chance, this beforeunload
+        // event is not fired, the timeout of a session will notice the disconnection.
+        session.setMaxInactiveInterval(MAX_INACTIVE_INTERVAL);
+
+        // destroy the session if exists. It actually destroys the current session which initiated
+        // the destruction request
+        final String sessionToDestroy = request.getParameter("logout");
+        if ((!isNullOrEmpty(sessionToDestroy)) && this.requests.containsKey(sessionId)) {
+            this.requests.remove(sessionId);
+            session.invalidate();
+            return;
         }
-        writer.flush();
-        writer.close();
+
+        final BlockingQueue<String> eventQueue = new LinkedBlockingQueue<>(MAX_SIZE_OF_QUEUE);
+        final ServletOutputStream outputStream = response.getOutputStream();
+        final PrintStream printStream = new PrintStream(outputStream);
+        final AtomicReference<Closeable> closeableReference = new AtomicReference<>(outputStream);
+
+        // the asynchronous task to retrieve and remove head element from the queue
+        final CompletableFuture<Void> elementRemovalFuture = CompletableFuture.runAsync(() -> {
+            try {
+                final boolean sessionExistsAndValid = this.requests.containsKey(sessionId)
+                        && this.requests.get(sessionId).isRequestedSessionIdValid();
+                // if session exists and valid, remove head element from the queue
+                while (sessionExistsAndValid) {
+                    final String data = eventQueue.poll(2, TimeUnit.SECONDS);
+                    if (isNull(data)) {
+                        printStream.print(":\n\n");
+                    } else {
+                        printStream.printf("data: %s%n%n", data);
+                    }
+                    printStream.flush();
+                }
+            } catch (final InterruptedException ex) {
+                logger.warn("Element removal timeout..." + stackTraceAsString(ex));
+                response.setStatus(HttpServletResponse.SC_OK);
+            }
+        });
+
+        final ServiceRegistration<?> registration = registerEventHandler(EMIT_EVENT_TOPIC, eventQueue, sessionId,
+                closeableReference, elementRemovalFuture);
+        // unregisters event handler service and close the servlet output stream
+        final CompletableFuture<Void> cleanupFuture = elementRemovalFuture.handle((ok, exeption) -> {
+            registration.unregister();
+            try {
+                outputStream.close();
+            } catch (final IOException ex) {
+                logger.warn("Outport Stream closing issue..." + stackTraceAsString(ex));
+            }
+            return null;
+        });
+
+        //returns the result if available or throw exception
+        try {
+            cleanupFuture.join();
+        } catch (RuntimeException re) {
+            //expecting if the queue is not processed
+        }
+    }
+
+    /**
+     * Registers the event handler
+     *
+     * @param topic
+     *            the topic to track changes
+     * @param eventQueue
+     *            the shared queue
+     * @param sessionId
+     *            the session ID to track
+     * @param output
+     *            the output stream
+     * @param future
+     *            the SSE task's future handle
+     * @return the service registration ID
+     * @throws NullPointerException
+     *             if any of the provided arguments is null
+     */
+    private ServiceRegistration<?> registerEventHandler(final String topic, final BlockingQueue<String> eventQueue,
+            final String sessionId, final AtomicReference<Closeable> output, final CompletableFuture<?> future) {
+        requireNonNull(topic, "Topic must not be null");
+        requireNonNull(eventQueue, "Provided Queue must not be null");
+        requireNonNull(sessionId, "Provided Session ID must not be null");
+        requireNonNull(output, "Output Stream must not be null");
+        requireNonNull(future, "Future reference must not be null");
+
+        // event handler properties
+        final Dictionary<String, Object> props = new Hashtable<>();
+        props.put(EVENT_TOPIC, topic);
+        // register the handler as a service instance
+        return this.bundleContext.registerService(EventHandler.class, event -> {
+            synchronized (EventHandlerServlet.class) {
+                final String eventData = String.valueOf(event.getProperty("emitter"));
+
+                // the event data can only be added to the queue if and only if it satisfies the following:
+                // 1. The initiated session request is valid (the session is not expired)
+                // 2. the queue has space to enqueue the event data
+                final HttpServletRequest servletRequest = this.requests.get(sessionId);
+                if (this.requests.containsKey(sessionId) && nonNull(servletRequest) && eventQueue.offer(eventData)) {
+                    return;
+                }
+
+                // session must be invalidated or it is already invalidated
+                this.requests.remove(sessionId);
+                if (nonNull(servletRequest)) {
+                    final HttpSession session = servletRequest.getSession();
+                    if (nonNull(session)) {
+                        session.invalidate();
+                    }
+                }
+
+                // It signifies that there exists no consumer thread (browser or tab closed).
+                // We must cancel the future instance that the current SSE worker thread
+                // working on.
+                final Closeable closeable = output.getAndSet(null);
+                if (isNull(closeable)) {
+                    // SSE worker thread already been killed
+                    return;
+                }
+                try {
+                    future.cancel(true);
+                    closeable.close();
+                } catch (final IOException ex) {
+                    logger.warn("Output Stream closing issue..." + ex);
+                }
+            }
+        }, props);
     }
 
 }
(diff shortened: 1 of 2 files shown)


amitjoy:
@ctron Done 👍

ctron (changes requested):
I am wondering if you have even thought this through or looked at the documentation of `LinkedBlockingQueue`.

1) The way you create an instance of `LinkedBlockingQueue` "limits" the queue to Integer.MAX_VALUE entries. I guess this could be seen as "unbounded" as well
2) What will happen in the case the queue is full? The OSGi Event mechanism will call a handler which just blocks forever.
3) It still doesn't solve the multi session issue

amitjoy:
Actually `LinkedBlockingQueue` is a complete bounded blocking queue (irrespective of its boundary condition which is limited to `Integer.MAX`). When the queue is full, it will not be blocked as the contract of `offer` method signifies that it will insert the specified element at the tail of this queue without exceeding the queue's capacity and if this queue is full, it will return false immediately. So, it would not incur any problem for that purpose.

ctron:
I missed that you are using 'offer' instead of add. Sorry for that.

Still issue 1 and 3 are still valid. A limit to Integer MAX is not a real limit, is it? And still one session will steal events from the other.

amitjoy:
The expected behaviour of this feature is to keep track of fired events by Producer Wire Components. The event handler will be able to track the fired events to show a graphical change if users are using the Kura Web Console. The system does not care if the user is not using the graphical user interface. In such case, the events can be rejected and will not be shown on the GUI.

The solution that I can think of is to use bounded queue to prevent OOM and the limit must be as minimum as possible. The multi-session issue is actually not applicable to this scenario as the system does not care when and how the events from the queue are consumed. Let's suppose, the user has created multi-sessions of Kura Web Console. This would result in creating multi event handlers. But in the OSGi runtime, there will only be one deployed Kura Wires graph that produces events so that these events will be put in the queue in order and the event handler servlets (in case of multi-session) will consume these events in order of insertion.

ctron:
So maybe you can explain how there will be multiple event handlers (servlets I suppose) be registered?

From what I am reading I thing you assume that the servlet is only registered when a session is active. And that there is one servlet instance per session. Is that assumption correct?

amitjoy:
The servlets are actually stored in memory and it is reused every time when the requested URL gets matched with the registered url pattern. Servlet container (in our scenario, the jetty bundle in Equinox) reuses the same servlet instance for every request to the server. This servlet is also shared among each request. To make the servlet execution thread safe, we must not assign session or request scoped data to servlet instance variables. According to user's personal intention, if the user extends this servlet to use `Single Thread Model`, there can exist multiple instances per JVM which could result in multiple events handlers to be registered. And also if you consider a distributed environment, actually multiple instances of a single servlet are created. A cluster of machines will be responsible for performing the request and hence, the machine can have a servlet instance of its own.

I actually thought you were talking about multiple servlet instances per JVM which could result in multiple sessions as well. As we do not need to process every event that is enqueued, the rejection of these events to the queue is acceptable. The consumer threads of the queue expect to process new events rather than the old events. That's the reason, I am also removing the old events as soon as the queue gets full.

ctron:
Yes but look at the location where you are removing events. This is in the constructor of the servlet. Which gets only called one the servlet is created.

I strongly suggest you check into the way servlets work and maybe sit down with Matteo to get it sorted out.

amitjoy:
My bad 😉! It should be in the `doGet()` so that before the elements from queue gets consumed, consumers should perform this check or if the queue is full and there are no consumers, the events will be rejected. This servlet is actually not a generic servlet that can be used with any event handlers. I developed this due to the need to apply Javascript communication with OSGi runtime as server sent events (SSE). The expected behaviour of the servlet is not to care about the missed or rejected events but to care about the ordering. Hence, for that purpose, the implementation must use a bounded queue as you advised (to prevent OOM) but at the same time, it should also conform to the behaviour of processing newer events rather than all the events triggered by the producers.

ctron:
So this change will now only trigger the cleanup when a client makes GET call. So we still run into an OOM situation since the cleanup will now only happen when a client is connected?!

So if you want to have a bounded FIFO is suggest using a simple linked list. And limiting the size when adding a new element.

But this all doesn't handle the multi session/multi tab issue. Since there is only one servlet instance per Kura process.

amitjoy:
To preserve the order, we still can use `LinkedList` as it is also retrofitted to work as a `Deque` implementation. Still it is not preferable as `LinkedList` is not thread-safe and its iterators are also fail-fast. On the other hand, `LinkedBlockingQueue` has snapshot iterators which can work in a multi-threaded environment without any issue.

As we are limiting the size of the `LinkedBlockingQueue`, only 10 elements at max can be added to the queue. The cleanup occurs if consumer thread (worker on `doGet()` method) consumes data from the queue. If there does not exist any consumer thread, it will not result in OOM as only max of 10 elements can be in the queue and any data or event produced by producer thread will be rejected if the queue is already full.

If we open multiple tabs, it would result in multiple consumer threads working on the same servlet instance. The clean up will still work as we are using `LinkedBlockingQueue` which would work in such multi-threaded environment.

The another way to clean up the queue can also be handled if we use a `ScheduledExecutorService` to schedule worker threads to clean up the queue in a time based fashion but I would still prefer cleaning up the queue if and only if there exists any consumer thread to consume event data from the queue. Otherwise the scheduled worker threads will always the active as long as Kura process is up and running.

kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/EventHandlerServlet.java
```diff
@@ -35,32 +38,36 @@
...
-            @Override
-            public void handleEvent(final Event event) {
-                synchronized (bundleContext) {
-                    final String eventData = String.valueOf(event.getProperty("emitter"));
-                    EventHandlerServlet.this.m_sharedCache.add(eventData);
-                }
+
+        // register the handler as a service instance
+        bundleContext.registerService(EventHandler.class, event -> {
+            synchronized (bundleContext) {
```
ctron: Maybe you can explain why you are synchronizing here.

kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/EventHandlerServlet.java
```diff
@@ -81,14 +88,20 @@ public void handleEvent(final Event event) {
     public void doGet(final HttpServletRequest request, final HttpServletResponse response)
             throws ServletException, IOException {
 
+        // if the queue is full, clear it as we don't need to care about rejected events
+        if (this.sharedCache.size() == MAX_SIZE_OF_QUEUE) {
+            this.sharedCache.clear();
```
ctron: If you clear the queue here. It will be empty a few lines below. Why that? Is the upper bound not enough?

ctron (changes requested):

ctron:
The cleanup will work, yes. But one tab/session would steal events from the other.

kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/EventHandlerServlet.java
amitjoy: You are right. It is not needed and a synchronization overhead as well.

kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/EventHandlerServlet.java
amitjoy: Let's suppose there does not exist any consumer thread but producers are producing event data continuously. So, max of 10 elements will be there in the queue. Now, a consumer thread appears and it now calls `doGet()` on the servlet. According to the requirement of this feature, the consumer thread is not responsible to consume stale event data that are already in the queue. It, hence, first clears the queue and waits for new event data so that it can consume thereafter.

The requirement is that as soon as an user creates a Kura Wires `Wire Graph` and deploys it in the runtime, the data produced by Producer nodes in the `Wire Graph` will fire an event which will be reflected in the `WireGraph` in real-time.  The connections or arrows in the `Wire Graph` will glow as the consumer thread simultaneously is consuming data from the queue.

In summary,

Producer node or `Wire Component` produces data -> it fires an event -> the event is inserted in the queue -> the consumer consumes the data from the queue -> the arrow in the `Wire Graph` glows

ctron:
Or in other words, when , between get calls, there are more than 10 events being generated nothing will "glow" at the moment.

ctron:
So a real solution to this could be to use the "GwtEventServiceImpl" for this as well.

Or, as you only seems to require the "last activity timestamp", just update an internal map of the latest activity and get back a state from that. Seems sufficient for your use case. And this would also solve the multi tab issue.

amitjoy:
`GwtEventServiceImpl` is mainly targeted to be used in conjunction with other Kura Specific Servlets which use XSRF tokens to validate a request. Kura Wires internally uses several bare JS codes which are outside the boundary of Kura Servlets (`OsgiRemoteServiceServlet`). That's why I had to use simple `HttpServlet` for that purpose.

amitjoy:
Applied some changes. Have a look at it.

kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/EventHandlerServlet.java
```diff
@@ -1,32 +1,43 @@
...
+ * Copyright (c) 2016, 2017 Eurotech and/or its affiliates and others
  *
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * which accompanies this distribution, and is available at
  * http://www.eclipse.org/legal/epl-v10.html
- * 
+ *
+ * Contributors:
+ *  Amit Kumar Mondal
```
ctron: I just saw this. I think we agreed to use "Eurotech" in your case here.

kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/EventHandlerServlet.java
```diff
@@ -80,18 +74,87 @@ public void handleEvent(final Event event) {
...
+                }
+                // the queue is getting full. It signifies that there exists no consumer
+                // thread (browser closed). We must kill the current SSE thread.
+                final Closeable closeable = output.getAndSet(null);
+                if (closeable == null) {
+                    // thread already killed
+                    return;
+                }
+                try {
+                    thread.interrupt();
```
ctron: This thread belongs to jetty an must not be killed.

kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/EventHandlerServlet.java
```diff
@@ -80,18 +74,87 @@ public void handleEvent(final Event event) {
...
+        final Thread thread = Thread.currentThread();
+        final BlockingQueue<String> eventQueue = new LinkedBlockingQueue<>(MAX_SIZE_OF_QUEUE);
+        final OutputStream outputStream = response.getOutputStream();
+        final PrintStream printStream = new PrintStream(outputStream);
+        final AtomicReference<Closeable> closeableReference = new AtomicReference<>(outputStream);
+        final ServiceRegistration<?> registration = register(EMIT_EVENT_TOPIC, eventQueue, closeableReference, thread);
+
+        try {
+            while (true) {
+                final String data = eventQueue.poll(2, TimeUnit.SECONDS);
```
ctron: This thread belongs to jetty and must not block.

ctron:
OK. I am sorry but that approach doesn't work. Servlets must not interfere with the threading of the servlet container. So you must not interrupt threads or block. What you can use is async servlet processing.

ctron (changes requested):

kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/EventHandlerServlet.java
amitjoy: The idea was to provide a servlet that allows javascript clients to see the Event Admin events which blocks the request thread until it gets killed. Asynchronous task processing will be beneficial in this scenario. As far as I am aware of `Asynchronous Servlet Processing` does not provide any option to cancel the ongoing task. I have to check it once. I am also thinking of using `CompletableFuture` which provides asynchronous processing with more advanced functionalities. I will make the necessary amendments and will let you know.

amitjoy:
The idea was to provide a servlet that allows javascript clients to see the Event Admin events which blocks the request thread until it gets killed. Asynchronous task processing will be beneficial in this scenario. As far as I am aware of Asynchronous Servlet Processing does not provide any option to cancel the ongoing task. I have to check it once. I am also thinking of using CompletableFuture which provides asynchronous processing with more advanced functionalities. I will make the necessary amendments and will let you know.

ctron:
Why do you want to cancel an ongoing task?

amitjoy:
If the browser or the tab is closed, there will not exist any SSE consumer thread. Hence, there is no need for that task to continue working.

ctron:
IIRC you will get an IOException in this case an can stop listening.

amitjoy:
I am canceling the task that is responsible for polling element from the queue. The true param would interrupt the task first so even if it is using Servlet Output Stream for that moment, it will be interrupted. And thereafter I am closing the stream. I have also checked with multiple browsers and multiple tabs to check how it behaves when multiple tasks are working in the background in asynchronous way and also if the browser or tab is closed, the respective task for that session is getting cancelled.

kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/EventHandlerServlet.java
amitjoy: This is because I have written this code as an Eurotech intern and I have now finished my internship. Right now I am contributing as an individual. According to our previous discussion, the copyright header signifies that the contribution has been made from Eurotech as well as others. And contributors section is an optional element that can comprise the name of the contributors.

ctron: Do then it would be both, Eurotech + you

ctron:
Wouldn't it be easier to directly write to the stream when an event occurs? Then you would not need a queue and let the web server handle all this.

Or use some existing SSE implementation instead creating your own.

ctron:
Or maybe we sockets?!

amitjoy:
SSE is an event source mechanism through which server can send data to a web page at any time, by pushing messages to the web page. It does not require any further server to set up as we need for web sockets. Yes, you are right. We can do it using Web Socket but due to an external server configuration and set up, I would prefer the default mechanism with SSE. SSE is actually very simple to handle in server side as well as in client side that we are also doing by providing necessary response parameters so that javascript event source can continuously receive the updates from server. But in our scenario, it is a bit different as we need to take care of the multi-session issue as you pointed in case the tab is closed or the browser is closed. In addition, the data that the server needs to push to client is generated continuously or in a time-based fashion. That is the reason, the implementation needed to consider such scenarios.

ctron:
If you close a tab the web socket will also be closed IIRC. 
For example you could use WebSockets like [1], there should also be the `@OnError` annotation which should indicate a closed tab.

In this latest changes it looks to me like as if you are filling up the session state now. Keeping `HttpRequest`s. For example you cannot rely on the fact that `beforeonunload` gets called. Because that may never happen.

But in the end I don't think I can help you here. I can only suggest to use something existing and tested instead of re-inventing what is already there.

[1] https://github.com/bndtools/bndtools/wiki/Websocket-with-jetty-9-and-osgi

amitjoy:
As you mentioned, there could be exceptional circumstances in which the browser event `beforeunload` in not fired and I am relying on it to cancel an ongoing task. In normal situation (the user uses browser(s) to open Kura Wires and closes the browser(s) anytime), the reliance on `beforeunload` event to invalidate a session will work as usual. In considering the exceptional scenario in which the `beforeunload` browser event is not fired by browser, I have also added session inactive timeout to the established session. It would be useful in such a scenario, in which the `beforeunload` event is not fired to invalidate a session. The session will, hence, wait for the event if the browser fires the event expectedly and also wait for the session inactivity for 10 minutes at max. If no activity is performed in that time duration, the session will be invalidated automatically. This will then trigger cancellation of the ongoing tasks that are not needed anymore.

You are right. Web Sockets is a great solution for such scenarios, but it depends on managing a Web Sockets server. On the other hand, Server Sent Events is an inbuilt feature in the browser so that it can receive updates easily from a server. For now, I would prefer SSE for event updates to be pushed to Javascript clients easily. In future, we can investigate more on such use cases for Web Sockets feature in Kura.

amitjoy:
I have developed a bundle to push OSGi events to Event Source. 

https://github.com/amitjoy/kura/commit/e6a307877574b4e0e8779fe34158726950a8c1fa

It uses Jetty Event Source implementation. I tried with Kura Wires use case but it does not work properly. The event data are delivered sometimes all at once for optimization which makes it difficult to process in javascript clients.

@ctron Could you kindly have a look at it and suggest something if you find any potential issue?

amitjoy:
As we are approaching towards the release of Kura Wires, I would suggest to have this merged as it is working as expected and it also solves the problems discussed earlier. Even though it is a ground-up implementation, it is a better solution than its previous version.

I would appreciate more feedbacks. @ctron @MMaiero @dwoodard1 @cdealti

MMaiero:
@ctron Any thoughts on this?
I had a look at the code and I've not found remarkable issues. I think we can merge and, eventually, apply patches or adopt a new solution later on.

ctron:
@MMaiero If you think it is ok, then merge it.

cdealti:
However, I believe that this cosmetic thing has already consumed too many cycles.
And that we probably need to address this in the more general context of a reactive web ui.
I don't know what is the best solution in general and I believe that @amitjoy implementation is clever:

1. If no browsers are observing the graph, no event handler is registered so the Event Admin does not need to deliver events. However note that an event handler is not unregistered if the browsers are closed. Also, it requires some session management and cleanup and I'm not sure it covers all the cases.

2. Streaming events to the client is cool but I would like to avoid the overhead of the extra executors

So, if we can remove the extra executors and make sure that we unregister the handlers I think we can temporarily accept it.

amitjoy:
@cdealti The cases that you pointed out have been considered as well. As soon as `elementRemovalFuture` (used to dequeue events) is complete, `cleanupFuture` (used to perform clean up actions needed if the event handler is inactive) is invoked which unregisters the event handler and closes the inactive output stream. The `elementRemovalFuture` can be complete either forcefully by cancelling it or if it is successful. As soon as browser is closed, the session is destroyed and the `elementRemovalFuture` is cancelled which triggers the deregistration of the event handler service.

This implementation only uses a single cached thread pool which handles all the operations internally to register event handlers and stream events to the browser. Without thread pool, we have to deal with low-level threading complexities of synchronization between removing elements from the queue and cleaning up event handlers.

cdealti:
> As soon as elementRemovalFuture (used to dequeue events) is complete, cleanupFuture (used to perform clean up actions needed if the event handler is inactive) is invoked which unregisters the event handler

You are right. I've missed the unregister.

> Without thread pool, we have to deal with low-level threading complexities of synchronization between removing elements from the queue and cleaning up event handlers.

I'm pretty sure that moving the loop outside the async task and surrounding it with a "try {} finally {}" will work. You wait for a while on the poll. When the poll returns you check your session again, if it's valid go ahead writing the stream and poll the queue, if not you are done.

As you know this stuff needs to run also on crappy hardware @433MHz and 512MB of RAM (or even less).
One thread requires a few hundreds kB of RAM for the stack which we can save (along with CPU cycles).

amitjoy:
If I move the async task and write the loop inside a try-catch would result in a blocking behaviour and the servlet would be blocked for further session processing. If one session is not completed before another gets initiated, the servlet would not be able to respond. That is the reason, it has been done in an async way.

cdealti:
your servlet doGet() method is already blocked waiting for ``cleanupFuture.get()``.
So what is the point to start an async execution if you then have to wait for its completion in the same method/stack?
Servlets already support concurrency and doGet methods can be called concurrently.
Then, unless a miss something obvious, the async execution is not needed.

amitjoy:
You are right. As `CompletableFuture` is also an implementation of Future API and according to the `get()` contract, it would block until any value gets available for the consumer thread to process. The motive of making it asynchronous here, is that the processing in the while loop can take longer to process if any data can not be polled from the queue. So, we needed to handle these operations asynchronously. Servlets process each session in single thread and that thread is a handle for a client. So, a client can initiate the servlet request and respective thread will be blocked if the data cannot be polled from the queue. Hence, the asynchronous approach has been considered for clients to process events from the queue in a non-blocking and reactive stream-oriented fashion. 

In short, every session or client with the servlet will asynchronously process events from the queue and no client will ever be blocked even if the associated servlet thread takes longer to process events from the queue.

The creation of extrac thread pools has been removed as JVM will handle all the asynchronous tasks now in a `ForkJoin` thread pool internally. The session invalidation timeout has also been reduced to 5 mins for performance. And the blocking wait has been replaced with asynchronous result delivery if any data is available in the queue.

cdealti:
Now it's really super complex. Can we restart from the beginning?

You want to stream OSGi events to Javascript creating for every session/browser a producer and a consumer synchronized through a queue.

As a first approximation you could neglect sessions and do this for every doGet() called.

How many concurrent doGet calls do you expect per browser instance?

In the doGet, you keep reading  from the queue (with timeout) and write the event to the response output stream **synchronously**.
I believe that if the browser is closed the socket will be closed too and writing an event to the stream should throw an exception breaking the loop and causing the doGet to return. 

Have you tried this?

Otherwise, if you really need to detect if this session is no longer valid, would be better to register an ``HttpSessionListener`` and cause the doGet to return cleaning up things?

amitjoy:
Few things that we have to consider to be on the same page:

1. The user open wires and creates a Wire Graph
2. This would trigger a session in the servlet
3. Now every client and the event producer are synchronized on their own instance of queue
4. This synchronization leverages multiple clients to not steal events from other queues
5. As soon as a WireGraph is created, a session is created, a queue is created and an event handler service is registered
6. The Wire Component emits events and the events are inserted into the queue and the respective event handler removes the element from the queue
7. This happens for every client connections
8. An user now opens a new tab or browser and open Wires Composer UI
9. The process is continued and all sessions are managed by the servlet and every session has respective event handler and a queue to poll data from it
10. Now, let's suppose, I have created multiple sessions, let's say 5
11. Now I close 2 of them. While closing, the browser will inform the servlet to close the respective sessions (this occurs through browser events - `browserUnload` )
12. This results in destroying the respective queue, stopping the respective task and deregistering the respective event handler service
13. Servlet is a client-server architecture so, to close a connection, the initiative **must** be taken by the client who want to be disconnected
14. If such initiative is not taken by the client, the data will continually be produced and kept in the queue and the event handler service will never be deregistered as these are handled by JVM threads in a `ForkJoin` thread pool which the servlet do not own. So, these thread will continue to work. And destroying the servlet will not even destroy these threads as they are not owned by the servlet.
15. That's why closing an already established communication is initiated by client and the respective queue for that connection is destroyed, the respective event handler is also deregistered and the asynchronous task is cancelled.
16. In extreme scenario, let's say the browser doesn't fire the `browserUnload` event which it always fires while closing a tab or browser.
17. In this scenario, closing an invalid session is handled through session timeouts (never encountered this situation though but has been taken into consideration)
18. `HttpSessionListener` would have helped but timeout is easy to implement and it works out of the box. As I mentioned, while closing browser or a tab, `browserUnload` event gets fired and javascript client informs the servlet to close the respective task and deregister the event handler service. It always happens in all the browser, Safari, Chrome and Opera (As I have tested in these 3 so far). `browserUnload` event is fired by all the browsers and that's why the session timeout is of no use so far. Still I have included session timeout to consider extreme cases such as a new browser in market which do not support `browserUnload` event.

I have throughly tested this application and benchmarked the creation and deletion of queues, cancellation of respective tasks while closing tab and browser and deregistration of respective event handlers. It has been implemented from scratch to stream events to javascript client code using server sent events. Streaming event admin events to javascript clients is a complex procedure as we have to synchronize properly with the producer and consumer queues and also we need to consider multiple sessions would have to be sysnchronized properly and no session will be able to steal events from other. Incorporation of such features resulted in asynchronous task in this servlets in which every session has its own associated task to be performed by JVM `ForkJoin` thread pool, own queue and event handler instance that consumes data from the associated queue.

cdealti:
So, 50 posts on this are enough. I give up.
I still believe your use cases do not exist. You will have 0 or 1 browser open. Usually no browser at all is open. Multiple browser should be supported but this use case should not be optimized to the point of sacrificing code clarity.
If you insist that is the proper fix to the original bug an since we need a fix this is going to be merged.

amitjoy:
🤣 Yeah, the discussions are on the rise! Yes, you are right. According to the Industrial IoT environment, we can expect only 2-3 browsers or tabs are opened to access Wires Composer UI. I believe we still need to develop a robust system that covers all the corner use cases in which we can also expect several tabs or browsers are opened and there will not be any event data stealing as well.  We have taken this asynchronous approach as asynchronous systems offer better performance and parallelism is easier to exploit. This approach, at the first glance, looks a bit complex but is better and more optimized than the  existing one.

MMaiero:
I don't want to reopen this, but I'm performing a long running test on a Kura installation and this is the result after few hours of working. Currently, there are no open browser sessions. 

schermata 2017-02-13 alle 11 50 13

The existing CompletableFuture and ForkJoinTask seem all related to the EventHandlerServlet. Current usage seems pretty high.

ctron:
I would re-open this.

PR #1005, opened 2017-01-13
Fix possible exception and service leak

ctron:


amitjoy:
It is definitely a good implementation 👍  as it uses the default java libraries to perform the same task. I am curious to know about the possible service leak that's lurking behind. According to the previous implementation, if we have strict bundle start levels for every bundle in which `CryptoService` will be registered before it is actually consumed by others. If we don't follow strict bundle level for every bundle, `CryptoService` can get registered before consumers initiate consuming. Hence, `ServiceLocator` cannot capture the `CryptoService` service reference.

ctron:
The service is get but never unget. In addition it solves the start level issue you described.

MMaiero:
I agree on using default java 8 base 64 since now available.
Regarding the old implementation, for what I remember, the web bundle has a DS dependency to the CryptoService. So the web bundle wouldn't start until the CryptoService dependency was resolved.

PR #1007, opened 2017-01-15
Fix #932 by splitting up implementation from API

### LINKED ISSUE #932: org.eclipse.kura.asset.provider provides API and service implementation
The bundle `org.eclipse.kura.asset.provider` provides base classes and a service implementation at the same time. That way it is impossible to exchange the service implementation with a different implementation.

As we are using OSGi we should follow the idea of splitting up API/base classes and internal service implementation.

Currently the "web" bundle requires a reference to `org.eclipse.kura.asset.provider` in order to use `BaseChannelDescriptor` and thus Kura an only be used with the single implementation of `org.eclipse.kura.asset.provider`.

ctron:


ctron:
According to @amitjoy this is ok to be merged. See https://github.com/eclipse/kura/issues/932#issuecomment-272694226

@dwoodard1 could you please perform the merge

PR #1011, opened 2017-01-16
Fix launch config to start ".web2" instead of ".web" bundle

ctron:


MMaiero:
Conflict here. Could you solve it? Thanks

MMaiero:
I think that this was already merged with commit 6bd0b93da91a2ab3f941ca67b8fd11356b1d47ad, right?

MMaiero:
@ctron Ping

PR #1019, opened 2017-01-16
Fix private members

ctron:


cdealti:
@dwoodard1 can we agree that this kind of changes do not require a PR?

MMaiero:
If the changes does not modify the code behavior, I agree.

dwoodard1:
I agree. However, if we are modifying code for this purpose, shouldn't we also clean up the other style issues. For example, the use of m_* for member names? But, that shouldn't need a PR either. Since it is not modifying code behavior.

cdealti:
I agree.

ctron:
Can we do this outside of a PR as Cristiano suggested?

dwoodard1:
Yes

PR #1025, opened 2017-01-18
Bring back Fedora distribution by adding Fedora 25

ctron:
This changes fixes the broken Fedora distribution by adding Fedora 25.

As Fedora 25 is out with official support for the Raspberry Pi and there
is no update anymore on the Fedberry distribution [1] I added proper
support for Fedora 25 based on the Fedberry release which we already
had.

[1] https://github.com/fedberry/fedberry/releases


ctron:
I think the next logical step would be to remove the `fedorapi` profile.

kura/distrib/src/main/resources/fedora25/README.md
```diff
@@ -0,0 +1,26 @@
...
+      xzcat ~/Downloads/fedora/Fedora-Minimal-armhfp-25-1.3-sda.raw.xz | sudo dd bs=1m of=/dev/rdisk2
+      ```
+1. Load the sdcard above and boot the RPi.
+1. Scan your network for the RPi (if using DHCP).
+      ```
+      sudo nmap -sP 192.168.0.0/24 | awk '/^Nmap/{ip=$NF}/B8:27:EB/{print ip}'
+      ```
+1. SSH into the RPi. You will need to be plugged into a wired ethernet network at this point,
+but after this step you could configure wifi and repeat the discovery step.
+1. Grow the root file system as described [here](https://fedoraproject.org/wiki/Architectures/ARM/F25/Installation#Resize_the_Root_Filesystem)
```
MMaiero, line 18: I would add a reference to the section named "How do I enlarge the root partition?" in the FAQ. It was useful for me as I was using the command-line and I had no clue on how to expand the root partition using only commands.

ctron: That should be in the last line, #18. Unless I didn't I missed something.

ctron: Unless I missed something. ... I hate auto correction.

MMaiero: Maybe I didn't get the point.
This is where I land when clicking the link in line 18:

schermata 2017-01-25 alle 14 49 59

This is what I needed to complete the step via command line. It is in the subsequent FAQ section:

schermata 2017-01-25 alle 14 50 15

PR #1028, opened 2017-01-18
Work around Kura sending empty string instead of null for Camel deps

ctron:
Once the service has been configured there is now way
with Kura to reset the value (to unset it) although it is optional.
In this case the value will be reported as "" (empty string)
instead of `null`.

This causes the Camel component to look for a component with an empty
name, which never exists.


kura/distrib/config/kura.build.properties
```diff
@@ -13,7 +13,7 @@ org.eclipse.kura.deployment.agent.version=1.0.7
 org.eclipse.kura.core.deployment.version=1.1.2
 org.eclipse.kura.camel.version=1.1.0
 org.eclipse.kura.camel.cloud.factory.version=1.0.0
-org.eclipse.kura.camel.xml.version=1.0.0
+org.eclipse.kura.camel.xml.version=1.0.2-SNAPSHOT
```
MMaiero: Why you need here to move and fetch 1.0.2-SNAPSHOT?
Since it was not distributed with 2.1.0, shouldn't it be 1.0.1-SNAPSHOT?

ctron:
Updated and rebased

PR #1030, opened 2017-01-18
Allow using JMX with default Camel XML router

ctron:
This change enables the default Camel XML router to activate the JMX
integration. This default is set to "enable JMX". It can be deactivated.


hekonsek:
Agreed. Many thanks, Jens!

PR #1031, opened 2017-01-18
Fix wire component provider

MMaiero:
Applied some changes in the way of getting the cloud service for the wire publisher and subscriber.


kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/cloud/CloudClient.java
```diff
@@ -124,6 +124,33 @@
...
+     * @param payload
+     *            An KuraPayload representing the message to be published
+     * @param qos
+     *            An integer specifying the quality of service the message was published on.
+     * @param retain
+     *            Whether or not the broker should retain the message
+     * @return The published message's ID.
+     * @since {@link org.eclipse.kura.cloud} 1.1.0
+     */
+    public int publish(String deviceId, String appTopic, KuraPayload payload, int qos, boolean retain)
```
amitjoy: Document the exception

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/cloud/CloudClient.java
```diff
@@ -162,6 +189,47 @@ public int publish(String appTopic, KuraPayload payload, int qos, boolean retain
...
+     *            An integer specifying the quality of service the message was published on.
+     * @param retain
+     *            Whether or not the broker should retain the message
+     * @param priority
+     *            Relative ordering of this message with other messages that may be currently queued for publishing.
+     * @return The published message's ID.
+     * @since {@link org.eclipse.kura.cloud} 1.1.0
+     */
+    public int publish(String deviceId, String appTopic, KuraPayload payload, int qos, boolean retain, int priority)
+            throws KuraException;
```
amitjoy: Document the exception

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/cloud/CloudClient.java
```diff
@@ -199,6 +267,47 @@ public int publish(String appTopic, KuraPayload payload, int qos, boolean retain
...
+     *            An integer specifying the quality of service the message was published on.
+     * @param retain
+     *            Whether or not the broker should retain the message
+     * @param priority
+     *            Relative ordering of this message with other messages that may be currently queued for publishing.
+     * @return The published message's ID.
+     * @since {@link org.eclipse.kura.cloud} 1.1.0
+     */
+    public int publish(String deviceId, String appTopic, byte[] payload, int qos, boolean retain, int priority)
+            throws KuraException;
```
amitjoy: Document the exception

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/cloud/CloudClient.java
```diff
@@ -338,6 +447,24 @@ public int controlPublish(String deviceId, String appTopic, byte[] payload, int
...
+     *
+     * @param deviceId
+     *            A String specifying the device ID.
+     * @param appTopic
+     *            A String object containing the application topic.
+     * @param qos
+     *            An int containing the Quality of Service.
+     * @since {@link org.eclipse.kura.cloud} 1.1.0
+     */
+    public void subscribe(String deviceId, String appTopic, int qos) throws KuraException;
```
amitjoy: Document the exception

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/cloud/CloudClient.java
```diff
@@ -366,6 +511,22 @@ public int controlPublish(String deviceId, String appTopic, byte[] payload, int
     public void unsubscribe(String appTopic) throws KuraException;
```
amitjoy, line 539: Document the exception

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/cloud/CloudClient.java
```diff
@@ -353,6 +480,24 @@ public int controlPublish(String deviceId, String appTopic, byte[] payload, int
...
+     *
+     * @param deviceId
+     *            A String specifying the device ID.
+     * @param appTopic
+     *            A String object containing the application topic.
+     * @param qos
+     *            An int containing the Quality of Service.
+     * @since {@link org.eclipse.kura.cloud} 1.1.0
+     */
+    public void controlSubscribe(String deviceId, String appTopic, int qos) throws KuraException;
```
amitjoy: Document the exception

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/cloud/CloudClient.java
```diff
@@ -366,6 +511,22 @@ public int controlPublish(String deviceId, String appTopic, byte[] payload, int
...
+     * This is a synchronous call. If the unsubscribe fails, an exception will be thrown
+     * that will contain information about the cause of the failure.
+     *
+     * @param deviceId
+     *            A String specifying the device ID.
+     * @param appTopic
+     *            A String object containing the application topic.
+     * @since {@link org.eclipse.kura.cloud} 1.1.0
+     */
+    public void unsubscribe(String deviceId, String appTopic) throws KuraException;
```
amitjoy: Document the exception

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/cloud/CloudClient.java
```diff
@@ -379,6 +540,22 @@ public int controlPublish(String deviceId, String appTopic, byte[] payload, int
...
+     * This is a synchronous call. If the unsubscribe fails, an exception will be thrown
+     * that will contain information about the cause of the failure.
+     *
+     * @param deviceId
+     *            A String specifying the device ID.
+     * @param appTopic
+     *            A String object containing the application topic.
+     * @since {@link org.eclipse.kura.cloud} 1.1.0
+     */
+    public void controlUnsubscribe(String deviceId, String appTopic) throws KuraException;
```
amitjoy: Document the exception

ctron:
Maybe you could explain a little bit about the new concepts which got introduced, how implementors of the CloudService API should handle this. I guess throwing an OPERATION_NOT_SUPPORTED exception is not a proper way to do it?!

amitjoy (changes requested):
Exception Documentation

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/cloud/CloudClient.java
MMaiero: It came out as a copy paste of the existing apis. Just adding the device Id. But I agree, we need to update the documentation with the description of the exception. I'll do that.

kura/org.eclipse.kura.wire.component.provider/src/main/java/org/eclipse/kura/internal/wire/publisher/CloudPublisher.java
```diff
@@ -54,44 +55,66 @@
...
+            }
+        }
+
+        @Override
+        public void removedService(ServiceReference<CloudService> reference, CloudService service) {
+            CloudPublisher.this.cloudService = null;
+        }
+    }
 
-    /** The Logger instance. */
```
amitjoy, line 62: Please add the relevant comments to the instance and class variables. It will be consistent with the other part of the code.

kura/org.eclipse.kura.wire.component.provider/src/main/java/org/eclipse/kura/internal/wire/subscriber/CloudSubscriberOptions.java
```diff
@@ -62,14 +78,59 @@ int getSubscribingQos() {
...
+
+    /**
+     * Returns the deviceId to be used for message subscription.
+     *
+     * @return the deviceId
+     */
+    String getSubscribingDeviceId() {
+        String subscribingDeviceId = DEFAULT_DEVICE_ID_TOPIC;
+        final Object deviceId = this.properties.get(CONF_DEVICE_ID_TOPIC);
+        if ((this.properties != null) && this.properties.containsKey(CONF_APP_TOPIC) && (deviceId != null)
```
amitjoy: Shouldn't it be CONF_DEVICE_ID_TOPIC?

amitjoy (changes requested):
Some non-time consuming fixes 😉

MMaiero:
The changes applied are described in the commits comments.
In brief, the CloudClient API was changed introducing an overloading of the existing methods, in order to have the possibility to specify the deviceId when, for example, publishing or subscribing. This allowed to uniform the API since there existed two methods `public int controlPublish(String deviceId, String appTopic, KuraPayload payload, int qos, boolean retain,
            int priority) throws KuraException;` and `public int controlPublish(String deviceId, String appTopic, byte[] payload, int qos, boolean retain, int priority)
            throws KuraException;` that allowed to specify the deviceId, while the others had not this option.

This change affected various components (cloud, camel, wires) so I've opened this PR a little before than usual in order to give you guys the opportunity to have a look and see if something needs to be changed.

Regarding implementors, as you have probably seen, I've proceeded modifying the CloudService API minor, that moved to 1.1. This required the change of all the CloudService API implementors, first of all moving imported version from [1.0,1.1) to [1.1,1.2). Furthermore I've implemented the proper code in `CloudServiceImpl` and `wire.component.provider`.
Regarding Camel implementation, the change in CloudService API affected `core.cloud bundle`. Here, for the new APIs, I've implemented all was possible, leveraging the existing code (see for example `public int publish(String deviceId, String appTopic, KuraPayload payload, int qos, boolean retain)`). For the parts that required a non trivial change, I've decided to not start changing code without knowing the rationale behind. I've preferred, instead, to insert the overriding methods, adding a note ` // TODO Check if this needs to be implemented`, that for sure you have seen, and return a `KuraException(OPERATION_NOT_SUPPORTED)`. 

I think this allows the code to compile, keeps the code working with existing clients (they are not aware of added methods) and updated consumers of CloudClient services will simply receive a KuraException. But, since this is specified in the method signature, we can suppose that those clients are properly designed to manage that sort of exceptions. 

Furthermore, I see that in 

- https://github.com/eclipse/kura/blob/develop/kura/org.eclipse.kura.camel/src/main/java/org/eclipse/kura/camel/internal/camelcloud/CamelCloudClient.java#L188 

- https://github.com/eclipse/kura/blob/develop/kura/org.eclipse.kura.camel/src/main/java/org/eclipse/kura/camel/internal/camelcloud/CamelCloudClient.java#L193

you used the same principle.

If you think is needed, we can work together and apply the needed changes in the newly added methods, in order to enable the Camel Cloud Client implementation to support deviceId in subscribe and unsubscribe methods.

kura/org.eclipse.kura.wire.component.provider/src/main/java/org/eclipse/kura/internal/wire/subscriber/CloudSubscriberOptions.java
MMaiero: My bad. I'll change it now

kura/org.eclipse.kura.wire.component.provider/src/main/java/org/eclipse/kura/internal/wire/subscriber/CloudSubscriberOptions.java
```diff
@@ -66,8 +66,7 @@
     int getSubscribingQos() {
         int subscribingQos = DEFAULT_QOS;
         final Object qos = this.properties.get(CONF_QOS);
```
amitjoy, line 75: A null check on `properties` would be beneficial either in all getters or in the constructor.

amitjoy (changes requested):

kura/org.eclipse.kura.wire.component.provider/src/main/java/org/eclipse/kura/internal/wire/subscriber/CloudSubscriberOptions.java
MMaiero: Since the configuration-policy is "require", properties will never be null.
I can agree with you that a best practice will be to have a check for properties==null but, as was the code before, if properties was null, every get method would end up with a NullPointer due to the this.properties.get call before properties!= null check the following line.

amitjoy: You are right but this is a custom class that the OSGi runtime does not have any hold onto. Consumers of this class can pass custom properties by creating a map and pass it directly to the instance of this class which is also acceptable. I believe, it would be better to specify null checks before we process the provided map otherwise it would result in NPE.

amitjoy: I have noticed that also. Previously it was misplaced, and the null check was performed after get on the map is called. It is due to extracting the piece of code to local variable using Eclipse feature and I forgot to add the null check thereafter. My proposal would be to add `requireNonNull` in the constructor which would save the verbosity of null checks in all the getters.

MMaiero: agreed.
I'll do that after the completion of another task I'm working on. I'll rebase my last commit accordingly.

MMaiero:
@amitjoy Please have a look

kura/org.eclipse.kura.wire.component.provider/src/main/java/org/eclipse/kura/internal/wire/publisher/CloudPublisher.java
amitjoy: Is this addressed? This enhances better readability as most of the methods are already commented.

kura/org.eclipse.kura.wire.component.provider/src/main/java/org/eclipse/kura/internal/wire/publisher/CloudPublisher.java
```diff
@@ -105,44 +128,46 @@ protected synchronized void activate(final ComponentContext componentContext,
...
+        // Update properties
+        this.options = new CloudPublisherOptions(properties);
+
+        if (this.cloudServiceTracker != null) {
+            this.cloudServiceTracker.close();
         }
+        initCloudServiceTracking();
+
+        s_logger.debug(s_message.updatingCloudPublisherDone());
     }
```
amitjoy, line 194: The tracker needs to be closed while deactivating in component

kura/org.eclipse.kura.wire.component.provider/src/main/java/org/eclipse/kura/internal/wire/subscriber/CloudSubscriber.java
```diff
@@ -90,43 +144,63 @@
...
+        this.componentContext = componentContext;
         this.wireSupport = this.wireHelperService.newWireSupport(this);
         this.options = new CloudSubscriberOptions(properties);
-        this.topic = this.options.getSubscribingTopic();
+        this.appTopic = this.options.getSubscribingAppTopic();
+        this.deviceId = this.options.getSubscribingDeviceId();
+        initCloudServiceTracking();
+        logger.debug(wireMessages.activatingCloudSubscriberDone());
+    }
+
```
amitjoy, line 185: the tracker needs to be closed while deactivating the component

kura/org.eclipse.kura.wire.component.provider/src/main/java/org/eclipse/kura/internal/wire/subscriber/CloudSubscriber.java
```diff
@@ -90,43 +144,63 @@
...
+        String filterString = String.format("(&(%s=%s)(kura.service.pid=%s))", Constants.OBJECTCLASS,
+                CloudService.class.getName(), selectedCloudServicePid);
+        Filter filter = null;
         try {
-            if (this.dataService.isConnected()) {
-                this.dataService.subscribe(this.topic, this.options.getSubscribingQos());
-            }
-        } catch (final KuraException e) {
-            s_logger.error(s_message.errorCreatingCloudClinet() + e);
+            filter = this.componentContext.getBundleContext().createFilter(filterString);
```
amitjoy: extract bundle context in an instance or member variable to save processing `getBundleContext` on `ComponentContext` multiple times

kura/org.eclipse.kura.wire.component.provider/src/main/java/org/eclipse/kura/internal/wire/subscriber/CloudSubscriber.java
```diff
@@ -54,32 +59,81 @@
...
+    private ServiceTracker<CloudService, CloudService> cloudServiceTracker;
 
     /** The cloud service. */
     private volatile CloudService cloudService;
 
-    /** The data service. */
-    private volatile DataService dataService;
+    private CloudClient cloudClient;
+
+    private final ServiceTrackerCustomizer<CloudService, CloudService> customizer = new CloudSubscriberServiceTrackerCustomizer();
```
amitjoy: the instance of the tracker must not be created eagerly because, if this bundle is not active (in case the dependencies are not satisfied), the instance of the tracker will still be created.

kura/org.eclipse.kura.wire.component.provider/src/main/java/org/eclipse/kura/internal/wire/publisher/CloudPublisher.java
```diff
@@ -54,44 +55,66 @@
...
 
-    /**
-     * Instantiates a new cloud publisher instance.
-     */
-    public CloudPublisher() {
-        this.monitor = new ReentrantLock();
-    }
+    private ComponentContext componentContext;
+
+    private final ServiceTrackerCustomizer<CloudService, CloudService> customizer = new CloudPublisherServiceTrackerCustomizer();
```
amitjoy: the instance of the tracker must not be created eagerly because, if this bundle is not active (in case the dependencies are not satisfied), the instance of the tracker will still be created. You can move this instantiation to `activate(...)` method.

amitjoy (changes requested):
last few change requests.

amitjoy:
@MMaiero Any update on this? I am curious as it would help me further make changes for restructuring the format of the Wire Envelop.

kura/org.eclipse.kura.wire.component.provider/src/main/java/org/eclipse/kura/internal/wire/publisher/CloudPublisher.java
MMaiero: The tracker was already instantiated in the activate method.
The trackerCustomizer, used by the tracker was initialized at the class instantiation. I've moved that into the activate method as is close to the initialization of the tracker, even if I don't believe we will save so much...

kura/org.eclipse.kura.wire.component.provider/src/main/java/org/eclipse/kura/internal/wire/publisher/CloudPublisher.java
MMaiero: Good catch!

kura/org.eclipse.kura.wire.component.provider/src/main/java/org/eclipse/kura/internal/wire/subscriber/CloudSubscriber.java
MMaiero: Agreed

kura/org.eclipse.kura.wire.component.provider/src/main/java/org/eclipse/kura/internal/wire/publisher/CloudPublisher.java
MMaiero: I, personally, don't believe variables declarations should be commented, if the names given are self-explanatory by themselves. For methods, I can agree on having comments, because we usually don't have (as theory requires) simple and plain methods.

kura/org.eclipse.kura.wire.component.provider/src/main/java/org/eclipse/kura/internal/wire/publisher/CloudPublisherOptions.java
```diff
@@ -126,39 +68,19 @@ int getDisconnectDelay() {
      *            the properties
      */
     CloudPublisherOptions(final Map<String, Object> properties) {
+        requireNonNull(properties, "Properties cannot be null.");
```
amitjoy: Add this to `WireMessages` localization resource

amitjoy (changes requested):
Everything looks pretty good. Just a minor change.

amitjoy (approved):

amitjoy:
@dwoodard1 @cdealti I have already approved the changes. Could you kindly merge this?

PR #1042, opened 2017-01-24
Fix data folder location for emulator.

dwoodard1:
This change moves the data and snapshot folders needed by the emulator to the ws directory instead of the kura/distrib directory.


cdealti:
+1 I believe there is also something created under /tmp/kura or /tmp/.kura I don't remember. Can you please check? Maybe something else should be moved in the ws...

ctron (approved):

dwoodard1:
@cdealti The /tmp folders are the same as always. I suppose we could move these to the ws location, but I have never had an issue with them being under /tmp.

amitjoy:
#1042, #1043 and #1044 failed due to the same reason - 

`Failed to execute goal org.apache.maven.plugins:maven-dependency-plugin:2.9:copy (copy-bundles-for-publishing) on project p2-repo-common: Unable to resolve artifact. Could not transfer artifact org.eclipse.kura:com.codeminders.hidapi:jar:1.1.2 from/to Eclipse Paho Repo (https://repo.eclipse.org/content/repositories/paho-releases/): Connect to repo.eclipse.org:443 [repo.eclipse.org/198.41.30.233] failed: Connection timed out`

@ctron @dwoodard1 @MMaiero @cdealti Could you kindly retrigger Travis CI execution?

MMaiero:
It's the repo.eclipse.org server that is not responding.
Unfortunately there is not so much we can do, apart notifying the Eclipse guys

PR #1043, opened 2017-01-24
Modified certificates policy in Cloud Service.

MMaiero:
Modified birth certs policy in Cloud Service supporting the disabling of the feature
or the publishing on connect or also at reconnect.
Solved minor issues.


MMaiero:
@mcarrer Could you check if this fits with your use case?

PR #1045, opened 2017-01-25
Changes to have can tests working.

MMaiero:
It provides also some improvements in order to have tests passing in OS X.


kura/org.eclipse.kura.protocol.can.test/src/main/java/org/eclipse/kura/protocol/can/test/CanSocketTest.java
```diff
@@ -117,6 +101,37 @@ public void updated(Map<String, Object> properties) {
...
+                // Ignore
+            }
+            this.m_pollThread = null;
+        }
+
+        this.m_pollThread = new Thread(new Runnable() {
+
+            @Override
+            public void run() {
+                if (CanSocketTest.this.m_canConnection != null && !IS_MAC) {
```
amitjoy: `CanSocketTest.this.m_canConnection `and `CanSocketTest.this.thread_done` must be volatile

kura/org.eclipse.kura.protocol.can.test/src/main/java/org/eclipse/kura/protocol/can/test/CanSocketTest.java
```diff
@@ -117,6 +101,37 @@ public void updated(Map<String, Object> properties) {
...
     }
 
+    @Before
+    public void setup() {
+        Assume.assumeFalse(IS_MAC);
+    }
+
+    @Test
+    public void startCanTestThread() {
+        if (this.m_pollThread != null) {
```
amitjoy: This code to let the thread finish if there already exists one is as same as in the `deactivate()` method. You can extract this code to a private method.

amitjoy (changes requested):
Some lurking threading issue and minor code refactoring

kura/org.eclipse.kura.protocol.can.test/src/main/java/org/eclipse/kura/protocol/can/test/CanSocketTest.java
MMaiero: Agreed.

kura/org.eclipse.kura.protocol.can.test/src/main/java/org/eclipse/kura/protocol/can/test/CanSocketTest.java
MMaiero: `CanSocketTest.this.thread_done` does not need to be a class variable. I've changed the code accordingly.

kura/org.eclipse.kura.protocol.can.test/src/main/java/org/eclipse/kura/protocol/can/test/CanSocketTest.java
```diff
@@ -108,42 +99,35 @@ public void setup() {
...
-        this.m_pollThread.start();
+        this.pollThread.start();
     }
 
     public boolean doCanTest() {
         byte[] b;
         CanMessage cm = null;
-        if (this.m_isMaster) {
-            if (this.m_orig >= 0) {
+        if (this.isMaster) {
```
amitjoy, line 123: `isMaster` is also a state variable shared among threads. It needs to be volatile as well. So is `this.orig`.

kura/org.eclipse.kura.protocol.can.test/src/main/java/org/eclipse/kura/protocol/can/test/CanSocketTest.java
```diff
@@ -108,42 +99,35 @@ public void setup() {
...
+            if (this.orig >= 0) {
                 try {
-                    testSendImpl(this.m_ifName, this.m_canId, this.m_orig);
+                    testSendImpl(this.interfaceName, this.canId, this.orig);
```
amitjoy, line 126: `interfaceName` and `canId` also need to be volatile.

kura/org.eclipse.kura.protocol.can.test/src/main/java/org/eclipse/kura/protocol/can/test/CanSocketTest.java
```diff
@@ -152,13 +136,11 @@ public boolean doCanTest() {
             } catch (InterruptedException e) {
             }
         } else {
-            s_logger.info("Wait for a request");
+            logger.info("Wait for a request");
             try {
-
-                cm = this.m_canConnection.receiveCanMessage(-1, 0x7FF);
-
+                cm = this.canConnection.receiveCanMessage(-1, 0x7FF);
```
amitjoy, line 142: `canConnection` too

amitjoy (changes requested):

kura/org.eclipse.kura.protocol.can.test/src/main/java/org/eclipse/kura/protocol/can/test/CanSocketTest.java
MMaiero: I can agree with you about `this.interfaceName`, `this.canId ` and `this.isMaster` that are variables that, in theory, can be affected by multithreading.
The others, in my opinion, are effectively managed by a single thread and do not need to be volatile.

amitjoy: `orig` should also be volatile as it determines a state on which thread works. `canConnection` should also be volatile. Let's suppose the `canConnection` is updated with new implementation at runtime but the thread already has acquired a reference to the old instance of `canConnection`, hence, the thread will not see the updated reference of `canConnection`.

MMaiero: canConnection **is** volatile.
In any case, due to the DS definition, if a new implementation of canConnection is available, taking the place of the old, the deactivate method will be invoked and the thread will be killed.

amitjoy: Actually in OSGi, it works a bit differently. During dynamic service replacement, the bind of the "new" service happens before the unbind of the "old" service. Hence, volatile is a best fit for such things.

kura/org.eclipse.kura.protocol.can.test/src/main/java/org/eclipse/kura/protocol/can/test/CanSocketTest.java
cdealti: @amitjoy I believe you are right on requiring volatile on all the fields that are modified in the ``updated`` method the way it's written now. However this ``updated`` method does not follow our tipical pattern of:

1. Stop all the threads
2. Modify internal state
3. Restart the threads (the run method will see the last modified stated)

But there is a more fundamental flaw. This OSGi (integration test) will not even run from Maven or Eclipse because it does not use the approach described in [1].

In my opinion the test will be ignored on the Mac (because of the assumption in the ``setup`` method) and will always succeed without testing anything because this.canConnection is always null in startCanTestThread.

@markoer what do you think?

[1] https://github.com/eclipse/kura/blob/develop/kura/org.eclipse.kura.core.test/src/main/java/org/eclipse/kura/core/test/ExampleTest.java#L41

amitjoy: You are right. It would not run successfully to inject the dependencies as in the OSGi container that maven initializes during the test-run, it has to be properly synchronized for dependencies using latches and static modifiers for dependencies.

markoer: I tend to agree. You might be lucky and dependency would be injected, or you might not and test would finish before it began.

cdealti:
Let's comment out the test module for the time being.

PR #1046, opened 2017-01-25
Kura Wires INTERNAL ERROR refs fixes

amitjoy:
1. Fixed Build Paths
2. Added Exception Details to Cloud Publisher Publish method
3. Asset Provider activation policy set to lazy
4. Added exception contract to Wire Service deleteConfiguration Javadoc

### DIFF
--- a/kura/org.eclipse.kura.util/src/main/java/org/eclipse/kura/util/base/ThrowableUtil.java
+++ b/kura/org.eclipse.kura.util/src/main/java/org/eclipse/kura/util/base/ThrowableUtil.java
@@ -1,19 +1,25 @@
 /*******************************************************************************
- * Copyright (c) 2016 Eurotech and/or its affiliates and others
+ * Copyright (c) 2016, 2017 Eurotech and/or its affiliates and others
  *
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * which accompanies this distribution, and is available at
  * http://www.eclipse.org/legal/epl-v10.html
- * 
+ *
+ * Contributors:
+ *  Eurotech
+ *  Amit Kumar Mondal
+ *
  *******************************************************************************/
 package org.eclipse.kura.util.base;
+
+import static java.util.Objects.requireNonNull;
 
 import java.io.PrintWriter;
 import java.io.StringWriter;
 
-import org.eclipse.kura.KuraErrorCode;
-import org.eclipse.kura.KuraRuntimeException;
+import org.eclipse.kura.localization.LocalizationAdapter;
+import org.eclipse.kura.localization.resources.UtilMessages;
 
 /**
  * The Class ThrowableUtil contains all necessary static factory methods for
@@ -21,19 +27,54 @@ import org.eclipse.kura.KuraRuntimeException;
  */
 public final class ThrowableUtil {
 
+    /** Localization Resource. */
+    private static final UtilMessages message = LocalizationAdapter.adapt(UtilMessages.class);
+
     /** Constructor */
     private ThrowableUtil() {
         // Static Factory Methods container. No need to instantiate.
     }
 
     /**
-     * Propagates a {@code throwable}.
+     * Propagates a {@code throwable} instance if the throwable is instance of the provided type
+     *
+     * Usage:
+     *
+     * <pre>
+     * <code>
+     * Throwable throwable = null;
+     *
+     * try {
+     *  ....
+     * } catch (Exception1 | Exception2 | Exception3 e) {
+     *   throwable = e;
+     *   ThrowableUtil.propagateIfInstanceOf(e, Exception3.class);
+     *   //if not an instance of Exception3
+     *   doSomethingForException1OrException2();
+     * }
+     *
+     * if (Objects.nonNull(throwable)) {
+     *  //now only propagate if instance of Exception2
+     *  ThrowableUtil.propagateIfInstanceOf(throwable, Exception2.class);
+     * }
+     * </code>
+     * </pre>
      *
      * @param throwable
      *            the throwable to propagate
+     * @param clazz
+     *            the type instance to check
+     * @throws NullPointerException
+     *             if any of the provided arguments is null
      */
-    public static void propagate(final Throwable throwable) {
-        throw new KuraRuntimeException(KuraErrorCode.INTERNAL_ERROR, throwable);
+    public static <E extends Throwable> void propagateIfInstanceOf(final Throwable throwable, final Class<E> clazz)
+            throws E {
+        requireNonNull(throwable, message.throwableNonNull());
+        requireNonNull(clazz, message.clazzNonNull());
+
+        if (clazz.isInstance(throwable)) {
+            throw clazz.cast(throwable);
+        }
     }
 
     /**
--- a/kura/org.eclipse.kura.wire.provider/src/main/java/org/eclipse/kura/internal/wire/WireServiceImpl.java
+++ b/kura/org.eclipse.kura.wire.provider/src/main/java/org/eclipse/kura/internal/wire/WireServiceImpl.java
@@ -1,10 +1,14 @@
 /*******************************************************************************
- * Copyright (c) 2016 Eurotech and/or its affiliates and others
+ * Copyright (c) 2016, 2017 Eurotech and/or its affiliates and others
  *
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * which accompanies this distribution, and is available at
  * http://www.eclipse.org/legal/epl-v10.html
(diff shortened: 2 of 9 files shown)


kura/org.eclipse.kura.wire.provider/src/main/java/org/eclipse/kura/internal/wire/WireServiceImpl.java
```diff
@@ -265,7 +268,7 @@ public void deleteWireConfiguration(final WireConfiguration wireConfiguration) {
                 }
             }
         } catch (final InvalidSyntaxException e) {
-            throw new KuraRuntimeException(KuraErrorCode.INTERNAL_ERROR, ThrowableUtil.stackTraceAsString(e));
+            throw new IllegalArgumentException(ThrowableUtil.stackTraceAsString(e));
```
ctron: Why don't you just throw in the exception as it is intended?

kura/org.eclipse.kura.util/src/main/java/org/eclipse/kura/util/base/ThrowableUtil.java
```diff
@@ -31,9 +32,11 @@ private ThrowableUtil() {
      *
      * @param throwable
      *            the throwable to propagate
+     * @throws RuntimeException
+     *             translates the provided exception as a runtime exception
      */
     public static void propagate(final Throwable throwable) {
-        throw new KuraRuntimeException(KuraErrorCode.INTERNAL_ERROR, throwable);
+        throw new RuntimeException(throwable);
```
ctron: I don't see a real benefit in such a wrapper method. Maybe you can explain this.

kura/org.eclipse.kura.wire.provider/src/main/java/org/eclipse/kura/internal/wire/WireServiceImpl.java
amitjoy: The client code cannot do anything about `InvalidSyntaxException` (a checked one). Especially if a checked exception is used for conditions from which the caller cannot recover, I think it is justified to translate that exception to a runtime exception.

kura/org.eclipse.kura.util/src/main/java/org/eclipse/kura/util/base/ThrowableUtil.java
amitjoy: This has been written for the same reason to translate a checked exception to an unchecked one. If the caller doesn't have any idea about how to recover, it is beneficial to propagate it as an unchecked one.

ctron: True. But you could simple inline that method. I see no real benefit in adding a method for that.

kura/org.eclipse.kura.wire.provider/src/main/java/org/eclipse/kura/internal/wire/WireServiceImpl.java
ctron: True. But passing in the original exception as a string is a bit against the idea of that. Normally in Java you would pass in the original exception and let Java handle the rest.

kura/org.eclipse.kura.util/src/main/java/org/eclipse/kura/util/base/ThrowableUtil.java
amitjoy: The downside that I can see now is that if `throwable` is a runtime exception, it would result in additional wrap. The idea is to provide useful utility method to deal with throwable instances. You are right, this method is not that useful. Let me think of something useful one.

kura/org.eclipse.kura.wire.provider/src/main/java/org/eclipse/kura/internal/wire/WireServiceImpl.java
amitjoy: I agree.

kura/org.eclipse.kura.util/src/main/java/org/eclipse/kura/util/base/ThrowableUtil.java
amitjoy: The method that I have added has a very good use case in Driver implementation for Kura Wires - soon to be available as a PR. The driver has to deal with all low-level complexities in which case it encounters several types of exceptions and based on specific exceptions, propagation is done but not for all. This could have been achieved using simply having multi-catch blocks but to save verbosity, I believe, this method is beneficial.

ctron: Just inline thr method.

amitjoy: Wouldn't it be flexible to extract it as a helper method so that it can be used in other places too? Even I am using it for different Driver implementations of Kura Wires (Modbus and OPC-UA Driver).

amitjoy:
@ctron @MMaiero Could you kindly have a look at it?

kura/org.eclipse.kura.wire.provider/src/main/java/org/eclipse/kura/internal/wire/WireServiceImpl.java
```diff
@@ -206,13 +209,13 @@ synchronized void createWires() throws KuraException {
                                 conf.setWire(wire);
                             }
                         } catch (final InvalidSyntaxException e) {
-                            s_logger.error(ThrowableUtil.stackTraceAsString(e));
+                            logger.error(ThrowableUtil.stackTraceAsString(e));
```
ctron: Same here as the passed arguments to exceptions. The logger can handle stack traces by itself and there is not need to pre-render this.

kura/org.eclipse.kura.wire.provider/src/main/java/org/eclipse/kura/internal/wire/WireServiceImpl.java
```diff
@@ -360,14 +363,14 @@ public synchronized void unbindWireHelperService(final WireHelperService wireHel
...
      */
     public synchronized void updated(final Map<String, Object> properties) {
-        s_logger.debug(s_message.updatingWireService() + properties);
+        logger.debug(message.updatingWireService() + properties);
         this.extractProperties(properties);
         try {
             this.createWires();
         } catch (final KuraException e) {
-            s_logger.error(ThrowableUtil.stackTraceAsString(e));
+            logger.error(ThrowableUtil.stackTraceAsString(e));
```
ctron: Same here as the passed arguments to exceptions. The logger can handle stack traces by itself and there is not need to pre-render this.

kura/org.eclipse.kura.wire.provider/src/main/java/org/eclipse/kura/internal/wire/WireServiceImpl.java
```diff
@@ -95,18 +98,18 @@ public WireServiceImpl() {
...
             this.trackerCustomizer = new WireComponentTrackerCustomizer(componentContext.getBundleContext(), this);
-            this.serviceTracker = new ServiceTracker<WireComponent, WireComponent>(componentContext.getBundleContext(),
-                    WireComponent.class, this.trackerCustomizer);
+            this.serviceTracker = new ServiceTracker<>(componentContext.getBundleContext(), WireComponent.class,
+                    this.trackerCustomizer);
             this.serviceTracker.open();
             this.createWires();
         } catch (final Exception exception) {
-            s_logger.error(ThrowableUtil.stackTraceAsString(exception));
+            logger.error(ThrowableUtil.stackTraceAsString(exception));
```
ctron: Same here as the passed arguments to exceptions. The logger can handle stack traces by itself and there is not need to pre-render this.

amitjoy: `org.slf4j.Logger` does not have any specific method which only takes a single object without passing a string. Hence, I needed to pass the stack trace as represented as stack trace.

ctron: So maybe it is good idea to pass in a string.

amitjoy:
Done 👍

cdealti:
@amitjoy This PR conflicts with #1081. It should be closed. Make sure that its contents are already addressed in #1081. Then please close it.

amitjoy:
@MMaiero Please address these fixes in #1081 where relevant.

MMaiero:
Merged in commit 7c94eb66d9c05344e7bd240c4e157c188c69ce06

PR #1070, opened 2017-02-02
Modified log message in ConfigurationServiceImpl.java

MMaiero:
Modified the logging message in registerComponentOCD to not log the OCD definition
when the logger is in info. Added a more complete message that will be logged when
the logger is in debug.


MMaiero:
@markoer Is there a particular requirement for the toStrings in OCD-related classes?

markoer:
Only that a few tests rely on it. If you delete toString()s, tests will fail. Nothing serious.

MMaiero:
Do you think we can have the tests working without that toString? I would prefer not exposing that stuff.

markoer:
Merge the PR and I'll update the tests. It seems to me that the build is configured so that failing tests don't fail the build itself.

PR #1075, opened 2017-02-03
Fixed a few cases in Configuration Service where NPE could be thrown

markoer:
Cases found during #782. Tests have been updated to reflect the changes.

### DIFF
--- a/kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/system/SystemService.java
+++ b/kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/system/SystemService.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2011, 2016 Eurotech and/or its affiliates
+ * Copyright (c) 2011, 2017 Eurotech and/or its affiliates
  *
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
@@ -298,7 +298,7 @@ public interface SystemService {
 
     /**
      * Returns the maximum number of snapshots to be retained in the file system.
-     * When the maximum number is reached, a garbage collector will deleted the older snapshots.
+     * When the maximum number is reached, a garbage collector will delete the older snapshots.
      *
      * @return maximum number of snapshots to be retained.
      */
--- a/kura/org.eclipse.kura.core.configuration/src/main/java/org/eclipse/kura/core/configuration/ConfigurationServiceImpl.java
+++ b/kura/org.eclipse.kura.core.configuration/src/main/java/org/eclipse/kura/core/configuration/ConfigurationServiceImpl.java
@@ -126,7 +126,7 @@ public class ConfigurationServiceImpl implements ConfigurationService {
     private BundleTracker<Bundle> bundleTracker;
 
     @SuppressWarnings("unused")
-    private MetaTypeService m_metaTypeService;
+    private MetaTypeService metaTypeService;
     private ConfigurationAdmin configurationAdmin;
     private SystemService systemService;
     private CryptoService cryptoService;
@@ -167,11 +167,11 @@ public class ConfigurationServiceImpl implements ConfigurationService {
     }
 
     public void setMetaTypeService(MetaTypeService metaTypeService) {
-        this.m_metaTypeService = metaTypeService;
+        this.metaTypeService = metaTypeService;
     }
 
     public void unsetMetaTypeService(MetaTypeService metaTypeService) {
-        this.m_metaTypeService = null;
+        this.metaTypeService = null;
     }
 
     public void setSystemService(SystemService systemService) {
@@ -576,22 +576,24 @@ public class ConfigurationServiceImpl implements ConfigurationService {
 
     @Override
     public List<ComponentConfiguration> getSnapshot(long sid) throws KuraException {
-        XmlComponentConfigurations xmlConfigs = loadEncryptedSnapshotFileContent(sid);
+        List<ComponentConfiguration> returnConfigs = new ArrayList<ComponentConfiguration>();
 
-        List<ComponentConfigurationImpl> configs = xmlConfigs.getConfigurations();
-        for (ComponentConfigurationImpl config : configs) {
-            if (config != null) {
-                try {
-                    Map<String, Object> decryptedProperties = decryptPasswords(config);
-                    config.setProperties(decryptedProperties);
-                } catch (Throwable t) {
-                    logger.warn("Error during snapshot password decryption");
+        XmlComponentConfigurations xmlConfigs = loadEncryptedSnapshotFileContent(sid);
+        if (xmlConfigs != null) {
+            List<ComponentConfigurationImpl> configs = xmlConfigs.getConfigurations();
+            for (ComponentConfigurationImpl config : configs) {
+                if (config != null) {
+                    try {
+                        Map<String, Object> decryptedProperties = decryptPasswords(config);
+                        config.setProperties(decryptedProperties);
+                    } catch (Throwable t) {
+                        logger.warn("Error during snapshot password decryption");
+                    }
                 }
             }
-        }
 
-        List<ComponentConfiguration> returnConfigs = new ArrayList<ComponentConfiguration>();
-        returnConfigs.addAll(xmlConfigs.getConfigurations());
+            returnConfigs.addAll(xmlConfigs.getConfigurations());
+        }
 
         return returnConfigs;
     }
@@ -856,6 +858,10 @@ public class ConfigurationServiceImpl implements ConfigurationService {
     }
 
     private void encryptConfigs(List<? extends ComponentConfiguration> configs) {
+        if (configs == null) {
+            return;
+        }
+
         for (ComponentConfiguration config : configs) {
             if (config instanceof ComponentConfigurationImpl) {
                 encryptPasswords(config);
@@ -895,8 +901,8 @@ public class ConfigurationServiceImpl implements ConfigurationService {
 
         for (Long snapshot : snapshots) {
             File fSnapshot = getSnapshotFile(snapshot);
-            if (!fSnapshot.exists()) {
-                throw new KuraException(KuraErrorCode.CONFIGURATION_SNAPSHOT_NOT_FOUND, snapshot);
+            if (fSnapshot == null || !fSnapshot.exists()) {
+                throw new KuraException(KuraErrorCode.CONFIGURATION_ERROR, snapshot);
             }
 
             //
@@ -967,6 +973,9 @@ public class ConfigurationServiceImpl implements ConfigurationService {
 
     private void writeSnapshot(long sid, XmlComponentConfigurations conf) throws KuraException {
         File fSnapshot = getSnapshotFile(sid);
+        if (fSnapshot == null) {
+            throw new KuraException(KuraErrorCode.CONFIGURATION_SNAPSHOT_NOT_FOUND);
+        }
 
         // Marshall the configuration into an XML
         String xmlResult;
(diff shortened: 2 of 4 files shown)


kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/configuration/Password.java
```diff
@@ -33,6 +33,10 @@ public Password(char[] password) {
 
     @Override
     public String toString() {
+        if (this.m_password == null) {
```
cdealti: In general I wonder why one should instantiate an (immutable) null Password.
I mean passing a null parameter to the constructor.
Should this be considered a bug instead and hence throw a runtime exception? 

Should we update the copyright date to 2017?

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/system/SystemService.java
```diff
@@ -298,7 +298,7 @@
 
     /**
      * Returns the maximum number of snapshots to be retained in the file system.
-     * When the maximum number is reached, a garbage collector will deleted the older snapshots.
+     * When the maximum number is reached, a garbage collector will delete the older snapshots.
```
cdealti, line 301: Should we update the copyright date to 2017?

kura/org.eclipse.kura.core.configuration/src/main/java/org/eclipse/kura/core/configuration/ConfigurationServiceImpl.java
```diff
@@ -577,7 +577,12 @@ public synchronized void rollback(long id) throws KuraException {
 
     @Override
     public List<ComponentConfiguration> getSnapshot(long sid) throws KuraException {
+        List<ComponentConfiguration> returnConfigs = new ArrayList<ComponentConfiguration>();
+
         XmlComponentConfigurations xmlConfigs = loadEncryptedSnapshotFileContent(sid);
+        if (xmlConfigs == null) {
```
cdealti: I would avoid the ``return`` in the middle of the method given the intention of the refactor favoring
something like:

```
if (xmlConfigs != null) {
        List<ComponentConfigurationImpl> configs = xmlConfigs.getConfigurations();
        for (ComponentConfigurationImpl config : configs) {
            ...
        }
}
```

kura/org.eclipse.kura.core.configuration/src/main/java/org/eclipse/kura/core/configuration/ConfigurationServiceImpl.java
```diff
@@ -678,7 +682,12 @@ synchronized void unregisterComponentConfiguration(String pid) {
 
     boolean mergeWithDefaults(OCD ocd, Map<String, Object> properties) throws KuraException {
         boolean changed = false;
-        Set<String> keys = properties.keySet();
+        Set<String> keys;
```
cdealti: I don't think it makes sense to call this method with null properties. The method modifies the instance referenced by the properties parameter :-( so I would consider this occurrence a bug and an NPE or any runtime exception should be appropriate.

kura/org.eclipse.kura.core.configuration/src/main/java/org/eclipse/kura/core/configuration/ConfigurationServiceImpl.java
```diff
@@ -856,6 +865,10 @@ private void encryptPassword(Map<String, Object> propertiesToUpdate, String key,
     }
 
     private void encryptConfigs(List<? extends ComponentConfiguration> configs) {
+        if (configs == null) {
```
cdealti, line 861: same as above

kura/org.eclipse.kura.core.configuration/src/main/java/org/eclipse/kura/core/configuration/ConfigurationServiceImpl.java
```diff
@@ -895,7 +908,7 @@ private void encryptPlainSnapshots() throws Exception {
 
         for (Long snapshot : snapshots) {
             File fSnapshot = getSnapshotFile(snapshot);
-            if (!fSnapshot.exists()) {
+            if (fSnapshot == null || !fSnapshot.exists()) {
```
cdealti: this looks to me a genuine internal error (maybe KuraErrorCode.CONFIGURATION_ERROR). I would return KuraErrorCode.CONFIGURATION_SNAPSHOT_NOT_FOUND if I had specified a snapshot ID has a parameter to a method and the snapshot could not be find

kura/org.eclipse.kura.core.configuration/src/main/java/org/eclipse/kura/core/configuration/ConfigurationServiceImpl.java
```diff
@@ -967,6 +980,9 @@ private synchronized long saveSnapshot(List<? extends ComponentConfiguration> co
 
     private void writeSnapshot(long sid, XmlComponentConfigurations conf) throws KuraException {
         File fSnapshot = getSnapshotFile(sid);
+        if (fSnapshot == null) {
+            throw new KuraException(KuraErrorCode.CONFIGURATION_ERROR);
```
cdealti: KuraErrorCode.CONFIGURATION_SNAPSHOT_NOT_FOUND in my opinion

kura/org.eclipse.kura.core.configuration/src/main/java/org/eclipse/kura/core/configuration/ConfigurationServiceImpl.java
```diff
@@ -1164,6 +1180,11 @@ String getSnapshotsDirectory() {
 
     private File getSnapshotFile(long id) {
         String configDir = getSnapshotsDirectory();
+
+        if (configDir == null) {
```
cdealti, line 1177: invalid parameter or something like this. But it's a private method so null will do

kura/org.eclipse.kura.core.configuration/src/main/java/org/eclipse/kura/core/configuration/ConfigurationServiceImpl.java
```diff
@@ -1312,7 +1333,11 @@ XmlComponentConfigurations loadEncryptedSnapshotFileContent(long snapshotID) thr
         }
 
         // File loaded, try to decrypt and unmarshall
-        String decryptedContent = new String(this.m_cryptoService.decryptAes(entireFile.toString().toCharArray()));
+        char[] decryptAes = this.m_cryptoService.decryptAes(entireFile.toString().toCharArray());
+        if (decryptAes == null) {
```
cdealti: So, if decryptAes fails to decrypt the char array it returns null?
Ideally it should throw an exception. If not (and we don't want to break the API) this should be documented in the Javadoc.

cdealti (changes requested):

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/configuration/Password.java
markoer: OK, it's a weird corner case and it does change the API. I'll drop this one and (c) can remain as is.

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/system/SystemService.java
markoer: Hmm... it's only a typo in docs, but you're right.

kura/org.eclipse.kura.core.configuration/src/main/java/org/eclipse/kura/core/configuration/ConfigurationServiceImpl.java
markoer: On the second thought, NPE is quite OK in this case.

kura/org.eclipse.kura.core.configuration/src/main/java/org/eclipse/kura/core/configuration/ConfigurationServiceImpl.java
markoer: XmlComponentConfigurations marshals the same with null and empty component configurations. Therefore it might be legal to store such a snapshot (although impossible in the current implementation). Meaning that this extra check is a safeguard from a wrong call that doesn't have to fail. No configs, no passwords, no change. And the same with null - no change.
If you still don't want the change, I'll remove it...

kura/org.eclipse.kura.core.configuration/src/main/java/org/eclipse/kura/core/configuration/ConfigurationServiceImpl.java
markoer: OK. Will update to the new code.

kura/org.eclipse.kura.core.configuration/src/main/java/org/eclipse/kura/core/configuration/ConfigurationServiceImpl.java
markoer: Agreed. The calling methods handle null, now, and react appropriately.

kura/org.eclipse.kura.core.configuration/src/main/java/org/eclipse/kura/core/configuration/ConfigurationServiceImpl.java
markoer: This scenario covers case with a mocked service that returns null (i.e. a bad/default service implementation). There's no immediate danger for it happening here.

kura/org.eclipse.kura.core.configuration/src/main/java/org/eclipse/kura/core/configuration/ConfigurationServiceImpl.java
markoer: It's not an uncommon practice in this service and reduces indentation level of the main part of the method, but that's just fine, too. Will update.

kura/org.eclipse.kura.core.configuration/src/main/java/org/eclipse/kura/core/configuration/ConfigurationServiceImpl.java
ctron: I agree with @markoer Unless Kura is full of assumptions which turn into NPEs every now and then. So checking for NULL doesn't hurt, unless Kura would adopt proper null pointer analysis.

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/system/SystemService.java
cdealti: @markoer please update the copyright date

kura/org.eclipse.kura.core.configuration/src/main/java/org/eclipse/kura/core/configuration/ConfigurationServiceImpl.java
cdealti: @markoer OK

kura/org.eclipse.kura.core.configuration/src/main/java/org/eclipse/kura/core/configuration/ConfigurationServiceImpl.java
cdealti: OK

kura/org.eclipse.kura.core.configuration/src/main/java/org/eclipse/kura/core/configuration/ConfigurationServiceImpl.java
cdealti: OK

cdealti:
Please update the copyright header above and rebase and this is ready to be merged

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/system/SystemService.java
markoer: Did that already, it's just not shown as a change here, because the comment is on another line.

cdealti (approved):

cdealti:
Ready to merge if the build passes

PR #1077, opened 2017-02-03
Updated CONTRIBUTING.md

markoer:
Added a few more hints for changes and updated (c) year

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -36,7 +36,7 @@ A proper header must be in place for any file contributed to Eclipse Kura. For a
 
 ```
 /*******************************************************************************
- * Copyright (c) 2016 <legal entity> and others
+ * Copyright (c) <year> <legal entity> and others
  *
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -36,7 +36,7 @@ A proper header must be in place for any file contributed to Eclipse Kura. For a
 
 ```
- * Copyright (c) 2016 <legal entity> and others
+ * Copyright (c) 2017 <legal entity> and others
```
amitjoy: Instead of mentioning 2016 or 2017, we could mention `<the current year> `

CONTRIBUTING.md
```diff
@@ -45,13 +45,13 @@ A proper header must be in place for any file contributed to Eclipse Kura. For a
  *
  *******************************************************************************/
  ```
- 
+
  Please ensure \<legal entity\> is replaced with the relevant information. If you are editing an existing contribution, feel free
  to create or add your legal entity to the contributors section as such:
- 
+
```
amitjoy: if we mention `<the current year>`, we can also mention here that <the current year> should be replaced with the current year in numerics, for example, 2016 or 2017.

markoer: It's already in the new line 86, but it couldn't hurt to add it here, too...

CONTRIBUTING.md
markoer: ... `<year>`?

amitjoy: I donno why it didn't appear before but when I used it as code in the comment it appeared.

CONTRIBUTING.md
amitjoy: Here too.

Issue #1079, opened 2017-02-06
[Testing] Tests for org.eclipse.kura.core.comm


djurod:
Writing tests for org.eclipse.kura.core.comm

MMaiero:
Merged the PR, can this be closed?

djurod:
Yes

PR #1080, opened 2017-02-06
#1079: Tests for org.eclipse.kura.core.comm

djurod:
- Prepared unit tests for class CommConnectionImpl from package
  org.eclipse.kura.core.comm
- Fixed the implementation of getBytesAsString method (extra space at
  the end of the result string)
- Ported the org.eclipse.kura.core.comm.test project from "kura" to
  "kura/test" directory.


djurod:
When I started working on tests for this project I first thought that it would be possible to prepare some more tests, but unfortunately it is not possible to do that without big effort.

In the end I just moved/ported the code from "kura" to "kura/test" directory.

It appears that the original contribution was from @ctron ( https://github.com/eclipse/kura/tree/develop/kura/org.eclipse.kura.core.comm.test ). I basically rewrote the tests (as can be seen there is not much you can do here), but I guess there could be some issues regarding the contribution noted in the files.

@ctron: If you would like something to be changed please let me know before this is merged.

ctron:
I think it is great to go with Java 8. But would try to work around creating another array of the same size. Instead try to do this more stream-like:

```java
Arrays.stream(bytes).map(b -> String.format("%02X", b)).join (Collectors.join(" "));
```

djurod:
This doesn't work because "Arrays.stream()" apparently doesn't support byte arrays...

ctron:
Then you need a different way around double-allocation.

djurod:
Sorry for the delay in answering this (I was on vacation).

As far as I see it is not really possible to avoid the double allocation, at least not without some complex solution that is either hard to understand or more effort would be needed to implement it that it is justifiable.

The solution that I prepared probably isn't any worse than the current implementation (but it should be better) and as far as I can see it is neither worse than your suggestion with the Array.stream() - while technically "smart" it is hard to understand what is the result of that line of code and as I have seen in different coding guidelines it is usually recommended to avoid this kind of solutions if possible.

ctron:
Is the following too hard to understand or too complex?

```java
StringJoiner sj = new StringJoiner(" ");
for ( byte b : bytes )  {
  sj.add ( String.format ("%02X", b) );
}
return sj.toString();
```

djurod:
I was not aware of this class and it does look better. I will update the pull request.

But I am not convinced that it actually reduces the memory usage...

kura/test/org.eclipse.kura.core.comm.test/build.properties
```diff
@@ -0,0 +1,19 @@
+#
+# Copyright (c) 2011, 2016 Eurotech and/or its affiliates
```
MMaiero: I think the copyright here needs to be updated

djurod: I also fixed the same issue in pom.xml

MMaiero:
The build seems to fail on Karaf

djurod:
From the log it appears that download of a file timed out...

> [ERROR] Failed to execute goal on project kura-runtime: Could not resolve dependencies for project org.eclipse.kura.karaf:kura-runtime:feature:0.0.1-SNAPSHOT: Failed to collect dependencies at org.apache.karaf.features:framework:kar:4.0.5 -> org.apache.karaf.features:base:jar:4.0.5: Failed to read artifact descriptor for org.apache.karaf.features:base:jar:4.0.5: Could not transfer artifact org.apache.karaf.features:base:pom:4.0.5 from/to central (http://repo.maven.apache.org/maven2): Failed to transfer file: http://repo.maven.apache.org/maven2/org/apache/karaf/features/base/4.0.5/base-4.0.5.pom. Return code is: 503 , ReasonPhrase:first byte timeout. -> [Help 1]

Should the build be manually restarted?

PR #1084, opened 2017-02-08
Wire Service Implementation and Wire Configuration Refactoring

amitjoy:


MMaiero:
Could you please rebase this into a single commit since are all related and cannot be separate one each other?
Furthermore, I would suggest you to keep consistency with the existing Kura code and not order methods in WireServiceImpl.java by name.

amitjoy:
@MMaiero You can also squash the commits while merging.

I personally prefer following naming convention that I find it useful to track in a class and that's why I order the methods in alphabetical order. And Eclipse Platform developers has surveyed this need and provided the option. I am not sure if Kura has such restrictions on method ordering. If yes, would you kindly refer me to it? I would then follow it from the next time onwards.

MMaiero:
Which is the option you used to setup that ordering?
It's better if you perform the squashing of your commits.

amitjoy:
`Member Sort Order` Option in Eclipse Preferences

MMaiero:
This one?

schermata 2017-02-14 alle 12 10 28

I have Mars and it is not reordering methods. Is it a new feature?

MMaiero:
However, I think that this was an implicit assumption that all the Kura developers followed to have the same structuring throughout all the project.
We had never the need to formalize it. (Maybe also there was no option at that time, but I may be wrong).

If needed we can formalize it, otherwise it becomes a mess.

amitjoy:
bildschirmfoto 2017-02-14 um 12 13 00 pm

Check in `Save Actions`

amitjoy:
It would also be good to formalize the placement of the methods in Kura documentation.

amitjoy:
Squashed it to single commit

MMaiero:
This is mine:
schermata 2017-02-14 alle 13 51 00

PR #1085, opened 2017-02-09
Wires refactor 2

MMaiero:
Few more minor changes to the wires.
Change descriptions in respective comments.


kura/org.eclipse.kura.asset.provider/src/main/java/org/eclipse/kura/asset/provider/BaseAsset.java
```diff
@@ -321,28 +341,31 @@ public ComponentConfiguration getConfiguration() throws KuraException {
...
-                    }
-                }
+                fillDriverSpecificChannelConfiguration(mainOcd, driverSpecificChannelConfiguration);
             }
         }
         return new ComponentConfigurationImpl(componentName, mainOcd, props);
     }
 
+    @SuppressWarnings("unchecked")
+    private void fillDriverSpecificChannelConfiguration(final Tocd mainOcd,
```
amitjoy: Add invariants

kura/org.eclipse.kura.asset.provider/src/main/java/org/eclipse/kura/asset/provider/BaseAsset.java
```diff
@@ -555,9 +580,14 @@ public void onDriverEvent(final DriverEvent event) {
         final Set<String> channelPrefixes = CollectionUtil.newHashSet();
         for (final Map.Entry<Long, Channel> entry : channels.entrySet()) {
             final Long key = entry.getKey();
-            final String prefix = key + CHANNEL_PROPERTY_POSTFIX.value() + CHANNEL_PROPERTY_PREFIX.value()
-                    + CHANNEL_PROPERTY_POSTFIX.value();
-            channelPrefixes.add(prefix);
+
+            final StringBuilder channelPrefix = new StringBuilder();
```
amitjoy, line 594: Not needed. Java 8 internally uses `StringBuilder` for string concatenation.

kura/org.eclipse.kura.asset.provider/src/main/java/org/eclipse/kura/internal/asset/provider/AssetOptions.java
```diff
@@ -103,11 +103,8 @@ private void addChannel(final Channel channel) {
      *
      * @param properties
      *            the properties to retrieve the channels from
-     * @throws NullPointerException
-     *             if the argument is null
      */
     private void checkChannelAvailability(final Map<String, Object> properties) {
-        requireNonNull(properties, message.propertiesNonNull());
```
amitjoy, line 110: The contract of this method is to use a map which is not null. Even though the map to be passed to this method in this class is not null, the contract should apply such preconditions.

kura/org.eclipse.kura.asset.provider/src/main/java/org/eclipse/kura/internal/asset/provider/AssetOptions.java
```diff
@@ -120,11 +117,8 @@ private void checkChannelAvailability(final Map<String, Object> properties) {
      *
      * @param properties
      *            the provided properties
-     * @throws NullPointerException
-     *             if the argument is null
      */
     private void extractProperties(final Map<String, Object> properties) {
-        requireNonNull(properties, message.propertiesNonNull());
```
amitjoy, line 127: Same here

kura/org.eclipse.kura.asset.provider/src/main/java/org/eclipse/kura/internal/asset/provider/AssetOptions.java
```diff
@@ -158,9 +152,10 @@ public AssetConfiguration getAssetConfiguration() {
      * @throws NullPointerException
      *             if any of the arguments is null
      */
-    private ChannelType getChannelType(final Map<String, Object> properties, final String channelTypePropertyKey) {
-        requireNonNull(properties, message.propertiesNonNull());
-        requireNonNull(channelTypePropertyKey, message.channelKeyNonNull());
+    private ChannelType getChannelType(final Map<String, Object> properties, final String channelKeyFormat) {
+        requireNonNull(channelKeyFormat, message.channelKeyNonNull());
```
amitjoy, line 150: Same

kura/org.eclipse.kura.asset.provider/src/main/java/org/eclipse/kura/internal/asset/provider/AssetOptions.java
```diff
@@ -243,50 +237,56 @@ private Channel retrieveChannel(final long channelId, final Map<String, Object>
         if (channelId <= 0) {
             throw new IllegalArgumentException(message.channelIdNotLessThanZero());
         }
-        requireNonNull(properties, message.propertiesNonNull());
```
amitjoy, line 246: This precondition is feasible here

kura/org.eclipse.kura.asset.provider/src/main/java/org/eclipse/kura/internal/asset/provider/AssetOptions.java
```diff
@@ -243,50 +237,56 @@ private Channel retrieveChannel(final long channelId, final Map<String, Object>
...
-        ChannelType channelType = null;
-        DataType dataType = null;
         Channel channel = null;
-        final Map<String, Object> channelConfig = CollectionUtil.newConcurrentHashMap();
 
         // All key names present is the properties
-        final String channelKeyContainment = CHANNEL_PROPERTY_POSTFIX.value() + CHANNEL_PROPERTY_PREFIX.value()
-                + CHANNEL_PROPERTY_POSTFIX.value();
-        final String channelKeyFormat = channelId + channelKeyContainment;
+        final StringBuilder channelPropertyKeyBuilder = new StringBuilder();
```
amitjoy: Java 8 uses `StringBuilder` by default

kura/org.eclipse.kura.asset.provider/src/main/java/org/eclipse/kura/internal/asset/provider/AssetOptions.java
```diff
@@ -243,50 +237,56 @@ private Channel retrieveChannel(final long channelId, final Map<String, Object>
...
-            }
-        }
         if (channelType != null && dataType != null) {
             channel = new Channel(channelId, channelName, channelType, dataType, channelConfig);
         }
         logger.debug(message.retrievingChannelDone());
         return channel;
     }
 
+    private Map<String, Object> retrieveChannelConfig(final long channelId, final Map<String, Object> properties) {
```
amitjoy, line 230: Add preconditions

kura/org.eclipse.kura.asset.provider/src/main/java/org/eclipse/kura/internal/asset/provider/AssetOptions.java
```diff
@@ -243,50 +237,56 @@ private Channel retrieveChannel(final long channelId, final Map<String, Object>
...
+                        + CHANNEL_PROPERTY_POSTFIX.value();
+                final String cKey = key
+                        .substring(key.indexOf(driverSpecificPropertyKey) + driverSpecificPropertyKey.length());
+                channelConfig.put(cKey, value);
+            }
+        }
+        return channelConfig;
+    }
+
+    private String retrieveChannelName(final Map<String, Object> properties, final String channelKeyFormat) {
```
amitjoy, line 248: Add preconditions

kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/client/ui/wires/PropertiesUi.java
```diff
@@ -628,10 +628,8 @@ private int findNextChannelId() {
 
             @Override
             public String getValue(final GwtChannelInfo object) {
-                final String value = object.get(id).toString();
-                return value == null ? "" : value;
+                return object.get(id).toString();
```
amitjoy: `object.get(id)` can return null and result in NPE

kura/org.eclipse.kura.wire.component.provider/src/main/java/org/eclipse/kura/internal/wire/common/DbServiceHelper.java
```diff
@@ -76,7 +76,6 @@ public static DbServiceHelper getInstance(final DbService dbService) {
      *             if argument is null
      */
     public void close(final Connection conn) {
-        requireNonNull(conn, message.connectionNonNull());
```
amitjoy: Shouldn't it be a precondition to this method? Even though `DbService` `close()` method doesn't work on null elements, my proposition would be too add invariant to this method. I believe this is a good practice.

amitjoy (changes requested):

kura/org.eclipse.kura.asset.provider/src/main/java/org/eclipse/kura/asset/provider/BaseAsset.java
MMaiero: Not really. The string concatenation, if performed, is done at different steps (at compile time or using JIT). The current Java8 for ARM is provided without JIT, so in our target hardware, it is fully interpreted and the concatenation optimizations are not performed.
Finally, having a look [at this research](http://www.pellegrino.link/2015/08/22/string-concatenation-with-java-8.html), using explicitly StringBuilder is still the best option.

kura/org.eclipse.kura.asset.provider/src/main/java/org/eclipse/kura/internal/asset/provider/AssetOptions.java
MMaiero: Those methods are private. The check is performed at the boundaries and we are assured that the map is not null. I completely agree in having contracts specified for public methods, since those define the interaction, and for these it's meaningful to check what is passed.
For private methods in a class it's a little bit an overkill, with an unneeded cost in terms of CPU time.

https://dev.eclipse.org/sonar/component/index#component=org.eclipse.kura%3Aorg.eclipse.kura.asset.provider%3Asrc%2Fmain%2Fjava%2Forg%2Feclipse%2Fkura%2Finternal%2Fasset%2Fprovider%2FAssetOptions.java&currentIssue=05c789f4-1d0e-4c9c-96dd-ebfa2bdcb17a

kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/client/ui/wires/PropertiesUi.java
MMaiero: if I try to execute this `final String value = object.get(id).toString();` and `object.get(id)` is null I would end in a NPE in any case when calling the `toString`.
Having `return value == null ? "" : value;` the following line would not prevent it.

amitjoy: Exactly. Previously also, it was erroneous. Just add a Null Check If condition on the get to avoid it.

kura/org.eclipse.kura.wire.component.provider/src/main/java/org/eclipse/kura/internal/wire/common/DbServiceHelper.java
MMaiero: The check if Connection is null is already performed in the Db implementation. So, having the check here is not needed. Furthermore, we should try to keep consistency and this is not the case if we throw an exception when trying to close the connection, while we accept to close a null statement.

kura/org.eclipse.kura.asset.provider/src/main/java/org/eclipse/kura/asset/provider/BaseAsset.java
amitjoy: Good source of knowledge but due to static concatenation, StringBuilder and String concatenation works equally. For dynamic concatenation, I completely agree that StringBuilder performs better than String concatenation.

kura/org.eclipse.kura.wire.component.provider/src/main/java/org/eclipse/kura/internal/wire/common/DbServiceHelper.java
amitjoy: According to my opinion, accepting a null is a violation of contract and it does not symbolize the nature of true object orientation. That's why such invariant is very helpful in implementing methods. I agree that `DbService.close()` method also accepts null but does not work on it because of its null check but I personally don't feel it's a good practice when it's a public API.

kura/org.eclipse.kura.asset.provider/src/main/java/org/eclipse/kura/internal/asset/provider/AssetOptions.java
amitjoy: SonarQube is just awesome 😉

MMaiero:
I've tried to do some more steps. Please have a look.
I think that the commented aspects are, in some way, minor elements that would not affect the overall correctness or performance. I believe that, if most of the elements are addressed, I would like to merge and, eventually, solve later on. In this way, we can speed up the M1 release.

kura/org.eclipse.kura.asset.cloudlet.provider/src/main/java/org/eclipse/kura/internal/asset/cloudlet/AssetCloudlet.java
```diff
@@ -427,27 +430,29 @@ private void wrapValue(final AssetRecord assetRecord, final String userValue, fi
         requireNonNull(userValue, message.valueNonNull());
         requireNonNull(userType, message.typeNonNull());
 
+        final DataType dataType = DataType.getDataType(userType);
```
amitjoy, line 435: Add this inside the `try` block as it might throw `IllegalArgumentException` and the exception must be propagated upwards. Also catch `IllegalArgumentException` in this method as well as in `doPutAssets()` in the cloudlet.

kura/org.eclipse.kura.asset.provider/src/main/java/org/eclipse/kura/asset/provider/BaseAsset.java
```diff
@@ -320,27 +339,41 @@ public ComponentConfiguration getConfiguration() throws KuraException {
...
+     * 
+     * @param mainOcd
+     *            a {@link Tocd} object.
+     * @param driverSpecificChannelConfiguration
+     *            the driver specific configuration.
+     */
+    @SuppressWarnings("unchecked")
+    private void fillDriverSpecificChannelConfiguration(final Tocd mainOcd,
+            final List<Tad> driverSpecificChannelConfiguration) {
+        if (mainOcd == null || driverSpecificChannelConfiguration == null) {
```
amitjoy, line 360: `mainOcd == null || driverSpecificChannelConfiguration == null || driverSpecificChannelConfiguration.isEmpty()`

kura/org.eclipse.kura.asset.provider/src/main/java/org/eclipse/kura/internal/asset/provider/AssetOptions.java
```diff
@@ -188,39 +170,17 @@ private ChannelType getChannelType(final Map<String, Object> properties, final S
      * @throws NullPointerException
      *             if any of the arguments is null
      */
-    private DataType getDataType(final Map<String, Object> properties, final String channelValueTypePropertyKey) {
+    private DataType getDataType(final Map<String, Object> properties, final String channelKeyFormat) {
```
amitjoy: `@throws IllegalArgumentException if the Data Type is not found`

amitjoy (changes requested):

kura/org.eclipse.kura.asset.cloudlet.provider/src/main/java/org/eclipse/kura/internal/asset/cloudlet/AssetCloudlet.java
```diff
@@ -292,9 +295,9 @@ private void doPutAssets(final KuraRequestPayload reqPayload, final KuraResponse
             boolean flag = true;
             try {
                 this.wrapValue(assetRecord, userValue, userType);
-            } catch (final NumberFormatException nfe) {
+            } catch (final IllegalArgumentException e) {
```
amitjoy, line 298: It should be `catch (final IllegalArgumentException | NumberFormatException  e)`

kura/org.eclipse.kura.asset.provider/src/main/java/org/eclipse/kura/asset/provider/BaseAsset.java
amitjoy: This `isEmpty` check reduces the unnecessary variable references creation that happens after this.

amitjoy (changes requested):

kura/org.eclipse.kura.asset.cloudlet.provider/src/main/java/org/eclipse/kura/internal/asset/cloudlet/AssetCloudlet.java
MMaiero: Please have a look at the javadoc https://docs.oracle.com/javase/7/docs/api/java/lang/NumberFormatException.html

amitjoy: 👍

kura/org.eclipse.kura.asset.provider/src/main/java/org/eclipse/kura/asset/provider/BaseAsset.java
MMaiero: I think that the code manages the case when `driverSpecificChannelConfiguration` is empty. 
Furthermore, having again a look at the code, I think that in another refactoring pass we can improve the methods structuring and have them more specific and more understandable. 
But this will go for M2. M1 is mostly closed now.

amitjoy:
ok cool. I will have a final look now and will merge by today if there is no further significant comment.

amitjoy (approved):

PR #1088, opened 2017-02-10
Fixed Web Ui services search bar

nicolatimeus:


MMaiero:
@nicolatimeus Do you mind if we perform this check on key press? Your implementation works pretty well but I think it's not easy, for an average user, to understand that the filtering is performed only after an enter. Probably, having the check performed at every key pressed, even if is slightly less efficient, is more easy to understand.

nicolatimeus:
No problem, I agree that doing the check at every keypress will improve usability.

PR #1090, opened 2017-02-13
Modified interval check in GwtNetworkServiceImpl.java.

MMaiero:
The interval allowed is now 1-32. If 0 is passed, the internal check will throw an exception.

### DIFF
--- a/kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/server/GwtNetworkServiceImpl.java
+++ b/kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/server/GwtNetworkServiceImpl.java
@@ -252,7 +252,7 @@ public class GwtNetworkServiceImpl extends OsgiRemoteServiceServlet implements G
                                         } else {
                                             gwtNetConfig.setIpAddress("");
                                         }
-                                        if (addressConfig.getNetworkPrefixLength() >= 0
+                                        if (addressConfig.getNetworkPrefixLength() > 0
                                                 && addressConfig.getNetworkPrefixLength() <= 32) {
                                             gwtNetConfig.setSubnetMask(NetworkUtil
                                                     .getNetmaskStringForm(addressConfig.getNetworkPrefixLength()));
@@ -723,7 +723,7 @@ public class GwtNetworkServiceImpl extends OsgiRemoteServiceServlet implements G
     @Override
     public void updateNetInterfaceConfigurations(GwtXSRFToken xsrfToken, GwtNetInterfaceConfig config)
             throws GwtKuraException {
-    	
+
         checkXSRFToken(xsrfToken);
         NetworkAdminService nas = ServiceLocator.getInstance().getService(NetworkAdminService.class);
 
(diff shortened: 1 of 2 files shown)


kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/server/GwtNetworkServiceImpl.java
```diff
@@ -252,7 +252,7 @@
                                         } else {
                                             gwtNetConfig.setIpAddress("");
                                         }
-                                        if (addressConfig.getNetworkPrefixLength() >= 0
+                                        if (addressConfig.getNetworkPrefixLength() > 0
```
MMaiero, line 255: @ibinshtok Do you remember why here we were allowing to pass the check also with 0? Because NetworkUtil.getNetmaskStringForm(addressConfig.getNetworkPrefixLength()) throws an exception if 0 is passed.

ibinshtok: @MMaiero I don't know why we are allowing /0 CIDR prefix there. May be because the NetworkUtil.getNetmaskStringForm(int prefix) takes it. As far as I know the 0.0.0.0/0 is a notation for    
default gateway in the routing table. So that the "show ip route" command return something like this:
Gateway of last resort is 10.11.24.253 to network 0.0.0.0

S*    0.0.0.0/0 [1/0] via 10.11.24.253
                [1/0] via 10.11.1.1
 So I think your fix is correct. Also note that if we pass 0 into the NetworkUtil. getNetmaskStringForm() it will return 255.255.255.255 which is not correct. It should return 255.255.255.255 for /32 (which it does) and 0.0.0.0 for /0.
Here is simple code to test this behavior:
public class MyTest {

	public static void main(String[] args) {
		
		System.out.println(getNetmaskStringForm(0));
		System.out.println(getNetmaskStringForm(8));
		System.out.println(getNetmaskStringForm(16));
		System.out.println(getNetmaskStringForm(24));
		System.out.println(getNetmaskStringForm(30));
		System.out.println(getNetmaskStringForm(32));
	}
	
	public static String getNetmaskStringForm(int prefix)  {
		if(prefix >= 0 && prefix <=32) {
			int mask = ~((1 << (32 - prefix)) - 1);
			return dottedQuad(mask);
		} else {
			System.out.println("invalid prefix");
			return null;
		}
	}
	
	public static String dottedQuad(int ip) {
		StringBuffer sb = new StringBuffer(15);
		for (int shift = 24; shift > 0; shift -= 8) {
			// process 3 bytes, from high order byte down.
			sb.append(Integer.toString((ip >>> shift) & 0xff));
			sb.append('.');
		}

		sb.append(Integer.toString(ip & 0xff));
		return sb.toString();
	}
}
and this is what I am getting back:

255.255.255.255
255.0.0.0
255.255.0.0
255.255.255.0
255.255.255.252
255.255.255.255

MMaiero: So, do you agree on merging this?

ibinshtok: @MMaiero yes, I do

PR #1095, opened 2017-02-14
Added Bundle to stream OSGi events to SSE clients

### LINKED ISSUE #1000: Memory leak and concurrency issue in 'EventHandlerServlet'

amitjoy:
This bundle is used to stream OSGi events and their associated properties to [Server Sent Events](https://www.w3.org/TR/2011/WD-eventsource-20110310/) (aka EventSource) clients. Consumers can make use of any Event Source supported language to build their own client application to receive OSGi event notifications. A simple javascript client can be written as follows:

```
var eventSource = new EventSource("http://<IP>:<PORT>/sse?topic=org/osgi/framework/*");
eventSource.onmessage = function(event) {
  var data = JSON.parse(event.data);
  console.log(data);
};
```

Kura Wires uses this feature to subscribe to Wire Component emit events to update the Composer Canvas UI in an asynchronous way.

This bundle uses the following Jetty subprojects:

- Jetty EventSource Servlet - https://github.com/jetty-project/jetty-eventsource-servlet
- Jetty Continuation - https://github.com/eclipse/jetty.project/tree/jetty-9.4.x/jetty-continuation

This fixes #1000 by removing `EventHandlerServlet` which operates on very low-level complexities and hence introduces bugs and memory issues.


kura/org.eclipse.kura.easse/build.properties
```diff
@@ -0,0 +1,16 @@
+#
+# Copyright (c) 2017 Eurotech and/or its affiliates and others
```
ctron: I thought we agreed upon keeping this in sync. So either both times "Amit Kumar Mondal" or "Eurotech".

kura/org.eclipse.kura.easse/build.properties
```diff
@@ -0,0 +1,16 @@
...
+#  which accompanies this distribution, and is available at
+#  http://www.eclipse.org/legal/epl-v10.html
+#
+#  Contributors:
+#	Amit Kumar Mondal
+#
+source.. = src/main/java/
+bin.includes = META-INF/,\
+               .,\
+               lib/jetty-eventsource-servlet-1.0.1-SNAPSHOT.jar,\
```
ctron: Why is it necessary to embed those JARs?

amitjoy: `Jetty EventSource Servlet` is not packaged as an OSGi bundle

kura/org.eclipse.kura.easse/META-INF/MANIFEST.MF
```diff
@@ -0,0 +1,23 @@
...
+ org.eclipse.kura.util.collection;version="[1.0,2.0)",
+ org.osgi.framework;version="[1.7.0,2.0.0)",
+ org.osgi.service.event;version="1.3.0",
+ org.osgi.service.http;version="1.2.0",
+ org.osgi.util.tracker;version="1.5.0",
+ org.slf4j;version="1.6.4"
+Export-Package: org.eclipse.kura.easse;version="1.0.0"
+Bundle-ClassPath: .,
+ lib/jetty-eventsource-servlet-1.0.1-SNAPSHOT.jar,
+ lib/jetty-continuation-9.4.0.v20161208.jar
```
MMaiero: Currently our target platform already has org.eclipse.jetty.continuation_9.3.9.v20160517. Is there a particular requirement for jetty-continuation-9.4.0?

MMaiero:
Few more comments:

- The Jetty EventSource bundle is in SNAPSHOT version. We cannot use it. 

- The Jetty EventSource is from a repository that states "Additional Jetty repositories not covered by the Eclipse Foundation IP policy". This means that a CQ will be required. Currently, only [1.0.0 is approved](https://dev.eclipse.org/ipzilla/show_bug.cgi?id=7036) but I think we will need a Piggyback CQ. ( @ctron please correct me, if I'm wrong)

- Have you done some memory test in order to verify that the leak is solved?

- I don't think we will manage to have this in M1. Why we cannot try to find the reason for the leak in the existing solution and just fix it?

kura/org.eclipse.kura.easse/META-INF/MANIFEST.MF
amitjoy: No, not really. Jetty continuation is an OSGi bundle and hence, it has been added to the runtime. On the other hand, Jetty EventSource is not an OSGi bundle but it requires Jetty continuation in its application classpath. Having Jetty Continuation is runtime will not be resolved by Jetty EventSource.

amitjoy:
I actually made the PR so that you can also test this application according to your need as well.

Jetty EventSource's snapshot version has been enhanced from the version that was released and hence, I have used it. I have also tested my bundle with the released version of this library and it didn't work due to a bug which has been fixed in the snapshot. 

Yes, you are right. I also believe we have to file a CQ for this.  @ctron Could you kindly enlighten us on this?

This library is also tested and I have found the reference to this library in many other applications. I have also tested this module and could not find any potential issue that might raise any memory problems. You can also test it and let me know if you find any such issue.

We had a very long discussion for the PR against #1000 and according to the valuable suggestions by @ctron, it's better to use third-party library to solve this problem instead of writing the whole new implementation from scratch. I had considered his suggestion and started developing an Event Admin SSE bundle to generalize this situation so that SSE clients can easily be notified with OSGi events. As lots of other applications are also using this Jetty EventSource library, I thought of giving it a try and it works. In addition, it also provides a very high-level design abstraction that refrains us to not deal with low-level complexities.

MMaiero:
I think that the project in that repository is definitely dead.
But I had a look at the jetty APIs and I've found this: 
http://download.eclipse.org/jetty/9.3.12.v20160915/apidocs/org/eclipse/jetty/servlets/EventSource.html
So, it seems that the APIs are there.
Looking at the code here: https://github.com/eclipse/jetty.project/tree/jetty-9.4.x/jetty-servlets/src/main/java/org/eclipse/jetty/servlets
you can find the current implementation.

It is also available in maven central:
https://mvnrepository.com/artifact/org.eclipse.jetty/jetty-servlets/9.3.9.v20160517

MMaiero:
In any case we prefer having a fix that tries to improve the current implementation. This fix will only exist in M1.
Later on, when discussed and, eventually, solved some corner cases of this solution, it will be merged. So, the target of this PR will be M2.

kura/org.eclipse.kura.easse/META-INF/MANIFEST.MF
ctron: But you can use a bundle or package import. There is no need to embed this!

kura/org.eclipse.kura.easse/build.properties
ctron: See above.

kura/org.eclipse.kura.easse/META-INF/MANIFEST.MF
amitjoy: I know. That's the OSGi way of modularity but the library is not provided as a OSGi bundle and we cannot use it as package import or require bundle.

amitjoy:
I am modifying this bundle to reference to the jetty-servlets which I believe, is a bundle and can be embedded in the runtime. If it goes well, I will update here.

kura/org.eclipse.kura.easse/META-INF/MANIFEST.MF
ctron: @amitjoy would you say the following looks like a valid OSGi manifest?!
```
Manifest-Version: 1.0
Require-Capability: osgi.ee;filter:="(&(osgi.ee=JavaSE)(version=1.8))"
Bundle-DocURL: http://www.eclipse.org/jetty
Implementation-Vendor: Eclipse.org - Jetty
Built-By: jesse
Bundle-SymbolicName: org.eclipse.jetty.continuation
Export-Package: org.eclipse.jetty.continuation;version="9.3.9";uses:="
 javax.servlet"
Bundle-Version: 9.3.9.v20160517
url: http://www.eclipse.org/jetty
Bundle-Description: Jetty module for Jetty :: Continuation
Build-Jdk: 1.8.0_77
Bundle-Classpath: .
Bundle-Vendor: Eclipse Jetty Project
Bnd-LastModified: 1463525267180
Bundle-Name: Jetty :: Continuation
```

amitjoy: @ctron Yes it is. But the scenario is different here. Jetty EventSource library is not an OSGi bundle and hence, it cannot be embedded in the OSGi runtime. This library has an internal dependency to Jetty Continuation and I have now 2 options:

1. Either embed Jetty Continuation in my bundle so that Jetty EventSource can resolve it in its application classpath
2. Use `Require-Bundle` or `Import-Package` in my bundle to resolve the Jetty Continuation bundle that is already present in the runtime

I embedded this as it is updated than the one in the runtime. I don't have any problem with removing this Jetty Continuation and use the one that is already present in the runtime but before that I have to check if Jetty EventSource is compatible with the Jetty Continuation that is in the container.

ctron: Perfect. So I would suggest to with option 2. then.

amitjoy:
@MMaiero Thanks for referring to the `Jetty Servlets` Subproject. The `Jetty EventSource` has been moved and merged with `Jetty Servlets` [1][2].

@ctron I have removed all embedded dependencies now as the `Jetty Servlets` is now an OSGi bundle and is embedded in the container.

[1] Binary: https://mvnrepository.com/artifact/org.eclipse.jetty/jetty-servlets/9.3.9.v20160517
[2] Sources: http://search.maven.org/remotecontent?filepath=org/eclipse/jetty/jetty-servlets/9.3.9.v20160517/jetty-servlets-9.3.9.v20160517-sources.jar

I will perform some stress tests and will update here if I find any issue.

amitjoy:
I have just completed running some tests using JProfiler and it seems this solution is very lightweight. I have also tried to subscribe to multiple events from browsers at the same time and it works pretty well.

amitjoy:
Superceded by #1151

PR #1097, opened 2017-02-14
Check modem up in resetModem method

pierantoniomerlino:
The resetModem method now uses lsusb to check if the modem is up. 
Moreover, the catch statement is now performed on specific exceptions, 
not the generic Exception class.

### DIFF
--- a/kura/org.eclipse.kura.linux.net/src/main/java/org/eclipse/kura/linux/net/modem/ModemDriver.java
+++ b/kura/org.eclipse.kura.linux.net/src/main/java/org/eclipse/kura/linux/net/modem/ModemDriver.java
@@ -281,16 +281,22 @@ public class ModemDriver {
                 echoSysfsResource(RELIAGATE_10_05_GSM_USB_PATH, true);
 
                 // wait until the modem is on again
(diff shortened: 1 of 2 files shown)


kura/org.eclipse.kura.linux.net/src/main/java/org/eclipse/kura/linux/net/modem/ModemDriver.java
```diff
@@ -281,16 +281,22 @@ public boolean resetModem() {
...
-                // int cnt = 10;
-                // while (!isOn() && cnt > 0) {
-                // sleep(1000);
-                // cnt--;
-                // }
-                // if (!isOn()) {
-                // retVal = false;
-                // }
-            } catch (Exception e) {
+                int cnt = 10;
```
MMaiero: I think that maybe having retVal defined as false, than in line 289 if (isOn) { retVal = true;}
This would prevent having so many retVal=false in the catches.

pierantoniomerlino: 👍

Issue #1096, opened 2017-02-14
[Testing] Tests for org.eclipse.kura.core.util


djurod:
Writing tests for org.eclipse.kura.core.util package in org.eclipse.kura.core project.

cdealti:
Manually merged. See #92f7ea9

PR #1098, opened 2017-02-14
#1096: Tests for org.eclipse.kura.core.util

djurod:
- Prepared unit tests for NetUtil, ProcessUtil, SafeProcess, and
  ValidationUtil classes.
- Improved input validation and conversion of hardware address from and
  to string in methods hardwareAddressToString() and
  hardwareAddressToBytes().
- Fixed some comments in ValidationUtil class. It was referencing
  EdcIllegalNullArgumentException instead of KuraException. Also
  documentation of notEmptyOrNull() method was extended to note that the
  exception is thrown also in case the input value contains only
  whitespace characters.
- Increased BREE from Java 1.6 to Java 1.8

### DIFF
--- a/kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/util/NetUtil.java
+++ b/kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/util/NetUtil.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2011, 2016 Eurotech and/or its affiliates
+ * Copyright (c) 2011, 2017 Eurotech and/or its affiliates
  *
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
@@ -14,6 +14,7 @@ package org.eclipse.kura.core.util;
 import java.net.InetAddress;
 import java.net.NetworkInterface;
 import java.util.Enumeration;
+import java.util.StringJoiner;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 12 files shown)


djurod:
As requested by @MMaiero the version of package org.eclipse.kura.core.util was increased. I also fixed the copyright date where appropriate.

kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/util/NetUtil.java
```diff
@@ -26,27 +27,45 @@ public static String hardwareAddressToString(byte[] macAddress) {
         if (macAddress == null) {
             return "N/A";
         }
-        StringBuilder sb = new StringBuilder();
-        for (int i = 0; i < macAddress.length; i++) {
-            sb.append(String.format("%02X%s", macAddress[i], i < macAddress.length - 1 ? ":" : ""));
+
+        if (macAddress.length != 6) {
+            throw new IllegalArgumentException("macAddress is invalid");
```
cdealti, line 32: As a reminder, the validation is correct but it will throw an runtime exception that was not previously thrown and that the caller will likely not catch.
I guess we will find problems, if any, during release QA

kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/util/NetUtil.java
```diff
@@ -26,27 +27,45 @@ public static String hardwareAddressToString(byte[] macAddress) {
...
     public static byte[] hardwareAddressToBytes(String macAddress) {
         if (macAddress == null || macAddress.isEmpty()) {
             return new byte[] { 0, 0, 0, 0, 0, 0 };
         }
+        
+        String[] items = macAddress.split("\\:");
 
-        macAddress = macAddress.replaceAll(":", "");
+        if (items.length != 6) {
+            throw new IllegalArgumentException("mac is invalid: " + macAddress);
```
cdealti, line 51: this is OK instead. The old code would have likely thrown an ArrayIndexOutOfBoundsException accessing characters past the end of the string

cdealti:
@djurod this will be merged manually. No need to update it

cdealti:
this has been merged manually

PR #1099, opened 2017-02-15
Set of changes in EventHandlerServlet to prevent memory leak.

MMaiero:
EventHandlerServlet.java is moved to the proper folder in web bundle.

Tested with a Java profiler, it seems able to close and release all the resources.

### DIFF
--- a/kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/server/servlet/EventHandlerServlet.java
+++ b/kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/server/servlet/EventHandlerServlet.java
@@ -11,7 +11,7 @@
  *  Amit Kumar Mondal
  *
  *******************************************************************************/
-package org.eclipse.kura.web;
+package org.eclipse.kura.web.server.servlet;
 
 import static java.util.Objects.isNull;
 import static java.util.Objects.nonNull;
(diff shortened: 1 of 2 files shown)


MMaiero:
@amitjoy We need to speed up a little bit the things. Did you have time to review this?

Would be ok for you to merge immediately and eventually step back if we find major issues? In this way we can start the QA process almost immediately.

kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/server/servlet/EventHandlerServlet.java
```diff
@@ -174,11 +176,26 @@ public void doGet(final HttpServletRequest request, final HttpServletResponse re
...
+        boolean sessionAvailable = this.requests.containsKey(sessionId);
+        if (sessionAvailable) {
+            HttpSession storedSession = this.requests.get(sessionId);
+            try {
+                storedSession.getCreationTime();
+                return true;
+            } catch (IllegalStateException ise) {
+                return false;
+            }
+        } else {
```
amitjoy, line 197: Sugar Coating 😉 You don't need this `else` condition. Simply `return false`.

MMaiero: Yes, I know, it's a possible improvement.

Issue #908, opened 2016-12-07
Web UI shows its own WLAN for connecting


ctron:
When performing test NET-WIFI-003 you are required to select a SSID for connecting. However the dialog for selecting the SSID also shows its own SSID, which should at least be marked as its own.

cdealti:
Do you mean its own SSID is listed in the scan results?

ctron:
Yes

cdealti:
Nice catch. I've never noticed this before. I guess that filtering out the self SSID is feasible but maybe not P1.

ctron:
True.

Issue #911, opened 2016-12-07
[Kura 2.1.0 QA] SSID selection not always shows all SSIDs


ctron:
When testing NET-WIFI-003 I noticed that the SSID not always shows all SSIDs. Sometimes some are missing.

cdealti:
Yes, channels above 11 are disabled. See the kura.properties file

ctron:
No it is a channel 9 SSID

cdealti:
Can you try with system level tools e.g. ``iwlist wlan0 scan`` and ``iw wlan0 scan``?

ctron:
So for both command:
a) it is way faster
b) the SSID is found

Just tried again using the Kura Web UI and the SSID is missing again.

ctron:
I will re-try this web Raspbian to see if there is a difference

ctron:
I am not sure if there is a real difference. It seems to be a bit quicker on Raspbian. And on Raspbian I most often get a full list. Where on Fedberry I most often get an incomplete list.

But in the end I have missing SSIDs on both Raspbian and Fedberry.

cdealti:
@pierantoniomerlino is taking a look

ibinshtok:
I think that it is wireless tool that doesn't always returns all SSIDs. If so, there is nothing we can really do in Kura except may be scanning more than once but this takes time.

pierantoniomerlino:
@ibinshtok I agree on your comment. However, Kura uses a hashmap to store the discovered access points, whose key is the SSID. In our office, we have several wireless nets and some of them transmit on multiple channels. So, only the last discovered SSID is added to the map. In our case, some network use channel 1 and 13 and the latter is not allowed by Kura. The result is that in some scanning the SSID is shown in the webUI, because the last discovered SSID is on channel 1, but in others is not, since the last discovered SSID is 13.
If the `kura.wifi.top.channel` in kura.properties is set to 13, all SSID are shown.
@ctron Setting  `kura.wifi.top.channel` to 13 improves the performance?

ctron:
The channel value was previous at 11, I changed it to 13.

That didn't make the scanning faster, but it made it more complete. Indeed I do sometimes get results now which include channel 13. Which probably where missing before.

So maybe that setting should be set to 13 by default?

pierantoniomerlino:
[...]

PR #1110, opened 2017-02-17
Fix for SSID search

pierantoniomerlino:
This PR fixes issue #908 and #911. 
The code is modified in order to filter the own SSID and to return a list
instead of a map during searching for SSIDs.


pierantoniomerlino:
@MMaiero could you please take a look on this? I did a refactor of the method for finding the SSIDs and changed the versions and apis.

kura/org.eclipse.kura.api/META-INF/MANIFEST.MF
```diff
@@ -29,13 +29,13 @@ Export-Package: org.eclipse.kura;version="1.3.0",
  org.eclipse.kura.gpio;version="1.0.1",
  org.eclipse.kura.linux.udev;version="1.0.0",
  org.eclipse.kura.message;version="1.1.0",
- org.eclipse.kura.net;version="1.2.1",
+ org.eclipse.kura.net;version="1.2.2",
```
MMaiero: I think that since we are adding methods and deprecating something, the micro change in the APIs is not sufficient. We need at least to change the minor.

kura/org.eclipse.kura.net.admin/src/main/java/org/eclipse/kura/net/admin/NetworkAdminServiceImpl.java
```diff
@@ -1234,124 +1213,75 @@ public void setFirewallNatConfiguration(List<FirewallNatConfig> natConfigs) thro
...
-                                groupCiphers.add(securityEntry);
-                            }
-                        }
+                    if (wap.getSSID() == null || wap.getSSID().length() == 0
+                            || isHotspotInList(channel, wap.getSSID(), wifiHotspotInfoList)) {
+                        s_logger.debug("Skipping hidden SSID");
+                        continue;
                     }
 
+                    // if (!wap.getSSID().matches(SSID_REGEXP)){
```
MMaiero: remove this comment if not needed.

kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/server/GwtNetworkServiceImpl.java
```diff
@@ -989,34 +988,32 @@ public void updateNetInterfaceConfigurations(GwtXSRFToken xsrfToken, GwtNetInter
...
 
         try {
-            Map<String, WifiHotspotInfo> wifiHotspotInfoMap = nas.getWifiHotspots(interfaceName);
-            if (wifiHotspotInfoMap != null && !wifiHotspotInfoMap.isEmpty()) {
-                Collection<WifiHotspotInfo> wifiHotspotInfoCollection = wifiHotspotInfoMap.values();
-                Iterator<WifiHotspotInfo> it = wifiHotspotInfoCollection.iterator();
-                while (it.hasNext()) {
-                    WifiHotspotInfo wifiHotspotInfo = it.next();
+            List<WifiHotspotInfo> wifiHotspotInfoList = nas.getWifiHotspotList(interfaceName);
+            if (wifiHotspotInfoList != null && !wifiHotspotInfoList.isEmpty()) {
```
MMaiero: Is this check `&& !wifiHotspotInfoList.isEmpty()` needed?
I think that the for loop manages this case.

kura/org.eclipse.kura.net.admin/src/main/java/org/eclipse/kura/net/admin/NetworkAdminServiceImpl.java
```diff
@@ -1714,10 +1644,168 @@ private int frequencyMhz2Channel(int frequency) {
...
+
+    private WifiMode getWifiMode(String ifaceName) throws KuraException {
+        WifiMode wifiMode = WifiMode.UNKNOWN;
+        List<? extends NetInterfaceConfig<? extends NetInterfaceAddressConfig>> netInterfaceConfigs = getNetworkInterfaceConfigs();
+        for (NetInterfaceConfig<? extends NetInterfaceAddressConfig> netInterfaceConfig : netInterfaceConfigs) {
+            if (netInterfaceConfig.getName().equals(ifaceName)) {
+                List<? extends NetInterfaceAddressConfig> netInterfaceAddresses = netInterfaceConfig
+                        .getNetInterfaceAddresses();
+                if (netInterfaceAddresses != null) {
+                    wifiMode = getWifiMode(netInterfaceAddresses);
```
MMaiero, line 1658: could the break be moved here?

kura/org.eclipse.kura.net.admin/src/main/java/org/eclipse/kura/net/admin/NetworkAdminServiceImpl.java
```diff
@@ -1714,10 +1644,168 @@ private int frequencyMhz2Channel(int frequency) {
...
+            }
+        }
+        return wifiMode;
+    }
+
+    private WifiMode getWifiMode(List<? extends NetInterfaceAddressConfig> netInterfaceAddresses) {
+        WifiMode wifiMode = WifiMode.UNKNOWN;
+        for (NetInterfaceAddressConfig netInterfaceAddress : netInterfaceAddresses) {
+            if (netInterfaceAddress instanceof WifiInterfaceAddressConfig) {
+                wifiMode = ((WifiInterfaceAddressConfig) netInterfaceAddress).getMode();
```
MMaiero, line 1670: maybe a break could be put here

kura/org.eclipse.kura.net.admin/src/main/java/org/eclipse/kura/net/admin/NetworkAdminServiceImpl.java
```diff
@@ -1714,10 +1644,168 @@ private int frequencyMhz2Channel(int frequency) {
...
+                    pairCiphers.add(securityEntry);
+                } else if (securityEntry == WifiSecurity.GROUP_CCMP || securityEntry == WifiSecurity.GROUP_TKIP) {
+                    groupCiphers.add(securityEntry);
+                }
+            }
+            Iterator<WifiSecurity> itRsnSecurity = esRsnSecurity.iterator();
+            while (itRsnSecurity.hasNext()) {
+                WifiSecurity securityEntry = itRsnSecurity.next();
+                if (securityEntry == WifiSecurity.PAIR_CCMP || securityEntry == WifiSecurity.PAIR_TKIP) {
+                    if (!pairCiphers.contains(securityEntry)) {
```
MMaiero: maybe this check can be merged to the previous one

kura/org.eclipse.kura.net.admin/src/main/java/org/eclipse/kura/net/admin/NetworkAdminServiceImpl.java
```diff
@@ -1714,10 +1644,168 @@ private int frequencyMhz2Channel(int frequency) {
...
+                        pairCiphers.add(securityEntry);
+                    }
+                } else if (securityEntry == WifiSecurity.GROUP_CCMP || securityEntry == WifiSecurity.GROUP_TKIP) {
+                    if (!groupCiphers.contains(securityEntry)) {
```
MMaiero: same here

kura/org.eclipse.kura.net.admin/src/main/java/org/eclipse/kura/net/admin/NetworkAdminServiceImpl.java
```diff
@@ -1714,10 +1644,168 @@ private int frequencyMhz2Channel(int frequency) {
...
+                wifiSecurity = WifiSecurity.SECURITY_WEP;
+            }
+        }
+
+        return wifiSecurity;
+
+    }
+
+    private void setCiphers(WifiHotspotInfo wifiHotspotInfo, WifiAccessPoint wap, WifiSecurity wifiSecurity) {
+        EnumSet<WifiSecurity> esWpaSecurity = wap.getWpaSecurity();
```
MMaiero: maybe you could try to simplify the things by merging the sets together and having a single iterator. Furthermore, the checks seem to be repetitive, so maybe they could be extracted.

kura/org.eclipse.kura.api/META-INF/MANIFEST.MF
pierantoniomerlino: 👍

kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/server/GwtNetworkServiceImpl.java
pierantoniomerlino: 👍

kura/org.eclipse.kura.net.admin/src/main/java/org/eclipse/kura/net/admin/NetworkAdminServiceImpl.java
pierantoniomerlino: 👍

kura/org.eclipse.kura.net.admin/src/main/java/org/eclipse/kura/net/admin/NetworkAdminServiceImpl.java
pierantoniomerlino: Yes, indeed this can simplify a lot the code...

MMaiero:
@pierantoniomerlino The travis build is failing due to a not correct bree. Could you please have a look?

kura/org.eclipse.kura.api/META-INF/MANIFEST.MF
kartben: Actually, adding a method to this interface is a breaking change, so you should increase the major. And in any case, this should also be reflected in the Bundle-Version as well.

As more and more people are starting to use and extend Kura, it would be nice to try to keep API breakage to a minimum (arguably, if everyone is extending Kura by means of importing packages rather than requiring bundles, which is a good practice anyways, this will minimize the impacts for them only to those packages they really are depending on, but still…)

cc @dwoodard1 @cdealti @ctron

ctron: Agreed with @kartben

You could alternatively use a "default" method on the Interface instead. Not sure this works though. Making such a breaking change would also allow to remove the method completely.

dwoodard1: First, I think a minor version increase is okay here. We try to follow the OSGi semantic versioning model. In this case, we are introducing a breaking change for the provider, not the consumer. The consumer is free to continue using the deprecated method, we are just indicating there is a better option. The consumer should be importing [1.2, 2), so they should be okay. I doubt there are many users of Kura re-implementing this interface. 

Second, while I agree with @kartben, didn't we decide that Kura v3.0.0 was not going to be backward compatible to v2.x? So, if we are going to make breaking changes to the API, now would be the time. As mentioned above, I don't think this is the case here, but just a general thought on the next release.

pierantoniomerlino: @kartben @ctron @dwoodard1 @MMaiero I agree with @dwoodard1 on the minor version. Indeed I changed it this commit : https://github.com/eclipse/kura/pull/1110/commits/5df8404debc49a73d0f166d596bfa08786751ac4
In general, I avoid to change the API, but in this case the only way to easily fix the issue was to add a new method and deprecate the old one. Moreover, it seems to me that v3.0.0 release is a good moment to add something to the API.

ctron: So in the past we the argument over and over that Kura follows semantic versioning ... now looking at the suggestions from "bndtools" about the API version from Kura 2.1.0 and this PR branch it shows:

```
===============================================================
* org.eclipse.kura.api 1.2.0.201702240825-1.1.0 suggests 2.0.0
===============================================================
  Package                                            Delta      New        Old        Suggest    If Prov.  
* org.eclipse.kura                                   MAJOR      1.3.0      1.2.1      2.0.0      -         
  org.eclipse.kura.annotation                        ADDED      1.0.0      -          ok         -         
  org.eclipse.kura.asset                             ADDED      1.0.0      -          ok         -         
  org.eclipse.kura.asset.listener                    ADDED      1.0.0      -          ok         -         
* org.eclipse.kura.cloud                             MAJOR      1.1.0      1.0.1      2.0.0      1.0.1     
  org.eclipse.kura.driver                            ADDED      1.0.0      -          ok         -         
  org.eclipse.kura.driver.listener                   ADDED      1.0.0      -          ok         -         
* org.eclipse.kura.gpio                              MAJOR      1.0.1      1.0.1      2.0.0      -         
* org.eclipse.kura.net                               MAJOR      1.3.0      1.2.1      2.0.0      1.2.1     
* org.eclipse.kura.net.wifi                          MAJOR      1.2.1      1.2.0      2.0.0      -         
* org.eclipse.kura.system                            MINOR      1.1.0      1.1.0      1.2.0      -         
  org.eclipse.kura.type                              ADDED      1.0.0      -          ok         -         
  org.eclipse.kura.wire                              ADDED      1.0.0      -          ok         -         
```

So the suggestion is to move to 2.0.0, which is what @kartben suggested. And I agree.

Unless of course you want to give up semantic versioning …

kartben: I think there is a "misalignment" between historical Kura (i.e a framework mostly targetting ESF use cases) and where Kura is going now. Correct me if I'm wrong but I believe @dwoodard1 and other ETH folks PoV is that while the interface is API, it is *implicitly* meant to be a "@noimplement", i.e people are only expected to consume implementations of this API, not to to provide their own to the framework. This is of course perfectly fine, and there are many cases in the Eclipse platform where that is indeed the case, and where for many reason it is not desirable, or even not supported, to have different implementations of an API.

That being said, the "problem" now that Eclipse Kura is getting more and more adoption, is that those not-to-be-implemented interfaces need to be properly documented (ideally annotations such as @noimplement are used too, to help drive automated tools such as PDE API tooling or bnd), because it is just not safe to assume that people won't want to extend the Kura framework to provide their own "tweaks" - eh, that's what OSGi is great for, right? :-)

Hope this makes sense!

kartben: Just to be clear, sticking to a minor increment would be absolutely fine and compatible with semantic versioning *if* the API is properly identified as a "for consumers only" (so I would disagree with @ctron here), unfortunately that is not the case today.

ctron: Ok, so after reading through the OSGi document again, I have to correct myself. *If* the API is considered to be implemented by "providers", then a minor increment would be sufficient. However in this case also the provider's import must be adapted to this, limiting the import to the minor version. Currently this is `[1.1,2.0)`.

However I also think that this API should not be considered "@noimplement", as the whole point of Kura is to be a modular framework, which allows users/consumers to replace modules with different implementations.

So yes, you are right, the minor version bump is enough. But the provider has to be adapted accordingly. Probably a problem with other Kura provider implementations as well. However this has to be documented. Like already described in [1], this interface should be annotated with e.g. `@ProviderType`.

Maybe it would be a good idea to start providing an "API policy" document like other Eclipse projects do, e.g. WTP [2], which is a bit more project specific than the page [1].

[1] https://github.com/eclipse/kura/wiki/Kura-Semantic-Versioning
[2] https://wiki.eclipse.org/WTP_API_Policy

cdealti: > this interface should be annotated with e.g. @ProviderType.

I have an fully annotated version of the Kura API that I intend to submit after the milestone release.

Issue #1053, opened 2017-01-27
dhclient is not being killed if interface is switched from DHCP to static.  


ibinshtok:
The LinuxNetworkUtil.disableInterface(interfaceName) call in the disableInterface() method of the EthernetMonitorServiceImpl class should be replaced by the m_netAdminService.disableInterface(interfaceName)

PR #1114, opened 2017-02-17
Fix a bug in the disableInterface() method of the EthernetMonitorServiceImpl class

ibinshtok:
This PR fixed a bug in the disableInterface() method of the EthernetMonitorServiceImpl class (Kura issue #1053). This bug prevents dhclient to be properly killed when ethernet interface is switched from DHCP to static mode. The fix is to use the m_netAdminService.disableInterface() method instead of LinuxNetworkUtil.disableInterface(). 
@MMaiero please review and merge when you have a chance.


MMaiero (approved):

MMaiero:
@cdealti I think we should merge this in M2. What do you think?

PR #1120, opened 2017-02-21
Fixed Internet Explorer EventSource streaming

amitjoy:


MMaiero:
Tested on IE 11: it does not work

amitjoy:
So far what I have found, [1] indicates the problem about IE 9. Not sure if it is different in IE10+.

[1] - https://blogs.msdn.microsoft.com/ieinternals/2010/04/05/comet-streaming-in-internet-explorer/

MMaiero:
Can we close this in favour of #1152?

amitjoy:
@MMaiero We can close this as #1151 provides generic support for streaming OSGi events to Server Sent Events clients in all the browsers inclusive of Internet Explorer.

PR #1122, opened 2017-02-22
Implement new Cloud Client API

ctron:
This change implements the new cloud client API and adds a few fixes.


MMaiero:
Could you please resolve conflicts with the apis? Could you also squash the commits? Thanks

ctron:
@MMaiero Is it ready to merge then? I would keep the number of rebases for this PR to a minimum.

You can also squash the commit during the merge process. So there is no need to do this right now.

MMaiero:
> Is it ready to merge then? I would keep the number of rebases for this PR to a minimum.

That was my purpose, and for that reason I've asked. 😄 

> You can also squash the commit during the merge process. So there is no need to do this right now.

I prefer if you do that because, being your commits, you know how to better squash them, so they are self-contained and separately make the build properly working.

ctron:
Done

Issue #1123, opened 2017-02-22
[Kura 3.0.0 M1 QA] Bugs in camel example publisher


djurod:
Two bugs were found in the org.eclipse.kura.example.camel.publisher:

1. Wrong period is applied for both int and double signals
In examplePublisher.makeSine() method the frequency needs to be doubled. It is currently `pi / T` but it should be `2 * pi / T`.

2. Amplitude for double signal is not read correctly from the configuration
Name of the configuration value in the xml is `double.amplitude` but it should be `double.amp`

MMaiero:
@djurod I've merged #1124. Can this be closed?

djurod:
Yes

PR #1124, opened 2017-02-22
#1123: Fix reported issues

ctron:

### DIFF
--- a/kura/examples/org.eclipse.kura.example.camel.publisher/src/org/eclipse/kura/example/camel/publisher/ExamplePublisher.java
+++ b/kura/examples/org.eclipse.kura.example.camel.publisher/src/org/eclipse/kura/example/camel/publisher/ExamplePublisher.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2016 Red Hat Inc and others.
+ * Copyright (c) 2016, 2017 Red Hat Inc and others.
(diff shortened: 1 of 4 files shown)


kura/examples/org.eclipse.kura.example.camel.publisher/src/org/eclipse/kura/example/camel/publisher/ExamplePublisher.java
```diff
@@ -117,16 +117,16 @@ protected PublisherConfiguration parseConfiguration(final Map<String, Object> pr
...
+                configuration.getPeriodDouble()));
 
         // return result
 
         return result;
     }
 
     private static double makeSine(final double amp, final double offset, final double period) {
-        final double freq = 1.0 / period * Math.PI;
+        final double freq = 1.0 / period * Math.PI * 2.0;
```
djurod: This could be written as:
`final double freq = 2.0 / period * Math.PI;`

Or even better as:
`final double freq = 2.0 * Math.PI / period;`

djurod, line 129: This could be written as:
`final double freq = 2.0 / period * Math.PI;`

Or even better as:
`final double freq = 2.0 * Math.PI / period;`

PR #1126, opened 2017-02-22
Added functionality to allow adding Wire Components after removing one

### LINKED ISSUE #1121: [Kura 3.0.0 M1 QA] Wires: cannot add component after deleting one
After a component is removed graph must be saved if you want to add a new component:
image

It is an annoying limitation that should be removed, if there is no specific need for it.
Option is for this dialog to allow saving the graph and the desired component is added immediately after the dialog is closed.

Low-priority issue.

Kura 3.0.0-SNAPSHOT, build 945.

amitjoy:
This allows adding Wire Components of same names to the Wire Graph even
after removing a Wire Component of the same name. This does not require
saving the Wire Graph before adding new Wire Components.

Fixes #1121

### DIFF
--- a/kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/client/ui/wires/WiresPanelUi.java
+++ b/kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/client/ui/wires/WiresPanelUi.java
@@ -13,6 +13,8 @@
  *******************************************************************************/
 package org.eclipse.kura.web.client.ui.wires;
 
+import static org.eclipse.kura.web.shared.service.GwtWireService.DELETED_WIRE_COMPONENT;
+
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.LinkedList;
@@ -203,7 +205,6 @@ public class WiresPanelUi extends Composite {
         exportJSNIUpdateWireConfig();
         exportJSNIUpdateSelection();
         exportJSNIShowDuplicatePidModal();
-        exportJSNIShowAddNotAllowedModal();
         exportJSNIshowCycleExistenceError();
         exportJSNImakeUiDirty();
         exportJSNIDeactivateNavPils();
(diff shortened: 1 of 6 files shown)


amitjoy:
@MMaiero If I'm not wrong this will go in KURA-3.0.0-M2, right?

MMaiero:
Yes

amitjoy:
@markoer If possible, could you kindly have a go at it? As far as I tested, this works as expected.

markoer:
It looks much better, now.

I noted one reproducible problem, however. I'm not sure whether this is the cause or not, but still. The TAB key's not working in the component adding dialog anymore. The result is that you can type-in the component name, but cannot tab to the Apply button.

One more I noted, but couldn't reproduce, was that at some point I got a warning that I'm trying to add a component named as one that's already there. The funny thing is that the graph was all but empty and definitely without a component named like that.

MMaiero:
> I noted one reproducible problem, however. I'm not sure whether this is the cause or not, but still. The TAB key's not working in the component adding dialog anymore. The result is that you can type-in the component name, but cannot tab to the Apply button.

Seen that too but, after a look at the code I couldn't find a possible reason. I will try with a develop build and see if the issue was already there.

> One more I noted, but couldn't reproduce, was that at some point I got a warning that I'm trying to add a component named as one that's already there. The funny thing is that the graph was all but empty and definitely without a component named like that.

Never happened to me.

amitjoy:
@markoer I have just tested the feature to check if this PR introduced any problem with the TAB key but I checked in Chrome and Safari and the TAB key switches to `Cancel` and `Apply` both after typing in the name of the Wire Component.

I would update it here if there is any issue that reports existent Wire Component while the graph does not contain it.

amitjoy:
@markoer I have also tested for issues if it reports for existent Wire Component while it is not at all present in the Wire Graph. Never encountered it so far. Let me know if you can reproduce it.

markoer:
Since I couldn't reproduce the issue with duplicate component names, I suppose the original intent of this PR is satisfied.

I tried tabbing to the buttons in Firefox 50.1 and Chrome 56.0. In FF it doesn't work, in Chrome it works. So it's probably really an unrelated issue.

MMaiero:
Ok, merging then

Issue #1131, opened 2017-02-25
ConfigurableComponentTracker calls bundleContext.getService(ref) when it shouldn't


scottslewis:
Currently the ConfigurationServiceImpl creates and starts a ConfigurableServiceTracker.  Here's the source for ConfigurableComponentTracker:

https://github.com/eclipse/kura/blob/develop/kura/org.eclipse.kura.core.configuration/src/main/java/org/eclipse/kura/core/configuration/ConfigurableComponentTracker.java

The ConfigurableServiceTracker is created with this constructor code:

        // TODO: find a better filter
        // This works but we track everything
        super(context, context.createFilter("(" + Constants.OBJECTCLASS + "=*)"), null);

As it says in the comment this filter of objectClass=* means that every created service is tracked.   One consequence of this is that all services registered will have their context.getService(ref) method called by this tracker, whether or not they have any actual consumers.    This is wasteful and defeats the declarative services laziness.

Another consequence is that for extenders like OSGi Remote Service Admin and Blueprint, the calls to getService(ref) will fail (return null) because these extenders create proxies for the service, and they are specified to use the Bundle consumer (in this case org.eclipse.kura.core.configuration) to load the service interfaces classes.   Because the kura configuration bundle classloader does/will not have access to some service interfaces classes (e.g. those services created by apps, etc) it will be unable to call context.getService(ref) when these extenders are being used.

Ideally the ConfigurableServiceTracker should not track all services, but rather some subset (e.g. based upon service properties or service interfaces) of services in order to avoid creating spurious errors when proxies are created.

ctron:
There currently is a way to achieve this, unfortunately this is not the default behavior of Kura at the moment.

You can set the system property `org.eclipse.kura.core.configuration.trackOnlyRelevantServices` to `true` and it will not use the `ConfigurableComponentTracker` but do proper tracking of services. However it was considered a breaking change back when I made this change and so the default sticks with the "track all" approach.

Also see: https://github.com/eclipse/kura/blob/develop/kura/org.eclipse.kura.core.configuration/src/main/java/org/eclipse/kura/core/configuration/ConfigurationServiceImpl.java#L85

ctron:
I would suggest to remove the old behavior as Kura 3.0.0 may do breaking changes.

scottslewis:
[...]

PR #1137, opened 2017-02-28
#1131: Only perform proper service tracking

ctron:
This PR removes the old behavior of tracking just all services. As Kura 3.0.0 may do breaking changes, we should go for a proper solution and drop the old code, causing serious issues in e.g. Apache Felix.


ctron:
I think this would be superseded by #1916

PR #1140, opened 2017-02-28
Improve separation of concerns for camel java example

apupier:


ctron:
I general I like the approach!

But that doesn't work 100%. The min/max fields are actually changed in `GatewayRouter` during runtime. Which would not update the values in the `RandomTemperatureGenerator` if I read if correctly.

ctron:
I had to re-base the PR on `develop` as this is the current development branch.

ctron (approved):

PR #1141, opened 2017-03-01
OPCUA driver fixes/improvements

nicolatimeus:
Added some missing package imports.
Replaced bundled jars with Milo 0.1.0 official Maven deps.
Improved request timeout handling.
Minor code cleanups.


kura/org.eclipse.kura.driver.opcua.provider/src/main/java/org/eclipse/kura/internal/driver/opcua/OpcUaDriver.java
```diff
@@ -451,61 +409,95 @@ public void unregisterDriverListener(final DriverListener listener) throws Conne
...
+        public static OpcUaRequestInfo extract(DriverRecord record) {
+            final Map<String, Object> channelConfig = record.getChannelConfig();
+            int nodeNamespaceIndex;
+            String nodeId;
+
+            try {
+                nodeId = channelConfig.get(NODE_ID).toString();
+            } catch (final Exception e) {
+                fail(record, message.errorRetrievingNodeId());
+                return null;
```
amitjoy: Just an improvement. Don't return `null`. use `Optional.ofNullable(..)` or `Optional.empty()`.

kura/org.eclipse.kura.driver.opcua.provider/src/main/java/org/eclipse/kura/internal/driver/opcua/OpcUaDriver.java
```diff
@@ -451,61 +409,95 @@ public void unregisterDriverListener(final DriverListener listener) throws Conne
...
+            }
+
+            try {
+                nodeNamespaceIndex = Integer.parseInt(channelConfig.get(NODE_NAMESPACE_INDEX).toString());
+            } catch (final Exception e) {
+                fail(record, message.errorRetrievingNodeNamespace());
+                return null;
```
amitjoy: Return `Optionals` in this class.

kura/org.eclipse.kura.driver.opcua.provider/src/main/java/org/eclipse/kura/internal/driver/opcua/OpcUaDriver.java
```diff
@@ -451,61 +409,95 @@ public void unregisterDriverListener(final DriverListener listener) throws Conne
...
+            this.nodeId = nodeId;
+        }
+
+        private static void fail(DriverRecord record, String message) {
+            record.setDriverStatus(new DriverStatus(DRIVER_ERROR_CHANNEL_NOT_ACCESSIBLE, message, null));
+            record.setTimestamp(System.currentTimeMillis());
+        }
+
+        public static OpcUaRequestInfo extract(DriverRecord record) {
+            final Map<String, Object> channelConfig = record.getChannelConfig();
```
amitjoy: Add a precondition to check if provided `DriverRecord` is null

kura/org.eclipse.kura.driver.opcua.provider/src/main/java/org/eclipse/kura/internal/driver/opcua/OpcUaDriver.java
```diff
@@ -451,61 +409,95 @@ public void unregisterDriverListener(final DriverListener listener) throws Conne
...
+        private final String nodeId;
+
+        public OpcUaRequestInfo(DataType dataType, int nodeNamespaceIndex, String nodeId) {
+            this.dataType = dataType;
+            this.nodeNamespaceIndex = nodeNamespaceIndex;
+            this.nodeId = nodeId;
+        }
+
+        private static void fail(DriverRecord record, String message) {
+            record.setDriverStatus(new DriverStatus(DRIVER_ERROR_CHANNEL_NOT_ACCESSIBLE, message, null));
```
amitjoy, line 458: Preconditions

kura/org.eclipse.kura.driver.opcua.provider/src/main/java/org/eclipse/kura/internal/driver/opcua/OpcUaDriver.java
```diff
@@ -451,61 +409,95 @@ public void unregisterDriverListener(final DriverListener listener) throws Conne
...
+        return Collections.unmodifiableList(records);
     }
 
+    private static class OpcUaRequestInfo {
+
+        private final DataType dataType;
+        private final int nodeNamespaceIndex;
+        private final String nodeId;
+
+        public OpcUaRequestInfo(DataType dataType, int nodeNamespaceIndex, String nodeId) {
```
amitjoy, line 451: Preconditions

kura/org.eclipse.kura.driver.opcua.provider/src/main/java/org/eclipse/kura/internal/driver/opcua/OpcUaDriver.java
```diff
@@ -350,72 +310,70 @@ public synchronized ChannelDescriptor getChannelDescriptor() {
...
+        if (result == null) {
+            throw new NoSuchElementException(message.errorNullResult());
+        }
+        return result;
+    }
+
+    private void checkStatus(StatusCode status) throws IOException {
+        logger.debug("result status code: {}", status);
+        if (status == null) {
+            throw new NoSuchElementException(message.errorNullResult());
```
amitjoy: Is there any specific reason of choosing `NoSuchElementException` over `NPE`?

kura/org.eclipse.kura.driver.opcua.provider/src/main/java/org/eclipse/kura/internal/driver/opcua/OpcUaDriver.java
```diff
@@ -350,72 +310,70 @@ public synchronized ChannelDescriptor getChannelDescriptor() {
         }
     }
 
+    private Object extractValue(DataValue value) throws IllegalArgumentException {
```
amitjoy, line 306: value can be null. Add Precondition

kura/org.eclipse.kura.driver.opcua.provider/src/main/java/org/eclipse/kura/internal/driver/opcua/OpcUaDriver.java
```diff
@@ -350,72 +310,70 @@ public synchronized ChannelDescriptor getChannelDescriptor() {
...
+        Variant variant = value.getValue();
+        if (variant == null) {
+            throw new NoSuchElementException(message.errorNullResult());
```
amitjoy: Is there any specific reason of choosing `NoSuchElementException` over `NPE`?

kura/org.eclipse.kura.driver.opcua.provider/src/main/java/org/eclipse/kura/internal/driver/opcua/OpcUaDriver.java
```diff
@@ -350,72 +310,70 @@ public synchronized ChannelDescriptor getChannelDescriptor() {
...
+        }
+        Object result = variant.getValue();
+        if (result == null) {
+            throw new NoSuchElementException(message.errorNullResult());
```
amitjoy: Is there any specific reason of choosing `NoSuchElementException` over `NPE`?

kura/org.eclipse.kura.driver.opcua.provider/src/main/java/org/eclipse/kura/internal/driver/opcua/OpcUaDriver.java
```diff
@@ -106,79 +102,24 @@
...
-            Filter filter = null;
-            try {
-                filter = ctx.createFilter(urlFilter);
-            } catch (final InvalidSyntaxException e) {
-                return Optional.empty();
-            }
-            tracker = new ServiceTracker<>(ctx, filter, null);
-            tracker.open();
-            urlConverterTrackers.put(protocol, tracker);
+    private <U> U runSafe(Future<U> future) throws ExecutionException, InterruptedException, TimeoutException {
```
amitjoy, line 114: Future instance can be null. Add Precondition.

kura/org.eclipse.kura.driver.opcua.provider/src/main/java/org/eclipse/kura/internal/driver/opcua/OpcUaDriver.java
```diff
@@ -106,79 +102,24 @@
     /** Node Namespace Index Property */
     private static final String NODE_NAMESPACE_INDEX = "node.namespace.index";
 
-    /** URL Converter Service Trackers */
-    private static Map<String, ServiceTracker<Object, URLConverter>> urlConverterTrackers = CollectionUtil
-            .newConcurrentHashMap();
-
-    // initialize Milo Stack class loader
-    static {
```
amitjoy, line 114: I believe the class loading issues with `Milo stack lib` should not be there anymore, right?

kura/org.eclipse.kura.driver.opcua.provider/META-INF/MANIFEST.MF
```diff
@@ -7,7 +7,10 @@ Bundle-Vendor: Eclipse Kura
 Bundle-License: Eclipse Public License v1.0
 Bundle-Category: Asset-Driver Management
 Bundle-ActivationPolicy: lazy
-Import-Package: org.eclipse.kura;version="[1.2,2.0)",
+Import-Package: javax.security.cert,
```
amitjoy, line 10: remove `org.eclipse.osgi.service.urlconversion` from import-package as it is not used any more. I used it for dynamic classloading.

kura/org.eclipse.kura.driver.opcua.provider/src/main/java/org/eclipse/kura/internal/driver/opcua/OpcUaDriver.java
```diff
@@ -451,61 +409,95 @@ public void unregisterDriverListener(final DriverListener listener) throws Conne
...
-            final NodeId nodeId = new NodeId(nodeNamespaceIndex, channelConfig.get(NODE_ID).toString());
+            final NodeId nodeId = new NodeId(requestInfo.nodeNamespaceIndex, requestInfo.nodeId);
             final VariableNode node = this.client.getAddressSpace().createVariableNode(nodeId);
             final DataValue newValue = new DataValue(new Variant(value.getValue()));
             try {
-                node.writeValue(newValue).get();
+                logger.debug("writing: {} to ns={};s={}..", value, requestInfo.nodeNamespaceIndex, requestInfo.nodeId);
+                checkStatus(runSafe(node.writeValue(newValue)));
                 record.setDriverStatus(new DriverStatus(WRITE_SUCCESSFUL));
+                logger.debug("write successful");
```
amitjoy: You can keep this string literal in localization resource.

kura/org.eclipse.kura.driver.opcua.provider/src/main/java/org/eclipse/kura/internal/driver/opcua/OpcUaDriver.java
```diff
@@ -225,7 +171,13 @@ public synchronized void connect() throws ConnectionException {
                 endPointBuilder.append("/").append(serverName);
             }
 
-            final EndpointDescription[] endpoints = UaTcpStackClient.getEndpoints(endPointBuilder.toString()).get();
+            String endpointString = endPointBuilder.toString();
+
+            logger.debug("Connecting to endpoint: {}", endpointString);
+
+            logger.debug("Fetching endpoint descriptions");
```
amitjoy, line 173: You can keep this string literal in localization resource.

kura/org.eclipse.kura.driver.opcua.provider/src/main/java/org/eclipse/kura/internal/driver/opcua/OpcUaDriver.java
```diff
@@ -225,7 +171,13 @@ public synchronized void connect() throws ConnectionException {
                 endPointBuilder.append("/").append(serverName);
             }
 
-            final EndpointDescription[] endpoints = UaTcpStackClient.getEndpoints(endPointBuilder.toString()).get();
+            String endpointString = endPointBuilder.toString();
+
+            logger.debug("Connecting to endpoint: {}", endpointString);
```
amitjoy, line 171: You can keep this string literal in localization resource.

amitjoy (changes requested):
Good changes 👍. Just few improvement comments.

kura/org.eclipse.kura.driver.opcua.provider/src/main/java/org/eclipse/kura/internal/driver/opcua/OpcUaDriver.java
MMaiero: It's debug, so it's not supposed to be localized. In this way we also prevent the cost of localization for strings that in 95% of cases are not shown.

kura/org.eclipse.kura.driver.opcua.provider/src/main/java/org/eclipse/kura/internal/driver/opcua/OpcUaDriver.java
nicolatimeus: Yes, the class loading issues were caused by some missing packages that were used by Milo deps. I've imported those packages and the issues should be solved.

kura/org.eclipse.kura.driver.opcua.provider/src/main/java/org/eclipse/kura/internal/driver/opcua/OpcUaDriver.java
amitjoy: My bad. I didn't notice it's a DEBUG log. Yes, you are right, it should not be localized.

kura/org.eclipse.kura.driver.opcua.provider/src/main/java/org/eclipse/kura/internal/driver/opcua/OpcUaDriver.java
nicolatimeus: runSafe() is a private method, so I think that preconditions are not needed if the rest of the code is able to ensure that this method never receives a null value.
In my opinion preconditions are good and needed only for public/API methods, if the method is private you have complete control on the arguments it receives and you can ensure that they are valid.

kura/org.eclipse.kura.driver.opcua.provider/META-INF/MANIFEST.MF
nicolatimeus: :thumbsup: I forgot about that

kura/org.eclipse.kura.driver.opcua.provider/src/main/java/org/eclipse/kura/internal/driver/opcua/OpcUaDriver.java
amitjoy: I tend to agree to some extent but if there is a precondition on a nonpublic method, then this signifies the author(s) of a class believes the precondition to hold no matter what a client does with the class, then assertions or preconditions are entirely appropriate.

nicolatimeus: Another argument could be that adding preconditions in private methods adds unnecessary overhead: you have an additional method invocation and null check for every requireNonNull(), this can be undesirable in particular for short and frequently called methods.
If the precondition is not satisfied you will end up with a NullPointerException, but the same happens if you do not add the precondition and you try to use a null pointer, the second case is faster however because you don't have unnecessary checks.

kura/org.eclipse.kura.driver.opcua.provider/src/main/java/org/eclipse/kura/internal/driver/opcua/OpcUaDriver.java
nicolatimeus: Ok, I can do that.

kura/org.eclipse.kura.driver.opcua.provider/src/main/java/org/eclipse/kura/internal/driver/opcua/OpcUaDriver.java
amitjoy: This is certainly a programmer's preference and I am a bit inclined to it as we are moving to Java 8 runtime, such overheads are considered good folks. It's upto you what you prefer.

kura/org.eclipse.kura.driver.opcua.provider/src/main/java/org/eclipse/kura/internal/driver/opcua/OpcUaDriver.java
nicolatimeus: Not really, If you prefer NullPointerException I can switch to it.

amitjoy: Since in most of the source base in Kura, null parameters throw NPE, I believe it is better to switch to NPE.

nicolatimeus: Ok, I agree.

MMaiero (approved):

kura/org.eclipse.kura.driver.opcua.provider/src/main/java/org/eclipse/kura/internal/driver/opcua/OpcUaDriver.java
```diff
@@ -451,61 +399,93 @@ public void unregisterDriverListener(final DriverListener listener) throws Conne
...
-                record.setDriverStatus(new DriverStatus(WRITE_SUCCESSFUL));
+                nodeNamespaceIndex = Integer.parseInt(channelConfig.get(NODE_NAMESPACE_INDEX).toString());
             } catch (final Exception e) {
-                record.setDriverStatus(new DriverStatus(WRITE_FAILURE, message.writeFailed(), e));
+                fail(record, message.errorRetrievingNodeNamespace());
+                return Optional.empty();
             }
-            record.setTimestamp(System.currentTimeMillis());
+
+            final DataType dataType = (DataType) record.getChannelConfig().get(CHANNEL_VALUE_TYPE.value());
```
amitjoy: replace `record.getChannelConfig()` with `channelConfig` local variable that is already present.

amitjoy (changes requested):

PR #1153, opened 2017-03-03
Implemented alphabetic sorting of dropdown lists in the Web Ui.

### LINKED ISSUE #244: [Web UI] Alphabetical ordering of metatype labels option in ListBoxes
In a custom bundle I have a configuration parameter setting the local timezone. This parameter proposes a long list of timezone names. In the current configuration this list is not ordered, the drop down list of labels is extracted from the xml file in a random order and will be very difficult to use by the user.

I propose to add an alphabetical ordering of the drop down list options . 

Do you think this can be generalized as a new feature of Kura web ui.
optionlist

nicolatimeus:
Closes #244


nicolatimeus:
This PR does not implement sorting for all dropdown lists in the Web UI but only for those included in sections that are derived from AbstractServicesUi (this includes at least all components under the "Services" menu and the configuration section of the Wire Components).
Sorting of the remaining lists will be eventually performed with future PRs.

PR #1158, opened 2017-03-06
Refactored System(Admin)ServiceImpl

markoer:
Extracted superclass with common methods

### DIFF
--- a/kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/system/SuperSystemService.java
+++ b/kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/system/SuperSystemService.java
@@ -1,0 +1,70 @@
+/*******************************************************************************
+ * Copyright (c) 2017 Eurotech and/or its affiliates and others
+ *
+ *   All rights reserved. This program and the accompanying materials
+ *   are made available under the terms of the Eclipse Public License v1.0
+ *   which accompanies this distribution, and is available at
+ *   http://www.eclipse.org/legal/epl-v10.html
+ ******************************************************************************/
+package org.eclipse.kura.core.system;
+
+import java.io.BufferedReader;
+import java.io.IOException;
+import java.io.InputStreamReader;
+
+import org.eclipse.kura.core.util.ProcessUtil;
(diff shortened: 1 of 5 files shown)


kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/system/SuperSystemService.java
```diff
@@ -0,0 +1,70 @@
...
+ */
+public class SuperSystemService {
+
+    private static final Logger logger = LoggerFactory.getLogger(SuperSystemService.class);
+
+    protected String runSystemCommand(String command) {
+        return runSystemCommand(command.split("\\s+"));
+    }
+
+    protected static String runSystemCommand(String[] commands) {
```
amitjoy, line 31: Instead of making it a statefull class, you can make it an utility class containing `runSystemCommand(String[])` and `runSystemCommand(String)` static methods. In such scenario, you don't have to extend this class, rather you can use the methods statically in `SystemAdminServiceImpl` and `SystemServiceImpl`.

markoer: I could and maybe would, if I didn't have to test the *Impl classes.

cdealti:
@markoer this will be merged manually. No need to rebase it

kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/system/SuperSystemService.java
ctron: I would mandate to use Commons Exec instead. Which already is in Kura. And remove code duplication.

cdealti:
@markoer I don't see tests depending on this refactor. Do you think this can be delayed a little bit?

markoer:
It can wait as long as you like.

cdealti:
@markoer do you have some tests ready for this refactor?

markoer:
There are no tests for this as the results are too OS-specific.

cdealti:
can we close it and reopen when you will have some tests?

markoer:
I'm just about to push some tests.

PR #1167, opened 2017-03-07
Fixed Wrong Order of Operation Execution

### LINKED ISSUE #1166: Wrong order of operations in WireAsset
In WireAsset.onWireReceive() method from org.eclipse.kura.internal.wire.asset package (from org.eclipse.kura.wire.component.provider bundle) the method documentation says that first all associated reading channels are read and after that all associated writing channels are written, but the method is implemented in a way that first the writing channels are written and then the reading channels are read.

Is this a bug in implementation or in documentation?

@amitjoy: since it appears that you originally wrote both the documentation and the implementation where this inconsistency was already present, maybe you could comment on the correct order?

amitjoy:
Fixes #1166

### DIFF
--- a/kura/org.eclipse.kura.wire.component.provider/src/main/java/org/eclipse/kura/internal/wire/asset/WireAsset.java
+++ b/kura/org.eclipse.kura.wire.component.provider/src/main/java/org/eclipse/kura/internal/wire/asset/WireAsset.java
@@ -209,8 +209,8 @@ public final class WireAsset extends BaseAsset implements WireEmitter, WireRecei
      * <li>Perform all write operations on associated writing channels</li>
      * <ul>
      *
-     * Both of the aforementioned operations are performed as soon as it timer
-     * wire component is also triggered.
+     * Both the aforementioned operations are performed as soon as this Wire Component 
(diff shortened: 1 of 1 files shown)


kura/org.eclipse.kura.wire.component.provider/src/main/java/org/eclipse/kura/internal/wire/asset/WireAsset.java
```diff
@@ -209,8 +210,8 @@ protected String getFactoryPid() {
      * <li>Perform all write operations on associated writing channels</li>
      * <ul>
      *
-     * Both of the aforementioned operations are performed as soon as it timer
-     * wire component is also triggered.
+     * Both the aforementioned operations are performed as soon as this Wire Component 
+     * receives emit events triggered by {@link Timer} Wire Component.
```
MMaiero: I don't think that the only source of events for an asset is the timer. In my opinion, we should stay general.

amitjoy: I agree 👍

PR #1170, opened 2017-03-07
Improved name check when creating new wire components

### LINKED ISSUE #1125: [Kura 3.0.0 M1 QA] Wires: Unexpected component configuration is shown
Steps to reproduce:
- Add a modbus service and name it "modbusDriver"
- Go to Wires and add any component and name it "modbusDriver"

Expected result: Appropriate component configuration for the selected component

Actual result: Content of the modbus service configuration is shown instead of the expected wires component configuration.

Note: the problem appears to be triggered when you name the Wires component the same as the modbus service.

nicolatimeus:
Closes #1125


kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/shared/service/GwtComponentService.java
```diff
@@ -28,6 +28,21 @@
 public interface GwtComponentService extends RemoteService {
 
     /**
+     * Return the PIDs for all the services that
+     * implements the ConfigurableComponent Maker interface and registered themselves
```
MMaiero: *marker

Maybe we can call it generally as findTrackedPids without exposing the underneath implementation.
What do you think?

PR #1174, opened 2017-03-09
Switch default behavior to only track relevant services

ctron:
Also change private fields to new Kura style.


cdealti:
Maybe we could use this PR to also remove some wrong registered services.
Examples need to only provide ConfigurableComponent. Registering an implementation class does not makes sense.
See https://github.com/eclipse/kura/commit/c88cce61fe9f7d3370a0f464d66fd61ecedd1f59

ctron:
Sure, I can amend this PR.

ctron:
On a second thought, maybe we keep this as is not to break accidentally anything. Because this change cannot be reversed by changing a system property.

ctron:
We will create a new PR for removing the double registration.

MMaiero:
Merging. Let's see what happens 😂

PR #1177, opened 2017-03-09
Fix Javadoc and Method Name in DbServiceHelper for better comprehension

### LINKED ISSUE #1176: Kura Wires: Inconsistent design of DbServiceHelper class
I found something in [DbServiceHelper](https://github.com/eclipse/kura/blob/develop/kura/org.eclipse.kura.wire.component.provider/src/main/java/org/eclipse/kura/internal/wire/common/DbServiceHelper.java) that is at least a bug in documentation. The getInstance() method is documented in a way that makes you think you are dealing with a singleton, but the only thing it actually does is that it calls the only (private) constructor in the class.

Probably this was meant to be a singleton at some point in the past and later changed basically to an ordinary class and the getInstance() method and its documentation remained. In this case I would recommend that the getInstance() method is removed and the constructor is made public.

But this could also be an incorrect attempt to make a singleton. This might be a bigger issue.

After taking another look at what the class does I would even say that it is probably even not needed. The only thing it does compared to the class it »wraps« is that it adds a few small features:
- execution of a SQL statement (with optional `int` parameters) in a transaction
- sanitation of SQL table and column names

From what I see it would make more sense to add these features directly in the DbService and just work directly with it instead of having this wrapper.

Also it might make more sense to replace the optional `int` parameters in the execute() method with optional `Object` parameters to make it more usable.

amitjoy:
Fixes #1176

### DIFF
--- a/kura/org.eclipse.kura.wire.component.provider/src/main/java/org/eclipse/kura/internal/wire/common/DbServiceHelper.java
+++ b/kura/org.eclipse.kura.wire.component.provider/src/main/java/org/eclipse/kura/internal/wire/common/DbServiceHelper.java
@@ -55,15 +55,15 @@ public final class DbServiceHelper {
     }
 
     /**
-     * Gets the single instance of DB Service Helper.
+     * Creates instance of {@link DbServiceHelper}
      *
      * @param dbService
-     *            the DB service
-     * @return single instance of Service Helper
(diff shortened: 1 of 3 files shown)


djurod:
As long as the build doesn't fail, the fix looks satisfactory to me.

PR #1181, opened 2017-03-10
Adding a "non-network" version for Fedora 25

ctron:


MMaiero:
@ctron I've solved the conflict and merged the PR.

I've also done some further changes to snapshot and other files that were not nn-compliant.

Please notice that, if I've not missed something, fedora will not expose port 80. You need to do that manually. Maybe we should point this out somewhere.

ctron:
> Please notice that, if I've not missed something, fedora will not expose port 80. You need to do that manually. Maybe we should point this out somewhere.

Good point!

PR #1182, opened 2017-03-10
Removed timeout based on magic number.

MMaiero:
Introduced a proper service tracker. This allows a proper component and ui management
with all the types of devices.

### DIFF
--- a/kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/server/GwtComponentServiceImpl.java
+++ b/kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/server/GwtComponentServiceImpl.java
@@ -38,6 +38,7 @@ import org.eclipse.kura.configuration.metatype.AD;
 import org.eclipse.kura.configuration.metatype.Icon;
 import org.eclipse.kura.configuration.metatype.OCD;
 import org.eclipse.kura.configuration.metatype.Option;
+import org.eclipse.kura.util.service.ServiceUtil;
 import org.eclipse.kura.web.server.util.GwtServerUtil;
 import org.eclipse.kura.web.server.util.KuraExceptionHandler;
 import org.eclipse.kura.web.server.util.ServiceLocator;
@@ -58,6 +59,8 @@ public class GwtComponentServiceImpl extends OsgiRemoteServiceServlet implements
     private static final String KURA_SERVICE_PID = ConfigurationService.KURA_SERVICE_PID;
     private static final String SERVICE_FACTORY_PID = "service.factoryPid";
     private static final String KURA_UI_SERVICE_HIDE = "kura.ui.service.hide";
+
+    private static final int SERVICE_WAIT_TIMEOUT = 60;
 
     private static final long serialVersionUID = -4176701819112753800L;
 
@@ -253,12 +256,15 @@ public class GwtComponentServiceImpl extends OsgiRemoteServiceServlet implements
                 if (conf != null) {
                     conf.getConfigurationProperties().put(ConfigurationAdmin.SERVICE_FACTORYPID, factoryPid);
                 }
-                if ((conf != null) && (conf.getDefinition() == null)) {
+                if (conf != null && conf.getDefinition() == null) {
                     String temporaryName = String.valueOf(System.nanoTime());
                     cs.createFactoryConfiguration(factoryPid, temporaryName, extraProps, false);
                     try {
-                        // wait for the services to be up
-                        TimeUnit.MILLISECONDS.sleep(500);
+
+                        // track and wait for the wire Component
+                        String filterString = "(" + ConfigurationService.KURA_SERVICE_PID + "=" + temporaryName + ")";
+                        ServiceUtil.waitForService(filterString, SERVICE_WAIT_TIMEOUT, TimeUnit.SECONDS);
+
                         conf = cs.getComponentConfiguration(temporaryName);
                         comp = createMetatypeOnlyGwtComponentConfiguration(conf);
                         return comp;
@@ -607,8 +613,8 @@ public class GwtComponentServiceImpl extends OsgiRemoteServiceServlet implements
                 gwtConfig.set(DRIVER_PID, props.get(DRIVER_PID));
             }
 
-            if ((props != null) && (props.get(SERVICE_FACTORY_PID) != null)) {
-                String pid = this.stripPidPrefix(config.getPid());
+            if (props != null && props.get(SERVICE_FACTORY_PID) != null) {
+                String pid = stripPidPrefix(config.getPid());
                 gwtConfig.setComponentName(pid);
                 gwtConfig.setFactoryComponent(true);
                 gwtConfig.setFactoryPid(String.valueOf(props.get(ConfigurationAdmin.SERVICE_FACTORYPID)));


kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/server/GwtComponentServiceImpl.java
```diff
@@ -59,6 +62,8 @@
     private static final String SERVICE_FACTORY_PID = "service.factoryPid";
     private static final String KURA_UI_SERVICE_HIDE = "kura.ui.service.hide";
 
+    private static final int SERVICE_WAIT_TIMEOUT = 60000;
```
amitjoy: Make it local to the method if it is only used in one place

kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/server/GwtComponentServiceImpl.java
```diff
@@ -244,12 +250,20 @@ public GwtConfigComponent findWireComponentConfigurationFromPid(GwtXSRFToken xsr
...
                     conf.getConfigurationProperties().put(ConfigurationAdmin.SERVICE_FACTORYPID, factoryPid);
                 }
-                if ((conf != null) && (conf.getDefinition() == null)) {
+                if (conf != null && conf.getDefinition() == null) {
                     String temporaryName = String.valueOf(System.nanoTime());
                     cs.createFactoryConfiguration(factoryPid, temporaryName, extraProps, false);
                     try {
-                        // wait for the services to be up
-                        TimeUnit.MILLISECONDS.sleep(500);
+
```
amitjoy, line 263: The initial idea was to have the worker thread sleep for 500 ms until all the temporary factory services are up and running. So, the execution gets halted here. In case of a Service Tracker, it spawns a new thread that waits for the service but the worker thread in this method will still continue and hence, sometimes it can work and sometimes it cannot.

That means, the asynchronous service call can return before the service tracker actually finds a matching service after waiting for the defined 60000 ms. This is because the service tracker spawns a new thread that waits for 60000 ms before it returns a valid service but asynchronous service call can return anytime.

One idea is to use cyclic barrier or countdown latch to check if the service is up and the barrier is closed or latch is counted down to zero before the worker thread continues to work.

MMaiero: Having a thread sleep, in this case, does not guarantee that the service will be up. This is especially true for low powered devices.
I'm not sure that the waitForService will be executed in a different thread. Looking at the code [here](http://grepcode.com/file/repo1.maven.org/maven2/org.apache.felix/org.apache.felix.framework/4.6.1/org/osgi/util/tracker/ServiceTracker.java#ServiceTracker.waitForService%28long%29), it should work as expected. Furthermore, the following line after the waitForService is a close and looking at the [javadoc](https://osgi.org/javadoc/r4v42/org/osgi/util/tracker/ServiceTracker.html#waitForService(long)) a close to the ServiceTracker, will determine the waitForService to return.

amitjoy: Yes, you are right. `Thread.sleep` will not guarantee that the services will be up and running, nor the `waitForService(long)`. As far as the feature requires, the best fit would be to go with `CountDownLatch` or `Cyclic Barrier` that would definitely guarantee that if and only if the service is up and running, the Worker Thread that is working in the asynchronous call in GWT, will halt and it will not continue if the temporary factory service is not up and running in the runtime.

MMaiero: Could you please point me out in the osgi code where a separate thread is spawn when managing the wait?

amitjoy: If you check `ServiceTracker` closely, it optimizes itself and checks if it can be run in separate thread. But in this scenario, it will not spawn new thread to wait for the service as it simply calls `Thread.wait` internally in the service instance to be tracked. Go for it 👍 . Just make the timeout local to the method.

kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/server/GwtComponentServiceImpl.java
```diff
@@ -244,12 +250,20 @@ public GwtConfigComponent findWireComponentConfigurationFromPid(GwtXSRFToken xsr
...
+                        // track and wait for the wire Component
+                        final BundleContext bundleContext = FrameworkUtil.getBundle(this.getClass()).getBundleContext();
+                        String filterString = "(" + ConfigurationService.KURA_SERVICE_PID + "=" + temporaryName + ")";
+                        Filter filter = bundleContext.createFilter(filterString);
+                        final ServiceTracker wireComponentTracker = new ServiceTracker(bundleContext, filter, null);
+                        wireComponentTracker.open();
+                        wireComponentTracker.waitForService(SERVICE_WAIT_TIMEOUT);
+                        wireComponentTracker.close();
+
                         conf = cs.getComponentConfiguration(temporaryName);
```
amitjoy: You can also do one more thing for enhancement.
```
T serviceInstance = (T) serviceTracker.waitForService(WAIT_FOR_SERVICE_SEC * 1000);
if (serviceInstance != null) { .. then invoke conf = cs.getComponentConfiguration(temporaryName); ...
```

I believe it's more compact this way.

kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/server/GwtComponentServiceImpl.java
markoer: This here is usually the place for constants. Besides, you don't want to have to initialize the variable every time you call the method.

amitjoy: I tend to agree if and only if it is used in multiple places. It is considered an anti-pattern to use a constant which is used only in one place.

amitjoy: I believe the appropriate solution is to have the constant variable in the inline method itself. Other methods won't be able to see it at all as it is only used in one place. If we consider encapsulation and clean code approach, if we define the constant in class level, you get a long list of class variables as well. In addition, we have to consider the unnecessary memory footprints as well. If the constant is inline, the variables are destroyed but in case of static constant, it will remain in memory all the time.

markoer: Then also consider this: either have one constant for all instances of the class that gets initialized once or have so many instances as there are instances of running method that get initialized every time the method is ran. And considering the time we wasted on this pointless debate, we already made up for all the initialization of constants in this and all other classes Kura will ever use.

amitjoy: 🤣 This is a suggestion from my end. You are right, every time the method gets executed the string literal instance will be created and destroyed after the completion but in case of class variable, it will remain in the memory. But it is also a fact that we are also using powerful JVM that hardly cares about such memory optimization. I believe it is just a clean coding approach. Anyways, I am approving this and leaving it to @MMaiero what he believes as a programmer.

amitjoy (approved):

Issue #889, opened 2016-12-05
[Kura 2.1.0 QA] Denali Configuration ExamplePublisher error descriptions


markoer:
Error messages are misleading for long, integer, short, double, float, byte: message says 'smaller', but should be 'smaller or equal' on the high limit, and 'greater' where it should be 'greater or equal' on low limit. It could otherwise specify a different (higher/lower) number (e.g. 4294967297 for long).

Test cases DEN-CONF-005 - DEN-CONF-029 from #788.

Other notes on these tests:
DEN-CONF-016 reports `For input string: "32768"` instead of expected `Value out of range. Value: 32768…`
DEN-CONF-027 mentions `Byte min -127 max 127.` which only confuses the users as those values are not allowed.
DEN-CONF-029 reports `For input string: "128"` instead of expected `Value out of range. Value: 128…`.

DEN-CONF-035 expects dirty flag on the fields. It is not visible in Firefox, Chrome or IE.

markoer:
Retested on Kura 2.1.0 build 19.

'Normal' range error messages are now OK.

DEN-CONF-027's message is a part of the component configuration and serves as a sample of the possible metrics description, so it's also OK.

Notes regarding DEN-CONF-016 and DEN-CONF-029 still remain open (see #915).

MMaiero:
I think this one can be closed, right @markoer?

markoer:
Verified on PR #1189's branch. Closing the issue.

PR #1189, opened 2017-03-14
Modified ui message for values out of range.

### LINKED ISSUE #915: [Kura 2.1.0 QA] DEN-CONF-005: Internal message when a "too big" number is entered
Entering 4294967297 shows a correct error message.

Entering 429496729733333333333 as a number shows the message: `For input string: "-429496729733333333333"`

MMaiero:

Closes #915 and #889

### DIFF
--- a/kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/client/ui/AbstractServicesUi.java
+++ b/kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/client/ui/AbstractServicesUi.java
@@ -58,6 +58,7 @@ public abstract class AbstractServicesUi extends Composite {
 
     private static final String CONFIG_MAX_VALUE = "configMaxValue";
     private static final String CONFIG_MIN_VALUE = "configMinValue";
+    private static final String INVALID_VALUE = "invalidValue";
 
     protected static final Logger logger = Logger.getLogger(ServicesUi.class.getSimpleName());
     protected static final Logger errorLogger = Logger.getLogger("ErrorLogger");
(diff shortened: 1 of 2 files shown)


markoer:
The new message would be 'Value X is out of range.'. But there is no check performed that the value is indeed a valid numeric value. This way 'foo' would produce the same result as an out-of-range numeric e.g. 9999999999 for integer.
With this approach (one message for all causes of NFE) a better message would be a more generic 'Value X is invalid.'.

kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/client/ui/AbstractServicesUi.java
```diff
@@ -607,7 +608,8 @@ public void onChange(ChangeEvent event) {
                     }
                 } catch (NumberFormatException e) {
                     this.valid.put(param.getId(), false);
-                    result.add(new BasicEditorError(box, box.getText(), e.getLocalizedMessage()));
+                    result.add(new BasicEditorError(box, box.getText(),
```
markoer, line 611: It might be better to check if the value is a valid numeric and then adjust the error message accordingly.

MMaiero: the try catches only issues for numeric fields https://github.com/eclipse/kura/pull/1189/files#diff-e55d221b480c9c188b8d72740e7d2e46R533
I don't see a problem if we say that "foo" is invalid because out of range. But we can simply write that Value "foo" is invalid. I don't think that is worth implementing so much logic for such an improper value.

markoer: As you wish. But in this case '... is invalid.' would be a better message.

MMaiero: I agree. I'll make the change as soon as possible. Thanks for the review!

MMaiero: Done

MMaiero:
Merging then.

PR #1191, opened 2017-03-14
Preventing possible NPE when MetaTypeInformation is null.

### LINKED ISSUE #696: FileServlet fails with NPE when there is no MetaTypeInformation
When the MetaTypeService returns a null pointer, the FileServlet fails with a NPE.

MMaiero:

Closes #696


kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/server/servlet/FileServlet.java
```diff
@@ -200,8 +200,11 @@ private void doGetIcon(HttpServletRequest req, HttpServletResponse resp) throws
                 }
                 MetaTypeInformation mti = mts.getMetaTypeInformation(b);
 
-                String[] pids = mti.getPids();
-                for (String p : pids) {
+                if (mti == null) {
```
markoer, line 203: Considering that API for the method is not explicit on the case where bundle doesn't contain metatype info, this is probably a good check.

markoer (approved):

PR #1192, opened 2017-03-14
WireEmitter/Receiver are ConsumerTypeS

cdealti:
The only way an application can consume these types is to
implement them.
An real-world graph will always include domain specific
WireComponentS (WireEmitterS and WireReceiverS) that
must developed by implementing and registering these
interfaces.


cdealti:
@amitjoy do you agree?

amitjoy:
I do 👍

amitjoy (approved):

PR #1218, opened 2017-03-16
ClockService ignores updates with long 'getDelay'

pierrepitiot:
In JavaNtpClockSyncProvider, add a test on TimeInfo.getDelay() to avoid an incorrect update of the clock.

Put ntpClient.close() in a finally block.


MMaiero:
@pierrepitiot I think you should apply the formatter and cleanup.

MMaiero:
@pstanton Could you please have a look and confirm that the issue is solved with this PR?

pstanton:
@MMaiero have checked the diff and it looks good to me, thanks.

Issue #1206, opened 2017-03-16
[Driver] Driver API optimizations


cdealti:
See https://github.com/eclipse/kura/wiki/Improvements-to-the-Driver-API

cdealti:
@nicolatimeus is assignded this task

PR #1225, opened 2017-03-17
First implementation of prepared reads for Asset and Driver

nicolatimeus:
This PR proposes a first implementation of prepared reads for Asset and Driver.

See #1206 

Modified Asset and Driver APIs to add support for prepared reads.
Adapted the OPCUA driver to use the new API.
WireAssets now always use prepared reads.

This is a breaking change, drivers that use the previous API will no longer work after this is merged.

### DIFF
--- a/kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/driver/Driver.java
+++ b/kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/driver/Driver.java
@@ -211,4 +211,26 @@ public interface Driver {
      */
     public List<DriverRecord> write(List<DriverRecord> records) throws ConnectionException;
 
+    /**
+     * This method allows the driver to perform protocol specific optimizations in order to accelerate the execution of
+     * batches of read requests having the same channel configuration.
+     * The result of this optimization will be returned by the driver as a {@link PreparedRead} instance that can be
+     * used to perform the requests.
+     * In order to improve efficiency a driver should validate the channel configuration of the provided channels during
+     * this method call.
+     * It is also permitted to the implementation of the {@link PreparedRead#execute()} and
+     * {@link PreparedRead#getDriverRecords()} methods to return the same {@link DriverRecord} instances provided as an
+     * argument to this method.
+     * If the validation of the channel configuration fails for some channels, the driver must not throw an exception
+     * but it is required to return driver records with proper error flags set as a result of the
+     * {@link PreparedRead#execute()} call.
+     * 
+     * @see PreparedRead
+     * @param records
+     *            The list of driver records that represent the request to be optimized.
+     * @return The {@link PreparedRead} instance
+     * @throws NullPointerException
+     *             if the provided list is null
+     */
+    public PreparedRead prepareRead(List<DriverRecord> records);
 }
--- a/kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/driver/PreparedRead.java
+++ b/kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/driver/PreparedRead.java
@@ -1,0 +1,53 @@
+/**
+ * Copyright (c) 2017 Eurotech and/or its affiliates and others
+ *
+ *  All rights reserved. This program and the accompanying materials
+ *  are made available under the terms of the Eclipse Public License v1.0
+ *  which accompanies this distribution, and is available at
+ *  http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *   Eurotech
+ */
+package org.eclipse.kura.driver;
+
+import java.util.List;
+
+import org.eclipse.kura.KuraException;
+import org.eclipse.kura.driver.Driver.ConnectionException;
+
+/**
+ * This interface represents an optimized request that can be performed by a driver.
+ * Implementations of this interface are returned by a driver as a result of
+ * a call to the {@link Driver#prepareRead(java.util.List)} method.
+ * 
+ * @see Driver#prepareRead(java.util.List)
+ */
+public interface PreparedRead extends AutoCloseable {
+
+    /**
+     * Performs the optimized read request described by this {@link PreparedRead} instance.
+     * In order to improve efficiency this method can return the same {@link DriverRecord} instances that were supplied
+     * as arguments to the {@link Driver#prepareRead(List)} call that created this {@link PreparedRead}.
+     * The returned records should not be modified while a valid (not closed) {@link PreparedRead} holds a
+     * reference to them, otherwise unpredictable behavior can occur.
+     * 
+     * @return the result of the request as a list of {@link DriverRecord} instances.
+     * @throws KuraException
+     *             if the provided {@link PreparedRead} is not valid (for example if it has been closed)
+     * @throws ConnectionException
+     *             if the connection to the field device is interrupted
+     */
+    public List<DriverRecord> execute() throws ConnectionException, KuraException;
+
+    /**
+     * Returns the list of driver records associated with this prepared read.
+     * In order to improve efficiency this method can return the same {@link DriverRecord} instances that were supplied
+     * as arguments to the {@link Driver#prepareRead(List)} call that created this {@link PreparedRead}.
+     * The returned records should not be modified while a valid (not closed) {@link PreparedRead} holds a
+     * reference to them, otherwise unpredictable behavior can occur.
+     * 
+     * @return The list of driver records associated with this prepared read.
+     */
+    public List<DriverRecord> getDriverRecords();
+}
--- a/kura/org.eclipse.kura.asset.provider/src/main/java/org/eclipse/kura/asset/provider/BaseAsset.java
+++ b/kura/org.eclipse.kura.asset.provider/src/main/java/org/eclipse/kura/asset/provider/BaseAsset.java
@@ -32,8 +32,10 @@ import static org.eclipse.kura.asset.ChannelType.WRITE;
 import static org.eclipse.kura.driver.DriverConstants.CHANNEL_ID;
 import static org.eclipse.kura.driver.DriverConstants.CHANNEL_VALUE_TYPE;
 
+import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
+import java.util.Map.Entry;
 import java.util.Set;
 import java.util.concurrent.locks.Lock;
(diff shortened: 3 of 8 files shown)


ctron:
Don't we agree on not making any more breaking changes?

nicolatimeus:
Well it breaks compatibility with drivers designed to work with Kura 3.0.0-M1, but that is a milestone release. Kura Wires is not yet part of an official release so I think we are still allowed to introduce braking changes in it.

ctron:
Ok .. so it is not breaking backwards compatibility with any previous release.

However all drivers published already have to be updated?!

MMaiero:
Exactly, it is not a breaking change with the previous released versions of Kura.
Currently, in the marketplace, we released only two drivers. We will update them as soon as possible with the merging of the new features.

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/driver/Driver.java
```diff
@@ -211,4 +211,24 @@ public void registerDriverListener(Map<String, Object> channelConfig, DriverList
...
+     * If the validation of the channel configuration fails for some channels, the driver must not throw an exception
+     * but it is required to return driver records with proper error flags set as a result of the
+     * {@link PreparedRead#execute()} call.
+     * 
+     * @see PreparedRead
+     * @param records
+     *            The list of driver records that represent the request to be optimized.
+     * @return The {@link PreparedRead} instance
+     */
+    public PreparedRead prepareRead(List<DriverRecord> records);
```
amitjoy: public API - records must not be null. `@throws NPE if records are null`

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/driver/PreparedRead.java
```diff
@@ -0,0 +1,53 @@
...
+     * @return the result of the request as a list of {@link DriverRecord} instances.
+     * @throws KuraException
+     *             if the provided {@link PreparedRead} is not valid (for example if it has been closed)
+     * @throws ConnectionException
+     *             if the connection to the field device is interrupted
+     */
+    public List<DriverRecord> execute() throws ConnectionException, KuraException;
+
+    /**
+     * Returns the list of driver records associated whit this prepared read.
```
amitjoy: with..

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/driver/PreparedRead.java
```diff
@@ -0,0 +1,53 @@
...
+     * In order to improve efficiency this method can return the same {@link DriverRecord} instances that were supplied
+     * as arguments to the {@link Driver#prepareRead(List)} call that created this {@link PreparedRead}.
+     * The returned records should therefore not be modified while a valid (not closed) {@link PreparedRead} holds a
```
amitjoy: not to be...

amitjoy (commented):
Just had a quick look over the changes and the changes are really good and extensible enough. A big thumbs up 👍 for the API Design. I will try to have another look pretty soon.

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/driver/Driver.java
nicolatimeus: agreed👍

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/driver/PreparedRead.java
amitjoy: the spelling

kura/org.eclipse.kura.driver.opcua.provider/src/main/java/org/eclipse/kura/internal/driver/opcua/OpcUaDriver.java
```diff
@@ -468,7 +477,52 @@ private static void fail(final DriverRecord record, final String message) {
...
             }
 
-            return Optional.of(new OpcUaRequestInfo(dataType, nodeNamespaceIndex, nodeId));
+            return Optional.of(new OpcUaRequestInfo(record, dataType, nodeNamespaceIndex, nodeId));
+        }
+    }
+
+    @Override
+    public PreparedRead prepareRead(List<DriverRecord> driverRecords) {
+        requireNonNull(driverRecords);
```
amitjoy: I would recommend a suitable localized message here in the precondition.

kura/org.eclipse.kura.asset.provider/src/main/java/org/eclipse/kura/asset/provider/BaseAsset.java
```diff
@@ -251,8 +265,13 @@ public AssetConfiguration getAssetConfiguration() {
         return this.assetConfiguration;
     }
 
-    public void setDriver(Driver driver) {
+    public synchronized void setDriver(Driver driver) {
         this.driver = driver;
+        if (driver != null) {
+            List<DriverRecord> readRecords = getAllReadRecords();
+            hasReadChannels = !readRecords.isEmpty();
+            tryPrepareRead(readRecords);
```
amitjoy, line 270: Is there any specific reason why the driver's prepared read is getting triggered while injecting the Driver instance?

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/driver/Driver.java
```diff
@@ -211,4 +211,26 @@ public void registerDriverListener(Map<String, Object> channelConfig, DriverList
...
+     * The result of this optimization will be returned by the driver as a {@link PreparedRead} instance that can be
+     * used to perform the requests.
+     * In order to improve efficiency a driver should validate the channel configuration of the provided channels during
+     * this method call.
+     * It is also permitted to the implementation of the {@link PreparedRead#execute()} and
+     * {@link PreparedRead#getDriverRecords()} methods to return the same {@link DriverRecord} instances provided as an
+     * argument to this method.
+     * If the validation of the channel configuration fails for some channels, the driver must not throw an exception
+     * but it is required to return driver records with proper error flags set as a result of the
+     * {@link PreparedRead#execute()} call.
```
amitjoy, line 226: As it is a very big descriptive javadoc, I would recommend using line breaks html tags (`<br/>`) for clear visibility.

kura/org.eclipse.kura.asset.provider/src/main/java/org/eclipse/kura/asset/provider/BaseAsset.java
nicolatimeus: tryPrepareRead() does not perform the actual read but just "prepares" it, it asks the driver to validate the current channel configuration and to perform eventual optimizations basing on it.
The idea behind prepared reads is to perform the operations above just once when the channel configuration changes and not on every read request.
To be able to do that we must have two things: the channel configuration and the driver instance attached.
This is the reason why tryPrepareRead() is called both in the updated() and in the setDriver() method: the first method provides the channel config and the second the driver instance, tryPrepareRead() checks whether we have both things and if yes it asks the driver to prepare the read.

amitjoy: 👍

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/driver/Driver.java
nicolatimeus: ok

kura/org.eclipse.kura.driver.opcua.provider/src/main/java/org/eclipse/kura/internal/driver/opcua/OpcUaDriver.java
nicolatimeus: ok

kura/org.eclipse.kura.driver.opcua.provider/src/main/java/org/eclipse/kura/internal/driver/opcua/OpcUaDriver.java
```diff
@@ -468,7 +477,52 @@ private static void fail(final DriverRecord record, final String message) {
...
+        for (DriverRecord record : driverRecords) {
+            OpcUaRequestInfo.extract(record).ifPresent(preparedRead.requestInfos::add);
+        }
+        return preparedRead;
+    }
+
+    private class OpcUaPreparedRead implements PreparedRead {
+
+        private List<OpcUaRequestInfo> requestInfos = new ArrayList<OpcUaRequestInfo>();
+        private List<DriverRecord> driverRecords;
```
amitjoy: I think this should be volatile. If there is only one thread acting in `execute()` method and another thread invokes `getDriverRecords()`, there can be situation when the `getDriverRecords()` calling thread doesn't get the updated `DriverRecord`s.

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/driver/Driver.java
amitjoy: Would you kindly apply this change request? It would be more readable.

nicolatimeus: Can we postpose this? looking at the javadocs there are also other places in which we could improve readability, maybe we can do this with a dedicated PR

amitjoy: Certainly we can but you can also make the javadocs in this design more readable in this PR itself. Wouldn't this be more timesaving?

kura/org.eclipse.kura.asset.provider/src/main/java/org/eclipse/kura/asset/provider/BaseAsset.java
nicolatimeus: I've made a small improvement here: tryPreparedRead() is no longer called in the updated() method but only in setDriver(), because of the fact that we are re-attaching the driver every time updated() is called and after the AssetConfiguration is refreshed.
This should guarantee that we have the channel config when setDriver() is called. I've made a couple of tests and everything seems to be working.

PR #1229, opened 2017-03-17
Update to Equinox Neon.3

ctron:
This fixes the build issue due to the archived Neon.1 release

### DIFF
--- a/target-platform/p2-repo-equinox_3.11.1/pom.xml
+++ b/target-platform/p2-repo-equinox_3.11.1/pom.xml
@@ -14,7 +14,9 @@
 	<packaging>pom</packaging>
     
     <properties>
-        <equinox.download.url>http://ftp-stud.fht-esslingen.de/pub/Mirrors/eclipse/equinox/drops/R-Neon.1-201609071200/equinox-SDK-Neon.1.zip</equinox.download.url>
+        <equinox.release>R-Neon.3-201703010400</equinox.release>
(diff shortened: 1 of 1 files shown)


MMaiero:
I personally would not like to change the equinox we tested so far.
Is there no possibility to keep the Neon.1?

cdealti:
@ctron obviously we cannot depend on these zip archives being "archived". Remember that you originally changed the build to fetch the target platform from "official" sources.
I would add the requirement that these sources are also "persistent".
The alternative is that we will have to upload the Equinox zip to one of our S3 buckets.

ctron:
We can fall back to the archived version. But maybe there is a good reason why Neon.1 was archived while Neon.2, Neon.3 and Neon where not: [1].

If you want to switch to the archived version, it is here: [2]

[1] http://download.eclipse.org/equinox/drops/?d
[2] http://archive.eclipse.org/equinox/drops/R-Neon.1-201609071200/download.php?dropFile=equinox-SDK-Neon.1.zip

ctron:
So, if you would like to stick to the archive version of Neon.1, then you can use PR #1235

cdealti:
@ctron we agreed to upgrade to Oxygen. Can you close this?

MMaiero:
@ctron Do you believe we can close this PR?

ctron:
Sure, close it.

PR #1230, opened 2017-03-17
Dropped support for native libraries that are not supported by our current profiles.

### LINKED ISSUE #1013: Drop support for ARMv5 SF/soft float
As all models of the Raspberry Pi back to model 1 support ARMv6 HF/hard float, there seems to be no need to maintain additional native binaries for the ARMv5 SF target.

MMaiero:

Closes #1013


pierantoniomerlino (approved):
Be aware that the x86_64 profile is still present.

MMaiero:
x86_64 is used by the APU profiles

ctron (approved):
LGTM: However, as explained yesterday, I have no clue how to specifically test this PR. I would suggest do simply get this tested in the final QA phase.

Issue #1208, opened 2017-03-16
[Asset] Driver read method should return void


cdealti:
As stated in the [Javadoc](https://github.com/eclipse/kura/blob/develop/kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/driver/Driver.java#L147) Driver.read modifies the ``DriverRecord``S parameter and returns the modified parameter.

An usage where the return value is discarded is in [BaseAsset](https://github.com/eclipse/kura/blob/develop/kura/org.eclipse.kura.asset.provider/src/main/java/org/eclipse/kura/asset/provider/BaseAsset.java#L485)

Returning void looks more natural to me.

cdealti:
@amitjoy we would like to change the API now (before Kura 3).

amitjoy:
According to my opinion, returning a value from a method is generally a cleaner approach. Passing a value and mutate it, is not very common practice in Java but there are few examples in JDK API where such approach has been taken, for example `Collection.sort(..)` that performs an in place sort of the provided list.

It is also a good design pratice to favor immutable objects in multi-threading application development.

So, we have to consider what our intention is. Previously while I started desiging the `Asset-Driver` model, the `read` and `write` methods returned `void` but later on during our design discussions, this has been changed but I preferred it to be `void` as it resembles the `Collections.sort(..)` scenario.

So, I upvote it 👍.

MMaiero:
@nicolatimeus Will take care of this

PR #1247, opened 2017-03-22
Changed return type of Driver.read() and Driver.write() to void

nicolatimeus:
Adapted BaseAsset and OPCUA driver to use the new API

See #1208 


amitjoy:
Testcases failed in `WireAssetTest.java`

PR #1248, opened 2017-03-22
Added CloudService support for JSON publishing.

### LINKED ISSUE #1239: Change wire publisher to remove the JSON/Kura payload serialization
This will be moved to the specific CloudService

### LINKED ISSUE #1240: Add JSON/Protobuf encoding option in CloudService
The applications using the CloudService/CloudClient should not be aware of the serialization used by the associated Cloud Service.

MMaiero:
Removed the payload encoding selection from the wire publisher.

Closes #1240 
Closes #1239


kura/org.eclipse.kura.core.cloud/src/main/java/org/eclipse/kura/core/cloud/CloudPayloadJsonDecoder.java
```diff
@@ -0,0 +1,170 @@
...
+import static org.eclipse.kura.core.cloud.CloudPayloadJsonFields.TIMESTAMP;
+import static org.eclipse.kura.core.cloud.CloudPayloadJsonFields.CloudPayloadJsonPositionFields.ALTITUDE;
+import static org.eclipse.kura.core.cloud.CloudPayloadJsonFields.CloudPayloadJsonPositionFields.HEADING;
+import static org.eclipse.kura.core.cloud.CloudPayloadJsonFields.CloudPayloadJsonPositionFields.LATITUDE;
+import static org.eclipse.kura.core.cloud.CloudPayloadJsonFields.CloudPayloadJsonPositionFields.LONGITUDE;
+import static org.eclipse.kura.core.cloud.CloudPayloadJsonFields.CloudPayloadJsonPositionFields.PRECISION;
+import static org.eclipse.kura.core.cloud.CloudPayloadJsonFields.CloudPayloadJsonPositionFields.SATELLITES;
+import static org.eclipse.kura.core.cloud.CloudPayloadJsonFields.CloudPayloadJsonPositionFields.SPEED;
+import static org.eclipse.kura.core.cloud.CloudPayloadJsonFields.CloudPayloadJsonPositionFields.STATUS;
+import static org.eclipse.kura.core.cloud.CloudPayloadJsonTypes.BOOLEAN;
```
amitjoy, line 25: Wouldn't it be better to reuse `org.eclipse.kura.type.DataType` instead of `org.eclipse.kura.core.cloud.CloudPayloadJsonTypes`?

MMaiero: If you have a look, currently there is not perfect match between the two. But I'll try to see how to improve it.

amitjoy: Instead of `CHARACTER`, everything remains same. The `Type` package has been made a top-level package as these types can be used in other places as well. You can add `CHARACTER` in `DataType` as well and use it here.

dwoodard1: +1 Though I believe it is missing FLOAT as well. Besides this, the PR looks good.

MMaiero: Float cannot be added due to some database limitations (it will be stored as double).
The real problem now is that the types in the wires do not match the ones supported by the Kura Protobuf and this can cause exceptions while trying to publish.
The thing now to understand is if we want to adhere to the Kura Protobuf or not and how to deal in wires with unsupported types.

dwoodard1: I understand. I suppose if we wanted to extend Kura Protobuf, now would be the time. But, do we need to? Can't we limit the data model of Wires to adhere to Kura Protobuf? Are there data types we really need that aren't included in Kura Protobuf?

MMaiero: That is the question. We need to decide before the release.
In any case the data types supported by wires cannot match perfectly the Kura protobuf types: float is not supported by the DB.

dwoodard1: I will merge this request and open a new issue to discuss data types within Kura.

dwoodard1 (approved):

Issue #1211, opened 2017-03-16
[Wires] Implement regex filter WireComponent


cdealti:
This WireComponent filter is used to filter out properties not matching a configurable regular expression.

amitjoy:
Instead of creating a Wire Component to filter out keys from the associated map of Wire Records, we can also use the filter defined in Wire Configuration. This way we can apply filters on wire connections. Users can click on the wire connection in Wire Graph either to delete it or to associate a regex filter to it. The implementation for such a filter at the wire level needs very minimal changes in internal classes (no API breaking changes). I have also created a branch with the needed changes for such filter at the wire level. We can also discuss about it further.

cdealti:
@amitjoy we need this for Kura 3.0. Can you share your preliminary work and list pros-and-cons of both approaches?

PR #1251, opened 2017-03-22
Filter functionality on WireRecord associated properties

amitjoy:
Proposal for #1211

This is a proposal PR and it introduces the filter functionality to the `WireRecord` associated properties. It filters keys based on regular expression.

The idea is that users can associate regular expression filter to the created wires either from Composer UI canvas or from Gogo Shell. The user can double click on the created wire in the `WireGraph` which will pop up a dialog box asking for regular expression filter. The UI changes have not been incorporated. On the other hand, I have added the necessary changes to the Gogo Shell commands to associate filter to `WireConfigurations`.

According to the Kura Wires Model,

Kura Wires delegates the requests to `WireAdmin` which internally manages a registry of created `Wires` and `WireAdmin` service pushes the data from the created `Wires` to the connected `Consumers` (`Consumer` and `Producer` are fixed terminologies in `WireAdmin`).

So, if we go for a Filter Wire Component, it will be as follows:

`Timer ----> Wire Admin ----> Filter Wire Component ----> Wire Admin ----> Logger`

On the other hand, according to this proposal,

`Timer ----> Wire Admin ----> Logger`

`--->` This signifies the data flow

1. The filtering mechanism occurs at the wire level and hence it does not incur multiple calls to `WireAdmin` to filter the `WireRecord` associated properties. 
2. In addition, in future we can also provide other `Filter Wire Components` to filter using `LDAP filter`, `DSL based filter` etc.

**Easy Testing Procedure**:

1. Add several properties of your choice in the map in `EmitJob.java` (in `Timer`)
2. Create a `Timer` instance (`T`) from the canvas
3. Create a `Logger` instance (`L`) from the canvas
4. execute `createWire T L \\wA\\w*` in console (syntax: `createWire emitterPid receiverPid filter`)

### DIFF
--- a/kura/org.eclipse.kura.wire.helper.provider/src/main/java/org/eclipse/kura/internal/wire/helper/WireSupportImpl.java
+++ b/kura/org.eclipse.kura.wire.helper.provider/src/main/java/org/eclipse/kura/internal/wire/helper/WireSupportImpl.java
@@ -19,11 +19,13 @@ import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 
 import org.eclipse.kura.localization.LocalizationAdapter;
 import org.eclipse.kura.localization.resources.WireMessages;
 import org.eclipse.kura.util.collection.CollectionUtil;
 import org.eclipse.kura.wire.WireComponent;
+import org.eclipse.kura.wire.WireConfiguration;
 import org.eclipse.kura.wire.WireEmitter;
 import org.eclipse.kura.wire.WireEnvelope;
 import org.eclipse.kura.wire.WireHelperService;
@@ -49,9 +51,9 @@ final class WireSupportImpl implements WireSupport {
 
     private final WireComponent wireSupporter;
 
-    private String emitterPid;
+    private final String emitterPid;
 
-    private String pid;
+    private final String pid;
 
     /**
      * Instantiates a new wire support implementation.
@@ -90,14 +92,47 @@ final class WireSupportImpl implements WireSupport {
     public synchronized void emit(final List<WireRecord> wireRecords) {
         requireNonNull(wireRecords, message.wireRecordsNonNull());
         if (this.wireSupporter instanceof WireEmitter) {
-            final WireEnvelope wei = new WireEnvelope(emitterPid, wireRecords);
-            for (final Wire wire : this.outgoingWires) {
-                wire.update(wei);
+            final String emitterPid = this.wireHelperService.getServicePid(this.wireSupporter);
+            final String pid = this.wireHelperService.getPid(this.wireSupporter);
+            final Set<WireConfiguration> wireConfigurations = this.wireHelperService
+                    .getWireConfigurationsByEmitterPid(emitterPid);
+            final WireEnvelope wireEnvelope = new WireEnvelope(emitterPid, wireRecords);
+
+            for (final WireConfiguration wc : wireConfigurations) {
+                final String filter = wc.getFilter();
+                final Wire wire = wc.getWire();
+                if (filter == null) {
+                    wire.update(wireEnvelope);
+                    continue;
+                }
+                final WireEnvelope filteredWireEnvelope = new WireEnvelope(emitterPid, filter(wireRecords, filter));
+                wire.update(filteredWireEnvelope);
             }
+
+            // fire OSGi event for every emit operation
             final Map<String, Object> properties = CollectionUtil.newHashMap();
             properties.put("emitter", pid);
             this.eventAdmin.postEvent(new Event(WireSupport.EMIT_EVENT_TOPIC, properties));
         }
+    }
+
+    /**
+     * Filters out the key from the map of provided {@link WireRecord}s that matches the provided filter
+     *
+     * @param wireRecords
+     *            the list of {@link WireRecord}s
+     * @param filter
+     *            the filter to match
+     * @return the filtered list of {@link WireRecord}s
+     * @throws NullPointerException
+     *             if any of the arguments is null
+     */
+    private List<WireRecord> filter(final List<WireRecord> wireRecords, final String filter) {
+        requireNonNull(wireRecords, message.wireRecordsNonNull());
+        requireNonNull(filter, message.filterNonNull());
+
+        // add filter logic
+        return Collections.emptyList();
     }
 
     /**
--- a/kura/org.eclipse.kura.wire.helper.provider/src/main/java/org/eclipse/kura/internal/wire/helper/WireSupportImpl.java
+++ b/kura/org.eclipse.kura.wire.helper.provider/src/main/java/org/eclipse/kura/internal/wire/helper/WireSupportImpl.java
@@ -14,15 +14,21 @@
 package org.eclipse.kura.internal.wire.helper;
 
 import static java.util.Objects.requireNonNull;
+import static java.util.stream.Collectors.toMap;
 
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
+import java.util.Map.Entry;
 import java.util.Set;
+import java.util.function.Supplier;
+import java.util.regex.Pattern;
+import java.util.stream.Stream;
 
 import org.eclipse.kura.localization.LocalizationAdapter;
 import org.eclipse.kura.localization.resources.WireMessages;
+import org.eclipse.kura.type.TypedValue;
 import org.eclipse.kura.util.collection.CollectionUtil;
 import org.eclipse.kura.wire.WireComponent;
 import org.eclipse.kura.wire.WireConfiguration;
@@ -92,8 +98,7 @@ final class WireSupportImpl implements WireSupport {
     public synchronized void emit(final List<WireRecord> wireRecords) {
         requireNonNull(wireRecords, message.wireRecordsNonNull());
         if (this.wireSupporter instanceof WireEmitter) {
-            final String emitterPid = this.wireHelperService.getServicePid(this.wireSupporter);
-            final String pid = this.wireHelperService.getPid(this.wireSupporter);
+            final String emitterPid = this.wireHelperService.getPid(this.wireSupporter);
             final Set<WireConfiguration> wireConfigurations = this.wireHelperService
                     .getWireConfigurationsByEmitterPid(emitterPid);
             final WireEnvelope wireEnvelope = new WireEnvelope(emitterPid, wireRecords);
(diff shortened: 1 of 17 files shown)


MMaiero:
Personally, I see a big drawback of this approach: when filtering is applied at the wire level, it is not immediate, by looking at the graph, to see if a filtering is applied.
A filtering component, instead, is immediately findable in the graph and you can easily guess that, from that point on, a filtering is applied.
Furthermore, this PR will require a further step in order to make configurable the wires from the web ui.
Having a filter component, instead, will require less work because we are reusing all the existing code.

amitjoy:
> Personally, I see a big drawback of this approach: when filtering is applied at the wire level, it is not immediate, by looking at the graph, to see if a filtering is applied.

That's just a sugar coating that needs to be applied in the javascript code. We just need to color the wire connections if filter is applied or else black. This is very small change.

> A filtering component, instead, is immediately findable in the graph and you can easily guess that, from that point on, a filtering is applied.

I agree but there involves performance hit as I have already explained in my previous comment.

> this PR will require a further step in order to make configurable the wires from the web ui.
Having a filter component, instead, will require less work because we are reusing all the existing code.

You are right. But there will be minor changes in the Web Bundle to incorporate this as user needs to double click on the wire to associate a filter. But it won't be a big change at all. As we are doing a revision on the complete wires for 3.0.0, we can apply these changes as well.

nicolatimeus:
I agree on the fact that a graph with no wire dedicated filtering components and with coloured wires would be more clean, but in my opinion implementing that would require a lot of effort in terms of changes to the web ui and more time spent on testing and maintaining the new code.

In my opinion this feature can be introduced in a more straightforward way by implementing a dedicated wire component, so without any changes to the framework.

It is true that having a filter component leads to more calls to the WireAdmin, but I think the CPU time spent doing this will be negligible if compared to the time spent matching the regular expression against all the properties, especially in case of big wire envelopes.
Moreover, this PR also introduces overhead because there is more work to be done inside the emit() method every time an envelope is sent, even if filtering is not needed.

amitjoy:
I would agree that it is more straight forward if you implement a Filter Wire Component but as I pointed out it incurs more overhead if you closely verify the WireAdmin design. The push and pull based design in WireAdmin is the Kura Wires' foundation and on the other hand, this incurs a bit of overhead as well. 

According to this functionality, I recommend to have a feature where user can associate filter to the connections by double clicking to it. If a filter has been associated to a wire, its color will be changed. This way we can clearly differentiate between filtered wires and non-filtered wires.

If we go for a filter wire component, we still have to write functionality to filter keys from the associated map using regex. This has been achieved by `filter()` method in this PR. This functionality to filter using regex can be improved as well. But as far as I have implemented, I have considered few cases for optimizations:

1. If no filter has been associated, just send the previously created WireEnvelope. No filter method will be invoked.
2. If filter has been associated, for every Wire Record in the Wire Envelope, check for a match in the map
2.1 If  the provided regular expression matches all the keys of the provided map, there is no need to create a new map instance, just return the old map

This also utilizes short-circuiting methods in Java 8 to further optimize a stream of map entries.

Check `filter()` and `lookup()` method optimizations but as I mentioned that this functionality can be improved as well. I am now more concerned about the design decision.

The web bundle changes are minimal and I have mostly implemented the feature and will like to push the complete feature by tomorrow.

amitjoy:
Just to update, I have implemented the complete filtering feature with a bit more optimizations while filtering and emitting data.

Have a look at it. In case you are testing, don't forget to fetch the updates as I amended a bug fix just right now.

nicolatimeus:
I've tried this, it looks good but I think I've found an issue, try performing the following steps starting from a fresh install:

* Create a Timer, connect it to a Logger and place a regexp on the wire between the two components (I tried with TIM*), click Apply.

* Create a second logger and connect it to the Timer without setting any regexp on the wire, click apply.

When I try this the first Logger behaves as expected and prints on the log the content of the envelopes it receives (they are empty because no property matches the regexp).
The second logger does not log anything (no envelopes at all, like if it wasn't connected the Timer), it should instead log envelopes containing the TIMER property as emitted by the Timer.

nicolatimeus:
We also found an use case for which this approach is not optimal: suppose you have a single wire emitter connected to multiple wire receivers, and you want to apply the same regexp to all wires between the emitter and the receiver.

If the filtering is performed by the wire you will need to apply the same regexp to all wires in the web UI, at runtime the same regexp will be matched multiple times on the same envelope, once per wire.

Introducing a filtering component you can define the regexp only once in the web ui and it will also matched only once at runtime.

amitjoy:
I have performed the steps as you said but I added several custom properties in `EmitJob` in Timer for further filtering testing.

I received the following:

```
{}
00:13:27,506 [DefaultQuartzScheduler_Worker-3] INFO  Logger:126  - WireMessages.wireEnvelopeReceived("T")
{LAMER=LongValue [value=1490739207503], TIMER=LongValue [value=1490739207503], LIMERs=LongValue [value=1490739207503]}
00:13:27,507 [DefaultQuartzScheduler_Worker-3] INFO  Logger:126  - WireMessages.wireEnvelopeReceived("T")
{}
00:13:37,505 [DefaultQuartzScheduler_Worker-4] INFO  Logger:126  - WireMessages.wireEnvelopeReceived("T")
{LAMER=LongValue [value=1490739217505], TIMER=LongValue [value=1490739217505], LIMERs=LongValue [value=1490739217505]}
00:13:37,505 [DefaultQuartzScheduler_Worker-4] INFO  Logger:126  - WireMessages.wireEnvelopeReceived("T")
{}
00:13:47,506 [DefaultQuartzScheduler_Worker-5] INFO  Logger:126  - WireMessages.wireEnvelopeReceived("T")
{LAMER=LongValue [value=1490739227506], TIMER=LongValue [value=1490739227506], LIMERs=LongValue [value=1490739227506]}
00:13:47,507 [DefaultQuartzScheduler_Worker-5] INFO  Logger:126  - WireMessages.wireEnvelopeReceived("T")
```

It shows the empty properties map as well as the whole map when received by two different Logger Wire Components.

bildschirmfoto 2017-03-29 um 12 16 36 am

> Introducing a filtering component you can define the regexp only once in the web ui and it will also matched only once at runtime.

You are right. As I have mentioned before, applying filters to wire level has lots of benefits. Along with it, we can also have a filtering wire component to do such thing if we want to aggregate filters. That can be considered as add-on.

kura/org.eclipse.kura.wire.helper.provider/src/main/java/org/eclipse/kura/internal/wire/helper/WireSupportImpl.java
```diff
@@ -41,80 +50,176 @@
...
     /** {@inheritDoc} */
     @Override
-    public synchronized void consumersConnected(final Wire[] wires) {
-        this.outgoingWires = Arrays.asList(wires);
+    public synchronized void emit(final List<WireRecord> wireRecords) {
+        requireNonNull(wireRecords, message.wireRecordsNonNull());
+        if (!(this.wireComponent instanceof WireEmitter)) {
+            return;
+        }
+        retrieveWireConfigurationsIfAbsent();
```
nicolatimeus: I've tested again and I was able to reproduce the issue once more, the problem should be here:
basically this.wireConfigurations is never modified after it is filled the first time by retrieveWireConfigurationsIfAbsent().
So if you create a wire graph, save it and then you attach more components to an existing emitter, these new components will never receive envelopes because the WireSupportImpl.wireConfigurations field will not be updated, and so it will not contain the new wires.
If you restart Kura everything will work because when retrieveWireConfigurationsIfAbsent() is called the first time it will find all the wires, including the new ones.

amitjoy: `WireSupportImpl.wireConfigurations` should be updated as it is again set in `consumersConnected(Wire)`. `consumersConnected(Wire)` callback is triggered as soon as new consumers are connected to the emitter. But it should be updated.

amitjoy: Fixed

amitjoy: There was a synchronization issue in WireAdmin callbacks. Fixed and tested it. It will work now.

MMaiero:
> Along with it, we can also have a filtering wire component to do such thing if we want to aggregate filters. That can be considered as add-on.

In my opinion, this will only make the graph more complicated and less user-friendly.

Personally, I continue to prefer having a separate wire component for filtering: a wire should be as simple as possible, while all the logic is contained in the wire components. In this way, the graph is logically simple and straight forward. Furthermore, having a wire component that performs the filtering will enable future expandability. The same level cannot be obtained with the filtering applied at the wire level.

amitjoy:
Filtering at the wire level is an integral part and that is why Wire Configuration contained a filter property since the beginning. Even Wire Admin believed that filtering should be done at the wire level and that's the reason Wire Admin also provides LDAP filtering. After my system design analysis, I could not find any scenario to use Wire Admin filtering but I kept it open for future revision to include filtering at wire level. And that's why this PR introduces complete feature that uses Wire Configuration filter property in efficient way. For simple filtering, assigning regex filter to the wire is also pretty simple as it doesn't involve multiple wire creations as you need it if you use another component for filtering. My opinion is to keep wire with filter mechanism for performance reasons. Filtering at the wire level incurs performance benefits so we still can use filter at the wire level. In addition, using a filter component introduces new wires to be maintained by Wire Admin as well as Wire Service. So, in worst case scenario, if we want to use different filters in a large graph of n no of elements, there will be x no of wires let's say but in case of filtering at the wire level, there will will be x/2 no of wires. Hence, there will be less Wire communications. But as I have mentioned before, a filter Wire component can be an add-on. It is upto the user to use it or not. I believe the integrated filter at the wire level will be performance-wise optimal. It's still upto the Wire Graph Developer to use it or not.

cdealti:
@amitjoy thanks for all your effort on this.

I'm not sure what approach is best but I believe and hope there will be many other filter types in the Kura Wires soon and these filters will be regular wire components.
Eventually, having filters both in wires and in components is confusing.

On top of that, if the filter is a regular component, the regex property can be changed in a simpler way using the Configuration Service (also remotely) than changing it if the filter is implemented in the Wire.

A wire component filter can also have a fan-out > 1 and the regex matching needs to be done only once per envelope.

If the component approach will prove to perform well enough this PR will be closed.

amitjoy:
> On top of that, if the filter is a regular component, the regex property can be changed in a simpler way using the Configuration Service (also remotely) than changing it if the filter is implemented in the Wire.

As every `Wire Configuration` comprises a `filter` property, I believe it would still be easily manageable from remote end. I agree that having a dedicated filter Wire Component will be easier to understand the graph topology.

> A wire component filter can also have a fan-out > 1 and the regex matching needs to be done only once per envelope.

This is the only scenario where a dedicated Wire Component proves to be optimal than filters in wires. But in scenarios, where we have different filters in different wires: there will be 2x more wires than the no of wires we need if filter is implemented directly in wires. That would eventually cost performance.

I believe both the approaches have pros and cons. I believe filters in wires are performance-wise optimal than dedicated wire component. On the other hand, a dedicated wire component will be flexible enough to understand a graph topology.

I will make a PR implementing the filter wire component and then we can asses both the approaches and compare their benefits.

amitjoy:
Closing this due to PR #1291

Issue #1216, opened 2017-03-16
[Wires] Sub-second Timer


cdealti:
We need a timer with sub-second resolution.

MMaiero:
@nicolatimeus Will take care of this

PR #1254, opened 2017-03-23
Implemented sub-second Timer in Wires

nicolatimeus:
Also updated TimerTest

See #1216  

### DIFF
--- a/kura/org.eclipse.kura.wire.component.provider/OSGI-INF/metatype/org.eclipse.kura.wire.Timer.xml
+++ b/kura/org.eclipse.kura.wire.component.provider/OSGI-INF/metatype/org.eclipse.kura.wire.Timer.xml
@@ -29,9 +29,23 @@
             cardinality="0"
             required="true"
             default="10"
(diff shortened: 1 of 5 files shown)


kura/org.eclipse.kura.wire.component.provider/OSGI-INF/metatype/org.eclipse.kura.wire.Timer.xml
```diff
@@ -28,8 +28,8 @@
             type="Integer"
             cardinality="0"
             required="true"
-            default="10"
-            description="Interval of the timer ticks, in seconds (Needs to be set if SIMPLE type is selected)"
+            default="10000"
```
amitjoy: I think it would be better to provide 1 Attribute Definition for only taking integer and 1 Attribute Definition that lets user choose the Unit of Time for the provided number. We can provide all the units as present in `java.util.concurrent.TimeUnit`. This way it will be more flexible and customizable.

Issue #1210, opened 2017-03-16
[Wires] Implement FIFO WireComponent


cdealti:
By default the wire graph is single threaded and the progression of a WireEnvelope it's a call stack on the thread that originally emiited the WireEnvelope.
As a result if a WireComponent hogs the CPU due some complex elaboration, all the application will be affected.

A FIFO  WireComponent allows to defer the execution of CPU-intensive WireComponents.

The FIFO  WireComponent has the following configuration:

* FIFO capacity
* Number of consumer threads

PR #1259, opened 2017-03-23
First implementation of FIFO component

nicolatimeus:
The FIFO component implemented in this PR dispatches envelopes using a single thread and in depth-first mode only.
Multithreading and breadth-first mode will be added in future.

See #1210 

### DIFF
--- a/kura/org.eclipse.kura.wire.component.provider/src/main/java/org/eclipse/kura/internal/wire/fifo/Fifo.java
+++ b/kura/org.eclipse.kura.wire.component.provider/src/main/java/org/eclipse/kura/internal/wire/fifo/Fifo.java
@@ -1,0 +1,237 @@
+/*******************************************************************************
+ * Copyright (c) 2017 Eurotech and/or its affiliates and others
+ *
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *  Eurotech
+ *
+ *******************************************************************************/
+package org.eclipse.kura.internal.wire.fifo;
+
+import static java.util.Objects.isNull;
+import static java.util.Objects.requireNonNull;
+
+import java.util.ArrayList;
+import java.util.Map;
+import java.util.concurrent.locks.Condition;
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReentrantLock;
+import java.util.function.Consumer;
+
+import org.eclipse.kura.configuration.ConfigurableComponent;
+import org.eclipse.kura.configuration.ConfigurationService;
+import org.eclipse.kura.localization.LocalizationAdapter;
+import org.eclipse.kura.localization.resources.WireMessages;
+import org.eclipse.kura.wire.WireEmitter;
+import org.eclipse.kura.wire.WireEnvelope;
+import org.eclipse.kura.wire.WireHelperService;
+import org.eclipse.kura.wire.WireReceiver;
+import org.eclipse.kura.wire.WireSupport;
+import org.osgi.service.wireadmin.Wire;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class Fifo implements WireEmitter, WireReceiver, ConfigurableComponent {
+
+    private static final String DISCARD_ENVELOPES_PROP_NAME = "discard.envelopes";
+    private static final String QUEUE_CAPACITY_PROP_NAME = "queue.capacity";
+
+    private static final Logger logger = LoggerFactory.getLogger(Fifo.class);
+    private static final WireMessages message = LocalizationAdapter.adapt(WireMessages.class);
+
+    private volatile WireHelperService wireHelperService;
+    private WireSupport wireSupport;
+
+    private FifoEmitterThread emitterThread;
+
+    public void bindWireHelperService(final WireHelperService wireHelperService) {
+        if (isNull(this.wireHelperService)) {
+            this.wireHelperService = wireHelperService;
+        }
+    }
+
+    public void unbindWireHelperService(final WireHelperService wireHelperService) {
+        if (this.wireHelperService == wireHelperService) {
+            this.wireHelperService = null;
+        }
+    }
+
+    public void activate(final Map<String, Object> properties) {
+        logger.info(message.activatingFifo());
+        wireSupport = this.wireHelperService.newWireSupport(this);
+        updated(properties);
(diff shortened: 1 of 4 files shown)


kura/org.eclipse.kura.wire.component.provider/src/main/java/org/eclipse/kura/internal/wire/fifo/Fifo.java
```diff
@@ -0,0 +1,180 @@
...
+        stopEmitterThread();
+
+        logger.debug("Creating new emitter thread: {}, queue capacity: {}, discard envelopes: {}", threadName,
+                queueCapacity, discardEnvelopes);
+        emitterThread = new FifoEmitterThread(threadName, queueCapacity, discardEnvelopes);
+        emitterThread.start();
+    }
+
+    @Override
+    public void onWireReceive(WireEnvelope wireEnvelope) {
```
amitjoy, line 106: Shouldn't it be `synchronized`?

kura/org.eclipse.kura.wire.component.provider/src/main/java/org/eclipse/kura/internal/wire/fifo/Fifo.java
```diff
@@ -0,0 +1,180 @@
...
+    public void activate(final Map<String, Object> properties) {
+        logger.info(message.activatingFifo());
+        wireSupport = this.wireHelperService.newWireSupport(this);
+        updated(properties);
+        logger.info(message.activatingFifoDone());
+    }
+
+    public void deactivate() {
+        logger.info(message.deactivatingFifo());
+        stopEmitterThread();
```
amitjoy, line 72: Shouldn't the `Emitter Thread` be killed while deactivating this component?

kura/org.eclipse.kura.wire.component.provider/src/main/java/org/eclipse/kura/internal/wire/fifo/Fifo.java
```diff
@@ -0,0 +1,180 @@
...
+        boolean discardEnvelopes = (Boolean) properties.getOrDefault(DISCARD_ENVELOPES_PROP_NAME, false);
+
+        restartEmitterThread(threadName, queueCapacity, discardEnvelopes);
+
+        logger.info(message.updatingFifoDone());
+    }
+
+    private synchronized void stopEmitterThread() {
+        if (emitterThread != null) {
+            emitterThread.quit();
```
amitjoy: I believe it should also be killed in here. `Thread.interrupt()`

amitjoy (commented):
Few minor ones

kura/org.eclipse.kura.wire.component.provider/src/main/java/org/eclipse/kura/internal/wire/fifo/Fifo.java
nicolatimeus: The synchronisation is performed by the FifoEmitterThread.submit() method.

kura/org.eclipse.kura.wire.component.provider/src/main/java/org/eclipse/kura/internal/wire/fifo/Fifo.java
nicolatimeus: We discussed about this and for now we decided to neither interrupt nor kill the emitter thread, because otherwise there is the risk of leaving downstream components in a bad state.
The current implementation notifies the emitter thread to stop using the FifoEmitterThead.shutdown() method, if this method is called the emitter dispatches at most one more envelope and then exits. Any threads blocked trying to add a new envelope to the queue will be notified of the fact that queue is being shut down and will unblock.
It is possible that the emitter thread never exits due to problems in the downstream components, but it's hard to recover in a clean way from such situation. Interrupting or killing the thread can cause trouble because we don't know what the components are doing.
Anyway we can change this in future if needed.

nicolatimeus:
I made a few changes to get more control over the shutdown phase of the queue in order to implement a cleaner shutdown.

kura/org.eclipse.kura.wire.component.provider/src/main/java/org/eclipse/kura/internal/wire/fifo/Fifo.java
```diff
@@ -0,0 +1,237 @@
...
+                lock.lock();
+                run = false;
+                producer.signalAll();
+                consumer.signalAll();
+            } finally {
+                lock.unlock();
+            }
+        }
+
+        public void sumit(WireEnvelope envelope) {
```
amitjoy: I am guessing you wanted this method name to be `submit` 😉

kura/org.eclipse.kura.wire.component.provider/src/main/java/org/eclipse/kura/internal/wire/fifo/Fifo.java
```diff
@@ -0,0 +1,237 @@
...
+                        }
+                        if (!run) {
+                            break;
+                        }
+                        next = queue.remove(queue.size() - 1);
+                        consumer.signal();
+                    } finally {
+                        lock.unlock();
+                    }
+                    wireSupport.emit(next.getRecords());
```
amitjoy, line 229: I believe you can also put `WireSupport.emit(..)` inside previous critical section and add finally to the outer `try..catch`

amitjoy:
The current implementation for clean shutdown is pretty good 👍

kura/org.eclipse.kura.wire.component.provider/src/main/java/org/eclipse/kura/internal/wire/fifo/Fifo.java
nicolatimeus: 😄  👍

kura/org.eclipse.kura.wire.component.provider/src/main/java/org/eclipse/kura/internal/wire/fifo/Fifo.java
nicolatimeus: The lock is just for synchronising access to the queue, if we keep the lock while dispatching envelopes we will cause the EnvelopeDiscardingSubmitter to become blocking, because the onWireReceive() method of the downstream components might take time.
If this happens we lose the decoupling that the Fifo component is meant to introduce.
I've put everything inside a catch(Exception) block because I don't want the emitter thread to die.

amitjoy: Oh! now I understand 👍

PR #1260, opened 2017-03-24
Updated the example publisher. 

### LINKED ISSUE #1241: Changes to the heater publisher
- It should be defined as a factory
- It should allow the user to configure the cloud service
- It should allow the user to configure appid
- It should allow the user to configure app topic

MMaiero:
Now the cloud service is configurable by the user, as well as the appID.
The example is now a factory that needs to be instantiated in order to work.

Closes #1241 

This PR need to be merged after the merge of #1248

### DIFF
--- a/kura/examples/org.eclipse.kura.demo.heater/OSGI-INF/heater.xml
+++ b/kura/examples/org.eclipse.kura.demo.heater/OSGI-INF/heater.xml
@@ -34,5 +34,6 @@
               bind="setCloudService"       
               unbind="unsetCloudService"       
               cardinality="1..1" 
-              interface="org.eclipse.kura.cloud.CloudService"/>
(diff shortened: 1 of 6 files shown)


MMaiero:
@dwoodard1 I've rebased the code. Now the commits are correct. Please have a look at the changes. Thanks.

dwoodard1:
@MMaiero As we discussed, some of the examples should be kept simple and not use the ServiceTrackerCustomizer. Let's change the heater demo so it uses the simpler, more static "target" in its component definition. The others can have the dynamic approach. We will need to update the documentation accordingly.

MMaiero:
@dwoodard1 Changes applied.

dwoodard1 (approved):

PR #1266, opened 2017-03-27
Fixed unnecessary allocations by WireAsset

nicolatimeus:


kura/org.eclipse.kura.asset.provider/src/main/java/org/eclipse/kura/asset/provider/BaseAsset.java
```diff
@@ -401,6 +401,15 @@ protected String getFactoryPid() {
         return CONF_PID;
     }
 
+    protected String getKuraServicePid() throws KuraException {
+        String kuraServicePid = (String) this.properties.get(ConfigurationService.KURA_SERVICE_PID);
```
amitjoy: cache the Kura Service PID otherwise everytime the map lookup occurs for every emit, the retrieval of Kura Service PID from the properties will happen.

PR #1272, opened 2017-03-28
Improved performance of WireSupportImpl.emit()

nicolatimeus:


nicolatimeus:
@amitjoy is it safe to cache service.pid and kura.service.pid like this?
These properties should be always available in the activate() method of a WireComponent and they should never change during its lifetime, right?

amitjoy:
If we are considering to merge #1251, such necessary elements' caching have been considered already in `WireSupport`

nicolatimeus:
Yes, you are right, I didn't notice it.

PR #1275, opened 2017-03-29
Filter ANSI escape in gatttool

pierantoniomerlino:
This PR is equal to #593 but all conflicts are solved. It fixes the #372 about the filtering of ANSI escape sequences in gatttool.


Erwan-S:
Thanks for completed that

PR #1276, opened 2017-03-29
Asset cloudlet now starts immediately.

MMaiero:

### DIFF
--- a/kura/org.eclipse.kura.asset.cloudlet.provider/OSGI-INF/AssetCloudletComponent.xml
+++ b/kura/org.eclipse.kura.asset.cloudlet.provider/OSGI-INF/AssetCloudletComponent.xml
@@ -1,16 +1,13 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!--
-    Copyright (c) 2016 Eurotech and/or its affiliates and others
+    Copyright (c) 2016, 2017 Eurotech and/or its affiliates and others
 
      All rights reserved. This program and the accompanying materials
      are made available under the terms of the Eclipse Public License v1.0
      which accompanies this distribution, and is available at
      http://www.eclipse.org/legal/epl-v10.html
 -->
-<scr:component xmlns:scr="http://www.osgi.org/xmlns/scr/v1.1.0" 
-	activate="activate" 
-	deactivate="deactivate" 
-	name="org.eclipse.kura.asset.cloudlet">
+<scr:component xmlns:scr="http://www.osgi.org/xmlns/scr/v1.1.0" activate="activate" deactivate="deactivate" immediate="true" name="org.eclipse.kura.asset.cloudlet">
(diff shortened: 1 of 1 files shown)


kura/org.eclipse.kura.asset.cloudlet.provider/OSGI-INF/AssetCloudletComponent.xml
```diff
@@ -1,16 +1,13 @@
...
      All rights reserved. This program and the accompanying materials
      are made available under the terms of the Eclipse Public License v1.0
      which accompanies this distribution, and is available at
      http://www.eclipse.org/legal/epl-v10.html
 -->
-<scr:component xmlns:scr="http://www.osgi.org/xmlns/scr/v1.1.0" 
-	activate="activate" 
-	deactivate="deactivate" 
-	name="org.eclipse.kura.asset.cloudlet">
+<scr:component xmlns:scr="http://www.osgi.org/xmlns/scr/v1.1.0" activate="activate" deactivate="deactivate" immediate="true" name="org.eclipse.kura.asset.cloudlet">
```
amitjoy, line 10: As far as I know the class loading mechanism and its instantiation of a delayed service component can be delayed until some consumer requests for the service. As `CloudLet`s are service instances (not only bare service components), `AssetCloudlet` has been made a delayed component.

MMaiero: I've setup a wire asset in a develop build and, as you can see in the following images, the cloudlet is not registered and cannot be invoked from remote.

schermata 2017-03-30 alle 13 53 20

schermata 2017-03-30 alle 13 52 13

This is, instead, what you get when you compile the branch of the suggested PR:
schermata 2017-03-30 alle 14 08 55
schermata 2017-03-30 alle 14 09 43

As you can see, in the first case, the ASSET-V1 cloudlet is not started and registered.
So, maybe I've missed something, but I think that with the current implementation, the cloudlet should be started immediately to have it working.

amitjoy: I don't know why but Asset-V1 cloudlet instance should have been registered as soon as this service instance is tracked by a consumer. This is the usual scenario of a delayed component. As soon as a consumer requests for a delayed component instance, it gets instantiated. And since beginning, it was delayed and during testing and our presentations, it worked even though it was delayed.

MMaiero: Yes, but you know that with the new component tracker system, changed a little the things.

amitjoy: Now I understand why it is not activated 👍

PR #1228, opened 2017-03-17
Azure Example

mcarrer:
Added markdown guide on how to configure the Cloud/Data/MQTT stack to connect to Azure IoT and sample Kura application which connects to Azure IoT.

PR #1279, opened 2017-03-30
Added throttle for message publishing

pierantoniomerlino:
The idea is to manage the rate and burst of publishing using the Token Bucket algorithm.
The user can configure in the DataService the "Maximum publish rate" and "Maximum number of messages in publish bursts". The first parameter is used to set the maximum messages per seconds that can be published. The second one, instead, sets the number of messages that can be published in a burst at max speed. After the burst, the publishing rate is limited to "Maximum publish rate" value.

If the "Maximum publish rate" is set to 0, no limits are set and the messaged are published at max speed. If "Maximum number of messages in publish bursts" is set to 0, the burst are disabled and the messages are published at "Maximum publish rate". The latter parameter is used only if the "Maximum publish rate" is set to a value different than 0.

Of course, this PR is preliminary and any suggestions are more than welcome!


kura/org.eclipse.kura.core/OSGI-INF/metatype/org.eclipse.kura.data.DataService.xml
```diff
@@ -98,7 +98,25 @@
             default="0"
             min="0"
             description="Timeouts the in-flight messages congestion condition. The service will force a disconnect attempting to reconnect (0 to disable)."/>
+
+        <AD id="max.publish.rate"
+            name="Maximum publish rate"
+            type="Double"
```
MMaiero: Why this is a double?

kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/data/DataServiceImpl.java
```diff
@@ -104,6 +109,14 @@ protected void activate(ComponentContext componentContext, Map<String, Object> p
 
         this.properties.putAll(properties);
 
+        if (this.properties.get(PUBLISH_RATE) != null && this.properties.get(BURST_LENGTH) != null) {
```
MMaiero: I think that this can be externalized to a separate method.

kura/org.eclipse.kura.core/OSGI-INF/metatype/org.eclipse.kura.data.DataService.xml
pierantoniomerlino: max.publish.rate is the number of messages that should be send in a second. So, if you want to publish 1 messages every 10 seconds, the value is 0.1. Float is sufficient.
Is there a better way to show it?

MMaiero: Maybe specifying a time unit.

kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/data/DataServiceImpl.java
```diff
@@ -526,6 +536,27 @@ public void run() {
         return autoConnect;
     }
 
+    private void getThrottle() {
+        if (this.properties.get(PUBLISH_RATE) != null && this.properties.get(BURST_LENGTH) != null) {
```
MMaiero: Isn't it more readable if we save `this.properties.get(PUBLISH_RATE)` and `this.properties.get(BURST_LENGTH)` into local variables?

MMaiero:
If we compare our implementation to a token bucket, I think that maybe we should stick with common terms like (https://en.wikipedia.org/wiki/Token_bucket): 

- `Average rate` instead of `Maximum publish rate`

- `Burst size` instead of `Maximum number of messages in publish bursts`

What do you think?

kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/data/DataServiceImpl.java
pierantoniomerlino: 👍

pierantoniomerlino:
IMHO, we have to think what is easier to understand from an user point of view.
Is it better to expose the "token bucket" terms or use terms related to the data service?

kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/data/DataServiceImpl.java
```diff
@@ -572,14 +598,7 @@ public void run() {
...
 
-                        publishInternal(message);
-
-                        // TODO: add a 'message throttle' configuration parameter to
-                        // slow down publish rate?
-
-                        // Notify the listeners
-                        DataServiceImpl.this.dataServiceListeners.onMessagePublished(message.getId(),
-                                message.getTopic());
+                        tryPublishInternal(message);
```
cdealti: why "try"?

kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/data/DataServiceImpl.java
```diff
@@ -593,6 +612,15 @@ public void run() {
         });
     }
 
+    private synchronized void tryPublishInternal(DataMessage message) throws KuraException {
+        if (this.throttle.getToken()) {
+            logger.debug("Throttle: got token - remaining token {}", this.throttle.getRemainingTokens());
```
cdealti: "remaining tokens"

kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/data/internal/TokenBucket.java
```diff
@@ -0,0 +1,79 @@
+/*******************************************************************************
+ * Copyright (c) 2011, 2017 Eurotech and/or its affiliates
```
cdealti: only 2017

kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/data/internal/TokenBucket.java
```diff
@@ -0,0 +1,79 @@
...
+ *
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *     Eurotech
+ *******************************************************************************/
+package org.eclipse.kura.core.data.internal;
```
cdealti: I believe it should be org.eclipse.kura.core.internal.data. Check the Wires implementation bundles and the Eclipse guidelines for package naming conventions.

kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/data/internal/TokenBucket.java
```diff
@@ -0,0 +1,79 @@
...
+    private long remainingTokens;
+    // period in mills between 1 token refill
+    private long refillPeriod;
+    private long lastRefillTime;
+
+    public TokenBucket(long capacity, long refillPeriod) {
+        this.capacity = (capacity == 0) ? 1 : capacity;
+        this.remainingTokens = capacity;
+        this.refillPeriod = refillPeriod;
+        this.lastRefillTime = 0L;
```
cdealti: Remove "L"

kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/data/internal/TokenBucket.java
```diff
@@ -0,0 +1,79 @@
...
+        this.refillPeriod = refillPeriod;
+    }
+
+    public long getLastRefill() {
+        return this.lastRefillTime;
+    }
+
+    public boolean getToken() {
+        boolean success;
+        if (this.refillPeriod == 0L) {
```
cdealti: Remove "L"

kura/org.eclipse.kura.core/OSGI-INF/metatype/org.eclipse.kura.data.DataService.xml
```diff
@@ -98,7 +98,36 @@
...
+            min="0"
+            description="The average message publish rate. The unit of measurement is set by the publish rate unit of measurement option. If set to 0 the rate limit is disabled."/>
+        
+        <AD id="average.publish.rate.unit"
+            name="publish rate unit of measurement"
+            type="String"
+            cardinality="0" 
+            required="true"
+            default="messages.second"
+            description="">
```
cdealti: Add description

kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/data/internal/TokenBucket.java
```diff
@@ -0,0 +1,79 @@
...
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *     Eurotech
+ *******************************************************************************/
+package org.eclipse.kura.core.data.internal;
+
+public class TokenBucket {
+
+    private long capacity;
```
cdealti: There will plenty of room just using ``int``.

kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/data/internal/TokenBucket.java
```diff
@@ -0,0 +1,79 @@
...
+    private long remainingTokens;
```
cdealti: Same

kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/data/internal/TokenBucket.java
```diff
@@ -0,0 +1,79 @@
...
+    // period in mills between 1 token refill
+    private long refillPeriod;
+    private long lastRefillTime;
+
+    public TokenBucket(long capacity, long refillPeriod) {
+        this.capacity = (capacity == 0) ? 1 : capacity;
```
cdealti: Unclear

kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/data/internal/TokenBucket.java
```diff
@@ -0,0 +1,79 @@
...
+        this.remainingTokens = capacity;
+        this.refillPeriod = refillPeriod;
+        this.lastRefillTime = 0L;
+    }
+
+    public long getCapacity() {
+        return this.capacity;
+    }
+
+    public void setCapacity(long capacity) {
```
cdealti: Remove setters. Create a new instance instead

kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/data/internal/TokenBucket.java
```diff
@@ -0,0 +1,79 @@
...
+
+    public void setRefillPeriod(long refillPeriod) {
+        this.refillPeriod = refillPeriod;
+    }
+
+    public long getLastRefill() {
+        return this.lastRefillTime;
+    }
+
+    public boolean getToken() {
```
cdealti: ``hasToken`` preferred

kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/data/internal/TokenBucket.java
```diff
@@ -0,0 +1,79 @@
...
+        boolean success;
```
cdealti: intialize to ``false``

kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/data/internal/TokenBucket.java
```diff
@@ -0,0 +1,79 @@
...
+        if (this.refillPeriod == 0L) {
+            success = true;
+        } else {
+            tryRefill();
+            if (this.remainingTokens > 0) {
+                this.remainingTokens--;
+                success = true;
+            } else {
```
cdealti: remove ``else``

kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/data/internal/TokenBucket.java
```diff
@@ -0,0 +1,79 @@
...
+                success = false;
+            }
+        }
+        return success;
+    }
+
+    private void tryRefill() {
```
cdealti: Just ``refill`` returning the number of remaining tokens.

However, we miss an ``waitToken`` method which blocks until one token is available.
Make sure that the wait can be properly interrupted.

kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/data/DataServiceImpl.java
```diff
@@ -156,6 +164,8 @@ public void updated(Map<String, Object> properties) {
         this.properties.clear();
         this.properties.putAll(properties);
 
+        getThrottle();
```
cdealti: ``createThrottle`` or something indicating that an instanciation  will occur

kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/data/DataServiceImpl.java
```diff
@@ -526,6 +536,28 @@ public void run() {
         return autoConnect;
     }
 
+    private void getThrottle() {
+        if (this.properties.get(PUBLISH_RATE) != null && this.properties.get(BURST_SIZE) != null) {
```
cdealti: Hence a new ``TokenBucket`` instance will be created

kura/org.eclipse.kura.core/OSGI-INF/metatype/org.eclipse.kura.data.DataService.xml
```diff
@@ -98,7 +98,36 @@
...
+        <AD id="average.publish.rate"
+            name="Average publish rate"
+            type="Integer"
+            cardinality="0"
+            required="true"
+            default="0"
+            min="0"
+            description="The average message publish rate. The unit of measurement is set by the publish rate unit of measurement option. If set to 0 the rate limit is disabled."/>
+        
+        <AD id="average.publish.rate.unit"
```
cdealti: I believe this is not needed.
I believe that messages/s is the right unit.
Realistically 1 [message/s] is probably what one will set.
Is it too fast for the broker? Setting 0.1 will do the trick

kura/org.eclipse.kura.core/OSGI-INF/metatype/org.eclipse.kura.data.DataService.xml
```diff
@@ -98,7 +98,36 @@
...
+        </AD>
+        
+        <AD id="publish.burst.size"
+            name="Message burst size"
+            type="Long"
+            cardinality="0"
+            required="true"
+            default="0"
+            min="0"
+            description="When the average publish rate is set, this option enables the publishing of messages in bursts at maximum speed. If set to 0 the bursts are disabled."/>
```
cdealti: Just to double check. The broker will not be happy with bursts. Setting this property to 1 disables bursts entirely?

kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/data/DataServiceImpl.java
```diff
@@ -593,6 +618,15 @@ public void run() {
         });
     }
 
+    private synchronized void tryPublishInternal(DataMessage message) throws KuraException {
+        if (this.throttle.getToken()) {
```
cdealti: It should just be ``this.throttle.waitToken()``. Otherwise if this method is called from the publishing task and the DB is full and you don't have tokens, the CPU will spin doing nothing.

With waitToken() at least the thread will block until a token is available.

cdealti:
About the priority of this one. It seems needed to limit the publish rate with Azure IoT but we need to recheck. See #1228.

@markoer  the DataService is a key component that if it fails will cause headaches with customers.
Assuming this will be merged, can you please add some new QA tests for validation covering the corner cases? You should pay attention to cases where the messages DB contains many messages, like after a disconnect and the client reconnects.

kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/data/DataServiceImpl.java
pierantoniomerlino: Because if there aren't tokens in the bucket, the Data service can't publish the message. So, it tries to publish something...

kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/data/DataServiceImpl.java
pierantoniomerlino: 👍

kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/data/internal/TokenBucket.java
pierantoniomerlino: 👍

kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/data/internal/TokenBucket.java
pierantoniomerlino: I'll add a comment about this.

kura/org.eclipse.kura.core/OSGI-INF/metatype/org.eclipse.kura.data.DataService.xml
pierantoniomerlino: Yes, but is it enough clear to the user set a rate of 0.1 message/second? If the maximum rate is 1 message/second, so we should change the unit to message/minute and get an integer value.

kura/org.eclipse.kura.core/OSGI-INF/metatype/org.eclipse.kura.data.DataService.xml
pierantoniomerlino: The burst size is the capacity of the token bucket. If it is 0, the token bucket constructor force the capacity to 1, so only a token per period can be extracted from the bucket and there will not be bursts.

kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/data/DataServiceImpl.java
pierantoniomerlino: Do we need a timeout for this?

kura/org.eclipse.kura.core/OSGI-INF/metatype/org.eclipse.kura.data.DataService.xml
cdealti: I suggest to add the 0.1 example to the description

kura/org.eclipse.kura.core/OSGI-INF/metatype/org.eclipse.kura.data.DataService.xml
cdealti: Then set min="1". WYSIWYG

kura/org.eclipse.kura.core/OSGI-INF/metatype/org.eclipse.kura.data.DataService.xml
pierantoniomerlino: I'll remove the unit option and set the rate as messages/second. I'll also update the description to give an example on how to use it. However, I don't like it.

kura/org.eclipse.kura.core/OSGI-INF/metatype/org.eclipse.kura.data.DataService.xml
pierantoniomerlino: Ok.

cdealti:
See #1416.

kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/data/DataServiceImpl.java
```diff
@@ -76,7 +70,7 @@
     private ScheduledFuture<?> reconnectFuture;
 
     // A dedicated executor for the publishing task
-    private ScheduledExecutorService publisherExecutor;
+    private ExecutorService publisherExecutor;
```
cdealti, line 74: It looks to me that a plain old Java Thread would me enough

kura/org.eclipse.kura.core/OSGI-INF/metatype/org.eclipse.kura.data.DataService.xml
```diff
@@ -98,7 +98,46 @@
             default="0"
             min="0"
             description="Timeouts the in-flight messages congestion condition. The service will force a disconnect attempting to reconnect (0 to disable)."/>
-            
+
+		<AD id="enable.rate.limit"
```
cdealti: Fix indentation

kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/data/DataServiceImpl.java
```diff
@@ -516,16 +523,27 @@ public void run() {
...
-                    TimeUnit.SECONDS);
+            }, initialDelay, reconnectInterval, TimeUnit.SECONDS);
         } else {
             // Change notification status to off. Connection is not expected to happen in the future
             this.cloudConnectionStatusService.updateStatus(this, CloudConnectionStatusEnum.OFF);
         }
         return autoConnect;
     }
 
+    private void createThrottle() {
```
cdealti, line 533: Should we add NPE checks for this.throttle everywhere?

kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/data/DataServiceImpl.java
```diff
@@ -680,4 +650,99 @@ public CloudConnectionStatusEnum getNotificationStatus() {
...
+                        DataMessage message = DataServiceImpl.this.store.getNextMessage();
+
+                        if (message != null) {
+                            checkInFlightMessages(message);
+
+                            boolean messagePublished;
+                            if (dataServiceOptions.isRateLimitEnabled() && message.getPriority() >= 5) {
+                                messagePublished = publishMessageTokenBucket(message);
+                            } else {
+                                messagePublished = publishMessageUnbound(message);
```
cdealti: this method alway returns true so set messagePublished to true explicitly or intialize it to true.

kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/data/DataServiceImpl.java
```diff
@@ -680,4 +650,99 @@ public CloudConnectionStatusEnum getNotificationStatus() {
...
+                    && DataServiceImpl.this.inFlightMsgIds.size() >= dataServiceOptions.getMaxInFlightMessages()) {
+                logger.warn("The configured maximum number of in-flight messages has been reached");
+                throw new KuraTooManyInflightMessagesException("Too many in-flight messages");
+            }
+        }
+
+        private void suspendPublisher(long timeout, TimeUnit timeUnit) throws InterruptedException {
+            try {
+                lock.lock();
+                if (!publisherEnabled) {
```
cdealti: Why ``publisherEnabled`` must be checked in a lock?
You don't need to throw an Exception, just return.

kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/data/DataServiceImpl.java
```diff
@@ -680,4 +650,99 @@ public CloudConnectionStatusEnum getNotificationStatus() {
...
+                        lockCondition.await(timeout, timeUnit);
+                    }
+                }
+                notifyPending = false;
+            } finally {
+                lock.unlock();
+            }
+        }
+
+        private boolean publishMessageUnbound(DataMessage message) throws KuraException {
```
cdealti: return void

kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/data/DataServiceOptions.java
```diff
@@ -0,0 +1,117 @@
+package org.eclipse.kura.core.data;
```
cdealti, line 12: Copyright header missing

kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/internal/data/TokenBucket.java
```diff
@@ -0,0 +1,58 @@
...
+ *
+ * Contributors:
+ *     Eurotech
+ *******************************************************************************/
+package org.eclipse.kura.core.internal.data;
+
+public class TokenBucket {
+
+    private final int capacity;
+    // period in mills between 1 token refill
```
cdealti: Check spelling or remove comment

kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/internal/data/TokenBucket.java
```diff
@@ -0,0 +1,58 @@
...
+        boolean result = false;
+        refill();
+        if (isTokenAvailable()) {
+            this.remainingTokens--;
+            result = true;
+        }
+        return result;
+    }
+
+    private boolean isTokenAvailable() {
```
cdealti, line 38: Move private methods at the bottom

kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/internal/data/TokenBucket.java
```diff
@@ -0,0 +1,58 @@
...
+        return this.remainingTokens != 0;
+    }
+
+    private void refill() {
+        long now = System.currentTimeMillis();
+        if (now - this.lastRefillTime >= this.refillPeriod) {
+            this.remainingTokens = (int) Math.min(this.capacity,
+                    this.remainingTokens + (now - this.lastRefillTime) / this.refillPeriod);
+            this.remainingTokens = Math.max(1, this.remainingTokens);
+            this.lastRefillTime = now;
```
cdealti: lastRefillTime is not accurate as you lose the elapsed part of the refillPeriod at the time refill() is called.
I believe it should be computed as:

lastRefillTime += ((now - lastRefillTime) / refillPeriod) * refillPeriod

Note that this does not equal ``now`` due to the calculation uses integer arithmetics.

kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/internal/data/TokenBucket.java
```diff
@@ -0,0 +1,58 @@
...
+        }
+    }
+
+    public long getNextRefillEta() {
```
cdealti: Maybe add a comment that you don't get an updated and meaningful value until getToken() is called. This usage is misleading and error prone however.
The method name should also indicate a "time to next refill" (relative to now) instead of an "ETA of the next refill" (relative to epoch).

cdealti (changes requested):

kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/data/DataServiceImpl.java
MMaiero: I believe in this context there is no real advantage in one or the other solution.

kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/data/DataServiceImpl.java
MMaiero: It's to prevent concurrent changes and reads of the shared variable.

kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/data/DataServiceOptions.java
MMaiero: Correct. Missed that.

kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/data/DataServiceImpl.java
MMaiero: Uhm, where do you think it is needed?

kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/data/DataServiceImpl.java
```diff
@@ -680,4 +649,106 @@ public CloudConnectionStatusEnum getNotificationStatus() {
...
+                }
+            }
+            logger.debug("Exited publisher loop.");
+        }
+
+        private boolean isExecutionEnabled() {
+            boolean executionEnabled;
+            lock.lock();
+            executionEnabled = publisherEnabled;
+            lock.unlock();
```
pierantoniomerlino: Is it possible to use an AtomicBoolean instead?

MMaiero: Agreed and done. I've done also some minimal testing and it seems to work.

PR #1289, opened 2017-04-03
Cleanup the payload encoder a bit

ctron:
* Iterate over the metrics map instead of iterating over the keys
  and fetching the values
* Fix the naming schema
* Check the "null" state instead of deliberately running into a second
  exception

### DIFF
--- a/kura/org.eclipse.kura.core.cloud/src/main/java/org/eclipse/kura/core/cloud/CloudPayloadProtoBufEncoderImpl.java
+++ b/kura/org.eclipse.kura.core.cloud/src/main/java/org/eclipse/kura/core/cloud/CloudPayloadProtoBufEncoderImpl.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2011, 2016 Eurotech and/or its affiliates
+ * Copyright (c) 2011, 2017 Eurotech and/or its affiliates and others
  *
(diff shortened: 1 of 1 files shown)


kura/org.eclipse.kura.core.cloud/src/main/java/org/eclipse/kura/core/cloud/CloudPayloadProtoBufEncoderImpl.java
```diff
@@ -46,46 +49,40 @@ public CloudPayloadProtoBufEncoderImpl(KuraPayload kuraPayload) {
...
             try {
-                KuraPayloadProto.KuraPayload.KuraMetric.Builder metricB = KuraPayloadProto.KuraPayload.KuraMetric
-                        .newBuilder();
+                KuraMetric.Builder metricB = KuraMetric.newBuilder();
                 metricB.setName(name);
 
                 boolean result = setProtoKuraMetricValue(metricB, value);
                 if (result) {
-                    metricB.build();
-
```
pierantoniomerlino, line 71: Why did you removed the metricB.build()?

ctron: Because it is not needed. In this context, not using the return value, it is a no-op.

PR #1291, opened 2017-04-03
Added Regular Expression Filter Wire Component

### LINKED ISSUE #1211: [Wires] Implement regex filter WireComponent

amitjoy:
Fixes #1211

### DIFF
--- a/kura/org.eclipse.kura.localization.resources/src/main/java/org/eclipse/kura/localization/resources/WireMessages.java
+++ b/kura/org.eclipse.kura.localization.resources/src/main/java/org/eclipse/kura/localization/resources/WireMessages.java
@@ -46,6 +46,12 @@ public interface WireMessages {
     @En("Activating Logger Wire Component...Done")
     public String activatingLoggerDone();
 
+    @En("Activating Regex Filter...")
+    public String activatingRegexFilter();
+
+    @En("Activating Regex Filter...Done")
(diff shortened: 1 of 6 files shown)


cdealti:
@amitjoy thanks. We will give this a try. Before testing it I'd like to check with you the typical scenario.
The WireEnvelope being filtered is emitted by an WireAsset configured with tenths of channels.
I'd like to configure the filter easily to remove all the properties matching a regex or to keep all the properties matching a regex. In some scenarios writing a regex to remove a single property is easier than writing one to keep all the rest. In some other cases it's the opposite: I want to only keep one property.
So there should be a radio box or something to control whether the regex is used to remove or keep.

amitjoy:
@cdealti I agree that providing this level of flexibility of keeping or removing entries that match the provided regex would be great. I will make the necessary changes to the PR.

MMaiero:
I would remove the messages like this one: `RegexFilter - filter - Regular Expression Filtering started..`
because they will fill the log without a real contribution for the end user.

Furthermore, I would suggest to improve the `filter.type` description: currently it is not clear.

amitjoy:
The message should be set as debug to check if the filter wire component ever triggered filtering operation.

I have also amended the filter type with necessary description.

Issue #1202, opened 2017-03-15
[Asset] Drop channel ID and use unique name


cdealti:
Having both channel ID and channel name is not needed.

The channel name will also be used in the ASSET-V1 Cloudlet topics and needs to be sanitized ('+', '#' and '/' characters are not allowed in MQTT topics.

PR #1292, opened 2017-04-04
Attempt to replace Asset channel id with channel name

nicolatimeus:
See #1202

Also updated WireAssetTest


MMaiero:
@amitjoy Any thoughts on this? I would like to merge it as soon as possible. Thanks

amitjoy:
@MMaiero Will do it soon.
@nicolatimeus Could you kindly resolve the conflicts?

MMaiero:
@amitjoy I've rebased @nicolatimeus branch

amitjoy:
closes #1202

amitjoy:
There are few places in which `short` and `byte` data types are mentioned. If you could trigger the CI again, it will show the errors. Could you kindly fix those?

MMaiero:
The rebase was before the merging of the types cleanup.
I will rebase it again, as soon as you have completed the review.

amitjoy:
👍

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/asset/Asset.java
```diff
@@ -117,7 +115,7 @@
      *             channel ID is less than or
      *             equal to zero
      */
-    public void registerAssetListener(long channelId, AssetListener assetListener) throws KuraException;
+    public void registerAssetListener(String channelName, AssetListener assetListener) throws KuraException;
```
amitjoy: throwable IllegalArgumentException description should be modified. This method should throw `IllegalArgumentException` if the provided channel name is not validated, that is, if the non-allowed characters are present and so on.

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/asset/Asset.java
```diff
@@ -69,7 +67,7 @@
      * @return the list of asset records which comprises the currently read
      *         value in case of success or the reason of failure
      */
-    public List<AssetRecord> read(List<Long> channelIds) throws KuraException;
+    public List<AssetRecord> read(Set<String> channelNames) throws KuraException;
```
amitjoy, line 70: Remove `IllegalArgumentException` as a throwable exception for this method as well

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/asset/AssetConfiguration.java
```diff
@@ -24,54 +24,49 @@
...
- * For example, 1.CH.name, 1.CH.value.type etc.<br/>
- * <br/>
- *
- * The representation in the provided properties as prepended by a number
- * signifies a single channel and it should conform to the following
- * specification.<br/>
- * <br/>
+ * <li>[name#property]</li> where name is a string denoting the channel's unique
+ * name and the {@code [property]} denotes the protocol specific properties.
+ * The name of a channel must be unique inside a channel configuration, and is not
```
amitjoy: The name of a channel must be unique in the channels' configurations of an Asset

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/asset/AssetConfiguration.java
```diff
@@ -24,54 +24,49 @@
...
+ * allowed to contain spaces or any of the following characters: '#', '_'.
```
amitjoy: Highlight the non-allowable characters in javadoc

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/asset/AssetRecord.java
```diff
@@ -49,48 +49,16 @@
     /**
      * Instantiates a new asset record.
      *
-     * @param channelId
+     * @param channelName
      *            the channel identifier
      * @throws IllegalArgumentException
```
amitjoy: Modify the `IllegalArgumentException` description

amitjoy: Add NPE to throw if channel name is null. It is a public API.

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/asset/AssetRecord.java
```diff
@@ -180,8 +136,49 @@ public void setValue(final TypedValue<?> value) {
     /** {@inheritDoc} */
     @Override
     public String toString() {
-        return "AssetRecord [assetStatus=" + this.assetStatus + ", channelId=" + this.channelId + ", timestamp="
+        return "AssetRecord [assetStatus=" + this.assetStatus + ", channelId=" + this.channelName + ", timestamp="
```
amitjoy: channelId => channelName

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/asset/Channel.java
```diff
@@ -68,60 +64,20 @@
      *             if channel ID is 0 or
```
amitjoy: Add a invariant to check if the channel name is valid and modify IllegalArgumentException description thereafter

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/driver/DriverConstants.java
```diff
@@ -16,10 +16,10 @@
 public enum DriverConstants {
```
amitjoy, line 16: Modify the copyright year

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/asset/AssetConstants.java
```diff
@@ -15,35 +15,29 @@
  */
 public enum AssetConstants {
```
amitjoy, line 16: Modify the copyright year

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/driver/DriverRecord.java
```diff
@@ -29,32 +30,18 @@
...
- * The channel ID must be set in the map as a key of :
- * {@link DriverConstants#CHANNEL_ID} or {@code channel.id}<br/>
- * <br/>
- *
- * In case you are using driver in isolation (i.e without using Assets), then
+ * If the channel configuration is generated by an {@link Asset}, it will also
+ * contain the {@link DriverConstants#CHANNEL_ID} property ({@code +name}),
+ * that identifies the channel name as configured in the asset.
+ * In case the driver is used directly (i.e without using Assets), then
  * the channel ID need not to be set.<br/>
```
amitjoy: Even though ID denotes identity, it would be better off with channel name instead.

kura/org.eclipse.kura.asset.cloudlet.provider/src/main/java/org/eclipse/kura/internal/asset/cloudlet/AssetCloudlet.java
```diff
@@ -14,13 +14,20 @@
...
 
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+
 import org.eclipse.kura.KuraException;
 import org.eclipse.kura.asset.Asset;
 import org.eclipse.kura.asset.AssetConfiguration;
```
amitjoy, line 32: make necessary amendments to the javadoc of this class to include `channel_name` instead of `channel_id`. It will be consistent everywhere.

kura/org.eclipse.kura.asset.cloudlet.provider/src/main/java/org/eclipse/kura/internal/asset/cloudlet/AssetCloudlet.java
```diff
@@ -328,42 +335,39 @@ private void findAssets() {
      *            the response paylaod to be prepared
      * @param assetPid
      *            the Asset PID
-     * @param channelId
+     * @param channelNameList
      *            the channel ID (might contain {@code -} for multiple reads)
```
amitjoy: channel name

kura/org.eclipse.kura.asset.cloudlet.provider/src/main/java/org/eclipse/kura/internal/asset/cloudlet/AssetCloudlet.java
```diff
@@ -328,42 +335,39 @@ private void findAssets() {
...
      * @throws NullPointerException
      *             if any of the arguments is null
      */
     private void readChannelsByIds(final KuraResponsePayload respPayload, final String assetPid,
```
amitjoy: readChannelsByNames

kura/org.eclipse.kura.asset.cloudlet.provider/src/main/java/org/eclipse/kura/internal/asset/cloudlet/AssetCloudlet.java
```diff
@@ -328,42 +335,39 @@ private void findAssets() {
...
-            final String channelId) {
+            final String channelNameList) {
         requireNonNull(respPayload, message.respPayloadNonNull());
         requireNonNull(assetPid, message.assetPidNonNull());
-        requireNonNull(channelId, message.channelIdNonNull());
+        requireNonNull(channelNameList, message.channelIdNonNull());
 
-        final String channelDelim = ",";
+        final String channelDelim = "#";
         Set<String> channelIds = null;
```
amitjoy: channelNames

kura/org.eclipse.kura.asset.provider/src/main/java/org/eclipse/kura/asset/provider/BaseAsset.java
```diff
@@ -235,11 +230,9 @@ private Tad cloneAd(final Tad oldAd, final String prefix) {
         requireNonNull(oldAd, message.oldAdNonNull());
```
amitjoy, line 225: change the prefix param javadoc accordingly

kura/org.eclipse.kura.asset.provider/src/main/java/org/eclipse/kura/asset/provider/BaseAsset.java
```diff
@@ -588,15 +557,13 @@ public boolean hasReadChannels() {
...
     /** {@inheritDoc} */
     @Override
-    public void registerAssetListener(final long channelId, final AssetListener assetListener) throws KuraException {
-        if (channelId <= 0) {
-            throw new IllegalArgumentException(message.channelIdNotLessThanZero());
-        }
+    public void registerAssetListener(final String channelName, final AssetListener assetListener)
+            throws KuraException {
         requireNonNull(assetListener, message.listenerNonNull());
         requireNonNull(this.driver, message.driverNonNull());
```
amitjoy: validate channel name

kura/org.eclipse.kura.asset.provider/src/main/java/org/eclipse/kura/internal/asset/provider/AssetOptions.java
```diff
@@ -133,172 +93,111 @@ public AssetConfiguration getAssetConfiguration() {
...
+        Set<String> alreadyProcessedChannelNames = new HashSet<>();
+        Map<String, Channel> result = new HashMap<>();
+
+        for (Entry<String, Object> e : properties.entrySet()) {
+            String channelName = extractChannelName(e.getKey());
+            if (channelName == null || alreadyProcessedChannelNames.contains(channelName)) {
+                continue;
+            }
+            alreadyProcessedChannelNames.add(channelName);
+            if (!Channel.isValidChannelName(channelName)) {
```
amitjoy, line 193: shouldn't this be added to the first if condition as an OR condition?

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/asset/AssetConfiguration.java
```diff
@@ -122,7 +117,7 @@ public AssetConfiguration(final String description, final String driverPid, fina
      *
      * @return the asset channels
      */
-    public Map<Long, Channel> getAssetChannels() {
+    public Map<String, Channel> getAssetChannels() {
         return this.assetChannels;
```
amitjoy, line 122: It's my mistake actually 😜. Could you kindly return an unmodifiable view of the map?

kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/client/ui/wires/WiresPanelUi.java
```diff
@@ -68,930 +68,931 @@
 
 public class WiresPanelUi extends Composite {
 
-    @UiField
```
amitjoy: I believe there is some issue with the formatter as it is showing the complete change in `WiresPanelUi`. As it is very large souce, it will be difficult to find out the real changes. Could you kindly check it once?

kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/shared/model/GwtChannelInfo.java
```diff
@@ -23,10 +23,6 @@
         return super.get("driver." + key);
```
amitjoy, line 23: Change the copyright year

kura/org.eclipse.kura.web2/src/main/webapp/denali.css
```diff
@@ -232,7 +232,7 @@ Upload panel END
 Channel Table Validation START
 ***********************/
```
amitjoy, line 233: change the copyright year

amitjoy (changes requested):
v1

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/asset/Asset.java
amitjoy, line 70: Are we considering case sensitivity? According to the contract of string equality, two strings with different cases are considered to be different strings and the set can contain two same strings with different cases.

nicolatimeus: Yes, but this shouldn't be a problem because the channel name is case-sensitive, right? Two strings with different cases refer to two different channels.

kura/org.eclipse.kura.asset.provider/src/main/java/org/eclipse/kura/internal/asset/provider/AssetOptions.java
nicolatimeus: It should be correct as it is now, because validating the channel is more expensive than checking if it is present in the alreadyProcessedChannelNames set.
In this way only the channels names that have not yet been discovered will be validated, and just once.

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/asset/Asset.java
amitjoy: From user perspective, this would create a bit of confusion. Otherwise fine.

kura/org.eclipse.kura.asset.provider/src/main/java/org/eclipse/kura/internal/asset/provider/AssetOptions.java
amitjoy: But even if the channel name is not valid, it can be inserted in `alreadyProcessedChannelNames`. That would require extra operation to be performed in case the channel name is invalid. You are right that validating a channel is more expensive than checking if it is present in `alreadyProcessedChannelNames` but it is not complete safe as I mentioned earlier in case the channel name is not valid, there will be extra operation to be performed. To make `Channel.isValidChannelName` efficient, you can use pattern matchers instead of looping through the chars.

nicolatimeus: Pattern matchers loop over the chars internally as well, plus you have the overhead introduced by the state machine that matches the regexp. The fact that an invalid name is added to alreadyProcessedChannelNames does not mean that it will be added to the returned channel map.

amitjoy: I agree that introducing regular expression state machine would be a bit of overhead but `Matchers` loop through chars in an efficient way. 

Wouldn't it be better to have it like this?

```
if (channelName == null || alreadyProcessedChannelNames.contains(channelName) || !Channel.isValidChannelName(channelName)) {
  continue;
}
```

This will be evaluated from left to right. So if a channel name is null, it will `continue` or if a channel name is not null but the channel name is already contained in the set, it will `continue` or if the channel name is not null and the channel name is not in the set, it will validate the channel name.

In the existing source, if the channel name is not in the set, it will always validate the channel name. In my suggested way if the channel name is not contained in set, it will also validate the channel name but the add operation on the set will not be invoked. It will only be invoked if and only if all the OR conditions return false.

This is the one I suggested at the first place. You can correct me If you believe it would not perform optimally.

nicolatimeus: The fact that invalid channel names enter the alreadyProcessedChannelNames list is not an issue but it is the expected behaviour. If we find a key in the map that contains a channel name that is already in this list means that we can completely skip the processing of that property because it has already been performed.
By processing I mean channel name validation, extraction of the configuration for that channel and creation of a Channel object.

nicolatimeus: The processing can also end by determining that a channel name is invalid, in this case no Channel object will be generated but we still need to add that name to alreadyProcessedChannelNames, in this way if we see that invalid name again the corresponding property will be immediately skipped.

amitjoy: If you expect to encounter same channel names multiple times (irrespective of it to be valid or invalid), `contains` on set will definitely be optimal as it will be performed in constant time. 👍

nicolatimeus: You will encounter it multiple times, once for each configuration property associated with that channel.

kura/org.eclipse.kura.asset.provider/src/main/java/org/eclipse/kura/asset/provider/BaseAsset.java
```diff
@@ -542,41 +507,45 @@ private AssetRecord createAssetRecordFromDriverRecord(DriverRecord driverRecord)
 
     /** {@inheritDoc} */
     @Override
-    public List<AssetRecord> read(final List<Long> channelIds) throws KuraException {
+    public List<AssetRecord> read(final Set<String> channelNames) throws KuraException {
         requireNonNull(this.driver, message.driverNonNull());
```
amitjoy, line 508: According to the Asset's `read` contract, if the `set` is empty, it would throw `IllegalArgumentException`

kura/org.eclipse.kura.asset.provider/src/main/java/org/eclipse/kura/asset/provider/BaseAsset.java
amitjoy: Please address this in `cloneAd(Tad, String) method`.

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/asset/Asset.java
```diff
@@ -39,21 +40,18 @@
...
+     * If the connection to the asset is interrupted, then any necessary resources
+     * that correspond to this connection should be cleaned up and a {@code KuraException} shall be
      * thrown.
      *
-     * @param channelIds
-     *            the list of channel identifiers which are to be read. The
-     *            channel identifiers for an asset must be unique for every
-     *            channel belonging to an asset.
+     * @param channelNames
+     *            the set of channel names which are to be read.
```
amitjoy: Could you kindly add a note saying that uniqueness of channel names also considers case sensitivity of strings?

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/asset/Asset.java
```diff
@@ -94,8 +92,8 @@
      * Registers asset listener for the provided channel name for a monitor
      * operation on it.
      *
-     * @param channelId
-     *            the channel identifier. The channel identifier for an asset
+     * @param channelName
+     *            the channel name. The channel name for an asset
```
amitjoy: Could you kindly add a note here as well saying that uniqueness of channel names also considers case sensitivity of strings?

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/asset/AssetConfiguration.java
amitjoy: Kindly address this.

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/asset/Asset.java
```diff
@@ -114,10 +112,10 @@
      * @throws NullPointerException
      *             if any of the arguments is null
      * @throws IllegalArgumentException
-     *             channel ID is less than or
-     *             equal to zero
+     *             If the provided channel name is no present in the configuration
```
amitjoy: no present => not present

kura/org.eclipse.kura.asset.cloudlet.provider/src/main/java/org/eclipse/kura/internal/asset/cloudlet/AssetCloudlet.java
```diff
@@ -327,42 +333,39 @@ private void findAssets() {
...
      * @throws NullPointerException
      *             if any of the arguments is null
      */
-    private void readChannelsByIds(final KuraResponsePayload respPayload, final String assetPid,
-            final String channelId) {
+    private void readChannelsByNames(final KuraResponsePayload respPayload, final String assetPid,
+            final String channelName) {
         requireNonNull(respPayload, message.respPayloadNonNull());
         requireNonNull(assetPid, message.assetPidNonNull());
-        requireNonNull(channelId, message.channelIdNonNull());
```
amitjoy, line 339: Change the localization literals and method names, such as `message.channelNameNonNull()`

kura/org.eclipse.kura.asset.cloudlet.provider/src/main/java/org/eclipse/kura/internal/asset/cloudlet/AssetCloudlet.java
```diff
@@ -327,42 +333,39 @@ private void findAssets() {
...
+        requireNonNull(channelName, message.channelIdNonNull());
 
-        final String channelDelim = ",";
+        final String channelDelim = "#";
         Set<String> channelIds = null;
```
amitjoy: channelIds => channelNames

kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/client/ui/wires/PropertiesUi.ui.xml
```diff
@@ -91,6 +101,27 @@
...
+						<g:FlowPanel>
+							<b:TextBox ui:field="newChannelNameInput" b:id="channelName" allowBlank="false"
+								autoComplete="false"/>
+						</g:FlowPanel>
+						<b:FormLabel addStyleNames="{style.channel-name-validation-label}" ui:field="newChannelNameError"/>
+					</b:FormGroup>
+				</b:Form>
+			</b:ModalBody>
+			<b:ModalFooter>
+                <b:Button b:id="btn-create-comp-cancel" icon="REMOVE" type="PRIMARY" dataDismiss="MODAL" ui:field="btnCancelCreatingNewChannel"/>
```
amitjoy: kindly change the ID of the button to something like `btn-create-channel-cancel`

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/asset/Asset.java
```diff
@@ -114,10 +112,10 @@
...
+     *             of this asset
      */
-    public void registerAssetListener(long channelId, AssetListener assetListener) throws KuraException;
+    public void registerAssetListener(String channelName, AssetListener assetListener) throws KuraException;
```
amitjoy: Shouldn't it be better to check if a channel is invalid first and if not throw `IllegalArgumentException`. Currently according to the contract, it will try to find the provided channel in the Asset Channel Configuration and if not found, it will throw `IllegalArgumentException`. In either way, it seems same to me but I believe from a good contract point of view we must check if the channel name is invalid. What do you think?

amitjoy (changes requested):

cdealti:
@nicolatimeus @amitjoy it would be great if you could complete the review by tomorrow. Our priority goes to the API, then to bugs and obvious performance bottlenecks. Implementation stylistic issues can be resolved in a second pass or service release.

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/asset/AssetConfiguration.java
nicolatimeus: It has been addressed in the constructor

kura/org.eclipse.kura.asset.provider/src/main/java/org/eclipse/kura/asset/provider/BaseAsset.java
nicolatimeus: If I'm not wrong the exception was not thrown even before this PR.
In my opinion there is no need to throw an exception in this case, returning an empty list is fine. I would prefer removing the IllegalArgumentException from the javadoc instead.

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/asset/Asset.java
nicolatimeus: It is the same also to me, in my opinion this can be postposed

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/asset/AssetConfiguration.java
amitjoy: cool 👍

amitjoy: Didn't notice 😉

kura/org.eclipse.kura.asset.provider/src/main/java/org/eclipse/kura/asset/provider/BaseAsset.java
amitjoy: Yeah you are right and that's why in my previous review, I suggested you to remove it.

amitjoy:
@cdealti I hope it will be done pretty soon.

MMaiero:
@amitjoy I've seen that @nicolatimeus has made some changes. Are we ok now on merging?

amitjoy:
@MMaiero I am doing final review

kura/org.eclipse.kura.asset.provider/src/main/java/org/eclipse/kura/asset/provider/BaseAsset.java
```diff
@@ -386,9 +373,8 @@ private void fillDriverSpecificChannelConfiguration(final Tocd mainOcd,
             List<Tad> channelConfiguration = (List<Tad>) baseChannelDescriptor;
             channelConfiguration.addAll(driverSpecificChannelConfiguration);
             for (final Tad attribute : channelConfiguration) {
-                final Set<String> channelPrefixes = retrieveChannelPrefixes(this.assetConfiguration.getAssetChannels());
-                for (final String prefix : channelPrefixes) {
-                    final Tad newAttribute = cloneAd(attribute, prefix);
+                for (final String channelName : this.assetConfiguration.getAssetChannels().keySet()) {
```
amitjoy: iteration over the `entrySet` is faster

amitjoy (changes requested):
Last Change Request

kura/org.eclipse.kura.asset.provider/src/main/java/org/eclipse/kura/asset/provider/BaseAsset.java
MMaiero: Change done. Ready to go?

amitjoy (approved):

Issue #1295, opened 2017-04-04
[Wires] Wires DataType lacks the KuraPayload float type


cdealti:
Wires DataType only supports double precision floating point numbers (why?).
The single precision floating point type of a KuraPayload cannot be removed for backward compatibility reasons.

* Add the float type to DataType or
* Document that metrics of type float received by a WirePublisher will be converted to a float DataType when converting the KuraPayload to a WireRecord

MMaiero:
The official HSQLDB documentation states (http://hsqldb.org/doc/1.8/guide/ch09.html):

> The recommended Java mapping for the JDBC datatype FLOAT is as a Java type "double". Because of the potential confusion it is recommended that DOUBLE is used instead of FLOAT.

amitjoy:
Single precision floating point doesn't provide large degree of precision and to retain accuracy in performing precise calculations, wires didn't include `floats`.

The idea was to include primitive types (except `floats`) to be used as the datatypes in wires. And keeping this in mind, the existent datatypes were included. Previously while designing, I was not aware of the backward compatibility issues of `floats` in `KuraPayload`. You are right, it should be introduced.

Issue #1294, opened 2017-04-04
[Wires] Remove ByteValue and ShortValue as they do not map to KuraPayload metric types


cdealti:
The above types do not map to the KuraPayload types potentially causing lossy type conversions translating an WireRecord to a KuraPayload and vice-versa.

Hence they should be removed.

PR #1296, opened 2017-04-04
Uniformed DataTypes to match the ones supported by kura protobuf.

MMaiero:


kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/type/TypedValue.java
```diff
@@ -22,7 +22,7 @@
  * @see TypedValues
  * @see BooleanValue
  * @see ByteArrayValue
- * @see ByteValue
+ * @see FloatValue
  * @see DoubleValue
  * @see IntegerValue
  * @see ShortValue
```
amitjoy: Remove `ShortValue` as well

kura/org.eclipse.kura.asset.cloudlet.provider/src/main/java/org/eclipse/kura/internal/asset/cloudlet/AssetCloudlet.java
```diff
@@ -442,18 +442,15 @@ private void wrapValue(final AssetRecord assetRecord, final String userValue, fi
             if (BOOLEAN == dataType) {
                 value = TypedValues.newBooleanValue(Boolean.parseBoolean(userValue));
             }
-            if (BYTE == dataType) {
-                value = TypedValues.newByteValue(Byte.parseByte(userValue));
+            if (FLOAT == dataType) {
+                value = TypedValues.newFloatValue(Byte.parseByte(userValue));
```
amitjoy: `Float.parseFloat`

amitjoy (changes requested):
Minor Change Requests

MMaiero:
As stated in HSQLDB only DOUBLE is the result of a query. I've not found a simple way to keep floats after a query.
I think that, for the database, we must point out that in the documentation. For the rest of the components (asset, publisher, etc) everything seems fine.

MMaiero:
Closes #1295

MMaiero:
Closes #1294

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/type/FloatValue.java
```diff
@@ -8,63 +8,62 @@
...
 
     /**
      * The actual contained value that will be represented as
      * {@link TypedValue}.
      */
-    private final byte value;
+    private final float value;
 
     /**
      * Instantiates a new byte value.
```
amitjoy: new float value

amitjoy (changes requested):

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/type/FloatValue.java
MMaiero: Done

amitjoy (approved):

PR #1300, opened 2017-04-05
Fix modem driver code

ibinshtok:
Fixed ModemDriver class to 
1. Read GPIO pin direction before setting it.
2. Check if modem is OFF/ON before attempting to tor it OFF/ON.
Fixed SupportedUsbodems class to use the "lsusb" system call instead of "lsusb -d vendor:product".


ibinshtok:
@dwoodard1 please take a look at this PR and merge it if it is ok with you. I tested it on RG 10-11-36, 10-11-33 and 10-20.

Issue #253, opened 2016-06-01
Check if bundle contexes correctly unget services after invoking getService


LucaDazi:
Several OSGi Services are acquired using the bundle context.
After the services are no more needed they should be released calling ungetService(...) in order to keep the reference count of said services consistant.

cdealti:
@LucaDazi do you remember what was the conclusion?

cdealti:
@ctron can you take a look. I noticed you fixed something related in #530.

ctron:
Yes #530 is exactly that.

Beside violating the spec :wink:  … 

Looking at Equinox's implementation this can happen:

```
if (useCount == Integer.MAX_VALUE) {
   throw new ServiceException(Msg.SERVICE_USE_OVERFLOW);
}
```

Now MAX_VALUE is quite a number, but running for some time, constantly getting called, you eventually reach it.

For ServiceFactory objects, they won't get garbage collected.

You probably won't run into any memory leaks, since everything seems to be reference counted.

But then again, cleaning up afterwards and following the spec seems like the right thing to do.

cdealti:
@ctron are you willing to find and fix all the unbalanced occurrences for Kura 3.0?

ctron:
Not really.

cdealti:
@nicolatimeus can you please find all occurrences of these unbalanced calls and provide a list of the unbalanced calls for estimating the effort?

nicolatimeus:
The table below reports the the invocations to `BundleContext.getService()`.

| Call Location | Released on all normal control paths | Always released on exception |
|----------|-----------------------------|---------------------|
|[ShellScriptResourceProcessorImpl.java:72](https://github.com/eclipse/kura/blob/1a3cf8e326a22250ec87ca31cf38f65d9c229d84/kura/org.eclipse.kura.deployment.customizer/src/main/java/org/eclipse/kura/deployment/rp/sh/impl/ShellScriptResourceProcessorImpl.java#L72)|yes|yes|
|[ServiceLocator.java:112](https://github.com/eclipse/kura/blob/0339ac787f90debdfc270c1dee0c16de16ea6f7e/kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/server/util/ServiceLocator.java#L112)|yes|yes|
[...]

PR #1303, opened 2017-04-07
Added functionality to efficiently acquire and release service instances

### LINKED ISSUE #797: Design of ServiceUtil is broken
I just wanted to check out the new `ServiceUtil` class and found out that it is completely broken:

* It doesn't compile anymore on Java 8

```java
final ServiceReference<?>[] refs = bundleContext.getServiceReferences(clazz.getName(), filter);
@SuppressWarnings("unchecked")
final ServiceReference<T>[] reference = (refs == null ? new ServiceReference[0] : refs);
```

OSGi has provides proper methods for returning `Collection<ServiceReference<T>>` which should be used instead.

* The call for `get` and `unget` is completely imbalenced:

A call to `getServiceReferences` returns all service **references** while a call to `ungetServiceReferences` actually **un**gets the services. However many calls to `getServiceReferences` ignore most of the returned services but do return other instances.

So assume you have the following services and their respective counters:

* Start
A=0, B=0, C=0

* call to get -> eval, return B -> unget
A=0, B=0, C=-1
Still B is returned and actively references in the code. Although its counter is zero now.

* call to get -> unget
A=-1, B=-1, C=-2

Even worse, assume that C actually did have a reference. Then the call to get -> unget would have actually reset that reference to zero.

amitjoy:
This PR comprises :

1. `AssetService` and `DriverService` API enhancements
2. Code Refactorings
3. New functionality to acquire and release service instances in an easy-to-use way
4. Removed NO-OP methods from Wire Service implementation

Also fixes #797


amitjoy:
@markoer Could you kindly check once why the `testCreateWires(..)` test case in this PR is failing? It seems to me that it should have passed.

kura/org.eclipse.kura.util/src/main/java/org/eclipse/kura/util/service/ServiceSupplier.java
```diff
@@ -0,0 +1,138 @@
...
+        return new ServiceSupplier<>(reference);
+    }
+
+    /**
+     * Retrieves the list of target service instances
+     *
+     * @return the list of service instances or empty list
+     */
+    @Override
+    public List<T> get() {
```
ctron: Calling the get method multiple times will cause the counter to be incremented multiple times. However the close method will only release once.

So again this implementation seems unbalanced to me.

amitjoy: You are right. I have just added a check to validate if the service instance has already been retrieved from the service reference.

amitjoy:
Any update?

cdealti:
@ctron can you please review? You debated a lot about this in #797.
@amitjoy can you please take a look to the table created by @nicolatimeus at the end of #253?
Just to be all on the same page...

MMaiero:
Closing, for now, as it is too far behind develop.
We will recover it, if needed

PR #1307, opened 2017-04-07
Add Everyware Cloud broker URL

cdealti:
It seems fair to mention Everyware Cloud.
After all it's the only platform where Kura device management is implemented.


cdealti:
@kartben are you OK with this change?
Both Everyware Cloud and Eclipse IoT are mentioned.

dwoodard1 (approved):

ctron:
Adding vendor specific information seems wrong to me. I would ask you to get in contact with EMO and get confirmation about that.

waynebeaton:
The general rule is to avoid preferring any particular vendor. You can, for example, list solution providers on a project website if you have a consistent set of rules defining how an organization can be listed. It's common, for example, to provide a list of companies that use the project's technology in a product.

In this particular case, my sense is that this is a violation of vendor neutrality in that one vendor is clearly preferred over other (albeit hypothetical) competitors. I don't see how this information can be reasonably represented in a vendor neutral manner. i.e. what happens when somebody shows up with an alternative? Does it even make sense to talk about there being a consistent means of listing multiple options in this context?

Note that the board is pretty particular that when it comes to listing contributors to the project, only those companies that have a least one committer who has been active in the last three months can be displayed.

ctron:
So from pure open source project perspective it doesn't make any sense to me adding a different URL other than "iot.eclipse.org" or maybe "kapua.eclipse.org" in the future. Because those two already provide everything necessary to run Kura.

dwoodard1:
Thanks @waynebeaton for the explanation. It really is just a convenience thing to allow copy/paste of the needed URL. We would have no issue including any cloud platforms that are compatible with Kura. It is a bit more difficult with AWS, Azure, etc. as the URLs are unique per account. 

Maybe we can consider linking to a separate web page that lists available brokers for connectivity.

ctron:
I like the idea of adding a link to a web page.

PR #1312, opened 2017-04-08
Applied minor changes to db store and filter.

MMaiero:
Db store wire component now reconciles the tables at activation or update.
DB filter emits an empty message in case the query fails.


kura/org.eclipse.kura.wire.component.provider/src/main/java/org/eclipse/kura/internal/wire/filter/DbWireRecordFilter.java
```diff
@@ -254,6 +254,8 @@ public void onWireReceive(final WireEnvelope wireEnvelope) {
 
         if (nonNull(this.lastRecords)) {
             this.wireSupport.emit(this.lastRecords);
+        } else {
+            this.wireSupport.emit(new ArrayList<WireRecord>());
```
amitjoy: Emit an immutable list instead.

amitjoy:
I believe the commits can be squashed as the first commit also refers to the enhancement of emitting empty data in case DB Filter fails.

PR #1313, opened 2017-04-09
Implemented simple json option for cloud service.

MMaiero:


kura/org.eclipse.kura.core.cloud/src/main/java/org/eclipse/kura/core/cloud/CloudPayloadJsonFields.java
```diff
@@ -16,19 +16,50 @@
...
+    }
+
+    /**
+     * Returns the string representation of the constant
+     *
+     * @return the string value
+     */
+    public String value() {
+        return this.value;
     }
```
pierantoniomerlino, line 64: It would probably be better to split the two enums.

amitjoy: I believe it is more cohesive to have `CloudPayloadJsonPositionFields` inside `CloudPayloadJsonFields` as `CloudPayloadJsonPositionFields` is an integral part of metrics in `CloudPayloadJsonFields`.

PR #1319, opened 2017-04-11
Moved Asset configuration constants and documentation out of the APIs.

nicolatimeus:

### DIFF
--- a/kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/asset/AssetConfiguration.java
+++ b/kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/asset/AssetConfiguration.java
@@ -31,47 +31,6 @@ import org.osgi.annotation.versioning.ProviderType;
  * Kura).<br>
  * <br>
  *
- * The properties as provided to an Asset must conform to the following
- * specifications.<br>
- * <br>
- *
- * <ul>
- * <li>The value associated with <b><i>driver.pid</i></b> key in the map denotes
- * the driver instance PID (kura.service.pid) to be consumed by this asset</li>
- * <li>A value associated with <b><i>asset.desc</i></b> key denotes the asset
- * description</li>
- * <li>[name#property]</li> where name is a string denoting the channel's unique
- * name and the {@code [property]} denotes the protocol specific properties.
- * The name of a channel must be unique in the channels configurations of an Asset, and is not
- * allowed to contain spaces or any of the following characters: <b>#</b>, <b>_</b>.
- * </ul>
- *
- * The configuration properties of a channel belong to one of this two groups: generic channel properties and
- * driver specific properties.
- * <br>
- * Generic channel properties begin with the '+' character, and are driver independent.
- * The following generic channel properties must always be present in the channel configuration:
- * <ul>
- * <li>{@code +type} identifies the channel type (READ, WRITE or READ_WRITE) as specified by {@link ChannelType}</li>
- * <li>{@code +value.type} identifies the {@link DataType} of the channel.</li>
- * </ul>
- * For example, the property keys above for a channel named channel1 would be encoded as channel1#+type and
- * channel1#+value.type<br>
- * 
- * The values of the <b>+value.type</b> and <b>+type</b> properties must me mappable
- * respectively to a {@link DataType} and {@link ChannelType} instance.
- * <br>
- * The value of these property can be either an instance of the corresponding type,
- * or a string representation that equals the value returned by calling the {@code toString()} method
- * on one of the enum variants of that type.
- * 
- * <br>
- * Driver specific properties are defined by the driver, their keys cannot begin with a '+' character.
- * For example, valid driver specific properties can be channel1#modbus.register,
- * channel1#modbus.unit.id etc.<br>
- * <br>
- *
- *
  * @see Channel
  * @see ChannelType
  * @see DataType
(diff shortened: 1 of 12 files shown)


MMaiero:
@amitjoy Could you please give us fast feedback on this PR? I need to merge it today to wrap the release branch. In any case, bugfixes will be done in the release branch.

nicolatimeus:
@amitjoy Since the Asset configuration format was only implemented by BaseAsset, we decided to move the javadoc describing such format and AssetConstants.java to the BaseAsset package.
This will allow to make changes to the configuration format without modifying the APIs, this will also allow other implementors of the Asset interface to define a custom configuration.

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/asset/AssetConfiguration.java
```diff
@@ -31,47 +31,6 @@
  * Kura).<br>
  * <br>
  *
- * The properties as provided to an Asset must conform to the following
```
amitjoy, line 34: I like the idea of not restricting future implementors to follow the current way of using `AssetConfiguration`. Future implementors can use it in different way as well. So, this enhanced and rich javadoc shouldn't be here. But truly this is not efficient way of allowing future implementors to extend the current model. If `AssetConfiguration` is supposed to be extended, we can extract it as an interface and keep its default implementation in Asset Provider. That's the usual way that OSGi Alliance follows. But due to release constraint let's keep it this way.

kura/org.eclipse.kura.asset.provider/src/main/java/org/eclipse/kura/asset/provider/AssetConstants.java
```diff
@@ -0,0 +1,107 @@
...
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ *******************************************************************************/
+package org.eclipse.kura.asset.provider;
+
+import org.eclipse.kura.type.DataType;
+
+/**
```
amitjoy: `AssetConstants` is an integral part of the current implementation and that's why I kept it in API to make it available for all the consumers. The same idea is followed by OSGi Alliance and that's why `Constants.java` has been kept in API Bundle in `org.osgi.framework` package. In addition, extracting `AssetConstants` resulted in code duplication in multiple bundles as well. I personally think that the documentation of configuring asset channels also is not suitable in `AssetConstants`. Due to release constraint, don't put the documentation here. You can keep this documentation in `BaseAsset` and in web bundle, use this `AssetConstants` too as web bundle also imports `org.eclipse.kura.asset.provider`

The ideal way is to extract `AssetConfiguration` as an interface and keep the default implementation in `Asset Provider` and the implementation will have this javadoc. 

To wrap up, have the documentation in `BaseAsset` and keep the AssetConstants in the exported package of `Asset Provider`. Use the exported `AssetConstants` in Web bundle too. Don't duplicate the constants in web bundle again.

nicolatimeus: I agree with you on the fact that moving AssetConstants out of the APIs will increase code duplication, but the constants in this class are all related to the particular configuration format used by BaseAsset, they are not related with the Asset itself.
The only bundles that are interested in these constants are the BaseAsset bundle and the web UI bundle. 
The code duplication in the web bundle was there even before this PR because the Asset related constants were hardcoded in the PropertiesUi.java, I think this duplication cannot be avoided if we use GWT.

amitjoy: You are right and I agreed with you. `AssetConstants` are now very specific to the default `AssetConfiguration`. The reason the duplication existed because GWT Client Parser sometimes failed in parsing Enum and it happened to me sometimes and that's why a bit of duplication existed in client classes. But you can try to import `AssetConstants` from `org.eclipse.kura.asset.provider`. If it works, it would be single point of change as well and I believe this will be a better alternative.

nicolatimeus: I will try that 👍

nicolatimeus: I've tried but I wasn't able to make the GWT compile see the `org.eclipse.kura.asset.provider.AssetConstants` class. I've tried adding`source` tags in denali.gwt.xml but no luck.

amitjoy: The same issue I confronted. Then let it be. Leave it. I will approve the changes.

amitjoy (approved):

kura/org.eclipse.kura.asset.provider/src/main/java/org/eclipse/kura/asset/provider/AssetConstants.java
amitjoy: Just one more thing, I believe it is better to have the javadoc in `BaseAsset` instead of `AssetConstant`

amitjoy (changes requested):
Could you kindly add the javadoc to BaseAsset instead of AssetConstants?

amitjoy (approved):

Issue #1321, opened 2017-04-12
64bit Linux version broken


ctron:
To me it looks like as is Kura is currently broken on all current 64bit Linux versions. Including Ubuntu, Fedora and CentOS/RHEL.

This is due to the change in PR #1230 which dropped the "libudev.so.1" compatible library of the "usb" bundle.

PR #1322, opened 2017-04-12
#1321: Use libudev.so.1 capable version of 'libEurotechLinuxUdev.so'

ctron:


cdealti:
@ctron this is not going to be merged on the release-3.0.0 branch.

It looks like the apu-debian and apu-debian-nn profiles are currently broken on the release-3.0.0 branch and nobody tested the intel-edison and intel-edison-nn profiles.

We should probably remove these profiles in the future because they are basically unmaintained.

And in that case there would not be any x86_64 supported platform in Kura.

Otherwise we should go back and compile native libraries for every arch even this is not one of the Kura supported platforms.

PR #1323, opened 2017-04-12
Fixed handling of special characters in Wires DB Store

nicolatimeus:
Adds support for adding any valid UTF-8 character in db table and column names.

### DIFF
--- a/kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/client/util/KuraBaseModel.java
+++ b/kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/client/util/KuraBaseModel.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2011, 2016 Eurotech and/or its affiliates
+ * Copyright (c) 2011, 2017 Eurotech and/or its affiliates
  *
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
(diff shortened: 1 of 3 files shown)


kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/client/util/KuraBaseModel.java
```diff
@@ -20,6 +20,8 @@
      */
     private static final long serialVersionUID = -2502956403624362215L;
 
+    private transient boolean unescape = false;
```
cdealti: I'd prefer ``unsecaped``. The initializer is not needed.
Update end date in the copyright header.

kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/client/util/KuraBaseModel.java
```diff
@@ -32,4 +34,21 @@ public void set(String name, Object value) {
         super.set(name, value);
     }
 
+    public void setUnescapingEnabled(boolean unescape) {
```
cdealti: I'd prefer ``setUnescaped``

kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/client/util/KuraBaseModel.java
```diff
@@ -32,4 +34,21 @@ public void set(String name, Object value) {
...
+        this.unescape = unescape;
+    }
+
+    @SuppressWarnings("unchecked")
+    public <X> X get(String key) {
+        if (unescape) {
+            Object value = super.get(key);
+            if (value != null && value instanceof String) {
```
cdealti: null check not needed with ``instanceof``

PR #1330, opened 2017-04-18
Modified Asset Cloudlet protocol

nicolatimeus:
Modified the Asset Cloudlet protocol in order to encode the channel name list as a metric in the request payload instead of having it in the request topic.

### DIFF
--- a/kura/org.eclipse.kura.asset.cloudlet.provider/src/main/java/org/eclipse/kura/internal/asset/cloudlet/AssetCloudlet.java
+++ b/kura/org.eclipse.kura.asset.cloudlet.provider/src/main/java/org/eclipse/kura/internal/asset/cloudlet/AssetCloudlet.java
@@ -14,22 +14,18 @@
 package org.eclipse.kura.internal.asset.cloudlet;
 
 import static java.util.Objects.requireNonNull;
-import static org.eclipse.kura.type.DataType.BOOLEAN;
-import static org.eclipse.kura.type.DataType.DOUBLE;
-import static org.eclipse.kura.type.DataType.FLOAT;
-import static org.eclipse.kura.type.DataType.INTEGER;
-import static org.eclipse.kura.type.DataType.LONG;
-import static org.eclipse.kura.type.DataType.STRING;
 
+import java.util.ArrayList;
 import java.util.Arrays;
-import java.util.Collections;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
+import java.util.Map.Entry;
+import java.util.Optional;
 import java.util.Set;
+import java.util.stream.Collectors;
 
-import org.eclipse.kura.KuraException;
 import org.eclipse.kura.asset.Asset;
-import org.eclipse.kura.asset.AssetConfiguration;
 import org.eclipse.kura.asset.AssetService;
 import org.eclipse.kura.channel.Channel;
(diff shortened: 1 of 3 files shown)


kura/org.eclipse.kura.asset.cloudlet.provider/src/main/java/org/eclipse/kura/internal/asset/cloudlet/AssetCloudlet.java
```diff
@@ -204,19 +257,31 @@ protected void doGet(final CloudletTopic reqTopic, final KuraRequestPayload reqP
             this.findAssets();
             if (reqTopic.getResources().length == 1) {
                 this.getAllAssets(respPayload);
+                return;
             }
             // Checks if the name of the asset is provided
-            if (reqTopic.getResources().length == 2) {
-                final String assetPid = reqTopic.getResources()[1];
-                this.getAllChannelsByAssetPid(respPayload, assetPid);
+            if (reqTopic.getResources().length != 2) {
```
amitjoy: You can extract `reqTopic.getResources()` as a variable so that you don't need to call it again thereafter.

kura/org.eclipse.kura.asset.cloudlet.provider/src/main/java/org/eclipse/kura/internal/asset/cloudlet/AssetCloudlet.java
```diff
@@ -58,27 +52,62 @@
  * asset. The application id is configured as {@code ASSET-V1}.
  *
  * The available {@code GET} commands are as follows
+ * <b>/assets</b><br>
+ * Retrieves the list of all assets assets present on this device.
```
MMaiero: only one assets

Maybe we can use the "available" term instead of "present"

kura/org.eclipse.kura.asset.cloudlet.provider/src/main/java/org/eclipse/kura/internal/asset/cloudlet/AssetCloudlet.java
```diff
@@ -58,27 +52,62 @@
...
+ * <b>/assets/${asset_name}</b><br>
+ * Can be used to perform the following operations on the Asset named ${asset_name},
+ * depending on the value of the <b>channel.names</b> metric in the payload:
```
MMaiero: `channel.names` should be replaced by `channel.name`

kura/org.eclipse.kura.asset.cloudlet.provider/src/main/java/org/eclipse/kura/internal/asset/cloudlet/AssetCloudlet.java
```diff
@@ -58,27 +52,62 @@
...
+ * <li>
+ * If a metric named <b>channel.name</b> of String type is present in the payload and its value is a
+ * comma separated list of channel names, the values for the specified channels will be returned.
+ * </li>
  * </ul>
+ * <br>
+ * 
+ * The available {@code PUT} commands are the following:<br>
+ * <b>/assets/asset_pid</b></br>
+ * Performs a write write operation on a specific set of channels.
```
MMaiero: only one write needed

Issue #1333, opened 2017-04-19
[Kura 3.0.0 QA] Wire DB store component maximum size


markoer:
Cleanup period was replaced with maximum table size (capacity) since 3.0.0-M1 and QA spreadsheet was not updated for this :(

The issue now is that the way cleanup is implemented, it will always allow 2 rows more in the table:
* the check happens when wire is received, but before it is stored,
* it checks > instead of >= max.

And you can probably never be sure whether the wire that triggered the cleanup will be inserted before the cleanup or after.

Log where this is clearly visible (maximum.table.size = 10):
```
2017-04-19 11:29:09,793 [DefaultQuartzScheduler_Worker-8] INFO  o.e.k.i.w.l.Logger - Received WireEnvelope from org.eclipse.kura.wire.DbWireRecordFilter-1492601127796-9
2017-04-19 11:29:09,794 [DefaultQuartzScheduler_Worker-8] INFO  o.e.k.i.w.l.Logger - Record List content: 
2017-04-19 11:29:09,795 [DefaultQuartzScheduler_Worker-8] INFO  o.e.k.i.w.l.Logger -   Record content: 
2017-04-19 11:29:09,796 [DefaultQuartzScheduler_Worker-8] INFO  o.e.k.i.w.l.Logger -     TIMESTAMP : 1492601349508
2017-04-19 11:29:09,796 [DefaultQuartzScheduler_Worker-8] INFO  o.e.k.i.w.l.Logger -     TIMER : 1492601349505
2017-04-19 11:29:09,797 [DefaultQuartzScheduler_Worker-8] INFO  o.e.k.i.w.l.Logger -   Record content: 
2017-04-19 11:29:09,797 [DefaultQuartzScheduler_Worker-8] INFO  o.e.k.i.w.l.Logger -     TIMESTAMP : 1492601345508
2017-04-19 11:29:09,798 [DefaultQuartzScheduler_Worker-8] INFO  o.e.k.i.w.l.Logger -     TIMER : 1492601345505
2017-04-19 11:29:09,798 [DefaultQuartzScheduler_Worker-8] INFO  o.e.k.i.w.l.Logger -   Record content: 
2017-04-19 11:29:09,799 [DefaultQuartzScheduler_Worker-8] INFO  o.e.k.i.w.l.Logger -     TIMESTAMP : 1492601341507
[...]

PR #1345, opened 2017-04-24
Modified wires db size check and cleanup.

MMaiero:


MMaiero:
@markoer It should solve the issue #1333. Could you give it a try?

amitjoy (approved):

PR #1347, opened 2017-04-24
Modified emulator position.

MMaiero:
Now it implements configurablecomponent interface. This should solve the exceptions in
dev-env.


MMaiero:
@dwoodard1 Uhm, I don't think I have introduced new external dependencies with this PR. Why is it complaining? In particular the tool refers to org.osgi:org.osgi.service.component.annotations 1.3.0.

cdealti:
The dependency check fails due to examples/bnd_examples/org.eclipse.kura.demo.bnd.heater.
@dwoodard1 any ideas why it's coming out now?

PR #1356, opened 2017-05-03
Fixed wires DBStore cleanup issue

nicolatimeus:

### DIFF
--- a/kura/org.eclipse.kura.wire.component.provider/src/main/java/org/eclipse/kura/internal/wire/store/DbWireRecordStore.java
+++ b/kura/org.eclipse.kura.wire.component.provider/src/main/java/org/eclipse/kura/internal/wire/store/DbWireRecordStore.java
@@ -224,8 +224,8 @@ public final class DbWireRecordStore implements WireEmitter, WireReceiver, Confi
                     this.dbHelper.execute(MessageFormat.format(SQL_TRUNCATE_TABLE, sqlTableName));
                 } else {
                     logger.info(message.partiallyEmptyingTable(sqlTableName));
-                    this.dbHelper
-                            .execute(MessageFormat.format(SQL_DELETE_RANGE_TABLE, sqlTableName, noOfRecordsToKeep));
+                    this.dbHelper.execute(MessageFormat.format(SQL_DELETE_RANGE_TABLE, sqlTableName,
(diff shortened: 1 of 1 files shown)


kura/org.eclipse.kura.wire.component.provider/src/main/java/org/eclipse/kura/internal/wire/store/DbWireRecordStore.java
```diff
@@ -224,8 +224,8 @@ private void truncate(final int noOfRecordsToKeep) {
                     this.dbHelper.execute(MessageFormat.format(SQL_TRUNCATE_TABLE, sqlTableName));
                 } else {
                     logger.info(message.partiallyEmptyingTable(sqlTableName));
-                    this.dbHelper
-                            .execute(MessageFormat.format(SQL_DELETE_RANGE_TABLE, sqlTableName, noOfRecordsToKeep));
+                    this.dbHelper.execute(MessageFormat.format(SQL_DELETE_RANGE_TABLE, sqlTableName,
+                            Integer.toString(noOfRecordsToKeep)));
```
amitjoy, line 228: Could you kindly elaborate what the issue is?

nicolatimeus: Basically `MessageFormat.format()` by default prints integers with commas as thousands separators, for example running`MessageFormat.format("{0}", 20000)` will return the string `20,000`.

So in this case if `noOfRecordsToKeep` is greater or equal than 1000 an illegal comma will be placed in the query and it won't compile. This causes the DB store cleanup to fail.

amitjoy: You could also do the following:

```java
private static final String SQL_DELETE_RANGE_TABLE = "DELETE FROM {0} WHERE rownum() <= (SELECT count(*) FROM {0}) - {1,number,#};";
```

This way you don't have to stringify it again and this is also suggested by message formatter.

cdealti:
@amitjoy your suggestion was the original implementation by @nicolatimeus. We changed this because to be consistent with other .toString() conversions in other queries. I need to merge this as is. It can be improved after the release

PR #1368, opened 2017-05-09
Wires Web Ui fixes

nicolatimeus:
Added support for triggering the wait modal to prevent the user from interacting with the graph when a RPC request is pending.
Removed unnecessary RPC call.

Addresses the following issues:

* Random XSRF token related exceptions showing up when editing a wire graph.
* The channels table for an asset is occasionally rendered as empty even if some channels are defined.


amitjoy (approved):
At least few RPC calls are no more 😉

PR #1381, opened 2017-05-19
[Testing] Added wires unit tests

markoer:
Added unit tests for wire components:
* CloudPublisher
* CloudSubscriber
* RegexFilter

### DIFF
--- a/kura/test/org.eclipse.kura.wire.component.provider.test/src/test/java/org/eclipse/kura/internal/wire/publisher/CloudPublisherTest.java
+++ b/kura/test/org.eclipse.kura.wire.component.provider.test/src/test/java/org/eclipse/kura/internal/wire/publisher/CloudPublisherTest.java
@@ -1,0 +1,214 @@
+/*******************************************************************************
+ * Copyright (c) 2017 Eurotech and/or its affiliates and others
+ *
+ *   All rights reserved. This program and the accompanying materials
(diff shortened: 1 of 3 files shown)


kura/test/org.eclipse.kura.wire.component.provider.test/src/test/java/org/eclipse/kura/internal/wire/publisher/CloudPublisherTest.java
```diff
@@ -0,0 +1,214 @@
...
+    @Test
+    public void testActivateFilterException() throws InvalidSyntaxException {
+        // test activation where exception during filter creation - tests null filter behavior
+        CloudPublisher cp = new CloudPublisher();
+
+        WireHelperService wireHelperServiceMock = mock(WireHelperService.class);
+        cp.bindWireHelperService(wireHelperServiceMock);
+
+        BundleContext bundleCtxMock = mock(BundleContext.class);
+        Filter filter = mock(Filter.class);
```
MMaiero: Does filter needs to be cleaned up, since it is declared but never used?

markoer: Done.

Issue #1384, opened 2017-05-23
OPC UA Driver channel descriptor property ids shouldn't be localised


nicolatimeus:
Localised property ids do not allow channel configuration to be ported across different locales.

MMaiero:
Closing as #1385 was merged

Issue #1383, opened 2017-05-23
OPC UA Driver only supports string node ids


nicolatimeus:
According to the OPC UA specification node ids can have different types (numeric, string, guid..). The OPC UA Driver only supports string node ids for now, support for more types should be added.

MMaiero:
Closing as #1385 was merged

PR #1385, opened 2017-05-23
Added support for more node id types to OPC UA driver

nicolatimeus:
Adds support for NUMERIC, GUID and OPAQUE node id types.
Also removes localisation from OPC UA channel descriptor ids.

Addresses #1384 and #1383


kura/org.eclipse.kura.driver.opcua.provider/src/main/java/org/eclipse/kura/internal/driver/opcua/OpcUaChannelDescriptor.java
```diff
@@ -33,33 +38,86 @@
...
         return elements;
     }
 
+    public static int getNodeNamespaceIndex(Map<String, Object> properties) {
+        return Integer.parseInt((String) properties.get(NODE_NAMESPACE_INDEX));
+    }
+
+    public static NodeIdType getNodeIdType(Map<String, Object> properties) {
+        String nodeIdType = (String) properties.get(NODE_ID_TYPE);
+        if (NodeIdType.NUMERIC.name().equals(nodeIdType)) {
```
amitjoy, line 99: You can use switch on strings for less verbosity.

kura/org.eclipse.kura.driver.opcua.provider/src/main/java/org/eclipse/kura/internal/driver/opcua/OpcUaChannelDescriptor.java
```diff
@@ -33,33 +38,86 @@
...
+            return NodeIdType.GUID;
+        } else if (NodeIdType.OPAQUE.name().equals(nodeIdType)) {
+            return NodeIdType.OPAQUE;
+        }
+        throw new IllegalArgumentException();
+    }
+
+    public static NodeId getNodeId(Map<String, Object> properties, int nodeNamespaceIndex, NodeIdType nodeIdType) {
+        String nodeIdString = (String) properties.get(NODE_ID);
+        if (nodeIdType == NodeIdType.NUMERIC) {
```
amitjoy: For less verbosity, you can use switch on NodeIdType

kura/org.eclipse.kura.driver.opcua.provider/src/main/java/org/eclipse/kura/internal/driver/opcua/OpcUaChannelDescriptor.java
nicolatimeus: That's not possible, because `NodeIdType.name()` is not considered to be a constant expression by the compiler

PR #1423, opened 2017-05-31
Allow launching BND Heater from User Workspace

cdealti:

### DIFF
--- a/kura/examples/bnd_examples/org.eclipse.kura.demo.bnd.heater/src/main/java/org/eclipse/kura/demo/bnd/heater/HeaterConfig.java
+++ b/kura/examples/bnd_examples/org.eclipse.kura.demo.bnd.heater/src/main/java/org/eclipse/kura/demo/bnd/heater/HeaterConfig.java
@@ -29,7 +29,7 @@ import org.osgi.service.metatype.annotations.*;
 @interface HeaterConfig {
 
     @AttributeDefinition(
-            name = "mode",
+            name = "Mode",
             type = AttributeType.STRING,
             defaultValue = "Program",
             description = "Operating mode for the heater. If operatng mode is Vacation, set point is automatiaclly set to 6.0C.",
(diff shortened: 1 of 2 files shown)


cdealti:
@dwoodard1 you should be able to import this as a Maven project after ``mvn clean install`` and launch it from the Kura User Workspace. Can you please check if it works for you?

kura/examples/bnd_examples/org.eclipse.kura.demo.bnd.heater/src/main/java/org/eclipse/kura/demo/bnd/heater/HeaterConfig.java
```diff
@@ -42,7 +42,7 @@
     String mode();
 
     @AttributeDefinition(
-            name = "program.startTime",
```
dwoodard1, line 45: Why did you change this (and all AD names)? The plugin will convert "Program start time" to "program.startTime".

cdealti: Look at the configuration in kura web ui

dwoodard1: Sorry, you are correct.

dwoodard1:
@cdealti In general it looks good. Two possible issues:

1. The icon doesn't resolve correctly. A simple solution would be adding a soft link in the project root directory to src/main/resources/heater.png.
2. The generated metatype file and component definition XML show up in the workspace. This may lead people to edit those files instead of the Java files, which defeats the purpose of the example.

Don't think there is anything we can do about 2 and not sure 1 is very important. If you are okay with these two issues, I can just merge.

cdealti:
I know, it's an hack.

How would you turn this in a standalone Tycho build without a p2 site? This leads us to the inevitable question...

Another problem of the BND project is that its dependencies do not match those of any Kura runtime (versions are different). It's dependencies are actually **provided** by the Kura runtime and may differ from the compile-time dependencies. How one knows what's in the runtime? The Kura API bundle is not enough; in another like example I had to add more dependencies.

PR #1402, opened 2017-05-29
[Testing][Refactoring]HostapdConfig writer and reader updates

markoer:
* added tests
* updated HostapdConfig reader and writer with possible new file locations
* refactored the HostapdConfigWriter and HostapdConfigReader

PR #1425, opened 2017-06-01
[Testing][Refactoring]IfcfgConfigReader/Writer refactoring and tests

markoer:
Refactored classes and added unit tests.

Depends on #1402.

@ibinshtok please review

### DIFF
--- a/kura/org.eclipse.kura.net.admin/src/main/java/org/eclipse/kura/net/admin/visitor/linux/IfcfgConfigReader.java
+++ b/kura/org.eclipse.kura.net.admin/src/main/java/org/eclipse/kura/net/admin/visitor/linux/IfcfgConfigReader.java
@@ -52,14 +52,14 @@ public class IfcfgConfigReader implements NetworkConfigurationVisitor {
 
     private static String OS_VERSION = System.getProperty("kura.os.version");
 
-    private static IfcfgConfigReader s_instance;
+    private static IfcfgConfigReader instance;
 
     public static IfcfgConfigReader getInstance() {
-        if (s_instance == null) {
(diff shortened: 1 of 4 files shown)


kura/org.eclipse.kura.net.admin/src/main/java/org/eclipse/kura/net/admin/visitor/linux/IfcfgConfigReader.java
```diff
@@ -413,10 +426,6 @@ else if (args[0].equals("iface") && args[1].equals(interfaceName)) {
 
         } catch (FileNotFoundException err) {
             throw new KuraException(KuraErrorCode.INTERNAL_ERROR, err);
-        } finally {
```
ibinshtok, line 416: @markoer Do we need to close the scanner?

ibinshtok (changes requested):
@markoer I think scanner needs to be closed, otherwise everything is ok.

kura/org.eclipse.kura.net.admin/src/main/java/org/eclipse/kura/net/admin/visitor/linux/IfcfgConfigReader.java
amitjoy: it's a good practice to close the scanner to avoid resource leak

markoer: `scanner` is declared as a resource in a try-with-resources that automatically closes it at the end of the statement.

ibinshtok (approved):

PR #1437, opened 2017-06-07
Added support for aarch64 arch

feclare:
Added new octeontx target, using aarch64(arm 64 bits)
Added aarch64 native libraries for:
org.eclipse.kura.linux.usb
com.codeminders.hidapi
jdk-dio
org.eclipse.soda.dk.comm

Updated usb version in kura.build.properties
Updated hidapi, jdk-dio, soda versions in kura.target-platform.build.properties

Signed-off-by: Federico Claramonte <fede.claramonte@caviumnetworks.com>


dwoodard1:
@feclare Thank you very much for the contribution. One important part of the targets we support is that we perform QA on the targets before each release. This allows us to catch potential hardware/OS specific issues inadvertently introduced by the development cycle. Would you be willing to perform this QA for future releases? We perform releases 3-4 times per year.

feclare:
Hi @dwoodard1, we can make the QA for each release on the octeontx. Is there a document describing the procedure?

feclare:
Created  #1461 with conflicts resolved.

PR #1439, opened 2017-06-07
Build updates

### LINKED ISSUE #1040: Clean up or remove 'pom_pom.xml'
Looking at the `pom_pom.xml` file it looks to me as if the only two modules which are still using this are `distrib` and `tools`. Digging deeper into the structure looks like that this could me massively improved. Which could also help to clean up the mess of two POM files in a single directory, renaming `manifest_pom.xml` back to `pom.xml`.

The `pom_pom.xml` could either be removed and only necessary section be ported over to either the tools or the `distrib` project or we could create a new parent module for both of them. But looking into the way the `distrib` module is currently being activated, this looks like all the profiles will always result in an activation of the `distrib` module and thus making the different profiles unnecessary.

MMaiero:
This PR closes #1040 merging manifest_pom and pom_pom.
Applied changes to APU profiles to match the latest changes applied to the other existing profiles.


MMaiero:
The **web** profile is now removed so, by default, the web bundle is built and placed into installers.
Two system properties are now defined:
- kura.skip.web (default false)
- kura.skip.can (default true)

These properties are valid in both the old manifest_pom and in the distrib pom and allow to specify if the web and/or can bundle have to be included into the distribution.

Brief set of changes:

- Made distrib pom a root pom.

- Defined two system properties, kura.skip.can and kura.skip.web, to deal with web and can bundle skipping.

- Made example pom as root pom.

- Now the tycho build points to the org.eclipse.kura-p2 repo.

- Updated build-all script with new invocations for new root poms.

cdealti:
Please rebase & update the PR as the build has been fixed by #1441

kura/features/pom.xml
```diff
@@ -8,7 +8,7 @@
 		<groupId>org.eclipse.kura</groupId>
 		<artifactId>kura</artifactId>
 		<version>3.1.0-SNAPSHOT</version>
-		<relativePath>../manifest_pom.xml</relativePath>
+		<relativePath>..</relativePath>
```
markoer: You can remove &lt;relativePath> because parent pom is now in the correct location. Here and in other pom.xml files.

pierantoniomerlino (approved):

pierantoniomerlino (approved):

MMaiero:
When merged, a change to the following page is needed: https://wiki.eclipse.org/Kura/Getting_Started
In the same page, the reference to the maven version should be updated (we currently require maven 3.3+)

MMaiero:
Going to change hudson and the documentation now

PR #1441, opened 2017-06-08
[Testing][Refactoring] WpaSupplicantConfigWriter

markoer:
* Added additional tests
* Refactored based on #1405 

@ibinshtok please review.

PR #1445, opened 2017-06-12
Added protection code to prevent null pointer in GwtWireServiceUtil.

MMaiero:
That null could cause the compose to not render.


cdealti:
Please rebase & update the PR as the build has been fixed by #1441

cdealti:
@MMaiero I believe we don't need to submit a PR if we only add an NPE check.

PR #1449, opened 2017-06-12
Fix net.admin tests on Travis

cdealti:
net.admin tests are failing on Travis due to missing dos2unix.
Install the dos2unix package.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,6 +18,11 @@ before_script:
   - export PATH=$PWD/maven/bin:${PATH}
   - hash -r
 
+addons:
+  apt:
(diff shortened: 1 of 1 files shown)


markoer:
The build is not failing because of the missing dos2unix binary. It's only that the log is full of printout because of that. It's failing because one of the necessary directories/files is created too late.

cdealti:
I see. Should I delete this PR?

markoer:
If dos2unix is present, it's even better. There will be less error logging. Please leave it. I just explained that missing dos2unix was not the cause.

Issue #1451, opened 2017-06-13
PppAuthSecrets entry removal


markoer:
`org.eclipse.kura.net.admin.visitor.linux.util.PppAuthSecrets.removeEntry(String type, String value)` skips an entry each time it deletes one. Therefore it doesn't remove all matching entries, unless they don't follow each other.

Kura 3.1.0-SNAPSHOT, develop branch

cdealti:
Fixed

PR #1452, opened 2017-06-13
[Testing][Refactoring] LinuxFileUtil and PppAuthSecrets

markoer:
Added tests and refactored:
* replaced symlink creation code in LinuxFileUtil with java code
* added PppAuthSecrets tests and fix for #1451

### DIFF
--- a/kura/org.eclipse.kura.net.admin/src/main/java/org/eclipse/kura/net/admin/util/LinuxFileUtil.java
+++ b/kura/org.eclipse.kura.net.admin/src/main/java/org/eclipse/kura/net/admin/util/LinuxFileUtil.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2011, 2016 Eurotech and/or its affiliates
+ * Copyright (c) 2011, 2017 Eurotech and/or its affiliates
  *
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
@@ -12,8 +12,9 @@
 package org.eclipse.kura.net.admin.util;
 
 import java.io.File;
+import java.nio.file.Files;
+import java.nio.file.Paths;
 
-import org.eclipse.kura.core.linux.util.LinuxProcessUtil;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -37,24 +38,13 @@ public class LinuxFileUtil {
         s_logger.debug("Creating symbolic link from " + targetFile + " to " + sourceFile);
 
         File f = new File(targetFile);
-        if (f.exists()) {
-            if (deleteOldLink) {
-                s_logger.debug("Deleting existing link");
-                f.delete();
-            }
+        if (f.exists() && deleteOldLink) {
+            s_logger.debug("Deleting existing link");
+            f.delete();
         }
 
         if (!f.exists()) {
-            StringBuffer buf = new StringBuffer("ln -s");
-            buf.append(' ');
-            buf.append(sourceFile);
-            buf.append(' ');
-            buf.append(targetFile);
-
-            int ret = LinuxProcessUtil.start(buf.toString());
-            if (ret != 0) {
-                throw new Exception("error executing command - " + buf.toString());
-            }
+            Files.createSymbolicLink(Paths.get(targetFile), Paths.get(sourceFile));
         }
     }
 }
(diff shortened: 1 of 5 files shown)


kura/org.eclipse.kura.net.admin/src/main/java/org/eclipse/kura/net/admin/util/LinuxFileUtil.java
```diff
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2011, 2016 Eurotech and/or its affiliates
+ * Copyright (c) 2011, 2017 Eurotech and/or its affiliates
```
cdealti, line 2: it looks to me that, using _nio_ this util class is no longer useful

kura/org.eclipse.kura.net.admin/src/main/java/org/eclipse/kura/net/admin/visitor/linux/util/PppAuthSecrets.java
```diff
@@ -234,51 +231,36 @@ public void removeEntry(String type, String value) throws Exception {
         if (type.compareTo("provider") == 0) {
```
cdealti: can you please reverse the comparison ``"a string constant".equals(anotherString)``

cdealti (changes requested):

kura/org.eclipse.kura.net.admin/src/main/java/org/eclipse/kura/net/admin/util/LinuxFileUtil.java
markoer: Well, the method is still good to have. I believe it may as well remain where it is. But if you wish, I can move it.

kura/test/org.eclipse.kura.net.admin.test/src/test/java/org/eclipse/kura/net/admin/util/LinuxFileUtilTest.java
```diff
@@ -0,0 +1,101 @@
...
+        LinuxFileUtil.createSymbolicLink(sourceFile, targetFile, false);
+
+        assertTrue(tgt.exists());
+
+        f.delete();
+        tgt.delete();
+    }
+
+    private boolean isWin() {
+        return System.getProperty("os.name").contains("indows");
```
amitjoy, line 49: even though this would definitely work, it would be good to check containment of `windows` instead

markoer: This accounts for both Windows and windows. If you insist on `windows`, I can refactor.

amitjoy: I was not aware of that. Leave it as it is.

ctron:
Wouldn't it be wise to use NIO2 instead of custom stuff?

markoer:
@ctron What exactly are you referring to?

kura/org.eclipse.kura.net.admin/src/main/java/org/eclipse/kura/net/admin/visitor/linux/util/PppAuthSecrets.java
```diff
@@ -97,7 +98,7 @@ public PppAuthSecrets(String secretsFilename) {
                 }
             } else {
                 // Create an empty file
-                s_logger.info("File does not exist - creating " + this.m_secretsFilename);
+                s_logger.info("File does not exist - creating " + this.secretsFilename);
```
MMaiero: Please change it as: logger.info("File does not exist - creating {}", this.secretsFilename);

kura/org.eclipse.kura.net.admin/src/main/java/org/eclipse/kura/net/admin/visitor/linux/util/PppAuthSecrets.java
```diff
@@ -231,63 +229,32 @@ private void writeToFile() throws Exception {
...
-            for (int i = 0; i < this.ipAddresses.size(); ++i) {
-                if (this.ipAddresses.get(i).compareTo(value) == 0) {
-                    try {
-                        this.removeEntry(i);
-                    } catch (Exception e) {
-                        throw e;
-                    }
-                }
-            }
+        if ("provider".equals(type)) {
```
MMaiero, line 232: I've not done a search on this, but do you think it would be worth moving those strings to an enum? Maybe we can reduce duplication and error possibilities.

markoer: These strings are not really used that much. So I extracted `provider` to a constant in the one class using it.

PR #1454, opened 2017-06-14
Cleanup org eclipse kura linux net wifi

ibinshtok:


ibinshtok:
@markoer please review this PR

kura/org.eclipse.kura.linux.net/src/main/java/org/eclipse/kura/linux/net/util/LinuxNetworkUtil.java
```diff
@@ -1547,36 +1578,20 @@ public static boolean isKernelModuleLoadedForMode(String interfaceName, WifiMode
         // Assume for now the interface name does not change.
         if (KuraConstants.ReliaGATE_10_05.getTargetName().equals(TARGET_NAME) && "wlan0".equals(interfaceName)) {
             SafeProcess proc = null;
-            BufferedReader br = null;
             String cmd = "systool -vm bcmdhd";
             try {
-                s_logger.debug("Executing '{}'", cmd);
+                s_logger.debug(EXECUTING_CMD_MSG, cmd);
                 proc = ProcessUtil.exec(cmd);
```
markoer, line 1584: It's already a nice refactoring job. Now there's a huge number of methods in this util that follow this pattern:
1. get command
1. execute command
1. parse the output and catch IOException
1. handle a few more exceptions
1. destroy the process

And the command is always (?) one producing only a few lines of output.

It might not be a bad idea to refactor it a bit further:
1. get command
1. do
   1. execute command
   1. store output in e.g. List&lt;String>
   1. handle exceptions
   1. destroy process
   1. return output
1. parse the output i.e. List&lt;String>

This way you could only test process execution once, to see it works, and test output parsing separately, with no reference to the process.

kura/org.eclipse.kura.linux.net/src/main/java/org/eclipse/kura/linux/net/util/iwScanTool.java
```diff
@@ -33,60 +33,58 @@
 
 public class iwScanTool extends ScanTool implements IScanTool {
 
-    private static final Logger s_logger = LoggerFactory.getLogger(iwScanTool.class);
+    private static final Logger logger = LoggerFactory.getLogger(iwScanTool.class);
     private static final String SCAN_THREAD_NAME = "iwScanThread";
+    private static final String FAILED_EXECUTE_SCAN_CMD_MSG = "failed to execute scan command ";
     private static final Object s_lock = new Object();
```
markoer, line 39: One more field to rename...

markoer (approved):
Refactoring is done carefully enough. From what I see these changes don't diminish the possibilities for testing, but rather make it easier. A few changes will still be necessary for writing nicer tests, but they can be added later on.

kura/org.eclipse.kura.linux.net/src/main/java/org/eclipse/kura/linux/net/util/KuraConstants.java
```diff
@@ -27,26 +27,26 @@
     Reliagate_20_26("rhel", "7.3", "reliagate-20-26"),
```
MMaiero, line 27: Just a brief question: do we still need to track the image version?

ibinshtok: @MMaiero I would keep it for now.

PR #1460, opened 2017-06-19
[Testing][Refactoring] NetworkAdminServiceImpl tests, refactoring, fix

markoer:
* added tests
* renamed fields
* fixed timeout that was shortened by half due to a wrong type

### DIFF
--- a/kura/org.eclipse.kura.net.admin/src/main/java/org/eclipse/kura/net/admin/NetworkAdminServiceImpl.java
+++ b/kura/org.eclipse.kura.net.admin/src/main/java/org/eclipse/kura/net/admin/NetworkAdminServiceImpl.java
@@ -1035,7 +1035,7 @@ public class NetworkAdminServiceImpl implements NetworkAdminService, EventHandle
     @Override
     public void disableInterface(String interfaceName) throws KuraException {
 
-        if (!interfaceName.equals("lo")) {
+        if (!"lo".equals(interfaceName)) {
             try {
                 if (LinuxNetworkUtil.hasAddress(interfaceName)) {
                     logger.info("bringing interface {} down", interfaceName);
@@ -1064,19 +1064,6 @@ public class NetworkAdminServiceImpl implements NetworkAdminService, EventHandle
     @Override
     public void manageDhcpClient(String interfaceName, boolean enable) throws KuraException {
         try {
-            /*
-             * int pid = LinuxProcessUtil.getPid(formDhclientCommand(interfaceName, false));
-             * if (pid > -1) {
-             * s_logger.debug("manageDhcpClient() :: killing {}", formDhclientCommand(interfaceName, false));
-             * LinuxProcessUtil.kill(pid);
-             * } else {
-             * pid = LinuxProcessUtil.getPid(formDhclientCommand(interfaceName, true));
-             * if (pid > -1) {
-             * s_logger.debug("manageDhcpClient() :: killing {}", formDhclientCommand(interfaceName, true));
-             * LinuxProcessUtil.kill(pid);
-             * }
-             * }
-             */
             DhcpClientManager.disable(interfaceName);
             if (enable) {
                 renewDhcpLease(interfaceName);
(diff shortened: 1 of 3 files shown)


markoer:
It might be better to merge #1454 before this one.

kura/org.eclipse.kura.net.admin/src/main/java/org/eclipse/kura/net/admin/NetworkAdminServiceImpl.java
```diff
@@ -1565,22 +1565,22 @@ private void submitNetworkConfiguration(List<String> modifiedInterfaceNames,
...
-            this.m_pendingNetworkConfigurationChange = false;
+        if (this.pendingNetworkConfigurationChange) {
+            logger.warn("Did not receive a network configuration change event");
+            this.pendingNetworkConfigurationChange = false;
         }
     }
 
     private void submitFirewallConfiguration() throws KuraException {
         // TODO
         short timeout = 30;		// in seconds
```
MMaiero: Shouldn't also this need to be set to float?

kura/org.eclipse.kura.net.admin/src/main/java/org/eclipse/kura/net/admin/NetworkAdminServiceImpl.java
```diff
@@ -1506,57 +1506,57 @@ private void enableWifiInterface(String ifaceName, NetInterfaceStatus status, Wi
...
+        logger.debug("Stopping hostapd and wpa_supplicant");
         HostapdManager.stop(ifaceName);
         WpaSupplicantManager.stop(ifaceName);
     }
 
     // Submit new configuration, waiting for network configuration change event before returning
     private void submitNetworkConfiguration(List<String> modifiedInterfaceNames,
             NetworkConfiguration networkConfiguration) throws KuraException {
-        short timeout = 30;		// in seconds
+        double timeout = 30;		// in seconds
```
MMaiero: float?

MMaiero: @cdealti I believe that this change modifies the code behaviour. Probably a deeper QA process should be done here.

ibinshtok: @cdealti @MMaiero I think the idea here was to have a timeout of 30 seconds but since timeout was declared as short and Thread.sleep was set to 500 msec we ended up having a timeout of 15 seconds since decrementing a short by 0.5 decrements it by 1. So I think it would be ok to declare timeout as float.

kura/org.eclipse.kura.net.admin/src/main/java/org/eclipse/kura/net/admin/NetworkAdminServiceImpl.java
ibinshtok: @MMaiero @cdealti yes, I think so. I will make a change and test.

ibinshtok:
@cdealti I rebased this branch to resolve conflicts and got to this point. If I now merge this PR, it will merge 68 commits. Is it ok? I think the reason ip-validation test has failed is because I did final 'git push' as was suggested by git instead of 'git push -s'.

markoer:
I believe it will be easier if I create a new up-to-date branch and apply my changes there. When that's done, I'll close this PR and prepare a new one.

Edit: I managed to rebase this one.

kura/org.eclipse.kura.net.admin/src/main/java/org/eclipse/kura/net/admin/NetworkAdminServiceImpl.java
markoer: Exactly as @ibinshtok explained. I now have a third solution I believe you will like best: `timeout` remains `short`, but starts at 30000 and decrements by 500. This way the code is more intuitive.

MMaiero:
Manually squashed and merged.

PR #1461, opened 2017-06-20
Added support for aarch64 arch

feclare:
Added new octeontx target, using aarch64(arm 64 bits)
    Added aarch64 native libraries for:
    org.eclipse.kura.linux.usb
    com.codeminders.hidapi
    jdk-dio
    org.eclipse.soda.dk.comm

    Updated usb version in kura.build.properties
    Updated hidapi, jdk-dio, soda versions in kura.target-platform.build.properties

    Signed-off-by: Federico Claramonte <fede.claramonte@caviumnetworks.com>


MMaiero:
@feclare Thanks again for your effort and your contribution.
I would ask if you could simply rebase your branch, instead of opening and closing PRs. This will make the process of analysis and all the discussion more consistent and complete.

Thanks.
Best regards,
Matteo

feclare:
Hi @MMaiero,
I was just trying to follow the rule of only one commit for each pull request.  Is it allowed to make more commits to fix issues in an ongoing pull request?

Thanks for the help and patience with the pull request.
Regards,
Fede

MMaiero:
Yes, the rule to have a single commit is correct. But you can get the same result by squashing two commits into one.

Best regards,
Matteo

feclare:
But, can I squash commits in a public branch? I thought it was not
recommended, I was squashing in private branch before pushing to a new
branch in github.

MMaiero:
It's from your fork, so you can do whatever you want there.

ctron:
I do think that squashing could be done right before the merge. That way discussions actually do have the full history of changes during the discussion on the PR. And still there will only be one commit when this is merged in the main branch.

MMaiero:
Actually, squashing immediately or right before merging is, in any case, ok. What is important is to keep the PR alive in order to track the ongoing discussion and not having it sparsed through many PRs.

feclare:
I think I understand now. It was really strange to create all those pull request for the same issue.

Will try do it right for the next time.
Thanks again for helping me.

kartben:
Looks like the PR is good to go? Will someone from the Kura team open a CQ for it? 
This looks like a great contribution so I am hoping this will make it into Kura sooner than later 😄

cdealti:
I'd like to establish some basic requirements that should be met when adding a new supported platform to Kura. Every new platform will cause a significant overhead in terms of maintenance, release QA and support.

For this particular pull request these are my concerns:

1. The target itself is a SoC and not a board like the RPI or the Beaglebone
2. How to obtain/purchase an Octeontx-based board like the RPI or BBB?
3. How to obtain the OS?
4. There is no active community using it, so no general interest

I'm not saying I'm not interested in seeing 64 bit support for ARM. If this PR was for an 64 bit Raspian I'd be less concerned.

This is my opinion. Of course others may disagree on that

feclare:
Some comments:
1. There is nothing really specific to the octeontx board or SOC. All the work was just compile the native libs for arm64. The name could be changed to arm64, aarch64, generic_arm64 or something like that if it make it easier to merge.
2. Any arm64 board should work, but we couldn't test as we don't have any other arm64 board. Sadly there is no raspian 64bits yet.
3. We used ubuntu 17.04 in our tests. Also tested with montavista linux.

ctron:
So from what I understood this should be targeted towards the aarch64 architecture. Not a specific board. So it should be possible to use other arm64 boards/SoCs/platforms as well. Maybe even a Raspberry Pi 3, which IIRC is arm64.

Ubuntu 17.04 should be easy to obtain. And I think I read something about aarch64 support for Fedora as well.

When it comes to the last point (4) I would suggest to be open, in order to foster a community, which might be helping Kura as well.

I think what is necessary to support this PR is a documented way of how to replicate this setup, so that it is possible for the community to test this. Like:
- Buy hardware X (and this should not be limited to a single vendor board)
- Install OS Y
- Perform the following steps

cdealti:
@markoer please add a link to the master QA spreadsheet describing the QA process for a supported Kura platform. It might help the discussion

feclare:
We will evaluate which other board could be used and will document how to run aarch64 kura on it.

markoer:
[Kura QA](https://eclipse.github.io/kura/assets/qa/system-test/qa_kura.ods)

garciagorka:
I work with Fede's (@feclare) at Cavium. 

I am afraid to say we cannot consider doing this development for other platforms. On the other hand, I think Fede's work is generic enough so that it could be tested in any AARCH64 platform, but we cannot take the responsibility of doing that. Adding Support for our competitors SoCs cannot be part of our work. 

Octeon TX development boards with all the needed software (OS, cross compilers, documentation, SDKs) are available through Cavium and there are also a number of customers providing these platforms as commercial products on a shape that will be very suitable for an IoT gateway (like the Nexcom DNA 1520).

We are also more than happy to donate Octeon TX reference platforms to the Eclipse Foundation if that helps in anyway and guarantee that Fede/Cavium will continue his collaboration with Eclipse IoT solutions regardless of what is decided. As Fede also mentioned, we are also OK with doing the QA process with our boards for each release.

MMaiero:
I believe that a limiting factor for this contribution is the fact that the profile name and other internal references are strictly related to a specific board/manufacturer (e.g. octeontx). In my opinion, this will make the profile less consumable by third parties that maybe have an ARM 64 board and think that this profile is specific only for this set of devices.
From what I see, the contribution does not require changes to the existing networking codebase, so maybe a possible enhancement would be to make this contribution more "agnostic", for example substituting the profile name from octeontx to armv8_64.

At the same time, in the project's documentation, we could create a document that states the supported boards, mapping them to the existing profiles. (This could also mean a change to the existing profiles to adhere to this new format)

@feclare Could this be an acceptable change for you?

feclare:
There is no problem renaming the profile name, but I suggest to use aarch64 instead of armv8_64. As it is the identifier used in the maven files.

MMaiero:
Fine for me. @dwoodard1 Ok also for you?

dwoodard1:
Absolutely. I think this is a good approach. @feclare Thanks for the PR and for working with us on the suggested changes!

feclare:
Renamed

dwoodard1:
@feclare For any file that you modified, can you add Cavium to the "Contributors" section of the header? I have given an example below. Files that were simply copy/pasted and not modified do not need this addition. I expect this should only be a few files.

@kartben We are going to merge this PR without a CQ even though it is > 1000 lines of code. The majority of the PR is simply copy/pasted code that is required to perform a build for a target. The PR does not change the behavior of Kura in any way nor does it add any new IP to Kura. Let me know if you see a problem with this.


\# Copyright (c) 2011, 2017 Eurotech and/or its affiliates and others
\#
\#  All rights reserved. This program and the accompanying materials
\#  are made available under the terms of the Eclipse Public License v1.0
\#  which accompanies this distribution, and is available at
\#  http://www.eclipse.org/legal/epl-v10.html
\#
\# Contributors:
\#   Eurotech
\#   Cavium

kura/distrib/src/main/resources/aarch64-nn/kura.properties
```diff
@@ -0,0 +1,156 @@
...
+#   Eurotech
+#
+
+## -----------------------------------------------------------------------------
+##  Kura Properties
+## -----------------------------------------------------------------------------
+kura.name=Everyware Software Framework
+kura.version=
+kura.company=ECLIPSE
+kura.project=aarch64-tx
```
MMaiero: The -tx is desired here?

kura/distrib/src/main/resources/aarch64-nn/kura.properties
```diff
@@ -0,0 +1,156 @@
...
+kura.primary.network.interface=eth0
+# kura.mac.address= Fetch from Java
+kura.home={kura.install.dir}/kura/kura
+kura.plugins={kura.install.dir}/kura/kura/plugins
+kura.packages={kura.install.dir}/kura/kura/packages
+kura.data={kura.install.dir}/kura/data
+kura.tmp=/tmp/.kura
+kura.snapshots={kura.install.dir}/kura/data/snapshots
+kura.style.dir={kura.install.dir}/kura/console/skin
+kura.have.net.admin=true
```
MMaiero: Since this is an NN profile, this should be set to false.

kura/distrib/src/main/resources/aarch64-nn/kura_install.sh
```diff
@@ -0,0 +1,111 @@
...
+    mkdir ${INSTALL_DIR}/kura/.data
+fi
+
+# setup /etc/sysconfig folder for iptables configuration file
+if [ ! -d /etc/sysconfig ]; then
+    mkdir /etc/sysconfig
+fi
+
+#set up default networking file
+cp ${INSTALL_DIR}/kura/install/network.interfaces.raspbian /etc/network/interfaces
```
MMaiero: Since this is an NN profile, I believe this set of changes that affect networking are not needed. Please refer, for example to an installer taken by another NN profile.

kura/distrib/src/main/resources/aarch64-nn/recover_default_config.init
```diff
@@ -0,0 +1,66 @@
...
+	    if [ $snapshot_file != $snapshots_dir"/snapshot_0.xml" ]; then 
+  		echo Removing the $snapshot_file configuration file ...
+		rm -f $snapshot_file
+		if [ "$?" = "0" ]; then
+            	    echo "  done"
+	        fi
+	    fi
+	done
+	
+	echo Copying the $dflt_data_dir/dhcpd-eth0.conf file to the /etc directory ...
```
MMaiero: Same here: it's an NN profile and the recover file should not touch networking related files.

kura/distrib/src/main/resources/aarch64-nn/snapshot_0.xml
```diff
@@ -0,0 +1,522 @@
...
+     are made available under the terms of the Eclipse Public License v1.0
+     which accompanies this distribution, and is available at
+     http://www.eclipse.org/legal/epl-v10.html
+
+    Contributors:
+      Eurotech
+
+-->
+<esf:configurations xmlns:ocd="http://www.osgi.org/xmlns/metatype/v1.2.0" xmlns:esf="http://eurotech.com/esf/2.0">
+    <esf:configuration pid="org.eclipse.kura.net.admin.FirewallConfigurationService">
```
MMaiero: Same here as above

kura/emulator/org.eclipse.kura.emulator/src/main/resources/log4j-to-file.properties
```diff
@@ -0,0 +1,29 @@
...
+#  All rights reserved. This program and the accompanying materials
+#  are made available under the terms of the Eclipse Public License v1.0
+#  which accompanies this distribution, and is available at
+#  http://www.eclipse.org/legal/epl-v10.html
+#
+# Contributors:
+#   Eurotech
+#
+
+### direct log messages to kura.log ###
```
MMaiero: Why this file needs to be added to the emulator?

MMaiero (commented):
Few more comments from my side.

kura/distrib/src/main/resources/aarch64-nn/jdk.dio.properties
```diff
@@ -0,0 +1,31 @@
+#
+# Copyright (c) 2011, 2014, 2017 Eurotech and/or its affiliates
```
markoer: CONTRIBUTING.md suggests the following date range format: year1, year2.

In this case it would be
```
# Copyright (c) 2011, 2017 Eurotech and/or its affiliates and others
...
```

Or, since it's actually a new file, this is more appropriate:

```
# Copyright (c) 2017 Eurotech and/or its affiliates and others
...
```

kura/distrib/src/main/resources/aarch64-nn/kura.properties
feclare: no, it was a problem with the replace I did. Will fix.

kura/distrib/src/main/resources/aarch64-nn/recover_default_config.init
feclare: About all the NN profile comments: I started with a non NN profile and then change to a NN. I didn't removed all this. I will fix them.

kura/distrib/src/main/resources/aarch64-nn/jdk.dio.properties
feclare: I can let it as you like, as most of the files were just copied from other directories I prefered to left the original copyrights.

feclare:
@dwoodard1 About the copyrights, almost all files were copied from other directories without any modification. I preferred to let the original copyright, as we did nothing to that files. But I have no problem adding the contributor if it is expected.

MMaiero:
Retargeted for merging into 3.1.0

MMaiero:
@feclare In order to be able to merge this PR in release, the existing conflicts need to be solved. Furthermore, please do not merge from develop or release branches. Use instead the rebase feature and squash your commits into one single, in order to be able to simplify the management and have the feature contributed with a single commit.

Thanks.

feclare:
@MMaiero Thanks for the fix. After your commit it compiles fine in my machine using build-all.sh. 

Do you know if org.eclipse.kura.linux.position.test failing is related to my changes?

MMaiero:
No, the test is indeed fragile. We will have to look at it after the release.

PR #1473, opened 2017-06-23
[Testing][Refactoring] ModemMonitorServiceImpl tests and refactoring

markoer:
Added service activation tests. Refactored so that testing was enabled.
Code review/actual test would be appreciated due to changed executor initialization.

@ibinshtok please review.

### DIFF
--- a/kura/org.eclipse.kura.net.admin/src/main/java/org/eclipse/kura/net/admin/monitor/ModemMonitorServiceImpl.java
+++ b/kura/org.eclipse.kura.net.admin/src/main/java/org/eclipse/kura/net/admin/monitor/ModemMonitorServiceImpl.java
@@ -84,9 +84,8 @@ import org.slf4j.LoggerFactory;
 
 public class ModemMonitorServiceImpl implements ModemMonitorService, ModemManagerService, EventHandler {
 
-    private static final Logger s_logger = LoggerFactory.getLogger(ModemMonitorServiceImpl.class);
+    private static final Logger logger = LoggerFactory.getLogger(ModemMonitorServiceImpl.class);
 
-    private ComponentContext m_ctx;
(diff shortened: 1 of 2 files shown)


ibinshtok (changes requested):
@markoer should we invoke the submitMonitorTask() from the processNetworkConfigurationChangeEvent() method as well?

kura/org.eclipse.kura.net.admin/src/main/java/org/eclipse/kura/net/admin/monitor/ModemMonitorServiceImpl.java
```diff
@@ -168,6 +168,9 @@ protected void activate(ComponentContext componentContext) {
 
         stopThread = new AtomicBoolean();
 
+        // must be initialized before trackModem() is called; risk of NPE otherwise
+        this.m_executor = Executors.newSingleThreadExecutor();
```
MMaiero: If it does not go ever 1k lines, I believe we can also rename the variables according to the new conventions.

markoer: I need to make updates anyway, so, yes.

MMaiero:
@ibinshtok Ok for you to merge?

ibinshtok (approved):

MMaiero:
@ibinshtok I believe something got broken with your additions: now we have double commits for the same thing by @markoer. Probably you needed to update your branch with upstream before pushing the new commits. Anyway, it should be working.

ibinshtok:
@MMaiero Yes, it should work. It happened because I forgot to do 'git pull origin' before adding a few more changes and  you have modified the code as well.

PR #1483, opened 2017-06-30
Added data transfer optimisation library for Drivers

nicolatimeus:
Adds a library bundle that provides the following functionalities:

* Provides an optimiser that allows to aggregate IO requests for contiguous or "close enough" data into larger requests, reducing IO times. 
* Allows to parse some raw binary data types into Java types (e.g. signed and unsigned integers and floating point types of different size and endianness).
* Provides some helper classes that ease the creation of Drivers that use the functionalities above.

Also includes some unit tests


MMaiero:
@markoer Please have a look at the unit tests added here and if new ones need to be created. I think that we can eventually open a related PR to track those new tests. What do you think?

kura/org.eclipse.kura.driver.block/src/main/java/org/eclipse/kura/driver/binary/Buffer.java
```diff
@@ -0,0 +1,36 @@
+package org.eclipse.kura.driver.binary;
```
MMaiero, line 13: Copyright header

kura/org.eclipse.kura.driver.block/src/main/java/org/eclipse/kura/driver/binary/ByteArrayBuffer.java
```diff
@@ -0,0 +1,39 @@
+package org.eclipse.kura.driver.binary;
```
MMaiero, line 13: Copyright header

kura/org.eclipse.kura.driver.block/pom.xml
```diff
@@ -0,0 +1,30 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+    Copyright (c) 2017 Eurotech and/or its affiliates and others
+
+     All rights reserved. This program and the accompanying materials
+     are made available under the terms of the Eclipse Public License v1.0
+     which accompanies this distribution, and is available at
+     http://www.eclipse.org/legal/epl-v10.html
+
+     Contributors:
```
amitjoy, line 10: `Contributors` is redundant

kura/org.eclipse.kura.driver.block/src/main/java/org/eclipse/kura/driver/block/task/AbstractBlockDriver.java
```diff
@@ -0,0 +1,320 @@
+/*******************************************************************************
```
markoer, line 1: No test coverage of this class...

kura/test/org.eclipse.kura.driver.block.test/src/test/java/org/eclipse/kura/driver/block/test/BlockTaskAggregatorTest.java
```diff
@@ -0,0 +1,293 @@
...
+                    try {
+                        next.run();
+                    } catch (IOException e) {
+                    }
+                    i++;
+                }
+                assertEquals(outputBlocks.length / 2, i);
+                assertEquals(inputBlocks.length / 2 + updateBlocks.length, childCount);
+            } else {
+                for (; blocks.hasNext(); blocks.next())
```
markoer: There's no assertion here. Unfinished test?

markoer (approved):
As far as tests are considered, it's OK. You could cover AbstractBlockDriver at least a bit, but the bundle is pretty well-covered already.

kura/test/org.eclipse.kura.driver.block.test/src/test/java/org/eclipse/kura/driver/block/test/BlockTaskAggregatorTest.java
nicolatimeus: This code was meant to be executed during the `shouldReportUnfeasibleProblem` tests, in this case it should be enough to trigger the aggregation process without checking anything until an exception is (or should be) thrown.
It is actually never executed because I forgot to remove the `.expect()` call, I will fix this.

kura/org.eclipse.kura.driver.block/src/main/java/org/eclipse/kura/driver/block/task/AbstractBlockDriver.java
nicolatimeus: I've added a few more tests that should cover the remaining parts of the bundle, can you please have a look to see if they are fine?
Feel free to modify/extend the tests if you want.

markoer: I suppose they're OK. Coverage is definitely high :)

PR #1485, opened 2017-07-03
First integration of Travis build stages

### LINKED ISSUE #1479: Travis-CI: Investigate build-stages to prevent build timeouts
https://docs.travis-ci.com/user/build-stages/

MMaiero:
Splitted kura build in two parts: Kura core first; distrib, example and tests in parallel
Added plugin to populate kura.build.version.
Removed stale wrl references in ant file.
Caching target-definition folder.
Build org.eclipse.kura-p2 before examples.

Closes #1479

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,9 +33,10 @@ jobs:
   include:
     - stage: Main-build
       script:
+        - rm -rf $HOME/.m2/repository/org/eclipse
         - $M2_HOME/bin/mvn -B -f target-platform/pom.xml clean install
-        - $M2_HOME/bin/mvn -B -f kura/pom.xml clean install 
-    - stage: distrib-examples-and-tests
+        - $M2_HOME/bin/mvn -B -f kura/pom.xml clean install -DskipTests
+    - stage: distrib-examples-tests-karaf
       script: $M2_HOME/bin/mvn -f kura/distrib/pom.xml clean install
     - # stage name not required, will continue to use the previous
       script:
@@ -43,3 +44,5 @@ jobs:
       - $M2_HOME/bin/mvn -f kura/examples/pom.xml clean install
     - # stage name not required, will continue to use the previous 
       script: $M2_HOME/bin/mvn -f kura/test/pom.xml clean install
+    - # stage name not required, will continue to use the previous
+      script: $M2_HOME/bin/mvn -f karaf/pom.xml clean install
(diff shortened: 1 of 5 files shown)


MMaiero:
@markoer

.travis.yml
```diff
@@ -24,4 +29,17 @@ addons:
       - dos2unix
       - oracle-java8-installer
 
-script: RUN_TESTS=true ./build-all.sh -Pbree -Dgwt.draftCompile=true -Dgwt.compiler.optimizationLevel=0 -Praspberry-pi-2-3
+jobs:
+  include:
+    - stage: Main-build
+      script:
+        - $M2_HOME/bin/mvn -B -f target-platform/pom.xml clean install
+        - $M2_HOME/bin/mvn -B -f kura/pom.xml clean install
```
markoer: See the comment at `pom.xml`.

kura/pom.xml
```diff
@@ -69,7 +69,6 @@
         <module>org.eclipse.kura.wire.provider</module>
         <module>features</module>
         <module>emulator</module>
-        <module>test</module>
```
markoer, line 72: I don't like that this update involves removing tests by default.

If you do this, please add another profile, that is activated by default:
```
        <profile>
            <id>tests</id>
            <activation>
                <property>
                    <name>!skipTests</name>
                </property>
            </activation>
            <modules>
                <module>test</module>
            </modules>
        </profile>
```

Then, use the following script for the first stage:
`- $M2_HOME/bin/mvn -B -f kura/pom.xml clean install -DskipTests`

This way the tests would still start when building manually.

Similar should probably be done for the `org.eclipse.kura-p2` module as it now builds twice.

markoer (approved):
I really like what you did, here. But I suggest you take into account the comment related to running tests.

kura/pom.xml
MMaiero: My original idea was to add that step in the hudson build and in the build-all script. But we can do this way as well

MMaiero: Same idea for the p2, even if its build will not be so much time consuming that we can decide to build twice

markoer: That's true, P2 builds rather fast.

MMaiero:
I've added another commit to include Karaf in the parallel build and to clean the eclipse folder in .m2 in order to try to prevent possible errors with stored Kura artifacts. 
I've finally added the profile for tests.

MMaiero:
After this PR, I will provide another to format, a bit, the pom files, since they seem a little bit messy.

nicolatimeus (approved):

Issue #1435, opened 2017-06-06
Snapshot problem on wires


aabramo:
If for some reason (i.e. a change of a package name of a wire component already present in a snapshot) the web browser cannot reload the snapshot, a GWT exception is generated.
In such a case, a previuos consistent snapshot should be reloaded instead, possibily up to snapshot_0.

dwoodard1:
I can confirm this is an issue. For me, steps to re-create the problem are:

1. Create a Wire component and give it a name
2. Since the name can't be edited, delete the component
3. Add a new component of the same type with the corrected name

The Wire graph will behave as expected, but the graph cannot be displayed in the UI (GWT Error).

MMaiero:
@dwoodard1 Which component did you use? Any save in the middle of the steps? I've tried with a nightly from develop both using a timer and an asset, but I was not able to recreate the issue.

MMaiero:
I've created #1445 that could mitigate the issue.

dwoodard1:
@MMaiero This happened to me with the Modbus Driver asset. I will try and re-create then test your fix.

cdealti:
@aabramo can you please build the develop branch and check if #1445 makes any difference?

MMaiero:
@dwoodard1 Any update on this?

dwoodard1:
@MMaiero I believe the usability is better, but there is still an issue. Steps to reproduce (note - I am using the emulator and ModbusPal):

* Install the Modbus driver
* Create a simple Wire Graph consisting of a Timer and Asset. The Asset should use the Modbus driver. Name the Asset "ModbusAsset". Save the graph.
* Delete the Asset. Add a new Asset using the Modbus driver and name it "modbusAsset". Save the graph.
* Leave the Wire Graph editor and then return. You will see an error like:
![image](https://user-images.githubusercontent.com/1392274/27753941-b0c5458e-5d9c-11e7-90b5-8d0fc11ea8d8.png)

The intention of the above steps is to be able to change the name of an Asset. I would think the easiest thing would be to make the name editable. Having to delete and re-add the asset is not ideal, especially if the Asset has many channels. I am happy to close this issue and open a new if you agree that a user should be able to edit a Wire Component name.

MMaiero:
I see what you mean. The issue do not happen if, between deleting and creating a new component, you perform a save. I believe that was the original process idea that Amit coded.

I will try to see how to fix temporary this issue, but I agree with you that giving the possibility to rename a component would be the best.
[...]

PR #1494, opened 2017-07-06
Fixed error happening when a renaming is performed

MMaiero:
Changed the equality check to deal with case.

### DIFF
--- a/kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/server/GwtWireServiceImpl.java
+++ b/kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/server/GwtWireServiceImpl.java
@@ -529,7 +529,7 @@ public final class GwtWireServiceImpl extends OsgiRemoteServiceServlet implement
             // check if jObj is an empty JSON. It means all the existing
             // wire components need to be deleted
             if (length == 0) {
-                logger.info("Deleting Wire Component: PID -> " + componentPid);
+                logger.info("Deleting Wire Component: PID -> {}", componentPid);
                 configService.deleteFactoryConfiguration(componentPid, false);
                 continue;
             }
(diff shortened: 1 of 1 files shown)


MMaiero:
Closes #1435

kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/server/GwtWireServiceImpl.java
```diff
@@ -529,21 +529,21 @@ private void removeDeletedFromWireGraphProperty(final String pid) throws GwtKura
...
+                logger.info("Deleting Wire Component: PID -> {}", componentPid);
                 configService.deleteFactoryConfiguration(componentPid, false);
                 continue;
             }
             boolean isFound = false;
             for (int i = 0; i < length; i++) {
                 final JsonObject jsonObject = jWireGraph.get(String.valueOf(i)).asObject();
                 final String component = jsonObject.getString("pid", null);
-                if (component.equalsIgnoreCase(componentPid)) {
+                if (component.equals(componentPid)) {
```
dwoodard1, line 540: @MMaiero I am fine merging this, but why was this originally equalsIgnoreCase? I just want to make sure we aren't breaking something more fundamental to fix this edge condition.

MMaiero: I believe I've not contributed it. So maybe we can ask @amitjoy to give some hints

amitjoy: Previously when I implemented this, I have added a constraint that names of wire components will be case-insesnsitive, that is, a wire component with a name `MyWC` would be as same as `mywc`. Therefore, no two components could co-exist with such names. Later on, I removed this constraint but its regression stayed. Now `equals` is more appropriate. Also note that, there are few more places where I used `equalsIgnoreCase` in a similar way. Make sure those also do not introduce any regression.

dwoodard1: Thanks Amit! @MMaiero Do you want to check for more instances of equalsIgnoreCase?

Issue #1023, opened 2017-01-17
[Test] Oomph Setup QA


dwoodard1:
The following scenarios should be tested to validate the Oomph based setup:

- [ ] Install a new developer's workspace on OSX and verify that all projects are present in the workspace, there are no errors in the workspace, and the emulator starts without error
- [ ] Repeat above for Linux

dwoodard1:
This may be user error on my part, but I am still having the below issues with a fresh Oomph setup:

- The emulator (org.eclipse.kura.emulator -> Kura_Emulator_OSX.launch) doesn't start. I can provide the errors, but there are many. Are others using a different approach to start the emulator?
- There are errors in the projects (org.eclipse.kura.protocol.can, org.eclipse.kura.windows.clock, etc). These are present because we are not including the projects in the Maven build. Not sure we can do anything about this, but it may raise questions when users see the red X on the projects. Maybe documenting this is enough.

ctron:
So for the ".can" protocol you need to actually download dependencies which seem to be optional. As the project it optional and only included in the build if the specific profile is enabled (default is disabled).

Of course we can enable the profile in the initial build.

The launch configuration I am using with the Oomph installer is "Eclipse Kura Emulator". And IIRC this should work on both Linux and Mac OS X.

pgufAtIBM:
tested the Oomph installer to setup a developers workspace on a Windows 7 machine (shown at https://www.youtube.com/watch?v=RwE3tLtejHM&feature=youtu.be). Last year in August I was able (after some problems) to setup the development environment on windows 7. Therefore I used now again this approach and faced now following problems:
**Problem 1** occurred when the installer wanted to download the source code of kura from git (error logs say that the authentication failed: java.lang.Exception: org.eclipse.jgit.api.errors.TransportException: git@github.com:eclipse/kura.git: Auth fail). I manually cloned the code to the path (no authentication was used) and then let Oomph the steps afterwards. Had this problem already last year, and solved it by cloning the github repo manually
**Problem 2** Plugin org.eclipse.tycho:tycho-maven-plugin:0.26.0 cannot be resolved:
[INFO] Scanning for projects...
[INFO] Downloading: https://repo.maven.apache.org/maven2/org/eclipse/tycho/tycho-maven-plugin/0.26.0/tycho-maven-plugin-0.26.0.pom
[ERROR] [ERROR] Some problems were encountered while processing the POMs:
[...]

PR #1495, opened 2017-07-06
Oomph update

markoer:
Updated working set creation a bit.
Updated emulator launcher for H2 DB.

See also #1023.


nicolatimeus:
I've tested it on OSX and seems fine: the build succeeds, the emulator starts out of the box, the Web Ui is reachable.

Issue #997, opened 2017-01-11
URI has an authority component


rita-zgheib:
Hey,
I am trying to build kura project using this command "mvn -Dmaven.test.skip=true -f manifest_pom.xml clean install"
but I have the same problem always URI has an authority component  Failed to load p2 repository with ID 'p2-repo-equinox_3.11.1' from location file:////C:/Users/rzgheib/git/kura/kura/target-definition/equinox_3.11.1/repository/

Any advice?
Thank you in advance,

Rita

dwoodard1:
I don't have a Windows machine to test on, but it looks like an extra "/" may have been added to the manifest_pom.xml file. Can you try changing:

file:////${kura.basedir}/target-definition/equinox_3.11.1/repository/

to 

file:///${kura.basedir}/target-definition/equinox_3.11.1/repository/

@pierrepitiot Do you have a Windows machine on which you could try building the develop branch?

rita-zgheib:
I tried to remove "/" and I tried all "/" options but it doesn't work. I followed the documentation and instructions in this link: https://wiki.eclipse.org/Kura/Getting_Started

It seems that it is a problem with equinox 3.11.1 because I found another version with kura-equinox_3.8.1 and it works. I am now following this documentation https://eclipse.github.io/kura/doc/kura-setup.html and i downloaded Kura with equinox_3.8.1 not 3.11.1

Anyway, do you recommend me to use Kura on raspberry?

ctron:
@rita-zgheib Did you use the manual way or the Eclipse installer way?

rita-zgheib:
Both of them !

dwoodard1:
Do you need to build Kura or do you simply want to run Kura on a Raspberry Pi? If you just want to try Kura on a Raspberry Pi, I would recommend following these instructions: http://eclipse.github.io/kura/doc/raspberry-pi-quick-start.html.

We will look into the issue of building Kura on Windows.

rita-zgheib:
I want to use Kura for my platform that I am developing during my phd. The target of my project is to exchange semantic messages over MQTT in an IoT applications. For that I want to use Kura and I want to add the specifications of my application.

I have already tested Kura on the raspberry and everything is OK!

dwoodard1:
[...]

PR #1496, opened 2017-07-06
Updated for Windows build

markoer:
* repository files: //// -> ///
* updated tests

Note: Emulator still doesn't run on Windows.

See also #997 and #1023.


markoer:
I removed the checkstyle changes and updated .gitattributes to force LF line endings.

Issue #1475, opened 2017-06-27
[Testing] Add H2DbService unit and integration tests


cdealti:
@markoer please consider this task after #1457 is merged.

MMaiero:
@markoer PR #1457 has been merged.

markoer:
Thanks for info.

cdealti:
@cdealti please review #1527

cdealti:
@markoer #1527 has been merged. Do you plan to add specific tests for the H2DbService?

markoer:
#1509 belongs here as well... I suppose this is what you meant, @cdealti?

MMaiero:
I believe we can close, right?

markoer:
I suppose so, yes.

PR #1503, opened 2017-07-10
[QA] Added H2 test procedures

markoer:
Added DB-H2-* test procedures.

Related to #1475.

@nicolatimeus, please review.


nicolatimeus:
Maybe you already considered this, but we could add some test procedures or integration tests to verify that existing components that use h2 support switching db service instances at runtime and that they don't break if they do not have a db service instance attached.
Something like this:

*  Attach a component to to an existing h2 db service instance and then switch to another existing instance.
*  Attach a component to an existing h2 db service instance, change the db service pid to an invalid value and then switch back to the existing instance.
* Attach a component to an existing h2 db service instance and then delete and recreate the db instance.

At the end of the tests above the components should be in a working state.

nicolatimeus:
Another corner case that could be covered is modifying a DB service that is in use by some components, for example by changing the db URL in order to switch a completely different database. The DataService and H2DbStore should be able to detect the update and recreate the tables if needed.

markoer:
* Attach a component to an existing h2 db service instance and then delete and recreate the db instance.

What exactly do you mean by that? To delete a file-backed DB (delete files)? Or to delete the h2 db service and create a new one with the same pid?

nicolatimeus:
The second one, deleting the h2 db service and creating a new one with the same pid.

markoer:
Done.

nicolatimeus (approved):

Issue #1506, opened 2017-07-11
H2 DB: Changing the db user does not work


nicolatimeus:
Currently if an H2 db service instance is active and the user is changed in the configuration, the connection pool will not be modified and the user switch will not be performed until the component is deactivated and reactivated.

MMaiero:
Closed by #1508

PR #1508, opened 2017-07-11
Fixed H2 Db user change

nicolatimeus:


nicolatimeus:
Changing the user for an existing instance should now close and reopen the connection pool with the new credentials.

This has the following implications:

* Changing the user for an active in-memory instance will result in the destruction and recreation of the instance (all data will be lost).
* Changing the user for a file-based instance may result in the destruction and recreation of the instance if the new user does not exist or the password is wrong.

MMaiero (approved):

MMaiero:
Closes #1506

PR #1525, opened 2017-07-13
Show the service description as well

ctron:
This change does show the service description as well as the
service name. In the menu on the left this is done by using the
title attribute, in the panel heading it uses bootstraps sub-title
feature.


nicolatimeus:
I've tested this and it's working, service descriptions are shown.
In order to improve consistency with the rest of the Web UI this could be modified in order to have the service description below the service name, like in the firewall configuration section, as depicted below. What do you think?
screen shot 2017-07-27 at 10 38 01

nicolatimeus:
@ctron any update on this? If you agree I can rebase this branch and implement the change proposed above.

MMaiero (approved):

Issue #1526, opened 2017-07-13
QA tests for the connectivity monitor of the DataServiceImpl


cdealti:
@markoer can you draft QA tests and **do a preliminary check**?
This has high priority.

See #1138 

- [ ] When enabled and properly configured it should cause a system reboot (via the Watchdog Service) **only**  if the DataService cannot connect to the broker for any reason but those described in [1] section **3.2.2.3 Connect Return code**. For example unplugging the Ethernet cable should cause a reboot but connecting with a bad username or password should not.

- [ ] Verify if changing the **connect.retry-interval** parameter could cause spurious reboots

- [ ] Use your imagination to try to cause spurious reboots

[1] http://docs.oasis-open.org/mqtt/mqtt/v3.1.1/mqtt-v3.1.1.html

markoer:
Currently I have the cases below.
**Bad username** currently causes a **reboot**.

* DATA-CRIT-001 (disconnect)

Prepare the following setup.
System->Network:
Disable all network interfaces except lo, eth0 and wlan0.
Setup eth0 as Enabled for WAN.
Setup wlan0 as Enabled for LAN and access point with a 'DHCP and NAT', 'Pass DNS Servers through DHCP' = false.

System->Cloud Services->DataService:
connect.auto-on-startup = true
connect.retry-interval = 60
enable.recovery.on.connection.failure = true
connection.recovery.max.failures = 3
Make sure device connects to the broker.

Enable Watchdog service:
Services->WatchdogService:
Watchdog enable = true

Connect to Kura through wlan0.
Observe the log: tail -f /var/log/kura.log

Disconnect the network cable. Make sure device cannot ping the outside network.

There will be messages that the eth0 interface is down and later that the client connection was lost. 3 reconnect attempts will be made. After the last one times out, there will be a warning message that connection failed, info 'Maximum number of connection attempts reached. Requested reboot' and in the end a Watchdog service's warning: 'Critical service DataServiceImpl failed -> SYSTEM REBOOT'.

After this the device should reboot.

* DATA-CRIT-002 (invalid credentials)

Prepare the same setup as DATA-CRIT-001.

Change the MQTT transport's username to an invalid one:
System->Cloud Services->MqttDataTransport:
username = &lt;invalid user>

Apply and observe the log.

After 3 attempts a reboot will be requested with similar log messages as before.

MMaiero:
Yes, currently that requirement is not implemented: I will provide a fix for that

markoer:
[...]

PR #1534, opened 2017-07-17
[QA] Added DATA-CRIT-* procedures

markoer:
Added new test procedures: DATA-CRIT-*.

Implementation of #1526.


MMaiero (approved):

MMaiero:
We must remember that, if PR #1557 is merged, we need to change DATA-CRIT-006

Issue #1513, opened 2017-07-12
H2 DB: Web app and TCP server are mutually exclusive


cdealti:
@nicolatimeus can we add more configuration options to enable the H2 Web app and TCP server independently?
I believe that replicating the existing settings for a another server instance would just work.

nicolatimeus:
We can achieve this in at least 3 ways:

1. Turning the H2DbServer into a factory and keeping the metatype as it is now.
2. Creating 3 dedicated H2DbServer components, one for each mode (regular ConfigurableComponents, not factories):
   * One for Web mode
   * One for TCP/JDBC mode
   * One for PG mode
  The metatype of each of these components will look like this:
   * **db.server.enabled**: BOOLEAN
   * **db.server.commandline**: STRING
3. Keeping a single H2DbServer component (not factory) and modifying the metatype in order to allow to enable and configure each mode separately, the metatype of this component will look like this:
   * **tcp.enabled**: BOOLEAN
   * **tcp.commandline**: STRING
   * **web.enabled**: BOOLEAN
   * **web.commandline**: STRING
   * **pg.enabled**: BOOLEAN
   * **pg.commandline**: STRING

cdealti:
1 looks cheaper in terms of work, right?

markoer:
1 looks most configurable, but has the potential to e.g. expose the database via web 2 or more times using different options. Do you want to allow such scenarios? If yes, go for it.
3 offers the most control and least chance for abuse while still compact.

cdealti:
@markoer we understand you point. If we go with 1 what are the testing implications?

markoer:
I suppose the only difference would be to test that multiple ports can provide the same service concurrently  vs. that that is not possible.

cdealti:
@nicolatimeus I also believe we should update the documentation in the Wiki to reflect the latest changes. It maybe easier if you do that now.

cdealti:
@nicolatimeus can we close this?

PR #1551, opened 2017-07-21
Implemented H2DbServer as a factory component

nicolatimeus:

### DIFF
--- a/kura/org.eclipse.kura.core/OSGI-INF/metatype/org.eclipse.kura.core.db.H2DbServer.xml
+++ b/kura/org.eclipse.kura.core/OSGI-INF/metatype/org.eclipse.kura.core.db.H2DbServer.xml
@@ -10,7 +10,7 @@
 
 -->
 <MetaData xmlns="http://www.osgi.org/xmlns/metatype/v1.2.0" localization="en_us">
-    <OCD id="org.eclipse.kura.core.db.H2DbHelper" 
+    <OCD id="org.eclipse.kura.core.db.H2DbServer" 
          name="H2DbServer" 
          description="H2 based database service.">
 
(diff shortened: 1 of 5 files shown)


nicolatimeus:
Closes #1513

kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/db/H2DbServer.java
```diff
@@ -0,0 +1,126 @@
...
+                logger.info("Shutting down DB server...");
+                server.stop();
+                logger.info("Shutting down DB server...done");
+            } catch (Exception e) {
+                logger.error("failed to shutdown DB server", e);
+            }
+        }
+    }
+
+    private static class ServerConfiguration {
```
cdealti: I believe we are standardising on using a public "Options" class for the configuration.
I also believe it would be safer to start using defensive code:
1. Inizialize with safe defaults if the property is missing or unexpected instanceof
2. Check instanceof before casting

kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/db/H2DbHelper.java
```diff
@@ -45,109 +32,24 @@ protected void unsetConfigurationService(ConfigurationService configurationServi
...
         try {
             if (configurationService.getComponentConfiguration(H2DbService.DEFAULT_INSTANCE_PID) == null) {
+                logger.info("Default H2DbService instance configuration not found, creating new instance...");
                 configurationService.createFactoryConfiguration(H2_DB_SERVICE_FACTORY_PID,
                         H2DbService.DEFAULT_INSTANCE_PID, null, true);
+            } else {
+                logger.info("Default H2DbService instance configuration found");
             }
         } catch (KuraException e) {
             throw new ComponentException(e);
```
cdealti: As far as I remember this causes the framework to retry component activation. Did you choose this for a specific reason?

cdealti (changes requested):

kura/org.eclipse.kura.core/OSGI-INF/metatype/org.eclipse.kura.core.db.H2DbServer.xml
```diff
@@ -43,7 +43,7 @@
             description="Specifies the parameters for the server, see http://www.h2database.com/javadoc/org/h2/tools/Server.html for more details."/>
```
markoer: I propose to add a note here, that port must be opened in the Firewall section - it's not opened automatically.

markoer (approved):
It works i.e. I was able to run multiple services/servers on multiple ports.

Issue #1530, opened 2017-07-14
Modify the wire composer to allow driver management


MMaiero:
According to what designed in https://github.com/eclipse/kura/wiki/Assets-and-Driver-Web-Ui-section-mockups the Wire composer should be updated to be able to manage the creation of a Driver directly from inside the composer.
The scope is to allow the user to manage and/or design a graph completely from the composer page, without the need to go out of it.

A first draft is available at the following link: https://github.com/MMaiero/kura/tree/enh_drivers-assets-refactor

MMaiero:
Attached few screenshots of the new prototyped design.

The configuration tab for a generic Wire component:
![Schermata 2017-07-14 alle 14.36.42.png](https://images.zenhubusercontent.com/572701dfbf93871ccd9a576a/f7ebcd01-7826-4b94-a7ec-b1212824743a)


When the dropped wire component is an Asset, the configuration tabs are 2: one for the asset configuration with the usual channels, the other for the related Driver instance that allows to manage the driver configuration.
![Schermata 2017-07-14 alle 14.40.22.png](https://images.zenhubusercontent.com/572701dfbf93871ccd9a576a/a7472433-b5f2-4e32-8878-824191f100cd)

![Schermata 2017-07-14 alle 14.40.31.png](https://images.zenhubusercontent.com/572701dfbf93871ccd9a576a/e75541b7-b225-4f31-980f-7d641eec4303)

New modal for the Asset creation.
![Schermata 2017-07-14 alle 15.36.49.png](https://images.zenhubusercontent.com/572701dfbf93871ccd9a576a/b0ba7b55-516e-42c4-a2ef-0324f4d81bea)

If the + driver button is clicked, a new modal is opened and allows to define a new driver instance selecting from the proper driver factories.
![Schermata 2017-07-14 alle 15.36.55.png](https://images.zenhubusercontent.com/572701dfbf93871ccd9a576a/d300095d-b9f8-4d11-ab34-eae65608c964)

![Schermata 2017-07-14 alle 15.37.00.png](https://images.zenhubusercontent.com/572701dfbf93871ccd9a576a/bfca4fa5-13c9-460a-abf7-0f93a15733db)

When done, a new driver instance is created and available to be used to associate with the new asset.
![Schermata 2017-07-14 alle 15.37.11.png](https://images.zenhubusercontent.com/572701dfbf93871ccd9a576a/cc3c8d2e-090b-4b72-8a53-7fc3d9fa00e9)

[...]

PR #1552, opened 2017-07-22
First set of changes to the wire composer

MMaiero:
This PR applies some changes to the wire tab presenting the configuration data as tabs.
For drivers, it's possible to create instances directly from the Wires page without the need to go out of the composer page. The driver configuration is displayed beside the asset one and is changeable as any other component in the composer.

Closes PR #1530


nicolatimeus (approved):

amitjoy (approved):
Got the time to review it completely. I like the changes.

PR #1557, opened 2017-07-25
Set the connect.retry-interval minimum value in metatype to 1.

### LINKED ISSUE #1556: connect.retry-interval=0 causes DataService activation failure
Setting the `connect.retry-interval` to 0 seems to work until reboot.
After a reboot, the DataService component starts continuously to try to activate, determining the impossibility to use the default cloud service and all the related stack.
I believe the root cause is related to the `IllegalArgumentException` thrown by the [scheduleAtFixedRate](https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/ScheduledExecutorService.html#scheduleAtFixedRate(java.lang.Runnable,%20long,%20long,%20java.util.concurrent.TimeUnit)) .

This determines that the `startConnectionMonitorTask` method throws that runtime exception till going up in the stack-trace at the `activate` level. This causes the component to continuously be restarted by the framework.

I've tested the same behaviour in Kura 3.1.0-SNAPSHOT and Kura 3.0.0. I've not done tests for previous versions of Kura.

MMaiero:
This should prevent the IllegalArgumentException in the startConnectionMonitorTask
and the continuous component activations during a reboot.

Closes #1556

### DIFF
--- a/kura/org.eclipse.kura.core/OSGI-INF/metatype/org.eclipse.kura.data.DataService.xml
+++ b/kura/org.eclipse.kura.core/OSGI-INF/metatype/org.eclipse.kura.data.DataService.xml
@@ -33,8 +33,8 @@
             cardinality="0" 
             required="true"
             default="60"
-            min="0"
-            description="Frequency in seconds to retry a connection of the Data Publishers after a disconnect (0 to disable)."/>
+            min="1"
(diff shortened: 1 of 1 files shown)


MMaiero:
Obviously, this change modifies a little bit the behaviour: now it is not possible to set to 0 the property and cause the framework to try reconnections. But I believe this last is a case that could not be applied for an IoT framework.

PR #1599, opened 2017-08-11
Fixed startup scripts and WatchdogService

pierantoniomerlino:
This PR takes care of :
- Missing creation of watchdog config file in /tmp when the WatchdogService is enabled
- Added ```sync```command execution before reboot
- Added checkin after stop in init scripts

### DIFF
--- a/kura/org.eclipse.kura.linux.watchdog/src/main/java/org/eclipse/kura/linux/watchdog/WatchdogServiceImpl.java
+++ b/kura/org.eclipse.kura.linux.watchdog/src/main/java/org/eclipse/kura/linux/watchdog/WatchdogServiceImpl.java
@@ -72,10 +72,10 @@ public class WatchdogServiceImpl implements WatchdogService, ConfigurableCompone
         if (this.options.isEnabled()) {
             this.rebootCauseWriter = new RebootCauseFileWriter(this.options.getRebootCauseFilePath());
 
-            try (PrintWriter wdWriter = new PrintWriter(this.options.getWatchdogDevice());) {
+            try (PrintWriter wdWriter = new PrintWriter(this.options.getWatchdogEnabledTemporaryFilePath())) {
                 wdWriter.write(this.options.getWatchdogDevice());
             } catch (IOException e) {
-                logger.error("Unable to write watchdog config file", e);
+                logger.error("Unable to write watchdog enabled temporary file", e);
             }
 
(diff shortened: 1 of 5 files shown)


kura/org.eclipse.kura.linux.watchdog/src/main/java/org/eclipse/kura/linux/watchdog/WatchdogServiceImpl.java
```diff
@@ -72,7 +74,7 @@ public void updated(Map<String, Object> properties) {
         if (this.options.isEnabled()) {
             this.rebootCauseWriter = new RebootCauseFileWriter(this.options.getRebootCauseFilePath());
 
-            try (PrintWriter wdWriter = new PrintWriter(this.options.getWatchdogDevice());) {
+            try (PrintWriter wdWriter = new PrintWriter(WD_FILE);) {
```
MMaiero: Just a clarification: the WD_FILE location is now somewhat hard coded in the implementation. But it is still modifiable in the service preferences.
There is still a reference to the `getWatchdogDevice` in the private method `writeWatchdogDevice`. Could this take to some sort of inconsistencies? What happens if the user changes that value in the service preferences?

nicolatimeus: The `watchdogDevice` configuration property and the `WD_FILE` constant refer to two different files:
* the first property contains the the path of the watchdog device to be used e.g. `/dev/watchdog`.
* The second constant defines the path of a file that is created if the watchdog is enabled and contains the path of the watchdog device, this file is used for example by the init scripts and the upgrader.

In my previous pull request I've refactored the configuration options handling by creating the `WatchdogServiceOptions` class and I've accidentally removed the references to WD_FILE, this PR fixes the issue.
Maybe we could define a more descriptive name for the `WD_FILE` constant in order to avoid confusing it with the `watchdogDevice` property.

MMaiero: Great. But at this point, since we have the properties class, I would move the WD_FILE there. Furthermore, in `WatchdogServiceOptions` the default for `PROPERTY_WD_DEVICE` is `"/tmp/watchdog"` that does not match the default value in the metatype.

MMaiero (approved):

PR #1628, opened 2017-08-31
Fix #1511: Add MQTT broker based on Artemis

### LINKED ISSUE #1511: Add Artemis broker to Kura core
As discussed last week during the F2F meeting, Artemis should get added to the Kura core distribution for providing a local gateway broker.

ctron:
This PR adds functionality to host Artemis based brokers in Kura and adds a simple MQTT based broker instance.


nicolatimeus:
Looks like the problem related to the missing dependency of `org.apache.activemq.artemis-amqp-protocol` is still present:
```
!ENTRY org.apache.activemq.artemis-amqp-protocol 4 0 2017-09-04 09:54:14.161
!MESSAGE FrameworkEvent ERROR
!STACK 0
org.osgi.framework.BundleException: Could not resolve module: org.apache.activemq.artemis-amqp-protocol [100]
  Unresolved requirement: Import-Package: org.apache.qpid.proton; version="[0.18.0,1.0.0)"
```
Any update on this?

ctron:
No, I can remove the AMQP protocol provider if we decide not to ship this.

nicolatimeus:
I agree, I believe we can just remove the AMQP protocol provider for now.

nicolatimeus:
This PR introduces two Artemis-related configurable components that are shown in the Web UI: **Simple Artemis MQTT Broker** and **ActiveMQ Artemis Broker**.
These two services both allow to start an Artemis instance, the first one is more straight forward for a basic user but is less configurable, the second is more powerful but it requires the user to have some more knowledge of Artemis configuration.

In my opinion having two different services that allow to start an Artemis broker can be a bit confusing for the user. Another approach can be keeping only the  **ActiveMQ Artemis Broker** service with a default configuration that matches the behaviour of the current **Simple Artemis MQTT Broker**.

In this way a basic user can get a working broker just by enabling the service without changing the configuration, and a more advanced user is still free to customise the instance, whithout having two services.
What do you think about this?

ctron:
AMQP supported is dropped as requested.

ctron:
First of all the "simple" and "xml" bundle can simply be dropped. Customizing the default behavior in that way. Also is there no direct conflict between them. You can run both instances at the same time (with different ports) if you want that. If you look into the source code you will see that the simple broker actually builds up the configuration based on the configuration options. So that isn't really possible with the plain XML approach and so you would loose the "simple" aspect.

I  also think it is a bit late in the process to start making bigger changes like that.

nicolatimeus:
Ok, I think we can merge then, can you please solve the conflicts?

nicolatimeus:
Can you also please sqash the commits?

ctron:
I did rebase the PR on the current develop branch.

You can squash the commits directly when merging the PR from GitHub using the option "Squash and merge". So the original history in this PR is preserved.

PR #1667, opened 2017-09-13
Support 'manual' mode for network interfaces

ibinshtok:
The idea is for Kura to support interface configuration mode where interface is up but boot protocol is not specified and IP address is not assigned. It would be ideal if Kura not just leaves interface alone, but actually configures interface in this mode as shown in examples below.
For Debian based systems:
auto eth1
iface eth1 inet manual

For Red Hat based system:
DEVICE=eth0
ONBOOT=yes
BOOTPROTO=none


kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/net/NetInterfaceConfigMode.java
```diff
@@ -0,0 +1,27 @@
+/*******************************************************************************
+ * Copyright (c) 2011, 2017 Eurotech and/or its affiliates
```
cdealti, line 2: Remove range, only 2017

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/net/NetInterfaceConfigMode.java
```diff
@@ -0,0 +1,27 @@
...
+ *
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
```
cdealti, line 9: Remove Contributors

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/net/NetInterfaceConfigMode.java
```diff
@@ -0,0 +1,27 @@
...
+ *     Eurotech
+ *******************************************************************************/
```
cdealti, line 11: You should enable "API Tools" in Eclipse and add the ``@since`` Javadoc tags.

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/net/NetInterfaceConfigMode.java
```diff
@@ -0,0 +1,27 @@
...
+package org.eclipse.kura.net;
+
+public enum NetInterfaceConfigMode {
+
+    /** network interface is configured as DHCP client **/
+    netIPv4ConfigModeDhcp,
```
cdealti, line 17: Enum values are constants and should be uppercase https://docs.oracle.com/javase/tutorial/java/javaOO/enum.html

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/net/NetInterfaceConfigMode.java
```diff
@@ -0,0 +1,27 @@
...
+
+    /** network interface is configured with static IP address **/
+    netIPv4ConfigModeStatic,
+
+    /**
+     * network interface is configured in 'manual' or 'BOOTPROTO=none' mode
```
cdealti, line 23: This API package should be OS independent. I'd remove references to Linux.
IMHO we should "unmanaged" instead of "manual". This interface is not managed by ESF.

And I'm not still convinced if "unmanaged" should be a value of the configuration mode.

If I think to the Web UI, I'd find more natural to set the interface **Status** to Unmanaged (as opposed to Enabled for LAN/WAN or Disabled). Selecting Unmanaged should disable the interface configuration widgets in the Web UI.

I believe that in this PR you can set the interface Status to e.g. Enabled for LAN and set the mode to unmanaged. I believe this is confusing... 

In a pinch, cannot we define the new value Unmanaged both for the interface status and the interface mode, the former implying the latter?

kura/org.eclipse.kura.net.admin/META-INF/MANIFEST.MF
```diff
@@ -26,7 +26,7 @@ Import-Package: javax.microedition.io,
  org.eclipse.kura.linux.net.route;version="[1.0,2.0)",
  org.eclipse.kura.linux.net.util;version="[1.0,2.0)",
  org.eclipse.kura.linux.net.wifi;version="[1.0,2.0)",
- org.eclipse.kura.net;version="[1.3,1.4)",
+ org.eclipse.kura.net;version="[1.3,1.5)",
```
cdealti, line 29: [1.4,1.5)

kura/org.eclipse.kura.net.admin/src/main/java/org/eclipse/kura/net/admin/NetworkConfigurationServiceImpl.java
```diff
@@ -599,6 +599,19 @@ private Tocd getDefinition() throws KuraException {
                     tad.setDescription(
                             NetworkAdminConfigurationMessages.getMessage(NetworkAdminConfiguration.CONFIG_AUTOCONNECT));
                     tocd.addAD(tad);
+
```
cdealti, line 602: BTW, do we have a good default if we start with an old snapshot which does not have this property?

I also think the many of the values of many of these attributes should be options, not any string. But this is for the next time ;)

kura/org.eclipse.kura.net.admin/src/main/java/org/eclipse/kura/net/admin/monitor/EthernetMonitorServiceImpl.java
```diff
@@ -235,7 +237,8 @@ private void monitor(String interfaceName) {
                 // (String interfaceName, boolean up, boolean link, IPAddress ipAddress)
                 // It will save a call to determine the iface type and it will keep InterfaceState
                 // as a state object as it should be. Maybe introduce an InterfaceStateBuilder.
-                currentInterfaceState = getEthernetInterfaceState(interfaceName);
+                NetInterfaceConfigMode ifaceConfigMode = this.getInterfaceConfigurationMode(currentInterfaceConfig);
```
cdealti, line 240: I would not mix interface state with mode

kura/org.eclipse.kura.net.admin/src/main/java/org/eclipse/kura/net/admin/monitor/EthernetMonitorServiceImpl.java
```diff
@@ -493,6 +496,25 @@ private NetInterfaceStatus getStatus(EthernetInterfaceConfigImpl ethernetInterfa
         return status;
     }
 
+    private NetInterfaceConfigMode getInterfaceConfigurationMode(EthernetInterfaceConfigImpl ethernetInterfaceConfig) {
```
cdealti, line 499: I guess this answers my question above on the default mode.
Isn't Sonar Lint complaining about the cyclomatic complexity of this method?

kura/org.eclipse.kura.net.admin/src/main/java/org/eclipse/kura/net/admin/monitor/InterfaceState.java
```diff
@@ -16,39 +16,47 @@
...
+        this.link = link;
+        this.ipAddress = ipAddress;
     }
 
-    public InterfaceState(NetInterfaceType type, String interfaceName) throws KuraException {
-        this.m_name = interfaceName;
-        this.m_up = LinuxNetworkUtil.hasAddress(interfaceName);
-        this.m_link = LinuxNetworkUtil.isLinkUp(type, interfaceName);
-
+    public InterfaceState(NetInterfaceType type, String interfaceName, NetInterfaceConfigMode ifaceConfigMode)
```
cdealti, line 40: See my comment above about mixing state and mode

kura/org.eclipse.kura.net.admin/src/main/java/org/eclipse/kura/net/admin/visitor/linux/IfcfgConfigReader.java
```diff
@@ -107,10 +108,9 @@ private void getConfig(NetInterfaceConfig<? extends NetInterfaceAddressConfig> n
             logger.debug("Setting NetInterfaceStatus to {} for {}", netInterfaceStatus, netInterfaceConfig.getName());
 
             boolean autoConnect = false;
-            // int mtu = -1; // MTU is not currently used
-            boolean dhcp = false;
+            NetInterfaceConfigMode netInterfaceConfigMode = NetInterfaceConfigMode.netIPv4ConfigModeManual;
             IP4Address address = null;
-            String ipAddress = null;
+            String ipAddress;
```
cdealti, line 113: why?

kura/org.eclipse.kura.net.admin/src/main/java/org/eclipse/kura/net/admin/visitor/linux/IfcfgConfigWriter.java
```diff
@@ -181,6 +180,9 @@ private void writeRedhatConfig(NetInterfaceConfig<? extends NetInterfaceAddressC
                                             .append(((NetConfigIP4) netConfig).getGateway().getHostAddress())
                                             .append("\n");
                                 }
+                            } else {
```
cdealti, line 183: Shouldn't we skip the interface configuration file (ifcfg-<ifname>) altogether? Isn't the file configured by another application or by the user?

kura/org.eclipse.kura.net.admin/src/main/java/org/eclipse/kura/net/admin/visitor/linux/IfcfgConfigWriter.java
```diff
@@ -490,12 +500,8 @@ private String debianWriteUtility(NetInterfaceConfig<? extends NetInterfaceAddre
...
                             }
                         } else {
-                            // DEFROUTE
-                            if (((NetConfigIP4) netConfig).getStatus() == NetInterfaceStatus.netIPv4StatusEnabledLAN) {
-                                sb.append("\tpost-up route del default dev ");
-                                sb.append(interfaceName);
-                                sb.append("\n");
-                            }
+                            // 'Manual' configuration
+                            sb.append("manual\n");
```
cdealti, line 504: Same here for Debian interface configuration file. Isn't this interface configured by someone else? Are we overwriting existing configuration?

kura/org.eclipse.kura.net.admin/src/main/java/org/eclipse/kura/net/admin/visitor/linux/IfcfgConfigWriter.java
```diff
@@ -653,11 +666,8 @@ private Properties parseNetInterfaceAddressConfig(NetInterfaceAddressConfig netI
                             props.setProperty("DEFROUTE", "no");
                         }
                     } else {
-                        if (((NetConfigIP4) netConfig).getStatus() == NetInterfaceStatus.netIPv4StatusEnabledWAN) {
-                            props.setProperty("DEFROUTE", "yes");
-                        } else {
-                            props.setProperty("DEFROUTE", "no");
-                        }
+                        // BOOTPROTO=none
+                        props.setProperty("BOOTPROTO", "manual");
```
cdealti, line 670: Same as above?

kura/org.eclipse.kura.net.admin/src/main/java/org/eclipse/kura/net/admin/visitor/linux/PppConfigReader.java
```diff
@@ -468,7 +469,8 @@ private NetConfigIP4 getNetConfigIP4(String interfaceName) throws KuraException
         }
         logger.debug("Setting NetInterfaceStatus to " + netInterfaceStatus + " for " + interfaceName);
 
-        NetConfigIP4 netConfigIP4 = new NetConfigIP4(netInterfaceStatus, true, true);
+        NetConfigIP4 netConfigIP4 = new NetConfigIP4(netInterfaceStatus, true);
+        netConfigIP4.setConfigMode(NetInterfaceConfigMode.netIPv4ConfigModeDhcp);
```
cdealti, line 473: Nothing else needed e.g. in the PPP monitor?

cdealti (changes requested):

cdealti:
@ibinshtok also please check why tests are failing.
It's not yet clear to me the purpose of this PR?
Is it to allow an external application to manage/configure a specific interface?

For example, shall I be able to edit /etc/network/interfaces, set the configuration of eth0 and configure ESF to not manage this interface?

kura/org.eclipse.kura.net.admin/src/main/java/org/eclipse/kura/net/admin/monitor/EthernetMonitorServiceImpl.java
ibinshtok: I will fix this Lint issue.

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/net/NetInterfaceConfigMode.java
ibinshtok: will fix in the next commit

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/net/NetInterfaceConfigMode.java
ibinshtok: will fix in the next commit

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/net/NetInterfaceConfigMode.java
ibinshtok: will fix in the next commit

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/net/NetInterfaceConfigMode.java
ibinshtok: will fix in the next commit

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/net/NetInterfaceConfigMode.java
ibinshtok: Yes, I will define the new value Unmanaged both for the interface status and the interface mode but I think we will have to discuss this further ...

kura/org.eclipse.kura.net.admin/META-INF/MANIFEST.MF
ibinshtok: will fix this in the next commit

kura/org.eclipse.kura.net.admin/src/main/java/org/eclipse/kura/net/admin/visitor/linux/IfcfgConfigWriter.java
ibinshtok: @cdealti I think we need to discuss this further ... On the one hand, setting either interface 'Status' or/and 'Configuration' to 'Unmanaged' implies that this interface will not be touched by Kura at all so that another network manager can handle it. And there may be network managers that are fully capable to configure ethernet and wireless interfaces. 
On the other hand, c5921 expects the following:
1. Interfaces controlled by c5921 (i.e. eth, wlan) are configured in the 'manual' (i.e. BOOTPROTO=none) mode in Linux but there is nothing within c5921 to do it. So it is desirable for Kura to be able configure those interfaces in 'manual' mode all the way to the /etc/network/interfaces file.
2. Wireless interface can't be configured by c5921, so that Kura needs to be able to configure hostapd and wpa_supplicant using the 'Wireless' tab.
3. We may also want to set NAT rules (but not DHCP server) in Linux for interfaces controlled by c5921. So that having NAT tab for "unmanaged" interface may be desirable. For example I could not make PPP work w/o masquerading. 

So maybe we need to handle both options? That is:
1. Leave interface alone and don't manage it at all.
2. Manage it but configure w/o boot protocol that is in 'manual' mode.

Let's discuss this further tomorrow ...

kura/org.eclipse.kura.net.admin/src/main/java/org/eclipse/kura/net/admin/visitor/linux/IfcfgConfigReader.java
ibinshtok: If I keep ipAddress is set to null on line 113, Lint suggests to "Remove this useless assignment to local variable "ipAddress"." I think this happens because ipAddress gets set from kuraProps at line 165 as shown below:
ipAddress = kuraProps.getProperty("IPADDR");

kura/org.eclipse.kura.net.admin/src/main/java/org/eclipse/kura/net/admin/visitor/linux/IfcfgConfigWriter.java
cdealti: @ibinshtok now I see what you mean!

However, manual mode may require more than setting e.g. ``iface eth1 inet manual`` (on Debian systems) so I'm not sure it's really worth in general, for example if one wants this in its /etc/network/interfaces:

```
auto eth0
iface eth0 inet manual
       up ifconfig $IFACE 0.0.0.0 up
       up /usr/local/bin/myconfigscript
       down ifconfig $IFACE down
```

would this PR allow this by setting the interface mode to "manual" and then preserve further edits to /etc/network/interfaces by the user?

Wouldn't be better to just not manage the interface configuration at all from ESF (Unmanaged) and ask the user to edit /etc/network/interfaces to add the manual configuration or whatever they like.

For production, you can seed a custom /etc/network/interfaces file and snapshot_0.xml as a post install phase.

ibinshtok: @cdealti No, my PR would not be able to handle advanced entries in the /etc/network/interfaces file. And you are right, those entries may be desired. So, if we leave the /etc/network/interfaces file alone, I think it would be the best to add the 'Unmanaged' option to interface 'Status' field only. So that if 'Status' is set to 'Unmanaged' all other fields in the 'TCP/IP' configuration tabs are disabled. In this case, interface are not going to be managed at all, except ensuring that it is up after 'Status' is switched to 'Unmanaged'. Do you agree?

ibinshtok: @cdealti Another question ... What are we going to do with 'Wireless' and 'DHCP & NAT' tabs if interface 'Status' is set to 'Unmanaged'? 

For the 'Wireless' tab:
Since IOS provides no way to configure hostapd and wpa_supplicant, it would be nice to allow Kura to do this if Kura's capabilities to configure wireless are sufficient. In other words, if all that customers want is WPA/WPA2 PSK, they can use Kura. If they need something like EAP-TLS, they need to handle it outside Kura. With this in mind, I recommend to add 'Unmanaged' option to the 'Wireless Mode' configuration field of the 'Wireless' tab. Do you agree with this?

For the 'DHCP & NAT' tab:
If the interface mode is set to 'Unmanaged', I would keep this tab enabled (at least for now) but default settings for the 'Router Mode' option would be 'Off'.

cdealti:
Closing this as it needs more thoughts on the requirements. Thanks

PR #1674, opened 2017-09-14
Soved issue with fillPropertiesFromConfiguration GwtServerUtil

### LINKED ISSUE #1661: The driver configuration is not applied
When a driver configuration is applied from the "Drivers and Assets" tab, the properties are indeed applied. 
screen shot 2017-09-11 at 10 25 31

However, if the configuration is applied from the "Driver" tab in the Wires graph, the new configuration is not applied to the driver.
screen shot 2017-09-11 at 10 25 52

It seems that it happens when the "id" and "name" attribute in the metatype are different. When they have the same name, all work fine. Probably, the "Drivers and Assets" tab reads the "id", while the Wires graph reads the "name" attribute causing this problem.

MMaiero:
It was taken the name value instead of the id.

Closes #1661
As a side note, I believe we should try to uniform the update code to reduce duplication and issues like this: wires update takes a completely different path than, for example, the usual Service configuration update.

### DIFF
--- a/kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/server/util/GwtServerUtil.java
+++ b/kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/server/util/GwtServerUtil.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2016 Eurotech and/or its affiliates and others
+ * Copyright (c) 2016, 2017 Eurotech and/or its affiliates and others
  *
  * All rights reserved. This program and the accompanying materials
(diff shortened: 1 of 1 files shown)


kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/server/util/GwtServerUtil.java
```diff
@@ -328,13 +328,13 @@ private GwtServerUtil() {
         if (backupCC == null) {
```
amitjoy, line 328: I believe you might need to update the copyright year for consistency.

PR #1691, opened 2017-09-20
Fixed possible NPE in DeploymentHookManager

nicolatimeus:

### DIFF
--- a/kura/org.eclipse.kura.core.deployment/src/main/java/org/eclipse/kura/core/deployment/hook/DeploymentHookManager.java
+++ b/kura/org.eclipse.kura.core.deployment/src/main/java/org/eclipse/kura/core/deployment/hook/DeploymentHookManager.java
@@ -18,8 +18,9 @@ import java.util.Properties;
 
 import org.eclipse.kura.configuration.ConfigurationService;
 import org.eclipse.kura.deployment.hook.DeploymentHook;
+import org.osgi.framework.BundleContext;
+import org.osgi.framework.FrameworkUtil;
 import org.osgi.framework.ServiceReference;
-import org.osgi.service.component.ComponentContext;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -29,7 +30,7 @@ public class DeploymentHookManager {
 
     private final Map<String, DeploymentHook> registeredHooks = new HashMap<>();
     private final Map<String, HookAssociation> associations = new HashMap<>();
-    private ComponentContext context;
+    private BundleContext bundleContext;
 
     public synchronized void bindHook(ServiceReference<DeploymentHook> hook) {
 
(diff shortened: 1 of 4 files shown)


kura/org.eclipse.kura.core.deployment/src/main/java/org/eclipse/kura/core/deployment/hook/DeploymentHookManager.java
```diff
@@ -67,16 +68,17 @@ public synchronized void unbindHook(ServiceReference<DeploymentHook> hook) {
...
             logger.info("Hook unregistered: {}", hookId);
         }
 
     }
 
-    public synchronized void activate(ComponentContext context) {
-        logger.info("activating...");
-        this.context = context;
-        logger.info("activating...done");
+    private BundleContext getBundleContext() {
```
amitjoy, line 77: instead of having to do it through FrameworkUtil, you can add an activate method with BundleContext as argument. BundleContext instance will be injected by the runtime.

nicolatimeus: The activate method was perviously there, with this PR I've removed it and switched to FrameworkUtil because DS can call bindHook() before activate(), this causes an NPE because the context field is null until the component is activated.

amitjoy: Yeah you are right. I missed that. `bindHook()` is a service binding callback which is always called before activate. Instead of using `bindHook(ServiceReference<DeploymentHook> hook)` you can also use `bindHook(ServiceReference<DeploymentHook> hook, DeploymentHook deploymentHook)`. That would be beneficial instead of calling `context.getService()` inside `bind()`. In this case, you don't have to worry about resetting the service count by invoking `BundleContext#ungetService(ServiceReference)`

nicolatimeus: Yes you are right, I'm using this signature for `bindHook()` becacuse I'm performing some validation on the `kura.service.pid` of the hook and I'm getting the service only if succeeds.

Switching to `bindHook(ServiceReference<DeploymentHook> hook, DeploymentHook deploymentHook)` would cause `getService()` to be called by DS even if the hook has an invalid `kura.service.pid`.

I'm performing this validation because `kura.service.pid` can be null for an hook if the property is not explicitly set in the component definition and `DeploymentHookManager` receives a reference to it before `ConfigurationService`, (this happened during my tests).

When you are done with the review can you please merge the PR? We are a bit in a hurry for the release.

MMaiero:
@nicolatimeus Please update the release notes

PR #1699, opened 2017-09-26
Setting an higher value for the GPIO pins.

MMaiero:


kura/examples/org.eclipse.kura.example.gpio/OSGI-INF/metatype/org.eclipse.kura.example.gpio.GpioComponent.xml
```diff
@@ -39,7 +39,7 @@
             required="false"
             default="1"
             min="0"
-            max="255"
+            max="1255"
```
cdealti, line 42: I believe we could remove max entirely

PR #1719, opened 2017-10-03
Changes to dp generation for examples and drivers.

MMaiero:
Drivers and wire script filter dps are generated in kura/feature.
Examples dps are generated in the specific project's pom files.

The changes are done to keep compatibility with existing dps published in Marketplace.
In this way we simplify updating.


MMaiero:
@nicolatimeus Please have a look if that fits better

nicolatimeus (approved):

Issue #1725, opened 2017-10-05
[Kura 3.1.0 QA] Cloud Services GUI crashes


markoer:
Certain conditions (e.g. removed deployment packages that still have some configuration) cause Cloud Services UI to crash with error:

![image](https://user-images.githubusercontent.com/22816717/31233665-6fc69dc4-a9ed-11e7-8c4e-ad245bfaee86.png)

The exception in the log is:
```
WARN  o.e.k.w.s.u.KuraExceptionHandler - RPC service non-application error
org.eclipse.kura.web.shared.GwtKuraException: INTERNAL_ERROR
	at org.eclipse.kura.web.server.util.ServiceLocator.applyToServiceOptionally(ServiceLocator.java:119)
	at org.eclipse.kura.web.server.GwtComponentServiceImpl.createMetatypeOnlyGwtComponentConfiguration(GwtComponentServiceImpl.java:536)
	at org.eclipse.kura.web.server.GwtComponentServiceImpl.findFilteredComponentConfigurationsInternal(GwtComponentServiceImpl.java:396)
	at org.eclipse.kura.web.server.GwtComponentServiceImpl.findFilteredComponentConfigurations(GwtComponentServiceImpl.java:100)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:497)
	at com.google.gwt.user.server.rpc.RPC.invokeAndEncodeResponse(RPC.java:587)
	at com.google.gwt.user.server.rpc.RemoteServiceServlet.processCall(RemoteServiceServlet.java:333)
	at com.google.gwt.user.server.rpc.RemoteServiceServlet.processCall(RemoteServiceServlet.java:303)
	at com.google.gwt.user.server.rpc.RemoteServiceServlet.processPost(RemoteServiceServlet.java:373)
	at com.google.gwt.user.server.rpc.AbstractRemoteServiceServlet.doPost(AbstractRemoteServiceServlet.java:62)
	at javax.servlet.http.HttpServlet.service(HttpServlet.java:707)
	at org.eclipse.kura.web.server.OsgiRemoteServiceServlet.service(OsgiRemoteServiceServlet.java:41)
[...]

PR #1726, opened 2017-10-05
Integration of PR #1697 in order to prevent possible NPE.

MMaiero:
NPE could be generated when scanning for component instances.
When, for example, a dp is deleted without killing first a factory instance, scanning of services
could lead to a NPE when analysing the component configuration.
This fix will simply skip those incomplete component configurations.

Closes also #1725


markoer (approved):

MMaiero:
Ok, merging then.

PR #1728, opened 2017-10-06
Fixed a bug in the parseDhclientLeaseBlock() method.

ibinshtok:

### DIFF
--- a/kura/org.eclipse.kura.linux.net/src/main/java/org/eclipse/kura/linux/net/dns/LinuxDns.java
+++ b/kura/org.eclipse.kura.linux.net/src/main/java/org/eclipse/kura/linux/net/dns/LinuxDns.java
@@ -200,7 +200,7 @@ public class LinuxDns {
             String fixedAddressMatch) throws UnknownHostException {
 
         if ((leaseBlock.size() > 2)
-                && (!leaseBlock.get(0).equals(interfaceMatch) || leaseBlock.get(1).equals(fixedAddressMatch))) {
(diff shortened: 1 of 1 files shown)


ibinshtok:
@cdealti I think I fixed DNS display issue. Please take a look. Note that more DNS related testing is required. As for the gateway issue, I believe the problem is that gateway information for an interface is obtained from respective /tmp/.kura/coninfo-<iface> file which is updated by /etc/network/if-down.d/ifdown-local and /etc/network/if-up.d/ifup-local scripts on 'ifdown' and 'ifup' commands. So the /tmp/.kura/coninfo-wlan0 file doesn't get updated properly since we are not doing ifup if interface is configured as DHCP client. I think we probably should not rely on the org.eclipse.kura.linux.net.ConnectionInfoImpl and obtain gateway IP address from 'ip route show dev <iface>', maybe for the next release. Let's discuss this on Monday.

cdealti (changes requested):
It look like this fixes a regression introduced with the refactoring made in 451370bac18cf527fa2733e43a8aa03ba77eb8ad and that the proposed fix somewhat reflects the intention of the original code. However I would take a look to the overall parser.
1. What happens if the size of leaseBlock is just two? The first check will be skipped but the parser may still find elements containing the string `domain-name-servers`. Is this correct?
2. Related to the above, why parsing all the elements looking for the string `domain-name-servers` instead of continuing from where you have just left?
3. Related to the above, the first two elements must be found at fixed positions of leaseBlock. I wonder if leaseBlock can contain empty string elements that should be skipped.  
4. Anything against using `String.contains` instead of `String.indexOf`?

PR #1729, opened 2017-10-09
Aligned dev-env kura.properties file to the one used in devices.

MMaiero:

### DIFF
--- a/kura/emulator/org.eclipse.kura.emulator/src/main/resources/kura.properties
+++ b/kura/emulator/org.eclipse.kura.emulator/src/main/resources/kura.properties
@@ -1,5 +1,5 @@
 #
-# Copyright (c) 2011, 2016 Eurotech and/or its affiliates
+# Copyright (c) 2011, 2017 Eurotech and/or its affiliates
 #
 #  All rights reserved. This program and the accompanying materials
 #  are made available under the terms of the Eclipse Public License v1.0
@@ -15,6 +15,7 @@
 ## -----------------------------------------------------------------------------
 kura.name=Kura
 kura.version=kura_emulator
+kura.marketplace.compatibility.version=
 kura.company=Eclipse
 kura.project=Dev
 kura.platform=dynacor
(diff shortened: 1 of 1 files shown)


kura/emulator/org.eclipse.kura.emulator/src/main/resources/kura.properties
```diff
@@ -25,6 +26,7 @@ kura.partNumber=DevPartNumber
 kura.serialNumber=DevSerialNumber
 kura.bios.version=DevBiosVersion
 kura.firmware.version=DevFirmwareVersion
+kura.primary.network.interface=eth0
```
cdealti: I believe this property is ignored on the Mac which always use en0 :-( Please add a comment.

kura/emulator/org.eclipse.kura.emulator/src/main/resources/kura.properties
```diff
@@ -34,6 +36,9 @@ kura.tmp=/tmp/kura/tmp
 kura.snapshots=/tmp/kura/snapshots
 kura.snapshots.count=10
 kura.have.net.admin=false
+# kura.keystore.password=password
+# kura.ssl.keyStorePassword=password
+kura.wifi.top.channel=11
```
cdealti: Why do we need to set kura.wifi.top.channel in the emulator?

kura/emulator/org.eclipse.kura.emulator/src/main/resources/kura.properties
```diff
@@ -34,6 +36,9 @@ kura.tmp=/tmp/kura/tmp
 kura.snapshots=/tmp/kura/snapshots
 kura.snapshots.count=10
 kura.have.net.admin=false
+# kura.keystore.password=password
+# kura.ssl.keyStorePassword=password
```
cdealti: Why adding these two properties?

kura/emulator/org.eclipse.kura.emulator/src/main/resources/kura.properties
MMaiero: In this case is just to align to the configurations we actually have.
For example, now the emulator configuration is close to the nn profiles we have. This will simplify a lot the maintenance.

kura/emulator/org.eclipse.kura.emulator/src/main/resources/kura.properties
MMaiero: Done

Issue #1268, opened 2017-03-27
[WebUI] Usability Improvements


mcarrer:
## Overall
- [x]  When making a selection in the navigation bar, can we avoid the little shift of the icons?
- [x]  Use the same font, weight and size for System and Services in the Navigation bar as the Title of each page
- [x]  For the title of each page, move the text a bit to the left (padding 3px); same for the instruction text underneath
- [x]  Rearrange the footer to be in a single line
- [x]  Move Refresh Button to last in the toolbar
- [x]  In the navigation bar, can we align the all the texts by giving the same width to all icons through extra padding if necessary? So move to a table-like layout?

## Device

- [x]  Command * Change the order of the buttons * Execute | Reset
- [x]  Profile | Threads | System Properties * keep consistent the style of the table
- [x]  Bundles * Move Refresh button last

## Network
- [x]  Align the left margin with the other panels. Now it is more to the right than the others
- [x]  Align the Interface Name panel top with the interface management box
- [x] Align the Help Text with the first widget on the left
- [x] The text in the "Help Text" box can be increased to 12px
- [x] Change "Help Text" to "Information" and make it size 14px like all the headers within a configuration box
- [x]  The size of the Labels in the TCP/IP tab * Status, Configure, ... * should be the same as the one in the Hardware Tab
- [x]  TCP/IP and Hardware tab font size should be 14px like in the Device and Firewall * let's keep the Tab name size consistent

## Firewall
- [x] There is an extra space above the toolbar * keep spacing consistent with the Device > Bundles UI
* I do not like the Apply button. If we batch the changes, we may need to have a Cancel as well. Or we can make we change effective after it is made.

## Cloud Services
- [x]  Add some instruction text at the top of the page
- [x]  Move Refresh Button to last
* In my build, I am not able to build a second CloudService instance
- [x]  In the new Cloud Service dialog, remove the icons from the buttons and just have them as Submit and Cancel
* Maybe add a Title over the Tabs for the currently selected CloudService?
* CloudService
    - [x]  device.display.name * Change tip to remove reference to Eurotech Everware Cloud. Say "Server defined relies on the remote management server to define a name."
    - [x]  topic.control-prefix * Topic prefix for system and device management messages
    - [x]  enable.default.subscriptions * Disables the default subscriptions to the gateway management MQTT topics. When disabled, the gateway will not be remotely manageable
* MQTTDataTransport
    - [x]  topic.context.account-name * For connections to remote management servers, this is generally the name of the server side account.
[...]

PR #1740, opened 2017-10-18
Web Usability Improvements

MMaiero:
Related to the changes requested in Issue #1268


nicolatimeus:
The remaining open points from #1268 should be:

* Maybe add a Title over the Tabs for the currently selected CloudService?
* Move Wire Components selection to a New (with drop down) button. This frees up space for the canvas

For the latter one, maybe we should perform a mockup session to define the desired result before proceeding with implementation.

MMaiero:
@nicolatimeus Let's keep those items on hold. I see them as low priority.

PR #1751, opened 2017-10-24
Blacklist network interfaces from the net admin

ibinshtok:


kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/system/SystemService.java
```diff
@@ -46,6 +46,10 @@
     public static final String KEY_KURA_NAME = "kura.name";
     public static final String KEY_KURA_VERSION = "kura.version";
     /**
+     * @since
```
MMaiero: I believe a version needs to be set here

kura/org.eclipse.kura.core/OSGI-INF/system.xml
```diff
@@ -18,10 +18,12 @@
       <provide interface="org.eclipse.kura.system.SystemService"/>
    </service>
    <property name="service.pid" value="org.eclipse.kura.system.SystemService"/>
+<!--
```
MMaiero: If it is not needed anymore, let's remove it

MMaiero:
Please have also a look why the tests are failing. Thanks

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/system/SystemService.java
ibinshtok: done

kura/org.eclipse.kura.core/OSGI-INF/system.xml
ibinshtok: done

ibinshtok:
I committed my Lint fixes and tested on RG 10-20 and 20-25. It worked fine but my changes created a few test errors in org.eclipse.kura.core.ssl.test and org.eclipse.kura.core.system.test. @markoer, could you take a look at this tomorrow if you have a chance, please?

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/system/SystemService.java
```diff
@@ -46,6 +46,10 @@
     public static final String KEY_KURA_NAME = "kura.name";
     public static final String KEY_KURA_VERSION = "kura.version";
     /**
+     * @since 3.2
```
MMaiero: This is usually related to the API package version. So, it should be 1.4.0

markoer (commented):
Which PR/functionality would cover Kura not configuring eth0 as DHCP during installation? If this one, then installation script would have to check this new property kura.net.admin.interface.blacklist and update the existing /etc/network/interfaces instead of overwriting it. Assuming, of course, that the installation package would already contain the updated kura.properties...
So, where does this PR end and some other begin?

markoer (approved):

MMaiero:
@markoer tests are still failing. Are we planning on having changes in them before merging?
Thanks

kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/ssl/SslManagerServiceImpl.java
```diff
@@ -240,17 +239,14 @@ public SSLSocketFactory getSSLSocketFactory(String protocol, String ciphers, Str
...
     public void installTrustCertificate(String alias, X509Certificate x509crt)
             throws GeneralSecurityException, IOException {
-        InputStream tsReadStream = null;
-        FileOutputStream tsOutStream = null;
-
-        try {
+        String trustStore = this.options.getSslKeyStore();
+        File fTrustStore = new File(trustStore);
+        try (InputStream tsReadStream = new FileInputStream(trustStore);
+                FileOutputStream tsOutStream = new FileOutputStream(trustStore)) {
```
markoer: It seems not to be a good idea to open the same file for reading and writing at the same time. If you move the output stream back down, it should work. You can add another try-with-resources, there.

kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/ssl/SslManagerServiceImpl.java
```diff
@@ -260,36 +256,26 @@ public void installTrustCertificate(String alias, X509Certificate x509crt)
...
 
     @Override
     public void deleteTrustCertificate(String alias) throws GeneralSecurityException, IOException {
-        InputStream tsReadStream = null;
-        OutputStream tsWriteStream = null;
 
-        try {
+        String trustStore = this.options.getSslKeyStore();
+        try (InputStream tsReadStream = new FileInputStream(trustStore);
+                OutputStream tsWriteStream = new FileOutputStream(trustStore)) {
```
markoer: See the comment above.

kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/ssl/SslManagerServiceImpl.java
```diff
@@ -298,17 +284,15 @@ public void installPrivateKey(String alias, PrivateKey privateKey, char[] passwo
             throws GeneralSecurityException, IOException {
         // Note that password parameter is unused
 
-        InputStream tsReadStream = null;
-        FileOutputStream tsOutStream = null;
-
-        try {
+        String keyStore = this.options.getSslKeyStore();
+        try (InputStream tsReadStream = new FileInputStream(keyStore);
+                FileOutputStream tsOutStream = new FileOutputStream(keyStore)) {
```
markoer: See the comment above.

markoer (changes requested):
Once these and Matteo's proposed changes are implemented, I approve.

markoer:
I see reviews are not really visible per-commit... So, I approve the changes after the code is fixed. The tests are still OK with the proposed code updates.
And the `@since` javadoc still needs to be updated, too.

kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/ssl/SslManagerServiceImpl.java
ibinshtok: Yes, I've already rolled this back on Friday w/o committing since I did not understand what exactly happens. Thanks!

ibinshtok:
@cdealti please take a loot at this PR. At this point, I believe, the only issue is in the org.eclipse.kura.core.test.AllCoreTests as shown below. It looks like the setConfigService() method is not being invoked and as a result the s_configService is null in the setUpClass() method. But I don't know why.

[INFO] org.eclipse.kura.core.system.test .................. SUCCESS [  7.036 s]
[INFO] org.eclipse.kura.core.test ......................... FAILURE [ 20.662 s]

ibinshtok@T560:~/eurotechdev/kura-ilya/kura/kura/test/org.eclipse.kura.core.test/target/surefire-reports$ more org.eclipse.kura.core.test.AllCoreTests.txt
-------------------------------------------------------------------------------
Test set: org.eclipse.kura.core.test.AllCoreTests
-------------------------------------------------------------------------------
Tests run: 2, Failures: 0, Errors: 2, Skipped: 0, Time elapsed: 11.223 sec <<< FAILURE! - in org.eclipse.kura.core.test.AllCoreTests
org.eclipse.kura.core.test.AllCoreTests  Time elapsed: 11.198 sec  <<< ERROR!
java.lang.Exception: Failed to reconfigure the broker settings - failing out
	at org.eclipse.kura.core.test.AllCoreTests.setUpClass(AllCoreTests.java:94)

org.eclipse.kura.core.test.AllCoreTests  Time elapsed: 11.222 sec  <<< ERROR!
java.lang.NullPointerException: null
	at org.eclipse.kura.core.test.AllCoreTests.tearDownClass(AllCoreTests.java:149)

cdealti:
@ibinshtok before starting or restarting the review, I'd like to restate the requirements.

1. If and interface is blacklisted, it will be filtered out by Kura at the lowest possible layer. In principle I would not expect changes to the Web UI or other parts of Kura. The interface is simply not available to the net admin or to applications.

2. Kura must preserve the configuration for the interfaces it does not manage. For example, if an interface usb0 is blacklisted, Kura must preserve its configuration in /etc/network/interfaces or other configuration files.

For the time being I would only focus on that.

markoer (approved):

markoer:
And Cristiano's requirements still stand, despite the approval of changes ;)

ibinshtok:
@cdealti The changes to the Web UI I made are not functional changes just an attempt to fig Lint issues. I touched Web UI files because of Lint issues addressed in the SystemServiceImpl. 
Functional changes are in the SystemServiceImpl and NetworkServiceImpl in the getAllNetworkInterfaceNames() method.

cdealti:
@ibinshtok I really appreciate the effort of writing the Python script for preserving the network configurations during installation. However I wonder if we could keep it on hold for a while.

I'd really like if we could avoid installing system configuration with the installer files and having Kura generating the them at the first boot.

The general idea is that Kura should be able to configure everything from the configuration snapshot.

So for the time being, I'd revert the Python addition.

cdealti:
@ibinshtok @markoer  this seems to be no longer needed.
Can we close it and reopen if needed?

markoer:
The only thing missing in the unmanaged interfaces is the script that merges the `/etc/network/interfaces` files during Kura installation. With it installation doesn't overwrite the interfaces, that are not expected on the system, e.g. additional usb2eth interfaces.

cdealti:
> The only thing missing in the unmanaged interfaces is the script that merges the /etc/network/interfaces files during Kura installation.

This requirement is quite unique. Usually the installer just replaces/overwrites existing configuration files. I'm not sure whether we should make an exception to this rule.
Maybe, if this is needed, this is the job of an external script

Issue #1700, opened 2017-09-27
[Kura 3.1.0 QA] Rollback doesn't restore deleted services


markoer:
Encountered during a variation of DEN-SNAP-002 on Kura 3.1.0 RC b40 on RPi3.

* Add a H2DbServer service. Configure it and apply changes.
* Make sure Java listens on the configured port.
* Delete it.
* Go to System->Settings and Rollback to the previous snapshot.
* Refresh page.
* The deleted service is not in the Services list and not started (no listener on the TCP port), but probably should be
* Restart Kura.
* Service is now back.

cdealti:
@markoer we will add this to the list of known issues.

markoer:
OK

markoer:
Very good. When the code is merged and build is available, I'll check it again and close.

markoer:
Verified in 3.2.0-SNAPSHOT b1242. Closing.

PR #1759, opened 2017-10-31
Improved snapshot rollback

### LINKED ISSUE #890: [Kura 2.1.0 QA] Snapshot management in web UI
DEN-SNAP-002 from #788 requires snapshot to be rolled back. The UI mentions possibility of a timeout after which manual refresh is necessary.

The issue is that the timeout probably didn't occur - it was definitely not indicated in any way in the UI, and the operation was rather fast - but UI data was not refreshed. ExamplePublisher's configuration remained as it was before the rollback until manual page refresh. And that happened every time I tried.

The same goes for snapshot file upload.

Kura 2.1.0 build 12, Raspberry Pi 3

nicolatimeus:
Closes #890 and #1700


nicolatimeus:
@markoer can you please have a look at the failing test? The code it triggers is no longer there after my changes so I wouldn't know how to fix it.. maybe can we just remove it?

markoer:
If the update I pushed works, the test can stay in there.

cdealti:
@markoer can you please review?

kura/test/org.eclipse.kura.core.test/src/main/java/org/eclipse/kura/core/test/ConfigurationServiceTest.java
```diff
@@ -35,6 +35,7 @@
 import org.eclipse.kura.core.configuration.XmlSnapshotIdResult;
```
markoer, line 35: Update the (c) years.

markoer (approved):
It fixes the issues it promises to fix. Just update the year.

Issue #1766, opened 2017-11-06
GSM modems not correctly detected after a reset


pierantoniomerlino:
A GSM modem that provides also net interfaces (i.e. usb0) are not correctly detected after a reset. This is due to the fact that the NetworkServiceImpl is able to detect only tty and block resources. When a net resource added event triggers the NetworkServiceImpl, the resource is added to the block list and the modem is not recognized.

PR #1767, opened 2017-11-06
Fixed modem detection in NetworkServiceImpl

pierantoniomerlino:
This PR fixes the #1766. Now the UsbDeviceEvent contains also the type (NET, TTY or BLOCK) of the device that triggers the event. When the NetworkServiceImpl is triggered, it is capable to recognize the type of device and add it to the correct list.


kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/usb/UsbDeviceAddedEvent.java
```diff
@@ -102,4 +102,14 @@ public String getUsbBusNumber() {
     public String getUsbDevicePath() {
         return (String) getProperty(USB_EVENT_DEVICE_PATH_PROPERTY);
     }
+
+    /**
+     * Returns the USB device type.
+     *
+     * @return
```
MMaiero: I know that it has not been specified before, but I would like to document that it returns null (if we don't want to manage that differently) if the property is not available.

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/usb/UsbDeviceRemovedEvent.java
```diff
@@ -102,4 +102,14 @@ public String getUsbBusNumber() {
     public String getUsbDevicePath() {
         return (String) getProperty(USB_EVENT_DEVICE_PATH_PROPERTY);
     }
+
+    /**
+     * Returns the USB device type.
+     *
+     * @return
```
MMaiero: Same as before

MMaiero (approved):

PR #1775, opened 2017-11-08
ISSUE_TEMPLATE proposal

MMaiero:
This is a proposal for the definition of an ISSUE_TEMPLATE file that should simplify the process of opening a new pull request and will simplify the process of issue classification.

### DIFF
--- a/ISSUE_TEMPLATE.md
+++ b/ISSUE_TEMPLATE.md
@@ -1,0 +1,35 @@
+Please describe the issue you are encountering in Kura.
+
+Fill the following form and provide a textual description if you are reporting a bug. 
+Provide only a textual description otherwise.
+
+- Component affected:
+  - [ ] Distribution
+    - Installer name (e.g. kura_3.1.0_raspberry-pi-2-3_installer.deb):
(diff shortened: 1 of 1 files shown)


ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,28 @@
+Please describe the issue you are encountering in Kura:
+
```
MMaiero: We can add an initial comment saying:
"Fill the following only in case you are reporting a **bug**. Provide a textual description otherwise."

markoer:
Maybe include browser and its version somewhere.

MMaiero:
@markoer Good idea. But I don't know where to place it in order to be meaningful.

markoer:
It could likely apply to all but Development environment and Oomph. Maybe just add another top-level item as a sibling to `Component affected`.

MMaiero:
I've tried to add it as a child of component affected. Have a look if it fits. Otherwise, like you suggested, we can move it as a sibling of Component affected.

markoer (approved):

PR #1779, opened 2017-11-09
Modified Web Ui to show status panel on load

### LINKED ISSUE #1758: [Web UI] Initial page view on a narrow screen
Open a browser and reduce width so that Web UI will show menu for narrow devices. Reload the page and the result is this:

image

Menu remains open with a modal over the content that makes one think it doesn't work. It might be better to close the menu initially, after Kura loads.

3.2.0-SNAPSHOT on RPi3 and Firefox.

nicolatimeus:
* The Status panel is now shown on page load
* The sidenav is now hidden by default on small screens

Closes #1758 


markoer (approved):
Looks great to me.

Issue #1747, opened 2017-10-20
[Kura 3.1.0] Build error due to exists-plugin


MMaiero:
Currently, trying to build Kura 3.1.0 from the ```KURA_3.1.0_RELEASE``` tag results in a build error like shown below:

```
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time: 25.772 s
[INFO] Finished at: 2017-10-16T15:15:45+00:00
[INFO] Final Memory: 24M/397M
[INFO] ------------------------------------------------------------------------
[ERROR] Failed to execute goal org.honton.chas:exists-maven-plugin:0.0.3:remote (default) on project target-platform: Artifact already exists in repository: org.eclipse.kura:target-platform:3.1.0/target-platform-3.1.0.pom -> [Help 1]
[ERROR] 
[ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch.
[ERROR] Re-run Maven using the -X switch to enable full debug logging.
[ERROR] 
[ERROR] For more information about the errors and possible solutions, please read the following articles:
[ERROR] [Help 1] http://cwiki.apache.org/confluence/display/MAVEN/MojoExecutionException
```

The workaround for this exception is to disable the ```exists-maven-plugin``` both from target-platform/pom.xml and from kura/pom.xml.

This will make the build working again.

kartben:
Is there an ETA for this? I would hope that this is high priority on the list for the Kura team, as not being able to build a *released* tag usually doesn't send the right message to the community. But it's good to know there is a workaround.

kartben:
I am not sure what to make of the silence on this issue :-)
Jens, maybe you can go ahead and merge your PR, tag the repo as 3.1.1, and someone else can take care of effectively releasing the bits? Would that work for everyone?

ctron:
Sure I can do that by tomorrow.

MMaiero:
Hello everyone,
I'll try to formalise here our idea on how to solve this issue. The constraints we see are the following:
- we would like to preserve the existing tag. That is the effective tag used to perform the 3.1.0 release and that should be kept as is.
- we would like to not modify the release branch. Indeed, since the changes are already merged into develop, the branch could be easily removed without loosing commits and preserving the linearity of the contributions.
[...]

PR #1782, opened 2017-11-13
#1747: Fix the build of a release version

ctron:
This change moves the "exists-maven-plugin" plugin into a separate
profile in order to not break the build when rebuilding a released
version of Kura.

### DIFF
--- a/kura/pom.xml
+++ b/kura/pom.xml
@@ -260,6 +260,31 @@
                 <module>org.eclipse.kura.web2</module>
             </modules>
         </profile>
+
+        <profile>
+            <id>release</id>
+            <build>
+                <plugins>
+                    <plugin>
+                        <groupId>org.honton.chas</groupId>
+                        <artifactId>exists-maven-plugin</artifactId>
+                        <version>0.0.3</version>
+                        <executions>
+                            <execution>
+                                <goals>
+                                    <goal>remote</goal>
+                                </goals>
+                                <configuration>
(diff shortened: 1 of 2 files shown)


MMaiero:
Hi @ctron,
we included a fix in develop that will leverage a specific profile to disable the exists plugin.
We, also, documented it here: https://github.com/eclipse/kura#exists-plugin-failure
Is it ok for you?
Do you think we can close the PR?

ctron:
No. I think a released version of Kura should be re-buildable without any flags.

dwoodard1:
@ctron What is your concern with using flags to reproduce the build? It is not clear if there is a technical advantage for using your approach.

ctron:
As far as I understood it is necessary to add additional flags to the build when the user wants to re-compile a released version of Kura.

In most cases people will probably simply to compile Kura. Run into problems. And then try to figure out why they ran into problems. Looking at the wiki it also doesn't mention anything about that: https://wiki.eclipse.org/Kura/Getting_Started 

My expectation of an open source project would be that a released version of an open source project can simple be rebuilt. The standard maven workflow for this is e.g. `mvn install`, Kura provides a `build-all.sh` script.

My suggestion would be to disable this plugin by default. And only enable this profile on machines which could possible run into the case of accidentally uploading already released versions.

MMaiero:
Agreed. We'll change the behaviour in 3.2.0

MMaiero:
Closing as already covered by #1925

PR #1784, opened 2017-11-13
Added refresh button in Wire Composer

### LINKED ISSUE #1776: [Wires] Add a refresh button in Wire Composer
Currently there is no way for refreshing the wire graph other that selecting another service tab and then switching back to the wire composer.
A refresh button can be added for this purpose, this button will also allow to undo non applied changes.

nicolatimeus:
Closes #1776 


MMaiero:
I've tried to have a look but it seems that, if you place a component and, before saving, reset the status, the apply button will keep having the *. And trying to move to other tabs will trigger the alert about the unsaved changes. But there the save button is disabled.

screen shot 2017-11-14 at 12 31 20

MMaiero (approved):

PR #1818, opened 2017-11-28
[Emulator] Changed default file.upload.size.max to -1 in kura.properties

samizel:
[Emulator] Changed default file.upload.size.max to -1 in kura.properties

Signed-off-by: Andreas Wojtok <Andreas.Wojtok@gmail.com>


dwoodard1:
@samizel Thanks for the fix!

PR #1842, opened 2017-12-19
Improved type conversions in OPCUA driver

### LINKED ISSUE #1840: [OPC-UA] Improve type conversions in OPC-UA driver
In case of a write, the OPC UA driver currently performs an automatic mapping between Java types and OPC UA types, this mapping is performed automatically by the Milo library.
The mapping is performed as follows:

Java  -> OPCUA

BOOLEAN -> Boolean
DOUBLE -> Double
FLOAT -> Float
INTEGER -> Int32
LONG -> Int64
STRING -> String
BYTE_ARRAY -> SByte[]

This implies that only a limited subset of the OPCUA types is mapped, and writing to not covered type variables will likely result in an error. For example this problem involves writing unsigned OPCUA types such as UInt16, UInt32 and UInt64.

Reading data using the BYTE_ARRAY Java type is currently implemented by converting the objects describing the read result received from Milo to a byte array using Java serialisation (ObjectOutputStream), this produces incorrect results.

nicolatimeus:
This PR adds an additional property in OPCUA driver channel configuration, named `opcua.type`.
This property is considered only in case of a write and is ignored otherwise, it allows to specify the type of the OPCUA variable to be written.
The driver will attempt to convert the Java value to be written to the specified type. The allowed values for this property are the following:

* DEFINED_BY_JAVA_TYPE
* BOOLEAN
* SBYTE
* INT16
* INT32
* INT64
* BYTE
* UINT16
* UINT32
* UINT64
* FLOAT
* DOUBLE
* STRING
* BYTE_STRING
* BYTE_ARRAY
* SBYTE_ARRAY

This PR should extend the set of types that can be written using the driver to the list above, types not included in the list are still unsupported.

The default value for this parameter is DEFINED_BY_JAVA_TYPE, if this value is selected, the driver will perform the conversion automatically as in the current implementation, as follows:

Java -> OPCUA
BOOLEAN -> Boolean
DOUBLE -> Double
FLOAT -> Float
INTEGER -> Int32
LONG -> Int64
STRING -> String
BYTE_ARRAY -> SByte[]

Backwards compatibility is preserved by the fact that DEFINED_BY_JAVA_TYPE is the default.

This PR also changes the way in which BYTE_ARRAY Java type is handled, now the driver should support reading and writing the following OPCUA types using the BYTE_ARRAY Java type:

ByteString,
Byte[] (array of unsigned bytes)
SByte[] (array of signed bytes)

This PR also includes some tests for type conversions.

Closes #1840 


MMaiero:
Backport this to Kura maintenance branch

MMaiero (approved):

PR #1844, opened 2017-12-19
Fixed regression in BaseChanelDescriptor

### LINKED ISSUE #1843: BaseChannelDescriptor contains some inconsistencies
* the value type options include `BYTE`, which is not a valid type for an Asset channel
* the value type options do not include the `FLOAT` type
* The default value type for a channel is `STRING`

After #1819 this is reflected to the Web ui, since BaseChannelDescriptor is now used to render Asset configuration.

nicolatimeus:
Closes #1843 


kura/org.eclipse.kura.asset.provider/src/main/java/org/eclipse/kura/asset/provider/BaseChannelDescriptor.java
```diff
@@ -66,6 +69,16 @@
 
     private final List<Tad> defaultElements;
 
+    private static void addOptions(Tad target, Enum<?>[] values) {
+        final List<Option> options = target.getOption();
+        for (Enum<?> value : values) {
+            Toption option = new Toption();
+            option.setLabel(value.name());
```
amitjoy: You can extract `value.name()` to a local var.

MMaiero (approved):

PR #1848, opened 2017-12-20
Disabled the kura.primary.network.interface property.

MMaiero:
This property needs to be enabled only on systems that support
Predictable Network Interface Names.


MMaiero:
This needs to be back ported to maintenance-3.1 branch

pierantoniomerlino (approved):

PR #1871, opened 2018-01-18
First implementation of Asset reacting to channel changes

### LINKED ISSUE #1861: Implement a WireAsset reacting to channel changes
In the current implementation an WireAsset can only be used to pull readings from its Driver.
This works well for some protocols, like Modbus which are strictly request/response but it cannot cope with those protocols that have notifications, like Bluetooth Low Energy.
Another example is a driver for the Linux GPIO API where an input can be configured to notify the state upon its changes (rising, falling or both edges).

In Kura there’s an Driver API to listen for Channel changes. The client (e.g. an WireAsset) of the Driver can register a listener for a specific channel.
Not all drivers support registering channel listeners, for example the ModbusDriver does not.

Upon registration of a channel listener the Driver will do what is needed to enable notifications from the physical asset(s).

There are two common uses cases:

1. A change in any channel enabled for listening is notified to the WireAsset which emits only that channel
2. A change in any channel enabled for listening is notified to the WireAsset which reads all the channels and emits them in a single envelope

Two achieve this we need to extend the configuration of the WireAsset:

a. A per-channel “listen” boolean property will be added to “name”, “type” and “type.value” in the channel configuration
b. A per-asset “emit.all.channels” boolean property

In the emitted WireEnvelope, per-channel timestamps are not always desirable. For example in the second use case above, it would be more useful if the WireAsset
set a single timestamp property in the envelope which is the timestamp of the channel whose status has changed.
In the WireAsset configuration, a new configuration “single.timestamp” property set to true will put the “asset.timestamp” property to the emitted WireEnvelope.

Finally, downstream components may be interested in channel read errors, for example a data logger application.
These applications must often generate an CSV file with the following structure:

timestamp,ch1,ch2,…

Every WireEnvelope will be logged as a line in the CSV file.

The requirement is that Driver read errors are logged too.
Another WireAsset configuration property “emit.errors” controls this behavior.

nicolatimeus:
Closes #1861 


kura/test/org.eclipse.kura.asset.provider.test/src/main/java/org/eclipse/kura/asset/provider/test/AssetTest.java
```diff
@@ -209,6 +212,129 @@ public void onChannelEvent(ChannelEvent event) {
...
+
+    /**
+     * Listeners should be removed from the driver when it is detached from the asset and added when it is attached.
+     */
+    @TestTarget(targetPlatforms = { TestTarget.PLATFORM_ALL })
+    @Test
+    public void testListenerAttachOnDriverChange() throws KuraException {
+        final Driver driver = ((BaseAsset) asset).getDriver();
+
+        final ArrayList<Boolean> attachSequence = new ArrayList<>();
```
amitjoy, line 222: program to an interface

kura/test/org.eclipse.kura.asset.provider.test/src/main/java/org/eclipse/kura/asset/provider/test/AssetTest.java
```diff
@@ -209,6 +212,129 @@ public void onChannelEvent(ChannelEvent event) {
...
+    /**
+     * It should be possible to add listeners to an asset even if the driver is not attached, the asset should attach
+     * them later on when the driver is tracked
+     */
+    @TestTarget(targetPlatforms = { TestTarget.PLATFORM_ALL })
+    @Test
+    public void testAttachListenerWhitoutDriver() throws KuraException {
+        final Driver driver = ((BaseAsset) asset).getDriver();
+
+        final ArrayList<Boolean> attachSequence = new ArrayList<>();
```
amitjoy, line 255: ==>

kura/test/org.eclipse.kura.asset.provider.test/src/main/java/org/eclipse/kura/asset/provider/test/AssetTest.java
```diff
@@ -209,6 +212,129 @@ public void onChannelEvent(ChannelEvent event) {
...
+    }
+
+    /**
+     * The same channel listener should not be attached multiple times.
+     */
+    @TestTarget(targetPlatforms = { TestTarget.PLATFORM_ALL })
+    @Test
+    public void testShouldNotReattachSameListener() throws KuraException {
+
+        final ArrayList<Boolean> attachSequence = new ArrayList<>();
```
amitjoy, line 288: ==>

Issue #1849, opened 2017-12-21
SSL Configuration - Keystore path change doesn't work


kandrej:
When I try to change the keystore path the property is saved, but the default java keystore is used.
There is a warning inside kura.log saying "null old password"

MMaiero:
Hi,
which version of Kura are you using?
What are the exact steps you did to perform the operation?

I've just done the change of keystore from /opt/eclipse/kura/security/cacerts.ks (the referenced file did not exist) to /opt/eclipse/security/cacerts.ks, setting also the proper password for the new keystore and this is the log I get in kura.log file:

```
2017-12-22 15:15:38,154 [qtp462890-78] WARN  o.e.k.n.a.v.l.WpaSupplicantConfigReader - WPA in client mode is not configured
2017-12-22 15:15:38,203 [qtp462890-78] INFO  o.e.k.c.c.ConfigurationServiceImpl - Loading init configurations from: 1513952753472...
2017-12-22 15:15:38,365 [qtp462890-78] INFO  o.e.k.c.c.ConfigurationServiceImpl - Merging configuration for pid: org.eclipse.kura.ssl.SslManagerService
2017-12-22 15:15:38,393 [qtp462890-78] INFO  o.e.k.c.c.ConfigurationServiceImpl - Updating Configuration of ConfigurableComponent org.eclipse.kura.ssl.SslManagerService ... Done.
2017-12-22 15:15:38,399 [Component Resolve Thread (Bundle 7)] INFO  o.e.k.c.s.SslManagerServiceImpl - updated...
2017-12-22 15:15:38,401 [Component Resolve Thread (Bundle 7)] INFO  o.e.k.c.d.DataServiceImpl - Notified DataTransportService configuration updating...
2017-12-22 15:15:38,401 [Component Resolve Thread (Bundle 7)] WARN  o.e.k.c.d.t.m.MqttDataTransport - MQTT client already disconnected
2017-12-22 15:15:38,401 [Component Resolve Thread (Bundle 7)] INFO  o.e.k.c.d.t.m.MqttDataTransport - Building new configuration...
2017-12-22 15:15:38,518 [Component Resolve Thread (Bundle 7)] INFO  o.e.k.c.d.DataServiceImpl - Notified DataTransportService configuration updated.
[...]

PR #1884, opened 2018-01-29
Modified the SSL password change method.

MMaiero:
Made it return void in SslManagerServiceImpl.java.

Closes also #1849


kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/ssl/SslManagerServiceImpl.java
```diff
@@ -629,27 +629,20 @@ private boolean changeKeyStorePassword() {
...
-            return false;
-        }
-
-        if (!Arrays.equals(oldPassword, newPassword)) {
+        if (oldPassword != null && !Arrays.equals(oldPassword, newPassword)) {
             try {
                 if (isKeyStoreAccessible(this.options.getSslKeyStore(), oldPassword)) {
                     changeKeyStorePassword(this.options.getSslKeyStore(), oldPassword, newPassword);
                 } else if (isKeyStoreAccessible(this.options.getSslKeyStore(), newPassword)) {
                     changeKeyStorePassword(this.options.getSslKeyStore(), newPassword, newPassword);
```
cdealti, line 637: What is the purpose of changing the keystore password with itself?

kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/ssl/SslManagerServiceImpl.java
```diff
@@ -629,27 +629,20 @@ private boolean changeKeyStorePassword() {
             }
         }
 
-        if (oldPassword == null) {
-            logger.warn("null old password");
-            return false;
-        }
-
-        if (!Arrays.equals(oldPassword, newPassword)) {
+        if (oldPassword != null && !Arrays.equals(oldPassword, newPassword)) {
```
cdealti: Can `newPassword` be null at this point? Should you check?

kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/ssl/SslManagerServiceImpl.java
MMaiero: If I'm not wrong was related to a keystore just provided through the web ui where you don't have an old password and you need the new password both to access the keystore and register it in the snapshot.

kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/ssl/SslManagerServiceImpl.java
MMaiero: new password is initialised as old password

kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/ssl/SslManagerServiceImpl.java
cdealti: OK, but I don't get what `changeKeyStorePassword(this.options.getSslKeyStore(), newPassword, newPassword);` does

kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/ssl/SslManagerServiceImpl.java
cdealti: and changed later on. It's unlikely to be null but just in case...

kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/ssl/SslManagerServiceImpl.java
MMaiero: checks that the keystore is accessible and stores the new password

cdealti: I mean this:
```
else if (isKeyStoreAccessible(this.options.getSslKeyStore(), newPassword)) {
                     changeKeyStorePassword(this.options.getSslKeyStore(), newPassword, newPassword);
```
It looks weird to me as it sounds "if the keystore is already accessible with newPassword, change the password (which is newPassword) to to newPassword"

MMaiero: it is definitely weird. But, if you remember, we need to structure the code in this way the last time we reviewed it together to deal with the different requirements we have. For sure it needs to be improved, but it was not the scope of this PR also because we are close to the release.

cdealti:
@MMaiero are there any other reasons for this PR other than preventing the warning reported in #1849? If not, and because we seem to agree that the code is weird, I'd prefer to leave with the current behavior.

MMaiero:
Agreed

PR #1888, opened 2018-02-02
Added tcp socket connection timeout

kandrej:
The Socket(this.ipAddress, this.port) constructor can block for over a minute if no timeout is specified

Signed-off-by: Andrej Krota <andrej.krota@gmail.com>


kura/org.eclipse.kura.protocol.modbus/src/main/java/org/eclipse/kura/protocol/modbus/ModbusProtocolDevice.java
```diff
@@ -669,7 +670,9 @@ public void connect() {
             } else {
                 if (!this.connected) {
                     try {
-                        this.socket = new Socket(this.ipAddress, this.port);
+                    	this.socket = new Socket();
```
amitjoy: It seems it is not indented correctly

kura/org.eclipse.kura.protocol.modbus/src/main/java/org/eclipse/kura/protocol/modbus/ModbusProtocolDevice.java
```diff
@@ -680,6 +683,7 @@ public void connect() {
                             s_logger.error("Failed to get socket streams: " + e);
                         }
                     } catch (IOException e) {
+                    	this.socket = null;
```
amitjoy: Similarly here as well

MMaiero (approved):

MMaiero:
@kandrej Thank you for your contribution.
I believe that you should also update the version in kura.build.properties

Issue #1870, opened 2018-01-18
Develop a GPIO Driver for Kura Wires


pierantoniomerlino:
Kura exposes a GPIO service to manage the digital GPIOs on the device. However, this feature is not available on Kura Wires. 

Kura should support a GPIO driver to be used in Kura Wires.

PR #1890, opened 2018-02-05
GPIO driver implementation and GPIO example modification

pierantoniomerlino:
This PR adds a new driver to manage on-board digital GPIOs.
Moreover, the classic GPIO example is modified to support also the GPIO names.


pierantoniomerlino:
Closes #1870

nicolatimeus:
Some copyright headers need to be updated. We should also mention in the documentation that defining multiple channels involving the same GPIO resource might cause some issues.

nicolatimeus (approved):

PR #1909, opened 2018-02-09
Fix gpio driver

pierantoniomerlino:
When the "#select resource" value is selected in the "resource.name" field in a GPIO asset and the user tries to read/write data, an exception is returned.
This PR fixes this behavior.
 
### DIFF
--- a/kura/org.eclipse.kura.driver.gpio.provider/src/main/java/org/eclipse/kura/internal/driver/gpio/GPIODriver.java
+++ b/kura/org.eclipse.kura.driver.gpio.provider/src/main/java/org/eclipse/kura/internal/driver/gpio/GPIODriver.java
@@ -73,6 +73,7 @@ public final class GPIODriver implements Driver, ConfigurableComponent {
 
     private static final Logger logger = LoggerFactory.getLogger(GPIODriver.class);
     private static final GPIOMessages message = LocalizationAdapter.adapt(GPIOMessages.class);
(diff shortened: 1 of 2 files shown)


kura/org.eclipse.kura.driver.gpio.provider/src/main/java/org/eclipse/kura/internal/driver/gpio/GPIODriver.java
```diff
@@ -245,8 +244,10 @@ private synchronized void runWriteRequest(GPIORequestInfo requestInfo) {
                 record.setChannelStatus(new ChannelStatus(ChannelFlag.FAILURE, message.readFailed(), null));
                 record.setTimestamp(System.currentTimeMillis());
                 logger.warn(message.writeFailed());
-                return;
             }
+        } else {
+            record.setChannelStatus(new ChannelStatus(ChannelFlag.FAILURE, message.readFailed(), null));
```
MMaiero: Can this code be externalised to a method to prevent code duplication?

pierantoniomerlino: 👍

pierantoniomerlino:
Fixed also a bug on GPIOChannelDescriptor about merging gpios from different GPIOservices.

MMaiero:
Could you please have a look at why CI tests fail?

MMaiero (approved):

PR #1913, opened 2018-02-13
Wires composer robustness and usability improvements

nicolatimeus:
* Should improve the robustness of the wire composer in the following scenarios, which previously led to glitches and/or rendered the composer unusable:
  1. The WireService references a wire component for which there is no configuration available (the component does not exist).
  2. The WireService references a component for which a configuration is present but the provider bundle is not installed.
  3. An Asset in the composer is not attached to its Driver instance.

  - In all scenarios affected components will be rendered in red and a message describing the problem will be shown instead of the component configuration.
  - In scenarios 1 and 2, saving the graph will be prohibited.
  - It should be possible to recover from scenarios 1 and 2 by deleting affected components.

* Port labels are now hidden by default.
* Port labels are shown when the port is hovered with the mouse or while a wire is being attached to the port (during wire creation).
* Moved port labels a bit to avoid overlapping with horizontal wires (overlapping is still possible in other configurations).
* Changed the way ports are highlighted during wire creation.
* Fixed regression in IE11 that prevented the graph from being displayed.

### DIFF
--- a/kura/org.eclipse.kura.asset.provider/src/main/java/org/eclipse/kura/asset/provider/BaseAsset.java
+++ b/kura/org.eclipse.kura.asset.provider/src/main/java/org/eclipse/kura/asset/provider/BaseAsset.java
@@ -299,7 +299,6 @@ public class BaseAsset implements Asset, SelfConfiguringComponent {
     }
 
     /** {@inheritDoc} */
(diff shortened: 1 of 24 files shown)


nicolatimeus:
@markoer can you please have a look at this?

MMaiero (approved):

PR #1919, opened 2018-02-15
Fix a few NPEs and log the cause exception

ctron:


kura/org.eclipse.kura.core.configuration/src/main/java/org/eclipse/kura/core/configuration/ConfigurationServiceImpl.java
```diff
@@ -1797,6 +1800,9 @@ private void ungetServiceReferences(final ServiceReference<?>[] refs) {
     protected String marshal(Object object) {
         String result = null;
         ServiceReference<Marshaller>[] marshallerSRs = getXmlMarshallers();
+        if (marshallerSRs == null) {
+            logger.warn("Failed to marshal configuration. No XML marshallers found");
```
amitjoy, line 1804: I believe we could make the same null scenarios in unmarshal(String, Class<T>) and marshal(Object) more consistent. In unmarshal(..), it throws KuraException whereas in marshal(..), it is logged with warning message.

ctron:
IIRC this is superseeded by a more recent PR.

PR #1922, opened 2018-02-16
Support 'Layer 2 Only' configuration for Ethernet and Wireless interfaces.

ibinshtok:
@MMaiero please take a look at this PR when you have a chance.


kura/org.eclipse.kura.linux.net/src/main/java/org/eclipse/kura/linux/net/util/KuraConstants.java
```diff
@@ -25,6 +25,7 @@
     Reliagate_10_11("yocto", "1.2.1", "reliagate-10-11"),
     Reliagate_20_25("yocto", "1.2.1", "reliagate-20-25"),
     BoltGATE_20_25("yocto", "1.2.1", "boltgate-20-25"),
+    BoltMAR_20_28("yocto", "1.2.1", "boltmar-20-28"),
```
MMaiero: Doesn't this change implies that the exported package in manifest needs to be updated?

kura/org.eclipse.kura.net.admin/src/main/java/org/eclipse/kura/net/admin/NetworkAdminConfiguration.java
```diff
@@ -16,6 +16,7 @@
     CONFIG_MTU,
     CONFIG_AUTOCONNECT,
     CONFIG_DRIVER,
+    CONFIG_IPV4_L2ONLY_ENABLED,
```
MMaiero: Doesn't this change imply an update of the exported package in manifest file?

kura/org.eclipse.kura.net.admin/src/main/java/org/eclipse/kura/net/admin/monitor/InterfaceState.java
```diff
@@ -17,38 +17,52 @@
...
+        this.name = interfaceName;
+        this.up = LinuxNetworkUtil.hasAddress(interfaceName);
+        this.link = LinuxNetworkUtil.isLinkUp(type, interfaceName);
+        logger.debug("InterfaceState() :: {} - link?={}", interfaceName, this.link);
+        logger.debug("InterfaceState() :: {} - up?={}", interfaceName, this.up);
+        ConnectionInfo connInfo = new ConnectionInfoImpl(interfaceName);
+        this.ipAddress = connInfo.getIpAddress();
+    }
 
+    public InterfaceState(NetInterfaceType type, String interfaceName, boolean isL2OnlyInterface) throws KuraException {
```
MMaiero, line 48: Doesn't this change imply an update of the exported package in manifest file?

kura/org.eclipse.kura.net.admin/src/main/java/org/eclipse/kura/net/admin/monitor/WifiInterfaceState.java
```diff
@@ -34,16 +34,33 @@
...
+     *
+     * @param interfaceName
+     *            - interface name as {@link String}
+     * @param wifiMode
+     *            - configured wifi mode as {@link WifiMode}
+     * @param isL2OnlyInterface
+     *            - is Layer 2 only interface
+     * @throws KuraException
+     */
+    public WifiInterfaceState(String interfaceName, WifiMode wifiMode, boolean isL2OnlyInterface) throws KuraException {
```
MMaiero, line 52: Doesn't this change imply an update of the exported package in manifest file?

kura/org.eclipse.kura.linux.net/src/main/java/org/eclipse/kura/linux/net/util/KuraConstants.java
kartben: Even more so, this doesn't look like it's even used in the Kura codebase, or is it? It would be great to not include (or at least try to stop adding new 😉 ) vendor-specific constants in the Kura OSS codebase.

kura/org.eclipse.kura.net.admin/src/main/java/org/eclipse/kura/net/admin/NetworkAdminConfiguration.java
ibinshtok: Made changes to the MANIFEST.MF

kura/org.eclipse.kura.net.admin/src/main/java/org/eclipse/kura/net/admin/monitor/InterfaceState.java
ibinshtok: Made changes to the MANIFEST.MF

kura/org.eclipse.kura.net.admin/src/main/java/org/eclipse/kura/net/admin/monitor/WifiInterfaceState.java
ibinshtok: Made changes to the MANIFEST.MF

kura/org.eclipse.kura.net.admin/META-INF/MANIFEST.MF
```diff
@@ -44,6 +44,6 @@ Import-Package: javax.microedition.io,
 Service-Component: OSGI-INF/*.xml
 Bundle-ClassPath: .
 Bundle-ActivationPolicy: lazy
-Export-Package: org.eclipse.kura.net.admin;version="1.2.0",
+Export-Package: org.eclipse.kura.net.admin;version="1.3.0",
  org.eclipse.kura.net.admin.modem;version="1.1.0",
- org.eclipse.kura.net.admin.monitor;version="1.0.11"
+ org.eclipse.kura.net.admin.monitor;version="1.1.11"
```
MMaiero: Shouldn't this be better set to 1.1.0 instead of 1.1.11?

ibinshtok: Yes, sorry. I'll change that.

cdealti:
@ibinshtok it looks good to me. Do you want me to retest due to the last change?

ibinshtok:
@cdealti I think I found a bug switching wireless interface into L2Only mode that I introduced yesterday. I'll try to fix it today and test some more.

ibinshtok:
@cdealti Please test the most recent changes when you have a chance.

cdealti:
It looks like that, when configured as "Layer 2 Only" mode, reloading the kernel module of Wi-Fi interfaces [1] is skipped. This is relevant for Wi-Fi interfaces that need to reload the firmware when switching between Access Point and Station Mode. I guess this is happening because interface **up** is not meaningful in L2 mode. Any ideas on how to fix it?

[1] https://github.com/ibinshtok/kura/blob/fed302b4ac699ee97aa892c5c75cc233e477908c/kura/org.eclipse.kura.net.admin/src/main/java/org/eclipse/kura/net/admin/monitor/WifiMonitorServiceImpl.java#L382

cdealti:
@ibinshtok to clarify more, the kernel module reload only works if the interface is temporarily disabled when it is reconfigured [1]. The monitor will then reload the kernel module and re-enable the interface [2]. Can you please debug whether this is not happening when the interface status is set to "Layer 2 Only" and you later on switch between station and access point mode?

[1] https://github.com/ibinshtok/kura/blob/fed302b4ac699ee97aa892c5c75cc233e477908c/kura/org.eclipse.kura.net.admin/src/main/java/org/eclipse/kura/net/admin/monitor/WifiMonitorServiceImpl.java#L239 
[2] https://github.com/ibinshtok/kura/blob/fed302b4ac699ee97aa892c5c75cc233e477908c/kura/org.eclipse.kura.net.admin/src/main/java/org/eclipse/kura/net/admin/monitor/WifiMonitorServiceImpl.java#L382

ibinshtok:
@cdealti please take a look at changes I made to fix reloading wireless kernel module. It seems to work when switching between "access point" and "Station" modes in the "L2 Only" mode w/o exceptions now. Switching in and out "L2 Only" mode seems to work as well.

MMaiero:
@ibinshtok Few findings I made: 
- when in Layer 2 mode, the changes for wifi and station mode are correctly persisted. Otherwise, when in AP mode, the changes are not registered: the hosted file is always the same and, for example, a change in the AP name has no effect.
- switching between TCP/Ip statuses determines that the wifi mode is always set to "Station" even if in the previous configuration it was in "AP" mode.

ibinshtok:
@MMaiero please take a look at my last two commits (one for each issue you have reported). For the second issue, I fixed a bug in the ChangeHandler for the this.tcpTab.status in the TabWirelessUi.java.  We have to use the MessageUtils.get(GwtNetIfStatus.netIPv4StatusEnabledWAN.name()) instead of GwtNetIfStatus.netIPv4StatusEnabledWAN.name() when comparing with the tcpIpStatus. Also I changed    
the logic to force the 'Wireless Mode' to 'Station' only if interface is switched to WAN. If interface is switched back to LAN or L2Only, the 'Wireless Mode' will be restored to one that is currently active. Please let me know is this logic looks correct to you.

MMaiero:
@ibinshtok It works fine for the wifi interface. Just one doubt: what about ppp?
Should the not managed and layer2only options be selectable for the ppp interface?
Do we have use cases?
Thanks

ibinshtok:
@MMaiero I think the 'Unmanaged' option should be present for ppp as for any other network interface.
As far as the 'L2 Only', I was not planning to have it for cellular interface, and believe it is a regression bug after I moved from having 'L2 Only' under 'Configure' to 'Status'. In my opinion, the 'L2 Only' makes sense if interface is controlled from the /etc/network/interfaces file, and we can set it to something like 'iface wlan0 inet manual'. So if ESF has support for CDC-NCM, it would be reasonable to include the 'L2 Only' option for the usb0 interface. I will work on fixing this bug today.

ibinshtok:
@MMaiero I pushed a commit that removes the 'L2 Only' option from cellular interface.

MMaiero:
Hi Ilya, now we don't have the layer 2 only option. But we do have the enabled for LAN option in the modem, which seems not correct for me.
Is that there on purpose?

MMaiero:
@ibinshtok I've tried to fix it. Please give it a look.

ibinshtok:
@MMaiero your changes look good; thank you for looking at this! I fixed a few minor lint issues in the TabTcpIpUi.java and retested.

PR #1925, opened 2018-02-19
Changed exists plugin default. Now it's disabled.

MMaiero:

Steps to be completed after merge: 

- [ ] add profile to Hudson
- [ ] document the new check-exists-plugin profile

### DIFF
--- a/kura/examples/pom.xml
+++ b/kura/examples/pom.xml
@@ -37,6 +37,7 @@
         <skip.can.example>true</skip.can.example>
 
         <osgi-dp-plugin-version>0.3.0</osgi-dp-plugin-version>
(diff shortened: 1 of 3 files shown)


cdealti:
@MMaiero is #1782 by @ctron doing the same thing?

MMaiero:
this is slightly more up to date dealing also with the examples

MMaiero:
@ctron Ok for you this version?

ctron:
So my understanding is that when I check out the released version (e.g. from a tag) I can build it locally even though the release has been made, without the requirement to make any changes to the checked out code.

In this case: +1

MMaiero:
Yes, that is the expected bahaviour

PR #1940, opened 2018-03-02
Solved some possible NPEs in web2 bundle.

MMaiero:


kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/client/ui/Network/TabTcpIpUi.java
```diff
@@ -378,21 +380,17 @@ private void initForm() {
         }
 
         // Populate status list
+        this.status.clear();
         if (this.selectedNetIfConfig != null && this.selectedNetIfConfig.getHwTypeEnum() == GwtNetIfType.MODEM) {
-            if (this.status != null) {
```
ibinshtok: @MMaiero do we need to remove other 'this.status == null' checks as well? (e.g. in the isLanEnabled() and isWanEnabled() methods) Can @UiField be null ?

nicolatimeus (approved):

Issue #1689, opened 2017-09-18
Instaling a dp can cause the WebUI to restart


nicolatimeus:
Installing a deployment package sometimes causes the WebUi and a few other services to be stopped and restarted:
```
2017-09-18 15:58:44,598 [qtp498690566-42] INFO  o.e.k.w.s.GwtPackageServiceImpl - Installing deployment package from Eclipse Marketplace, URL https://s3-us-west-2.amazonaws.com/kura-repo/drivers/3.0.0-RELEASE/org.eclipse.kura.driver.opcua_1.0.0.dp...
2017-09-18 15:58:44,622 [DeploymentAgent] INFO  o.e.k.d.a.i.DeploymentAgent - About to install package at URL https://s3-us-west-2.amazonaws.com/kura-repo/drivers/3.0.0-RELEASE/org.eclipse.kura.driver.opcua_1.0.0.dp
2017-09-18 15:59:17,202 [Refresh Thread: Equinox Container: 707ba579-7e9c-0017-10ad-a1314df20a22] INFO  o.e.k.c.c.ConfigurationServiceImpl - Removing component configuration for pid org.eclipse.kura.web.Console
2017-09-18 15:59:17,212 [Refresh Thread: Equinox Container: 707ba579-7e9c-0017-10ad-a1314df20a22] INFO  o.e.k.w.Console - deactivate...
2017-09-18 15:59:17,239 [Refresh Thread: Equinox Container: 707ba579-7e9c-0017-10ad-a1314df20a22] INFO  o.e.k.w.s.s.FileServlet - Servlet org.eclipse.kura.web.server.servlet.FileServlet destroyed
2017-09-18 15:59:17,288 [Refresh Thread: Equinox Container: 707ba579-7e9c-0017-10ad-a1314df20a22] INFO  o.e.k.c.c.ConfigurationServiceImpl - Removing component configuration for pid org.eclipse.kura.wire.WireService
2017-09-18 15:59:17,308 [Refresh Thread: Equinox Container: 707ba579-7e9c-0017-10ad-a1314df20a22] INFO  o.e.k.c.Cloudlet - Releasing CloudApplicationClient for ASSET-V1...
2017-09-18 15:59:17,862 [Component Resolve Thread (Bundle 6)] INFO  o.e.k.c.Cloudlet - Getting CloudApplicationClient for ASSET-V1...
[...]

Issue #1928, opened 2018-02-19
Removing and reinstalling the Script Filter dp causes a lot of components to be restarted


nicolatimeus:
- Component affected:
  - [X] Distribution
    - Installer name kura_3.2.0-SNAPSHOT_raspberry-pi-2-3_installer.deb:
    - Platform: observed on Raspberry pi 3 and emulator
    - Kura version affected Kura 3.2.0 snapshot (probably also previous versions)
    - Java version: java version "1.8.0_65"
Java(TM) SE Runtime Environment (build 1.8.0_65-b17)
Java HotSpot(TM) Client VM (build 25.65-b01, mixed mode)
    - Host OS (e.g. Debian): raspbian
    - Host OS version

  - [ ] Development environment
    - Version: 
    - Operating system (Mac OS, Windows, Linux):
    - OS Version:
    - IDE Version:
    - Host System Java Version:
      
  - [ ] Oomph installer
  - [X] Add-ons
    - Name: Script Filter
    - Version: 1.0.100-SNAPSHOT and 1.0.0
    - URL:
  - [ ] Eclipse Kura Marketplace
  - [ ] Kura Website
  - [ ] Kura Documentation

  - If needed to replicate the issue, please provide references for the browser used:
    - Name:
    - Version:
     
- Textual description of the issue:

Uninstalling and reinstalling the Script Filter dp causes a lot of framework components to be restarted by the Refresh Thread.

```
2018-02-19 14:37:22,860 [Refresh Thread: Equinox Container: 2001f15f-7f15-0018-14db-d15a9290fe56] INFO  o.e.k.c.c.ConfigurationServiceImpl - Registering metatype pid: org.eclipse.kura.wire.CloudPublisher ...
2018-02-19 14:37:22,860 [Refresh Thread: Equinox Container: 2001f15f-7f15-0018-14db-d15a9290fe56] INFO  o.e.k.c.c.ConfigurationServiceImpl - Registering metatype pid: org.eclipse.kura.wire.DbWireRecordStore ...
```

The refresh process takes some time, when it finishes, a new component instance is created for each existing factory component configuration. The new instance will have the same `kura.service.pid` as the pre-esisting instace and a different `service.pid`, this can be observed using the OSGi console:

```
osgi> ls | grep CloudService
12	Active		org.eclipse.kura.cloud.CloudService			org.eclipse.kura.core.cloud(bid=63)
[...]

PR #1942, opened 2018-03-05
Removed localisation from Kura core components

nicolatimeus:
See #1928 


nicolatimeus:
Closes #1689

PR #1945, opened 2018-03-06
Fixed db cleanup issue

### LINKED ISSUE #1941: [DbDataStore] Stale messages not purged with timezone other than UTC
Kura uses a HouseKeeper thread to purge messages from the internal data store once they have been published. The query to find these messages (QoS == 0) is of the form:

`DELETE FROM <table_name> WHERE qos = 0 AND publishedOn <= DATEADD('ss', 5, NOW()) AND publishedOn IS NOT NULL;"`

The NOW() function returns the current time from the underlying OS. However, when updating the "publishedOn" column once a message is published, the Java code uses Date().getTime(), which always returns UTC. Because of this, systems with a timezone set to anything west of UTC will never clear messages from the datastore.

Kura should be able to handle systems with a timezone set to something other than UTC.

nicolatimeus:
Closes #1941 

#1941 is likely a regression introduced after the switch to H2.
Previously in the DataStore all timestamps were passed from Java to H2 in UTC. After H2 PR was merged some timestamps are computed using the NOW() function, which uses the JVM timezone.
This PR restores the old behaviour.


nicolatimeus:
@dwoodard1  can you please have a look if this fixes the issue?

dwoodard1 (approved):

dwoodard1:
@nicolatimeus Looks good to me.

Issue #1950, opened 2018-03-09
Service description missing in the Web UI


ctron:
- Component affected:
  - [X] Emulator container
    - Kura version affected: 3.1.0

  - If needed to replicate the issue, please provide references for the browser used:
    - Name: Firefox
    - Version: 58.0.2
     
- Textual description of the issue:

A configurable service, having a description, will only have its name shown in the Web UI, although the Kura Web UI supports this in general.

IIRC this worked in the past, so this is probably a regression from a previous Kura version.

MMaiero:
@markoer Could you please do a triage of this issue?

ctron:
I created PR #1951 which fixes this issue.

MMaiero:
Hi Jens,
maybe I've missed something. If so, I'm sorry. But, isn't the fix already in the 3.2.0 codebase?
I believe we added with PR #1525 and in particular, adapted and made uniform with the rest of UI with commit e2de4847e4ca66181fa84c121ac9514d71c243dc.

This is what I see with the latest build from develop and Firefox 58.0.2:
<img width="1432" alt="screen shot 2018-03-09 at 12 30 24 pm" src="https://user-images.githubusercontent.com/10448126/37205760-4b5c71f2-2396-11e8-8c94-01c40dbffc86.png">

This is what we get from your PR:
<img width="1432" alt="screen shot 2018-03-09 at 12 19 09 pm" src="https://user-images.githubusercontent.com/10448126/37205782-62d4cb86-2396-11e8-856c-26e51f667cef.png">

ctron:
Which is interesting, because I didn't see anything like that before making the PR. Thus my assumption was that there was some issue with the "div" being between the two "panel" elements. Which probably isn't the way bootstrap wants it to be.

But I can give it another try, using the `develop` branch.

ctron:
I guess it was some kind of refresh issue. Switching back to "develop" of doing a full rebuild shows the description. Sorry.

PR #1951, opened 2018-03-09
[#1950] Fix missing service description

ctron:
This change fixes the missing service description in the Web UI.


ctron:
No need to merge this.

PR #1963, opened 2018-03-12
Prevent getters from creating new object instances

ctron:
This change does create the unmodifiable wrapper in the constructor
and not in the getter. So that the getter can return the same object
instance without create a new instance every time the getter is called.


amitjoy (approved):
Looks good to me.

PR #1967, opened 2018-03-14
Improved Web Ui validation for new component pid fields

### LINKED ISSUE #1955: [Wires] New Wire Component name entry field is broken
- Component affected:
  - [X] Kura Wires

  - If needed to replicate the issue, please provide references for the browser used:
    - Name: Firefox
    - Version: 58.0.2
     
- Textual description of the issue:

The name entry field of the "New Wire Component" has several defects:
* It is not possible to move the cursor using the arrow keys on the keyboar
* It is not possible to insert "special" characters such as "ä", which this may be on purpose, there is not explanation about this
* Still it is possible to copy & paste in special characters
* The placeholder text says "an user friendly name", I am not 100% sure, but I guess it should be "a user friendly name".
* It is possible to enter other special characters e.g. backticks (`), spaces

### LINKED ISSUE #1959: [Web] Escaping doesn't seem to work properly
Using the "new component" dialog it is possible to create components with names like `)`, and `"`.

While `)` can be deleted afterwards the `"` (and also others like `&` or `'`) gets expanded to `&quot;` and then can no longer be deleted as a component named `&quot;` doesn't exist. I guess at some point this gets escaped twice.

nicolatimeus:
Improves Web Ui validation for input fields that allow to enter a component name (`kura.service.pid`).

The performed validation enforces that the provided names are valid OSGi `service.pid`s (only non empty sequences of alphanumeric characters, '_' or '-' separated by '.').

This `kura.service.pid` parameter in theory can allow more freedom than the `service.pid`, but its values currently need to be used at least in the following contexts:

* In LDAP filters
* As part of a topic name (e.g wires publisher allows to build an application topic from an asset name).

Validating the `kura.service.pid` in the same way as the `service.pid` should be safe at least for the scenarios above even if maybe too restrictive. This can be eventually changed in the future, performing a less strict validation should not be a breaking change.

This PR does not introduce validations on the platform side at this point of the release cycle, as this might cause compatibility issues.

Closes #1959
Closes #1955

### DIFF
--- a/kura/org.eclipse.kura.web2/src/main/resources/org/eclipse/kura/web/client/messages/Messages.properties
+++ b/kura/org.eclipse.kura.web2/src/main/resources/org/eclipse/kura/web/client/messages/Messages.properties
@@ -547,7 +547,7 @@ servicesComponentFactorySelectorIdle=--- Select Component Factory---
 servicesComponentFactoryNew=New Component
 servicesComponentFactoryFactory=Factory
 servicesComponentFactoryName=Name
-servicesComponentFactoryNamePlaceholder=Please enter an user friendly name
+servicesComponentFactoryNamePlaceholder=Please enter a user friendly name
 servicesComponentFactoryAlertNotSelected=Component must be selected and the name must be non-empty
 
 driversAssetsTabIntro=Create and manage your Drivers and Assets instances. Inspect and change the Assets values.
--- a/kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/client/ui/EntryClassUi.java
+++ b/kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/client/ui/EntryClassUi.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2011, 2017 Eurotech and/or its affiliates and others
+ * Copyright (c) 2011, 2018 Eurotech and/or its affiliates and others
  *
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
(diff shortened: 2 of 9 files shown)


kura/org.eclipse.kura.web2/src/main/resources/org/eclipse/kura/web/client/messages/Messages.properties
```diff
@@ -547,7 +547,7 @@ servicesComponentFactorySelectorIdle=--- Select Component Factory---
 servicesComponentFactoryNew=New Component
 servicesComponentFactoryFactory=Factory
 servicesComponentFactoryName=Name
-servicesComponentFactoryNamePlaceholder=Please enter an user friendly name
+servicesComponentFactoryNamePlaceholder=Please enter a user friendly name
```
amitjoy, line 550: user-friendly

kura/org.eclipse.kura.web2/src/main/resources/org/eclipse/kura/web/client/messages/Messages.properties
```diff
@@ -547,7 +547,7 @@ servicesComponentFactorySelectorIdle=--- Select Component Factory---
...
 servicesComponentFactoryAlertNotSelected=Component must be selected and the name must be non-empty
```
amitjoy, line 551: Shouldn't it make more sense if we write as "the name must not be empty"?. Just thinking.

nicolatimeus:
@ctron can you have a look at this?

kura/org.eclipse.kura.web2/src/main/resources/org/eclipse/kura/web/client/messages/Messages.properties
ctron: This isn't a validation message, but a text placeholder. So I think the text which is there is more appropriate.

ctron:
@nicolatimeus Did you have a look in the OSGi spec what a "valid" PID can be?

nicolatimeus:
On the R6 spec I've found the following information:

> The service.pid property optionally identifies a persistent, unique identifier for the service object. See Persistent Identifier (PID) on page 133.

> The format of the PID should be:
pid ::= symbolic-name // See 1.3.2

> symbolic-name ::= token ( '.' token )*
> token ::= ( alphanum | '_' | '-' )+
> alphanum ::= alpha | digit
> alpha ::= [a..zA..Z]
> digit ::= [0..9]

I've expressed this with the following Javascript RegExp

`^[\w\-]+([\.][\w\-]+)*$`

`\w` should match `alphanum |  _` [1]

Does this make sense for you?

[1] https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp

darionct (approved):
Automated tests confirmed

Issue #1971, opened 2018-03-20
EthernetMonitor can't monitor more than two ethernet interfaces


ibinshtok:
Please describe the issue you are encountering in Kura.

Fill the following form and provide a textual description if you are reporting a bug. 
Provide only a textual description otherwise.

- Component affected:
  - [ ] Distribution
    - Installer name (e.g. kura_3.1.0_raspberry-pi-2-3_installer.deb):
    - Platform (e.g. Raspberry pi 2):
    - Kura version affected (e.g. Kura 3.1.0): 
    - Java version (paste the result of **java -version**:
    - Host OS (e.g. Debian): 
    - Host OS version

  - [ ] Development environment
    - Version: 
    - Operating system (Mac OS, Windows, Linux):
    - OS Version:
    - IDE Version:
    - Host System Java Version:
      
  - [ ] Oomph installer
  - [ ] Add-ons
    - Name:
    - Version:
    - URL:
  - [ ] Eclipse Kura Marketplace
  - [ ] Kura Website
  - [ ] Kura Documentation

  - If needed to replicate the issue, please provide references for the browser used:
    - Name:
    - Version:
     
- Textual description of the issue: The Ethernet Monitor can't monitor more then two Ethernrt interfaces as hardcoded in the this.executor = Executors.newFixedThreadPool(getNumberOfInterfacesToMonitor()); line of its activete() method. If a platform has more than 2 Ethernet interfaces or some Ethernet-like tap interfaces, those extra interfaces will be configurable with Kura but not properly monitored.

PR #1972, opened 2018-03-20
Allow to monitor more than two Ethernet interfaces

ibinshtok:


ibinshtok:
@MMaiero please take a look at this PR that is an attempt to fix Kura issue #1971 that I created today.

MMaiero:
@ibinshtok I have some concerns on this solution: it is not enough general and based on the fact that the interfaces will be named eth* or tun*. But that could be not the case in nowadays operating systems with the consistent device naming. Couldn't we find another idea to achieve this?

MMaiero:
Furthermore, there are test failures

ibinshtok:
@MMaiero I have replaced the newFixedThreadPool with the newCachedThread pool in the activate() method of the EthernetMonitorServiceImpl. It seems to work fine; please take a look.

MMaiero (approved):

Issue #1970, opened 2018-03-19
[Kura 3.2.0 QA] Multiple handling of PositionLocked


MMaiero:
Having the position service running and the cloud service set to manage the position fix could determine an increasing number go Handlers for the PositionLockedEvent.

Having a look at the log, you can end up with something like:
```
2018-03-19 13:44:29,329 [EventAdmin Async Event Dispatcher Thread] INFO  o.e.k.l.p.PositionServiceImpl - Updating... Done.
2018-03-19 13:44:29,329 [EventAdmin Async Event Dispatcher Thread] INFO  o.e.k.c.c.CloudServiceImpl - Handling PositionLockedEvent
2018-03-19 13:44:29,335 [EventAdmin Async Event Dispatcher Thread] INFO  o.e.k.c.c.CloudServiceImpl - Handling PositionLockedEvent
2018-03-19 13:44:29,340 [EventAdmin Async Event Dispatcher Thread] INFO  o.e.k.c.c.CloudServiceImpl - Handling PositionLockedEvent
2018-03-19 13:44:29,345 [EventAdmin Async Event Dispatcher Thread] INFO  o.e.k.c.c.CloudServiceImpl - Handling PositionLockedEvent
2018-03-19 13:44:29,350 [EventAdmin Async Event Dispatcher Thread] INFO  o.e.k.c.c.CloudServiceImpl - Handling PositionLockedEvent
2018-03-19 13:44:29,355 [EventAdmin Async Event Dispatcher Thread] INFO  o.e.k.c.c.CloudServiceImpl - Handling PositionLockedEvent
2018-03-19 13:44:29,360 [EventAdmin Async Event Dispatcher Thread] INFO  o.e.k.c.c.CloudServiceImpl - Handling PositionLockedEvent
2018-03-19 13:44:29,365 [EventAdmin Async Event Dispatcher Thread] INFO  o.e.k.c.c.CloudServiceImpl - Handling PositionLockedEvent
2018-03-19 13:44:29,369 [EventAdmin Async Event Dispatcher Thread] INFO  o.e.k.c.c.CloudServiceImpl - Handling PositionLockedEvent
2018-03-19 13:44:29,374 [EventAdmin Async Event Dispatcher Thread] INFO  o.e.k.c.c.CloudServiceImpl - Handling PositionLockedEvent
[...]

PR #1979, opened 2018-03-22
Fixed multiple events received in cloud service.

MMaiero:
The bug was caused when multiple cloud stacks were created or updated
causing multiple handling of the same event by those stacks.


MMaiero:
Closes #1970

PR #2023, opened 2018-04-04
Removed 'Search Domains' field.

ibinshtok:


MMaiero:
I believe it's better to remove the code. We can simply recall it, when needed, by looking at the repo

Issue #2009, opened 2018-04-03
DATA-LIMIT-001 fails


ctron:
Setting the period to 1000 ms causes the publisher to spit out messages way faster:

~~~
2018-04-03 14:46:04,226 [pool-11-thread-1] INFO  o.e.k.c.d.DataServiceImpl - Storing message on topic :#account-name/#client-id/EXAMPLE_PUBLISHER/data/metrics, priority: 5
2018-04-03 14:46:04,231 [pool-11-thread-1] INFO  o.e.k.c.d.DataServiceImpl - Stored message on topic :#account-name/#client-id/EXAMPLE_PUBLISHER/data/metrics, priority: 5
2018-04-03 14:46:04,231 [pool-11-thread-1] INFO  o.e.k.e.p.ExamplePublisher - Published to data/metrics message: org.eclipse.kura.message.KuraPayload@1e142d6 with ID: 2236
2018-04-03 14:46:04,234 [DataServiceImpl:Submit] INFO  o.e.k.c.d.t.m.MqttDataTransport - Publishing message on topic: account-name/B8:27:EB:97:D8:D5/EXAMPLE_PUBLISHER/data/metrics with QoS: 0
2018-04-03 14:46:04,239 [DataServiceImpl:Submit] INFO  o.e.k.e.p.ExamplePublisher - Published message with ID: 2236 on application topic: data/metrics
2018-04-03 14:46:04,240 [DataServiceImpl:Submit] INFO  o.e.k.e.p.ExamplePublisher - Published message with ID: 2236 on application topic: data/metrics
2018-04-03 14:46:04,240 [DataServiceImpl:Submit] INFO  o.e.k.e.p.ExamplePublisher - Published message with ID: 2236 on application topic: data/metrics
2018-04-03 14:46:04,240 [DataServiceImpl:Submit] INFO  o.e.k.e.p.ExamplePublisher - Published message with ID: 2236 on application topic: data/metrics
2018-04-03 14:46:04,241 [DataServiceImpl:Submit] INFO  o.e.k.e.p.ExamplePublisher - Published message with ID: 2236 on application topic: data/metrics
2018-04-03 14:46:04,242 [DataServiceImpl:Submit] INFO  o.e.k.e.p.ExamplePublisher - Published message with ID: 2236 on application topic: data/metrics
[...]

PR #2025, opened 2018-04-05
Fixed example publisher activation

markoer:
Fixed activation in case cloud client is not available, yet. Reported in #2009.


MMaiero (approved):

MMaiero:
I would change the check in this way:
`if (this.cloudClient != null && this.cloudClient.isConnected()) {`

Then you need to subscribe also on the onConnected callback (if not already done. I've not checked)

MMaiero (approved):

PR #2026, opened 2018-04-06
Fix for array validation

### LINKED ISSUE #2007: Array validation doesn't work properly
This is related to DEN-CONF-036, I know it is not part of the test, still I think it is a bug.

The field `metric.short.array` states that is has a range of "Metric range min -32760 max 32760". Entering "32761" causes a proper error to appear, but selecting the next field in the array makes the error disappear again. The (out of range) value can be saves and gets applied to the service:

~~~
2018-04-03 14:22:11,581 [Component Resolve Thread (Bundle 7)] INFO  o.e.k.e.p.ExamplePublisher - Update - metric.short.array: [10, 20, 30, 32761]
~~~

I do think that validation checks should be done by Kura configuration service **in addition** to the UI, in order to fix issues like this.

markoer:
Should fix #2007.
Now all the text fields in the group are validated whenever any of them changes.

### DIFF
--- a/kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/client/ui/AbstractServicesUi.java
+++ b/kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/client/ui/AbstractServicesUi.java
@@ -8,7 +8,7 @@
  *
  * Contributors:
  *     Eurotech
- *     Red Hat 
+ *     Red Hat
  *******************************************************************************/
 package org.eclipse.kura.web.client.ui;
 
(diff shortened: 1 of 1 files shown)


MMaiero:
I believe that the specific issue is solved, but I've found a variant there: place an invalid value in the array, the error pops up. Then change another field with a proper value. The top buttons will be enabled and perform the save.
Before saving we should check the validation status

MMaiero:
Maybe a fix could be to have the `this.valid.put(param.getId(), true);` outside the for cycle.

MMaiero:
Otherwise the risk is that, if one of the fields is correct, in the array, will set to true the valid entry for all the array and the valid check passed.

markoer:
You beat me to it by a few seconds :)
And, of course, you're correct. It does work better this way.

MMaiero:
> You beat me to it by a few seconds :)

:-)

If it's ok for you, after the CI build, I'll merge it

MMaiero (approved):

markoer:
Of course.

PR #2029, opened 2018-04-06
Added a page refresh on component delete.

### LINKED ISSUE #2014: Deleting a component keeps tab open
Tested with Kura 3.2 RC:

* Create factory service instance
* Select service
* Delete service -> tab remains open
* Make changes to configuration
* Press "Apply" -> Error

image

I think there are two things wrong here:

1. The service tab should be closed when the service is deleted
1. A proper error message should be shown when the "apply" action fails

MMaiero:

Closes #2014

### DIFF
--- a/kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/client/ui/ServicesUi.java
+++ b/kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/client/ui/ServicesUi.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2011, 2017 Eurotech and/or its affiliates and others
+ * Copyright (c) 2011, 2018 Eurotech and/or its affiliates and others
  *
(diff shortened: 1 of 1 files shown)


kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/client/ui/ServicesUi.java
```diff
@@ -252,6 +257,7 @@ public void onSuccess(Void result) {
                             setDirty(false);
                             ServicesUi.this.entryClass.fetchAvailableServices(null);
                             EntryClassUi.hideWaitModal();
+                            Window.Location.reload();
```
nicolatimeus: Since we want to refresh the entire page, I think we can safely clear the body of the `onSuccess()` callback here, leaving only `Window.Location.reload()` inside it. This should spare a couple of RPC calls performed by `fetchAvailableServices()`

nicolatimeus (approved):

PR #2034, opened 2018-04-10
Added stop watchdogd step to WatchdogService

nicolatimeus:
Allows the `WatchdogService` to take control of the watchdog device from `watchdogd` if it is running.

### DIFF
--- a/kura/org.eclipse.kura.linux.watchdog/src/main/java/org/eclipse/kura/linux/watchdog/WatchdogServiceImpl.java
+++ b/kura/org.eclipse.kura.linux.watchdog/src/main/java/org/eclipse/kura/linux/watchdog/WatchdogServiceImpl.java
@@ -37,6 +37,9 @@ import org.slf4j.LoggerFactory;
 
 public class WatchdogServiceImpl implements WatchdogService, ConfigurableComponent {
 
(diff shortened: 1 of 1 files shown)


kura/org.eclipse.kura.linux.watchdog/src/main/java/org/eclipse/kura/linux/watchdog/WatchdogServiceImpl.java
```diff
@@ -77,6 +80,31 @@ public void updated(Map<String, Object> properties) {
         });
     }
 
+    private void stopWatchdogd() {
```
pierrepitiot, line 83: I don't remember what was exactly the issue and you have probably tried to find the best solution but would it be possible to determine which system is running and execute the corresponding command rather to try all possible commands ?

MMaiero: We are trying to step back from this habit, in order to prevent the hardcoding of platforms in Kura code. For now we try to support the maximum. The plan is to be able to define a better solution for the next release, not only for watchdog but also for networking.

pierrepitiot: This is ok for me, no problem

Issue #520, opened 2016-09-07
[Net] PppFactory is not a factory and its design is wrong


cdealti:
`PppFactory` is used by `ModemMonitorServiceImpl` to both create new instances and return already created instances. There are several defects:
1. It's not a factory. It's more similar to the [Multiton Pattern](https://en.wikipedia.org/wiki/Multiton_pattern).
2. `obtainPppService(String iface, String port)` is misleading and error prone. Internally only the interface name, for example `ppp0` is used as the key.

The above can lead to the following scenario:
- Initially `obtainPppService(ppp0, /dev/ttyACM0)` creates a new `IModemLinkService` instance for the interface `ppp0` running on the modem serial port `/dev/ttyACM0`. This instance is tracked internally by PppFactory by the interface name `ppp0`. The monitor will spawn `pppd` on `/dev/ttyACM0`
- The modem gets reset/unplugged (for any reason). On Linux is entirely possible that the serial port device name changes for example to `/dev/ttyACM1`.
- In the next loop iteration the monitor will call `obtainPppService(ppp0, /dev/ttyACM1)` again but this will return the previous instance since the serial port name does not matter. The monitor will spawn `pppd` on `/dev/ttyACM0` which is clearly wrong.

pierantoniomerlino:
A quick and dirty fix is to release the PppService after a modem reset:

```
modem.reset();
PppFactory.releasePppService(ifaceName);
pppSt = PppState.NOT_CONNECTED;
```

The PppService will be created again by the PppFactoy with the updated port at the next monitor run.

PR #2036, opened 2018-04-11
Fix for modem serial ports detection and ppp

pierantoniomerlino:
This PR fixes a potential issue on USB modem detection. The USB modems expose a set of serial ports that can be used for ppp connection, AT commands or get NMEA strings from the GPS module. 
Kura identifies the right serial port based on the tty device name. However, the modem serial ports can be mapped to tty devices in an unpredictable order and change after a modem reset.

So, this PR changes the way the tty devices are identified based on the bInterfaceNumber of the USB tty interface. Moreover, the issue reported in #520 is fixed (the PppFactory should be refactored, however...).


kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/net/modem/ModemDevice.java
```diff
@@ -21,6 +21,11 @@
 @ProviderType
 public interface ModemDevice {
 
+    /**
+     * The list of serial ports available on the device
+     *
+     * @return a list of serial ports names
+     */
```
MMaiero: `@since` missing

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/usb/UsbModemDevice.java
```diff
@@ -21,80 +21,133 @@
...
+
+    /**
+     * Adds a tty device identified by its name and USB interface number (bInterfaceNumber).
+     * The devices will be sorted by the interface number. If this is missing, the name will be used.
+     *
+     * @param ttyDev
+     *            the name of the tty device
+     * @param interfaceNumber
+     *            the number of the interface as described by the bInterfaceNumber property
+     */
```
MMaiero, line 91: the since is right here :-)

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/usb/UsbModemDevice.java
```diff
@@ -21,80 +21,133 @@
...
+    public void addTtyDev(String ttyDev, Integer interfaceNumber) {
```
cdealti, line 92: I believe that we agree that interfaceNumber **should not be null**. Otherwise `addTtyDev("foo", null)` would be equivalent to `addTtyDev("foo")` which is now deprecated.
Forcing the interface number to be non null would be the right decision but, paradoxically, the current implementation can recover such a broken modem if we let the interface number to be null.

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/usb/UsbModemDevice.java
```diff
@@ -21,80 +21,133 @@
...
 
+    /**
+     * @deprecated this method is deprecated in favor of addTtyDev(String ttyDev, Integer interfaceNumber)
+     */
+    @Deprecated
     public void addTtyDev(String ttyDev) {
-        if (!this.m_ttyDevs.contains(ttyDev)) {
-            this.m_ttyDevs.add(ttyDev);
-            Collections.sort(this.m_ttyDevs, new DevNameComparator());
+        TtyDev dev = new TtyDev(ttyDev);
```
cdealti: just call `addTtyDev(ttyDev, null)`

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/usb/UsbModemDevice.java
```diff
@@ -197,4 +250,87 @@ private int getDigitPosition(String devName) {
             return pos;
         }
     }
+
+    private class TtyDev {
+
+        @Override
```
cdealti: reorder: first come fields then constructor and getters.

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/usb/UsbModemDevice.java
```diff
@@ -197,4 +250,87 @@ private int getDigitPosition(String devName) {
...
+        public int hashCode() {
+            final int prime = 31;
+            int result = 1;
+            result = prime * result + getOuterType().hashCode();
+            result = prime * result + (this.portName == null ? 0 : this.portName.hashCode());
+            return result;
+        }
+
+        @Override
+        public boolean equals(Object obj) {
```
cdealti, line 289: This is a little bit asymmetrical with the comparator. The test for equality only considers the port name. This is a trick to support removing an `TtyDev` from the internal list by its name only. The comparator, instead, considers first interface numbers.
However, the Java doc states:
> It is generally the case, but not strictly required that (compare(x, y)==0) == (x.equals(y)). Generally speaking, any comparator that violates this condition should clearly indicate this fact. The recommended language is "Note: this comparator imposes orderings that are inconsistent with equals."

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/usb/UsbTtyDevice.java
```diff
@@ -15,30 +15,43 @@
...
     public UsbTtyDevice(String vendorId, String productId, String manufacturerName, String productName,
             String usbBusNumber, String usbDevicePath, String deviceNode) {
         super(vendorId, productId, manufacturerName, productName, usbBusNumber, usbDevicePath);
-        this.m_deviceNode = deviceNode;
+        this.deviceNode = deviceNode;
+        this.interfaceNumber = null;
+    }
+
+    public UsbTtyDevice(String vendorId, String productId, String manufacturerName, String productName,
+            String usbBusNumber, String usbDevicePath, String deviceNode, Integer interfaceNumber) {
```
cdealti: Same as above.

kura/org.eclipse.kura.linux.usb/src/main/java/org/eclipse/kura/linux/usb/UsbTtyInterface.java
```diff
@@ -0,0 +1,66 @@
...
+ *******************************************************************************/
+package org.eclipse.kura.linux.usb;
+
+import java.io.ByteArrayInputStream;
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.nio.charset.Charset;
+import java.util.List;
+
+import org.apache.commons.exec.CommandLine;
```
cdealti: Not sure if adding two new dependencies (from two different bundles) is worth for a temporary fix.  This util will be removed and the interface number returned by the native code.

kura/org.eclipse.kura.linux.usb/src/main/java/org/eclipse/kura/linux/usb/UsbTtyInterface.java
```diff
@@ -0,0 +1,66 @@
...
+import org.apache.commons.exec.DefaultExecutor;
+import org.apache.commons.exec.PumpStreamHandler;
+import org.apache.commons.io.IOUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class UsbTtyInterface {
```
cdealti: `LinuxUdevUtil`

kura/org.eclipse.kura.linux.usb/src/main/java/org/eclipse/kura/linux/usb/UsbTtyInterface.java
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+    public static Integer getInterfaceNumber(String deviceNode) {
+        Integer interfaceNumber = null;
+        String command = "udevadm info --query=property --name=" + deviceNode;
+        logger.debug("Executing: {}", command);
+        try {
+            return parse(execute(command));
+        } catch (IOException e) {
+            logger.error("Failed to execute command {}", command, e);
```
cdealti: As a general rule, it would be better to let the caller catch the exception.

cdealti (changes requested):

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/usb/UsbModemDevice.java
```diff
@@ -197,4 +248,87 @@ private int getDigitPosition(String devName) {
             return pos;
         }
     }
+
+    private class TtyDev {
```
cdealti: You want this nested class to be static (you don't want to mess with the outer type)

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/usb/UsbModemDevice.java
```diff
@@ -197,4 +248,87 @@ private int getDigitPosition(String devName) {
...
+            return this.interfaceNumber;
+        }
+
+        private UsbModemDevice getOuterType() {
+            return UsbModemDevice.this;
+        }
+
+        @Override
+        public String toString() {
+            return "TtyDev [portName=" + this.portName + ", interfaceNumber=" + this.interfaceNumber + "]";
```
cdealti: NPE if interface number is null

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/usb/UsbModemDevice.java
pierantoniomerlino: @cdealti Ok, I can add a comment on the comparator/equals methods.

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/usb/UsbModemDevice.java
pierantoniomerlino: @cdealti I agree.

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/usb/UsbModemDevice.java
pierantoniomerlino: @cdealti If we decide that the interface number can be null (and this is the case for this fix), we have to take care of this NPE.

PR #2059, opened 2018-04-20
Enh log4j2

MMaiero:
Will require CQs


MMaiero:
CQs filed

pierantoniomerlino:
On the OSGI terminal the `getkuraloglevel` returns:

```
Default log level: INFO
: INFO
org.eclipse.kura: INFO
```

Moreover, setting the log level to DEBUG seems that has no effect. I tried to modified a bundle using explicitly the new log4j with the same result.

pierantoniomerlino:
I run several tests on a RPi. It seems that the configuration using .property files has some flaws. I re-write the configuration using a .xml file and it seems to work better. Here my example:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<!--
#
# Copyright (c) 2011, 2018 Eurotech and/or its affiliates
#
#  All rights reserved. This program and the accompanying materials
#  are made available under the terms of the Eclipse Public License v1.0
#  which accompanies this distribution, and is available at
#  http://www.eclipse.org/legal/epl-v10.html
#
# Contributors:
#   Eurotech
#
-->
<Configuration status="info" strict="true" name="KuraConfig" monitorInterval="30">

    <Properties>
        <Property name="filename">/var/log/kura</Property>
    </Properties>
    <Filter type="ThresholdFilter" level="trace"/>

    <Appenders>
        <RollingFile name="RollingFile" fileName="${filename}.log" filePattern="${filename}-%d{MM-dd-yy-HH-mm-ss}-%i.log.gz">
            <PatternLayout>
                <Pattern>%d{ISO8601} [%t] %-5p %c{1.} - %m%n</Pattern>
            </PatternLayout>
            <Policies>
                <SizeBasedTriggeringPolicy size="20 MB"/>
            </Policies>
            <DefaultRolloverStrategy max="10"/>
        </RollingFile>
    </Appenders>

    <Loggers>
        <Logger name="org.eclipse" level="info" additivity="false">
            <AppenderRef ref="RollingFile"/>
        </Logger>

        <Logger name="org.eclipse.kura.example.ble.tisensortag.tinyb.BluetoothLe" level="info" additivity="false">
            <AppenderRef ref="RollingFile"/>
        </Logger>

        <Root level="info">
            <AppenderRef ref="RollingFile"/>
        </Root>
  </Loggers>

</Configuration>
```

Here is also a custom debug level for a specific package.

pierantoniomerlino:
Using the configuration file shown above, the kura-console.log is flooded by messages like these:

```
2018-08-09 10:16:10,661 Start Level: Equinox Container: 7731e3fc-74ae-4f7b-aefb-0bfe65d15cdf INFO Log4j appears to be running in a Servlet environment, but there's no log4j-web module available. If you want better web container support, please add the log4j-web JAR to your web archive or server lib directory.
```

and

```
2018-08-09 10:15:26,152 Log4j2-TF-391-Scheduled-353 INFO File /opt/eclipse/kura/kura/log4j.xml was modified on 1533809700000, previous modification was 1533809660000
```

The first one is explained in [1], while the second one is related to the re-loading of the configuration file.
To avoid that, simply change the first line of the file as follows:

```
<Configuration status="warn" strict="true" name="KuraConfig" monitorInterval="30">
```

where the status property (that sets the log level of the logger itself) is changed from `info` to `warn`.

[1] https://stackoverflow.com/questions/37344431/why-log4j-thinks-my-project-run-in-servlet-environment/37345105

PR #2063, opened 2018-04-23
Fix a NPE and provide proper cause of error.

ctron:
Currently, when no marshaller/unmarshaller is found, the code simply throws a NPE. In addition it will wrap all root causes into a generic INTERNAL_ERROR.

This change now throws a proper exception (ENCODER_ERROR) and provides a helpful error message.

### DIFF
--- a/kura/org.eclipse.kura.core.configuration/src/main/java/org/eclipse/kura/core/configuration/ConfigurationServiceImpl.java
+++ b/kura/org.eclipse.kura.core.configuration/src/main/java/org/eclipse/kura/core/configuration/ConfigurationServiceImpl.java
@@ -1089,11 +1089,13 @@ public class ConfigurationServiceImpl implements ConfigurationService, OCDServic
         String xmlResult;
         try {
             xmlResult = marshal(conf);
-            if (xmlResult.trim().isEmpty()) {
+            if (xmlResult == null || xmlResult.trim().isEmpty()) {
                 throw new KuraException(KuraErrorCode.INVALID_PARAMETER, conf);
             }
-        } catch (Exception e1) {
-            throw new KuraException(KuraErrorCode.INTERNAL_ERROR, e1);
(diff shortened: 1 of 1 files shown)


ctron:
@dwoodard1 @cdealti @MMaiero could someone of your review this, I think it is a small patch, so it really shouldn't require much time.

kura/org.eclipse.kura.core.configuration/src/main/java/org/eclipse/kura/core/configuration/ConfigurationServiceImpl.java
```diff
@@ -1089,11 +1089,13 @@ private void writeSnapshot(long sid, XmlComponentConfigurations conf) throws Kur
         String xmlResult;
         try {
             xmlResult = marshal(conf);
-            if (xmlResult.trim().isEmpty()) {
+            if (xmlResult == null || xmlResult.trim().isEmpty()) {
```
MMaiero, line 1092: Am I wrong or having now marshaller throwing an exception, if result is null, makes this check for null redundant?

MMaiero (approved):
Hi Jens, sorry for the delay but we had few days where we were not at the office. The PR seems ok, I made only one comment.

kura/org.eclipse.kura.core.configuration/src/main/java/org/eclipse/kura/core/configuration/ConfigurationServiceImpl.java
ctron: It is indeed. But I would keep it in for stability. Having and additional null check prevents you from unforeseen things in the future.

MMaiero: Agreed

PR #2064, opened 2018-04-23
Properly declare required/provided services

ctron:
As the ConfigurationServiceImpl doesn't properly work when the XML marshaller/unmarshaller is missing, it should declare a dependency on this service.


ctron:
**Note:** This PR requires Tycho 1.2.0 (which will be released in a few days). A simple travis re-build should suffice to re-test. Until that has been tested this PR should not be merged.

ctron:
Ok now that tycho 1.2.0 is released this actually works as expected. The problem is that the Kura XML marshaller bundle is using internal data structure from Kura Core Configuration, which creating cyclic dependencies. I do think that there should be some proper abstraction. But I don't have the time to work on this. So let's just live with the fact that service dependencies are not properly declared and startups might fail due to that.

PR #2073, opened 2018-05-03
Added Driver for Eddystones

pierantoniomerlino:
This PR adds a new Driver that supports Eddystones. Since they are a sort of beacons, only the listen feature is supported. Read and writes operations have no effect. Moreover, the output of the Driver is a semi-colon separated string like this:

URL;url;txPower;rssi;distance

for URL frame type and

UID;namespace;instance;txPower;rssi;distance

for UID frame type. So, the value.type is not used.


kura/org.eclipse.kura.driver.eddystone.provider/src/main/java/org/eclipse/kura/internal/driver/eddystone/EddystoneDriver.java
```diff
@@ -0,0 +1,319 @@
...
+            EddystoneListener sensorListener = iterator.next();
+            if (sensorListener.getListener().equals(listener)) {
+                iterator.remove();
+            }
+        }
+    }
+
+    @Override
+    public void write(final List<ChannelRecord> records) throws ConnectionException {
+        logger.warn("Write operation not supported");
```
amitjoy, line 233: According to contract, it should throw `KuraRuntimeException` with `KuraErrorCode#OPERATION_NOT_SUPPORTED`

kura/org.eclipse.kura.driver.eddystone.provider/src/main/java/org/eclipse/kura/internal/driver/eddystone/EddystoneDriver.java
```diff
@@ -0,0 +1,319 @@
...
+
+        @Override
+        public void close() throws Exception {
+            // Not needed
+        }
+
+        @Override
+        public List<ChannelRecord> execute() throws ConnectionException, KuraException {
+            logger.warn("Read operation not supported");
+            return new ArrayList<>();
```
amitjoy: You can also return `Collections.emptyList()`

kura/org.eclipse.kura.driver.eddystone.provider/src/main/java/org/eclipse/kura/internal/driver/eddystone/EddystoneDriver.java
```diff
@@ -0,0 +1,319 @@
...
+        }
+
+        @Override
+        public List<ChannelRecord> getChannelRecords() {
+            return new ArrayList<>();
```
amitjoy: You can also return `Collections.emptyList()`

kura/org.eclipse.kura.driver.eddystone.provider/src/main/java/org/eclipse/kura/internal/driver/eddystone/EddystoneDriver.java
pierantoniomerlino: 👍

kura/org.eclipse.kura.driver.eddystone.provider/src/main/java/org/eclipse/kura/internal/driver/eddystone/EddystoneDriver.java
pierantoniomerlino: 👍

kura/org.eclipse.kura.driver.eddystone.provider/src/main/resources/org.eclipse.kura.driver.eddystone.provider.dpp
```diff
@@ -0,0 +1,20 @@
+#Deployment Plugin Project File
```
MMaiero: I don't believe we need the dpp file. Maybe add a git ignore or delete it.

MMaiero:
We don't have tests for this driver?

kura/org.eclipse.kura.driver.eddystone.provider/src/main/resources/org.eclipse.kura.driver.eddystone.provider.dpp
pierantoniomerlino: Mhh... so we should remove also that for the SensorTag driver.

pierantoniomerlino:
@MMaiero No, we don't have tests for this driver.

MMaiero:
We should try to have some black box tests that obviously simulate a device or what a device shares, in order to catch possible issues/ simplify an eventual process of refactoring.

Issue #2070, opened 2018-04-30
Icon not working for factory components


ctron:
It looks like the "Icon" from the MetaType model is not working when the component is a factory component. It only shows a "broken image" image.

ctron:
It looks like the Web UI is requesting the icon by the "service PID" and not by the "factory PID". Which of course won't work.

PR #2076, opened 2018-05-04
Fix a bunch of issues when loading the service icon

ctron:
This change fixes a bunch of issues when loading the icons for services.
Including #2070, a possible file handle resource leak, corrupted images
and unnecessary lookups.


MMaiero:
Manually merged

PR #2077, opened 2018-05-04
Fix Artemis version to support Kura build infrastructure

ctron:


ctron:
@MMaiero did you have a chance to look into the PR?

target-platform/config/kura.target-platform.build.properties
```diff
@@ -13,7 +13,9 @@ org.eclipse.kura.windows.system.version=1.0.0-SNAPSHOT
 org.eclipse.kura.windows.launcher.version=1.0.0-SNAPSHOT
 org.eclipse.kura.windows.service.version=1.0.0-SNAPSHOT
 org.moka7.version=1.0.100
-org.apache.activemq.artemis.version=2.5.0
+
+org.apache.activemq.artemis.version=2.5.100-SNAPSHOT
```
MMaiero: Why not having this as 2.5.0-SNAPSHOT?

MMaiero:
Please also keep in touch with EMO to remove the artifacts erroneously pushed to the release repo.

target-platform/config/kura.target-platform.build.properties
ctron: Because version 2.5.0 has been pushed out. So the next version would be 2.5.100-SNAPSHOT from my understanding.

MMaiero: 2.5.0 has never been pushed out officially. So we can have EMO to clean the repo and work with SNAPSHOt and finally have the official 2.5.0 pushed to the repo when we will release Kura 4.0.0. What do you think?

ctron: If you prefer that. I will change the PR and get in contact with EMO.

MMaiero: Thanks.

ctron:
Updated the PR to use version 2.5.0-SNAPSHOT. Sent out e-mail to the Eclipse Webmaster, CCing @MMaiero  and @dwoodard1

MMaiero:
Thank you very much. Merging after travis build.

PR #2081, opened 2018-05-07
Fix Wire component labels

ctron:
This PR fixes the Wire Component labels by using the "name" attribute of the OCD where possible.

For other services this is already present, however for the wire components this is still the old behavior of using the last segment of the PID, and falling back to a hardcoded list of of alias labels.

This change prefers the provided name as label (if present) and falls back to the PID handling, like with the services. All components, except the WiresAsset already did support that.


nicolatimeus:
@ctron please update `kura.build.properties` where needed.

MMaiero (approved):

nicolatimeus:
No need to update `kura.build.properties`, it is ok after recent pull request merges.

I've noticed the following glitch:

If you create a new wire component the label of the tab above configuration section is correctly
displayed as `<component name> - <kura.service.pid>.`

screen shot 2018-05-10 at 14 48 30

If you save the graph and reload the page, the label becomes `<kura.service.pid> - <kura.service.pid>`:

screen shot 2018-05-10 at 14 48 54

ctron:
@nicolatimeus Ok .. I will look at this problem.

ctron:
Can you please re-check. I think this should be fixed now.

nicolatimeus (approved):

PR #2083, opened 2018-05-08
Improve debug output of known wires

ctron:
This change zero pads the wire numbers and also shows if a wire
is actually present for the configuration.


nicolatimeus (approved):

nicolatimeus:
@ctron please update the bundle version in `kura.build.properties`, then we can merge

ctron:
Done

PR #2090, opened 2018-05-08
Fixed CellularModem instances state persisting across resets

### LINKED ISSUE #2089: [NetAdmin] Some state of the logic that manages modems persists across resets
A part of the logic that manages modems is provided by the `CellularModem` subclasses, that are instantiated by a `CellularModemFactory` and are cached.
Currently a `CellularModemFactory` always returns the same `CellularModem` instance for a given modem, even if the managed modem has been reset.
This behaviour can cause problems because some of the `CellularModem` implementations keep internal state.

For example, the `TelitModem` class defines a boolean flag that specifies if the GPS is enabled or not, if the modem is reset with GPS enabled and the `NetAdmin` fails to disable GPS, the value of this flag will be retained. This can cause the NetAdmin to incorrectly assume that the GPS is still enabled after the reset even if it is not.

The state of the `CellularModem` instances should be cleared after a modem reset.

nicolatimeus:
Closes #2089 


MMaiero (approved):

nicolatimeus:
Yes good catch, I forgot to update the version of the api package. I guess the version will be updated with the Update Equinox PR.

PR #2092, opened 2018-05-09
[#2091] Compile with a dependency to GLIBCXX_3.4 only

ctron:
Before:

~~~
Version References:
  required from libgcc_s.so.1:
    0x0b792650 0x00 08 GCC_3.0
  required from ld-linux-x86-64.so.2:
    0x0d696913 0x00 07 GLIBC_2.3
  required from libc.so.6:
    0x09691a75 0x00 06 GLIBC_2.2.5
  required from libstdc++.so.6:
    0x0bafd177 0x00 05 CXXABI_1.3.7
    0x056bafd3 0x00 04 CXXABI_1.3
    0x08922974 0x00 03 GLIBCXX_3.4
    0x0297f870 0x00 02 GLIBCXX_3.4.20
~~~

After:

~~~
Version References:
  required from libgcc_s.so.1:
    0x0b792650 0x00 07 GCC_3.0
  required from ld-linux-x86-64.so.2:
    0x0d696913 0x00 06 GLIBC_2.3
  required from libc.so.6:
    0x09691a75 0x00 05 GLIBC_2.2.5
  required from libstdc++.so.6:
    0x0bafd177 0x00 04 CXXABI_1.3.7
    0x056bafd3 0x00 03 CXXABI_1.3
    0x08922974 0x00 02 GLIBCXX_3.4
~~~

The shared library also loads on Ubuntu 16.04 LTS.

Issue #2091, opened 2018-05-08
version `GLIBCXX_3.4.20' not found


ctron:
It looks like `libjavatinyb.so` is linked against a libstdc++ which will not be available on CentOS/RHEL.

~~~
/git/kura/target-platform/tinyb-parent/tinyb.x86_64/src/main/lib/libjavatinyb.so: /lib64/libstdc++.so.6: version `GLIBCXX_3.4.20' not found (required by /h/git/kura/target-platform/tinyb-parent/tinyb.x86_64/src/main/lib/libjavatinyb.so)
	linux-vdso.so.1 =>  (0x00007ffddb7da000)
	libtinyb.so.0.5.0-11-g76c0e10 => not found
	libstdc++.so.6 => /lib64/libstdc++.so.6 (0x00007fb52782c000)
	libgcc_s.so.1 => /lib64/libgcc_s.so.1 (0x00007fb527616000)
	libc.so.6 => /lib64/libc.so.6 (0x00007fb527249000)
	/lib64/ld-linux-x86-64.so.2 (0x00007fb527d56000)
	libm.so.6 => /lib64/libm.so.6 (0x00007fb526f47000)

~~~

ctron:
I think is was closed by @MMaiero's merge, right?

PR #2093, opened 2018-05-09
Add Intel UP2 support with CentOS 7

ctron:
Add support for Intel UP² based on CentOS 7

**Note:** This PR also requires #2091 to be fixed. Possibly by merging #2092


kura/distrib/src/main/resources/intel-up2-centos-7/kura.properties
```diff
@@ -0,0 +1,157 @@
+#
+# Copyright (c) 2011, 2017 Eurotech and others
```
MMaiero: 2018?

kura/distrib/src/main/resources/intel-up2-centos-7/kura_install.sh
```diff
@@ -0,0 +1,121 @@
+#!/bin/sh
+#
+# Copyright (c) 2016 Red Hat Inc and others
```
MMaiero: 2018?

Same for other files contributed here.

MMaiero:
@dwoodard1 I don't have a board to do some testing. Could you do a check with yours?

ctron:
@MMaiero I added 2018 to the copyright headers. But only when they existed and I actually modified the file.

MMaiero:
@ctron Do you have any hurry on this? I would like to keep on hold until we get feedback from Dave. Just to verify that everything works as expected with a proper QA validation.

ctron:
I don't see anything which speaks against merging that now. It doesn't change Kura itself, just adds a new distribution profile. And letting the PR wait and longer, only causes more trouble later on when it has to be re-based and possible conflicts being resolved.

And I also don't really understand what you mean by proper QA validation at this point. That is normally done right before the release of Kura. And when Kura enters that phase, then changes like that won't get merged as they are deemed too risky to be merged. At least that was the argument in the past.

So unless anything speaks against merging this, then I would say, merge it now.

dwoodard1:
Sorry, just now getting around to this. Question, why are we targeting CentOS? If I'm not wrong, the UP2 comes with Ubuntu (at least mine did). Isn't installing CentOS before installing Kura an unnecessary step?

ctron:
Well you may be targeting Ubuntu as well. But my focus is on targeting CentOS. And I think in the spirit of being open source, there is nothing wrong with that approach.

dwoodard1:
Of course I have no issue supporting CentOS. I think it is a nice addition. I just didn't understand starting here, since unlike the Raspberry Pi, the UP2 ships with an installed OS. I will create a separate PR for Ubuntu support, since I assume that will be the simplest path for new Kura/UP2 users.

Since this doesn't impact the framework directly, only adds a new distribution, I think we can merge. If we find issues we can create separate issues. I can test this after I have tested the default Ubuntu installation. @ctron did you want to make any changes before I merge?

ctron:
Agreed. No please go ahead and merge. As you said, if we find issues we can always improve.

Issue #1638, opened 2017-09-05
Kura does not support Consistent Network Device Naming


pierantoniomerlino:
The last Raspian Stretch image (Release date: 2017-08-16) [1] supports by default the Consistent Network Device Naming feature [2], [3]. Kura uses the eth0 interface as primary network interface and uses it to set up several components. If the CNDN is enabled, Kura fails to start and the webUI is not shown in the browser.

Since this new feature will become common in new distros, Kura should support it.
In the meanwhile, a simple workaround is to disable it, following this procedure:
1. edit the ```/boot/cmdline.txt```
2. add at the end of the line ```net.ifnames=0```
3. reboot the device

[1] https://www.raspberrypi.org/downloads/raspbian/
[2] https://access.redhat.com/documentation/en-US/Red_Hat_Enterprise_Linux/7/html/Networking_Guide/ch-Consistent_Network_Device_Naming.html
[3] https://en.wikipedia.org/wiki/Consistent_Network_Device_Naming

Amar82:
Hi Pier,

nice to meet you again :) .

my scenario is my sensors use i2c to connect to kura, and kura should connect  via wifi to internet router to connect to broker in cloud.
I am trying to connect Kura ( 3.1), via wifi ( station mode) to my home router ( router is connected to Internet).  But I am not able to do so.
I am using the wifi driver of Pi3 B+, there is NO additional wifi driver.

I have following network settings:
 ethernet setting : disabled
lo setting disabled

wlan settings - 
WAN
manual ip setting : 192.168.0.7 
gateway : to local router 192.168.0.1
corresponding subnet masks & DNS addresses

but I am getting errors & broker is not getting connected.

Please suggest the settings , or what I am doing wrong.
Should I use separate wifi driver ?

MMaiero:
@Amar82 I believe your question is not related to the issue opened by @pierantoniomerlino.
I've also seen you opened a forum issue here with the same question: https://www.eclipse.org/forums/index.php/m/1776149/#msg_1776149

I believe it's better if you don't cross post. In this way, we will not have duplication. 
Or, if needed, please open a specific issue in GitHub.

Thanks.

Best regards,
Matteo

cdealti:
There are hardcoded "eth0", "eth1", "wlan0" and "wlan1" occurrences in the source code

PR #2094, opened 2018-05-09
Sort interfaces alphabetically

ctron:


kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/server/GwtStatusServiceImpl.java
```diff
@@ -226,6 +226,22 @@ private void fillFromDataService(List<GwtGroupedNVPair> pairs, final String data
                 return Collections.emptyList();
             }
 
+            Collections.sort(gwtNetInterfaceConfigs, new Comparator<GwtNetInterfaceConfig>() {
```
amitjoy: should not it be easier with Java 8 Comparators? Just a proposal.

```java
Collections.sort(gwtNetInterfaceConfigs, Comparators.comparing(GwtNetInterfaceConfig::getName));
```

ctron: As of now Java 8 isn't supported as you might remember. But as we agreed on yesterday's call we will upgrade to GWT 2.8.x and then we can go ahead and use Java 8 APIs.

MMaiero: If I'm not wrong, java 8 apis can be used in the server part, not in the client part.
This is server part, so it can be used. https://github.com/eclipse/kura/blob/6eaec61ca692a7bded9db1846eb31570626e175c/kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/server/GwtComponentServiceImpl.java#L108

amitjoy: I agree with @MMaiero. Since it is the server part, we have been using Java 8 and we can also do it here as well.

ctron: Well I am not sure if it is a good practice to mix Java 7 and 8 targets in a single project. But assuming that we will get GWT 2.8.x anyway, lets move forward to use Java 8 and simple ignore Java 7 from now on.

MMaiero:
Maybe a dumb question: why do we need this?

ctron:
You want me to explain the benefit of an alphabetical sort order? Or did I somehow misunderstand the question?

MMaiero:
No, for what I remember, currently there is a specific ordering: lo first, then ethernet interfaces (ordered alphabetically), then wifi (ordered alphabetically), then ppp interfaces (ordered alphabetically)

ctron:
Well in my case `enp2s0`appeared after `enp3s0`, so my assumption was that this is not sorted alphabetically.

MMaiero:
I believe the it is due to the fact that currently the Kura networking is not working properly with Consistent Device Naming https://github.com/eclipse/kura/issues/1638
Maybe @ibinshtok can give us some hints.

ctron:
Well I am not sure how this would interfere with alphabetically sorting device names. But I think that this PR would help in providing a stable, understandable sort order.

MMaiero (approved):

PR #2103, opened 2018-05-10
Changed exception type in WireAsset

### LINKED ISSUE #2102: emitAllReadChannels method doesn't handle Runtime exceptions
The `emitAllReadChannels` method in `WireAsset` class doesn't handle the unchecked exceptions, but only `KuraException`. So, if an unchecked exception as `RuntimeException` is thrown, it is simply passed to the upper layer and this can stop the current thread.

For example, if a Driver doesn't support readings, the `emitChannelRecords` throws an `IllegalArgumentException` that is not caught by the `emitAllReadChannels` method. The results is that the current thread on Wires is closed.

pierantoniomerlino:
Closes #2102.


nicolatimeus (approved):

nicolatimeus:
Maybe we can do the same in the `writeChannels()` method

PR #2104, opened 2018-05-10
Added Driver for iBeacons

pierantoniomerlino:
This PR adds a new Driver that supports iBeacons. As done with the Driver for Eddystones, only the listen feature is supported. Read and writes operations have no effect. Moreover, the output of the Driver is a semi-colon separated string like this:

uuid;txPower;rssi;major;minor;distance

So, the value.type is not used.


kura/features/org.eclipse.kura.driver.ibeacon/pom.xml
```diff
@@ -0,0 +1,34 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
MMaiero, line 1: Please add copyright

kura/org.eclipse.kura.driver.ibeacon.provider/OSGI-INF/IBeaconDriverComponent.xml
```diff
@@ -0,0 +1,28 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+    Copyright (c) 2017 Eurotech and/or its affiliates
```
MMaiero, line 3: 2018

PR #2111, opened 2018-05-11
Added Wires/Driver development components

### LINKED ISSUE #2110: Add Dummy Driver
A Dummy Driver can be added to the codebase, serving as a starting point that can be used to quickly start writing a new Driver.

This driver can also be useful when used together with a `WireAsset` to inject arbitrary values on the wire.

Its configuration contains a single configuration parameter, **value**. The value entered in this field as a `String` will be converted to the value type specified in the **value.type** field and then emitted on the wire.

The driver also supports writes. If a value is written to a channel, the written value will be emitted in successive reads instead of the value hardcoded in channel config. Written values can be cleared by updating the driver configuration.

The driver also showcases the use of `ChannelListener`s and is able to simulate channel events.

It also allows to simulate connection delays and some error cases like returning invalid `DriverDesrcriptor`, useful for framework debugging.

nicolatimeus:
Provides a bunch of components useful during wire graph development:

* The `DummyDriver` described in #2110 
* The `Sink` wire component, that allows to measure timings between two consecutive envelopes.
* The `Delay` wire component, that allows to introduce random delays in envelope processing, simulating slow components.

Closes #2110 


kura/distrib/config/kura.build.properties
```diff
@@ -71,6 +71,7 @@ org.eclipse.kura.example.camel.quickstart.version=1.1.200
 org.eclipse.kura.example.can.version=1.0.100
 org.eclipse.kura.example.gpio.version=1.0.200
 org.eclipse.kura.example.driver.sensehat.version=1.0.0
+org.eclipse.kura.wire.devel.component.provider.version=1.0.0-SNAPSHOT
```
amitjoy, line 74: I am just wondering if `.devel` could be succinctly changed to `develop` for better comprehension

kura/examples/org.eclipse.kura.wire.devel.component.provider/META-INF/MANIFEST.MF
```diff
@@ -0,0 +1,24 @@
...
+Bundle-SymbolicName: org.eclipse.kura.wire.devel.component.provider
+Bundle-Version: 1.0.0.qualifier
+Bundle-ActivationPolicy: lazy
+Import-Package: org.eclipse.kura;version="[1.0,2.0)",
+ org.eclipse.kura.channel;version="[1.0,2.0)",
+ org.eclipse.kura.channel.listener;version="[1.0,2.0)",
+ org.eclipse.kura.configuration;version="1.1.1",
+ org.eclipse.kura.configuration.metatype;version="1.0.2",
+ org.eclipse.kura.core.configuration.metatype,
+ org.eclipse.kura.db;version="[1.0.0,2.0.0)",
```
amitjoy: for consistency, `[1.0,2.0)`

kura/examples/org.eclipse.kura.wire.devel.component.provider/src/org/eclipse/kura/wire/devel/DataTypeHelper.java
```diff
@@ -0,0 +1,46 @@
...
+
+import org.eclipse.kura.type.DataType;
+import org.eclipse.kura.type.TypedValue;
+import org.eclipse.kura.type.TypedValues;
+
+public final class DataTypeHelper {
+
+    private static final Decoder decoder = Base64.getDecoder();
+
+    private DataTypeHelper() {
```
amitjoy, line 23: You could also throw IllegalAccessError since private constructors can also be invoked using reflection

kura/examples/org.eclipse.kura.wire.devel.component.provider/src/org/eclipse/kura/wire/devel/DataTypeHelper.java
```diff
@@ -0,0 +1,46 @@
...
+    }
+
+    public static TypedValue<?> parseTypedValue(DataType type, String valueString) {
+        switch (type) {
+        case BOOLEAN:
```
amitjoy: I believe the switch block is not properly formatted

kura/examples/org.eclipse.kura.wire.devel.component.provider/src/org/eclipse/kura/wire/devel/DataTypeHelper.java
```diff
@@ -0,0 +1,46 @@
...
+        case BYTE_ARRAY:
+            return TypedValues.newByteArrayValue(decoder.decode(valueString));
+        case DOUBLE:
+            return TypedValues.newDoubleValue(Double.parseDouble(valueString));
+        case STRING:
+            return TypedValues.newStringValue(valueString);
+        case INTEGER:
+            return TypedValues.newIntegerValue(Integer.parseInt(valueString));
+        default:
+            throw new IllegalArgumentException();
```
amitjoy: You could also write a meaningful message. Just a proposal.

kura/examples/org.eclipse.kura.wire.devel.component.provider/src/org/eclipse/kura/wire/devel/DataTypeHelper.java
```diff
@@ -0,0 +1,46 @@
...
+import org.eclipse.kura.type.TypedValues;
+
+public final class DataTypeHelper {
+
+    private static final Decoder decoder = Base64.getDecoder();
+
+    private DataTypeHelper() {
+    }
+
+    public static TypedValue<?> parseTypedValue(DataType type, String valueString) {
```
amitjoy: to be consistent with the practice you are also following in this PR - use `final`

MMaiero (approved):

Issue #1195, opened 2017-03-14
[Net] Firewall IP Forwarding rules entered in the Web UI lost on reboot


cdealti:
Affects Kura 2.0.2. Not checked against Kura 2.1.0.

The use case is configure IP forwarding between wlan0 and ppp0 but **without** NAT.
This can be achieved by setting IP Forwarding rules in the Web UI.

When you save the IP Forwarding rules, they are correctly saved in the snapshot and in the firewall configuration file ``/etc/sysconfig/iptables``. They are also correctly listed with ``iptables --list``.
The above network configuration works nicely.

Then two things happen:

1. In the Web UI if you click outside the IP Forwarding tab and you go back to it, the rules are no longer there.

2. If you change the network configuration, the /etc/sysconfig/iptables and the snapshot are rewritten and the IP forwarding rules are deleted. The above network configuration stops working.

As a workaround we added the IP forwarding rules to the firewall custom script which triggered another bug fixed by #1194.

There is another issue, difficult to describe:
If you use the firewall custom script, ``/etc/init.d/firewall_cust``, on the next configuration change, the custom rules are saved also /etc/sysconfig/iptables.
If you change the configuration again the custom rules are removed from /etc/sysconfig/iptables.
So, if the custom rules are not present in /etc/sysconfig/iptables, they are added. If they are present then they are removed.
This is weird but it does not cause any troubles

markoer:
I can confirm that the behavior is similar in 3.2.0-SNAPSHOT on RPi3. And it happens both if NAT is enabled on wlan0 and if it is not.

Note that the initial forwarding rule from wlan0 to eth0 that is set up in `/etc/sysconfig/iptables` is not set up in snapshot 0 and is not visible in Firewall->IP Forwarding/Masquerading tab.

I can proceed to adding an additional rule that forwards from eth0 to wlan0 (ppp0 is not available). The change is visible in the new snapshot, in `/etc/sysconfig/iptables` and in output of `iptables -L -n -v`. Now for the part where the behaviors differ: I can leave the firewall page and return and the rule will still be visible. When I reload the page the rule changes a bit - source and destination networks are now displayed (0.0.0.0/0). If I restart Kura, the rule disappears from the table, but it's still in the snapshot and file and is applied on the gateway.

Change of eth0 from Using DHCP to Manually will remove the rule from the file and the system.

As far as the custom script is concerned, the observed behavior is this:
[...]

PR #2119, opened 2018-05-17
Fix ip forwarding

ibinshtok:


ibinshtok:
@MMaiero, @pierantoniomerlino I created this PR, but still working on it ...

MMaiero:
@ibinshtok As far as you remember, is this PR ready for merge?
Do we have a corresponding issue?

Thanks

ibinshtok:
@MMaiero sorry for not replying, I have not seen your comment. This PR was an attempt to fix the Kura issue #1195. These changes worked for me, but only on platforms that support rule tagging (e.g. -m comment --comment "pf1") added to 'Port Forward' and 'IP Forward' rules. If platform doesn't support this, firewall implementation would be broken. If I remember right, Eurotech platforms supported it, but not Raspberry Pi.

nicolatimeus:
@ibinshtok just a quick question about your last comment:
When you say that this might break firewall on Raspberry Pi, do you mean that firewall will work with this PR, but it will still have issues described in #1195 or that merging this PR will result in a completely not working firewall implementation for that platform?

ibinshtok:
@nicolatimeus No, the issue described in #1195 will be fixed, but if tagging is not supported on a particular platform, an iptables rule that contains something like '-m comment --comment "pf1"' will fail to apply. So that iptables, iptables-save, and iptables-restore commands will fail. If I remember right, I developed it on BoltMAR 20-28, so that it will work with Eurotech most recent bsp versions. But I believe it failed on Pi.

cdealti:
@ibinshtok would you merge this if the only supported platform was the RPI?

ibinshtok:
@cdealti I would not merge it unless all platforms support it.

pierantoniomerlino:
The feature was added in #3334

PR #2136, opened 2018-05-29
Add a non-networked profile for the Intel UP2 with CentOS 7

ctron:
Also switch to "bash" explicitly and detect script errors.


MMaiero:
Ok for me. I'm not sure about the dates in the copyright headers.

MMaiero (approved):

ctron:
So this can be merged now?!

PR #2137, opened 2018-05-29
Enhancement platform segregation

MMaiero:


kura/org.eclipse.kura.linux.net/src/main/java/org/eclipse/kura/linux/net/dns/LinuxNamed.java
```diff
@@ -178,39 +159,34 @@ public boolean enable() throws KuraException {
...
-                    || OS_VERSION.equals(KuraConstants.ReliaGATE_50_21_Ubuntu.getImageName() + "_"
-                            + KuraConstants.ReliaGATE_50_21_Ubuntu.getImageVersion())) {
+            } else if (OS_VERSION.equals(KuraSupportedPlatforms.RASPBIAN_100.getImageName())
+                    || OS_VERSION.equals(KuraSupportedPlatforms.DEBIAN_100.getImageName())) {
                 result = LinuxProcessUtil.start("/etc/init.d/bind9 start");
-            } else if (OS_VERSION.equals(KuraConstants.Fedora_Pi.getImageName())
-                    || OS_VERSION.equals(KuraConstants.Intel_Up2.getImageName())
-                    || OS_VERSION.equals(KuraConstants.Reliagate_20_26.getImageName() + "_"
-                            + KuraConstants.Reliagate_20_26.getImageVersion())) {
+            } else if (OS_VERSION.equals(KuraSupportedPlatforms.FEDORA_2X.getImageName())
```
ctron, line 180: Wouldn't it make sense to provide some kind of `NameServerController` interface and then implement that by some `SystemdNameServerController` … instead of keeping that if-nightmare alive?!

MMaiero: That is the plan. This is just a first step. I believe I've talked about this in one of the last weekly meetings.

PR #2139, opened 2018-05-31
Channels Upload/Download from csv file

LucaDazi:
- Added new feature to download the channel table for a Asset to a csv file.
- Added new feature to upload the channel table for a Asset from a csv file.
  CSV parsing and basic validation already in place.
- Enanched the web UI Error label to consider line breaks and carriage returns.


MMaiero:
@LucaDazi Hi Luca,
I've made a commit on top of your work.
For what I've tested, it works.
Maybe the only point that do not convince me is the fact that the channels download is enabled when the configuration is dirty: in this case you download the previous saved configuration (if any) and not the one shown.
Maybe, in this case, it's better to simply disable the download button if the configuration is dirty.

MMaiero:
@LucaDazi Please confirm that the changes are ok.
Then we can merge.

LucaDazi:
Thanks @MMaiero. The changes are perfectly good for me.

PR #2140, opened 2018-06-01
Ignore the created JAR file from Git

ctron:
Currently a JAR file named `org.eclipse.kura.driver.block.jar` lingers in my git status and wants to be committed. I think this file should be ignored.


kartben (approved):

kartben:
Ya this has been bothering me as well :)

PR #2159, opened 2018-07-13
fix for Camel example publisher

darionct:
payload timestamped

Signed-off-by: Dario Novakovic <dario.novakovic@comtrade.com>


markoer (approved):

MMaiero:
I believe you should also update kura.build.properties file to reference the updated bundle.

MMaiero (approved):

PR #2180, opened 2018-08-10
Simple fix to hide server version from response headers.

MMaiero:


MMaiero:
Closing, since we have now the jetty customizer. It's better to implement it there.

PR #2206, opened 2018-08-29
Fixed OPCUA driver authentication

### LINKED ISSUE #2205: [OPC-UA Driver] Authentication not working
Currently certificate based authentication does not work in OPC UA driver because certificates are never loaded from keystore.

nicolatimeus:
Closes #2205 


kura/org.eclipse.kura.driver.opcua.provider/src/main/java/org/eclipse/kura/internal/driver/opcua/auth/ExternalKeystoreCertificateManager.java
```diff
@@ -0,0 +1,91 @@
...
+import javax.net.ssl.X509TrustManager;
+
+public class ExternalKeystoreCertificateManager extends CertificateManager {
+
+    private final String keyStorePath;
+    private final String keyStoreType;
+    private final char[] keyStorePassword;
+    private final String clientAlias;
+
+    private boolean loaded;
```
amitjoy: For memory visibility, we should use `volatile`

MMaiero (approved):

PR #2238, opened 2018-09-18
Fix Kura for Intel Up2 board with CentOS

pierantoniomerlino:
This PR tries to fix CentOS profile for Up2 board.
The modifications are the following:

- updated TinyB library 
- updated Selinux profile to add permissions for bluetooth
- changed network interfaces configuration
- modified the installer to avoid errors during installation


ctron:
I gave it a quick test run using docker, and that seems to work fine. The GLIBC issue is gone as is the segmentation fault.

I will continue testing with a real Intel UP2 once my local build and the installation is complete.

MMaiero:
Let's merge it since it seems to work in your and our environment and let's open other issues if something is still broken.

PR #2246, opened 2018-09-19
Provide new container build

ctron:
---
Please note: This isn't completely tested yet.


dwoodard1:
@ctron Thanks for putting this together! I have a few questions/comments:

1. I am not sure that we should be using individual names/emails for the maintainer tag. It appears [Docker has deprecated the maintainer tag](https://docs.docker.com/engine/reference/builder/#cmd) in favor of labels anyway. Maybe we could use labels with the emails of the committers, or use the Kura mail list email. Thoughts?
2. It appears you are adding several camel packages to the image. Don't these need proper CQ approval? My concern is we would have a standard Kura release that is compatible with the EPL and this separate Docker release that is a non EPL compatible superset of features. We had a similar mess some time ago when we were changing the Kura web UI. 
3. If we are hosting this in the Eclipse Github repository, I think we should also use an Eclipse Kura Dockerhub repository. The project handbook defines this very well [here](https://www.eclipse.org/projects/handbook/#resources-dockerhub). We can either ask to be part of the Eclipse repository or create our own with the eclipse-kura convention. Of course, this would require the docker image to be compatible with EPL (see point 2).

Let me know your thoughts.

ctron:
1. Yes, I know the tag is deprecated, but it also is still in widespread use (e.g. https://hub.docker.com/r/prom/prometheus/~/dockerfile/). I haven't seen a benefit yet to remove it. But I agree, the kura mailing list might be a more appropriate e-mail to put in.
2. Camel packages are not added to the mail container image, only to one extended image. I had a chat with Wayne about that, and the essence of that was: you don't need a CQ, if this is distributed from e.g. dockerhub (not from eclipse download servers) and everyone else could easy re-produce the same container image. Simply think of the fact that we are adding all kinds of things to the container image, like `gzip`, `tar`, `openssl`, … I am also not sure if it makes sense to distribute a pre-built version of the camel image. I would use it as a showcase *how* you can do it, but only publish the main Kura image. In any way, I wouldn't really rely on a pre-built container image for production anyway. So that is all just a convenient way to test something.
3. Yes, I think it makes sense to have Kura distributed under the "eclipse/kura" repository. That works fine for most Eclipse projects, and Kura isn't that big. So that may be enough. I am not sure where you read that requirement about the EPL regarding DockerHub, I couldn't find it in the project handbook. Maybe I overlooked it. But the Eclipse organization itself hosts a version of MySQL, https://hub.docker.com/r/eclipse/mysql/ So I think this goes back to the results of 2., it is fine, as long as other people can (technically and legally) do it the same way.

MMaiero:
What is our plan for Kura container build?
Is it legal to build those docker containers in Eclipse's Hudson?

Furthermore, I would prefer, since we have two possible images, to have one plain image that matches the default distribution (in this case the one for the UP squared board) and another that could be enhanced with Camel plugins and/or Apache File Install or other.

MMaiero:
I also think that we should have the container build executed in Travis, just to prevent possible errors.

MMaiero:
Any thoughts on the comments made?

ctron:
I think it makes sense to keep the Felix FileInstall, as it provides a real benefit if you want to work with Kura. I have no experience with building Docker on Travis or Jenkins. So I can't help you there.

MMaiero:
I have tried to involve Wayne into this discussion but it seems a long question that requires time. And in any case some CQ seem to be preferred.

At this point, to try to speed up things and stay on the safe side, I would ask you if you could create a contribution that simply wraps Kura in a Docker container, without further changes. This will be merged as soon as possible.

Once Eclipse defines those aspects, we can merge this PR.

ctron:
Then I am not really sure how I can help anymore. As I I did have a conversations with Wayne, and other projects act the same way.

However if you prefer, then we can of course close this PR for the moment. And I would simple suggest that you create a container build of what you had in mind. We can always resurrect this one.

MMaiero:
I've made few changes to have the Dockerfile build and the resulting image able to run in my environment. If them are ok, we can merge.

PR #2255, opened 2018-09-28
Added the interfaceName as argument in monitor listeners; fixed NPE

pierantoniomerlino:
This PR adds the interface name as argument to monitor listener.
Moreover a potential NPE is fixed in to the NetworkServiceImpl.

### DIFF
--- a/kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/net/modem/ModemMonitorListener.java
+++ b/kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/net/modem/ModemMonitorListener.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2011, 2016 Eurotech and/or its affiliates
+ * Copyright (c) 2011, 2018 Eurotech and/or its affiliates
  *
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
@@ -16,5 +16,5 @@ import org.osgi.annotation.versioning.ConsumerType;
 @ConsumerType
 public interface ModemMonitorListener {
 
-    public void setCellularSignalLevel(int signalLevel);
+    public void setCellularSignalLevel(String interfaceName, int signalLevel);
 }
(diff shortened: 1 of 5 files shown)


kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/net/modem/ModemMonitorListener.java
```diff
@@ -16,5 +16,5 @@
 @ConsumerType
 public interface ModemMonitorListener {
 
-    public void setCellularSignalLevel(int signalLevel);
+    public void setCellularSignalLevel(String interfaceName, int signalLevel);
```
MMaiero, line 22: I believe this requires an `@since 2.0` due to the fact that is a breaking change

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/net/wifi/WifiClientMonitorListener.java
```diff
@@ -16,5 +16,5 @@
 @ConsumerType
 public interface WifiClientMonitorListener {
 
-    public void setWifiSignalLevel(int signalLevel);
+    public void setWifiSignalLevel(String interfaceName, int signalLevel);
```
MMaiero, line 22: same here

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/net/modem/ModemMonitorListener.java
pierantoniomerlino: 👍

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/net/wifi/WifiClientMonitorListener.java
pierantoniomerlino: 👍

MMaiero (approved):

PR #2258, opened 2018-10-02
Fixed since annotation in BT APIs

pierantoniomerlino:
This PR fixes some wrong since annotations in BT APIs.


MMaiero:
According to this https://stackoverflow.com/questions/52621961/nofileassignedexception-exception-when-generating-karaf-feature-with-maven it seems an issue related to a maven plugin that is incompatible with Kara features.
Merging then, since the rest of the builds succeed.

MMaiero (approved):

PR #2261, opened 2018-10-04
Removed pid example in validation message.

MMaiero:


MMaiero:
A possible fix to close #2262

markoer (approved):
This solution will deprive users of service pid naming hints in all other cases where those would be OK, but it does solve this particular issue.

PR #2265, opened 2018-10-05
Fix PDP auth with CID 1 in 4G.

### LINKED ISSUE #2264: [Cellular] PDP Auth failed using context #1 in 4G network (Telit LE910-EU1 and LE910-V2 EU, possibly NA versions)
Please describe the issue you are encountering in Kura.

Fill the following form and provide a textual description if you are reporting a bug. 
Provide only a textual description otherwise.

- Component affected:
  - [ ] Distribution
    - Installer name (e.g. kura_3.1.0_raspberry-pi-2-3_installer.deb): All
    - Platform (e.g. Raspberry pi 2): All
    - Kura version affected (e.g. Kura 3.1.0): 3.2.0 and earlier releases
    - Java version (paste the result of **java -version**: NA
    - Host OS (e.g. Debian):  NA
    - Host OS version NA

  - [ ] Development environment
    - Version: 
    - Operating system (Mac OS, Windows, Linux):
    - OS Version:
    - IDE Version:
    - Host System Java Version:
      
  - [ ] Oomph installer
  - [ ] Add-ons
    - Name:
    - Version:
    - URL:
  - [ ] Eclipse Kura Marketplace
  - [ ] Kura Website
  - [ ] Kura Documentation

  - If needed to replicate the issue, please provide references for the browser used:
    - Name:
    - Version:
     
- Textual description of the issue:

With  Kura 3.2.0 and earlier versions, network registration fails using context #1 in a 4G network.
Found in the following products:
* Telit LE910-EU1 fw 20.00.413
* Telit LE910-EU V2 fw 20.00.402

If the 4G network is **not** available, the setup generated by Kura works fine. In this case the modem can automatically update to 4G if the network becomes available after the initial authentication in 2G or 3G.

However, by default, the above Telit modems are configured to always prefer a 4G network if this is available and in this case the Kura setup is such to cause a network registration reject (User Authentication failed) by the network.

The following manual configuration has proven to work reliably with an TIM M2M SIM and PAP authentication, in a 4G, 3G and 2G network (the mobile operator usually provides all three access technologies).

```
AT+CREG=2 # For debug purposes: enable network registration URC
AT#SIMDET=1 # Needed after a power cycle if Kura has not already done this
AT+COPS=2 # Deregister from network
AT+CGDCONT? # List PDP contexts
AT+CGDCONT=<CID> # Delete all CIDs but CID 1
AT+CGDCONT=1,"IP",<APN> # Set APN for CID 1
AT#PDPAUTH=1,1,< username>,<password> # Set credentials for PDP auth for CID 1
AT+COPS=0 # Register to the network. You should see URCs in the terminal
```
CID 1 (APN and auth) and operator selection are saved to the NVM so this would be only needed once, when the configuration changes. Network de-registration and registration takes some time and are asynchronous to the AT command. It does not make sense to do this in the pppd chat script.

Telit technical support suggests to avoid setting again CID 1 in the pppd chat script.
Remove the line

```
OK      AT+CGDCONT=<CID>,"IP",<APN>
```

Also make sure to specify CID 1 in the dial string

```
ATD*99***1#
```

Note that specifying the PDP authentication with the `AT#PDPAUTH` command is mandatory to register to a 4G network and in this case the PDP authentication credentials in `/etc/ppp/peer/pppX` are redundant (indeed they could be removed). However they are still necessary for a 3G or 2G network where `AT#PDPAUTH` alone does not allow to establish the PPP link.

## Notes
Telit modems can be forced using network access technologies using the `AT+WS46` command. For example, for the LE910-EU1 (which is 4G/2G modem)

```
AT+WS46=30 # 4G preferred with fallback to 2G
AT+WS46=12 # 2G only
AT+WS46=28 # 4G only
```
**Note** that switching between selection using the above command may cause the serial port used by pppd to stop responding to AT commands or return to command mode (even using `ATH` or `AT+CHUP` from the other serial port). This issue has been observed to also happen in other circumstances that need to be further investigated with Telit. Luckily Kura will reset the modem if this occurs.

The AT command `AT#REBOOT` can be used to reboot the modem (at least during system level testing). It is faster than the reset by pin but maybe less reliable.

cdealti:
This is an attempt to fix #2264.

It is not perfect and relies on the modem monitor to recover the modem
if something goes wrong.
It applies to Telit products LE910-EU1 and LE910-V2 EU/NA.
PDP context 1 is rarely used and there should not be any
performance impact for the other contexts.
See also comment in TelitLe910v2.java.
* AT+CGDCONT removed from chat script
* PDP context and auth set only if config changes
* Improved tip in UI for LTE modems
Configuring context 1 it's likely to require a modem reset before
the changes take effect.
NOTE: changes to PDP auth type and credentials only take effect
if the APN is changed too.

### DIFF
--- a/kura/org.eclipse.kura.net.admin/src/main/java/org/eclipse/kura/net/admin/modem/telit/de910/TelitDe910.java
+++ b/kura/org.eclipse.kura.net.admin/src/main/java/org/eclipse/kura/net/admin/modem/telit/de910/TelitDe910.java
@@ -396,7 +396,7 @@ public class TelitDe910 extends TelitModem implements EvdoCellularModem {
     }
 
     @Override
-    public boolean isSimCardReady() throws KuraException {
+    public boolean isTelitSimCardReady() throws KuraException {
         return true;
     }
 
--- a/kura/org.eclipse.kura.net.admin/src/main/java/org/eclipse/kura/net/admin/modem/telit/generic/TelitModem.java
+++ b/kura/org.eclipse.kura.net.admin/src/main/java/org/eclipse/kura/net/admin/modem/telit/generic/TelitModem.java
@@ -20,7 +20,6 @@ import org.eclipse.kura.KuraErrorCode;
 import org.eclipse.kura.KuraException;
 import org.eclipse.kura.comm.CommConnection;
 import org.eclipse.kura.comm.CommURI;
-import org.eclipse.kura.internal.board.BoardPowerState;
 import org.eclipse.kura.linux.net.modem.SupportedUsbModemInfo;
 import org.eclipse.kura.linux.net.modem.SupportedUsbModemsInfo;
 import org.eclipse.kura.linux.net.modem.UsbModemDriver;
@@ -453,7 +452,7 @@ public abstract class TelitModem {
     public String getMobileSubscriberIdentity() throws KuraException {
         synchronized (this.atLock) {
             if (this.imsi == null) {
-                if (isSimCardReady()) {
+                if (isTelitSimCardReady()) {
                     logger.debug("sendCommand getIMSI :: {}", TelitModemAtCommands.getIMSI.getCommand());
                     byte[] reply;
                     CommConnection commAtConnection = openSerialPort(getAtPort());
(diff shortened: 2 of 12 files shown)


pierantoniomerlino:
Successfully tested with Telit LE910 EU v2.
I deleted the context 2 before every tests. Works with Vodafone SIM (without authentication and 4G) and Arkessa SIM (with authentication but 2G).

cdealti:
@dwoodard1 this is ready to be merged if you think it's OK.

dwoodard1:
@cdealti This is ok for me.

dwoodard1 (approved):

PR #2281, opened 2018-10-15
Fixed issue where metatype names were used instead of ids.

### LINKED ISSUE #2272: Asset CSV upload/download issues if AD name and id differs
If a Driver provides a channel descriptor with ADs defining a `name` different than the `id`, the corresponding columns in the CSV downloaded for an Asset will be empty.

MMaiero:

Closes #2272


kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/server/servlet/ChannelServlet.java
```diff
@@ -94,11 +94,11 @@ private boolean fillCsvFields(CSVPrinter printer, String assetPid, String driver
         try {
             WIRE_ASSET_CHANNEL_DESCRIPTOR.getParameters().forEach(i -> {
                 try {
-                    printer.print(i.getName());
+                    printer.print(i.getId().substring(1));
```
amitjoy, line 97: Out of curiosity: Why substring? Can't we simply use the returned ID?

nicolatimeus (approved):

PR #2283, opened 2018-10-16
Set dependency as dynamic to prevent service disappear.

### LINKED ISSUE #2279: [Kura 4.0.0 QA] Service disappears from menu when reconfigured
ExamplePublisher disappears when it's reconfigured in a certain way:
* install ExamplePublisher
* add service
* service appears in the menu
* click on it and configure publisher and/or subscriber
* apply and confirm
* service disappears from the menu

It seems to be related to the selection of cloud connection(s). If I update some other parameter, it’s OK. When I change either one of those 2, it happens. Page refresh solves the issue that doesn’t reappear in the next save. The service also reappears if I reconfigure the same or other parameters again. And doesn’t disappear afterwards.
 
e.g. DB server doesn’t have this issue – no cloud connection to configure.

  - Kura version affected: Kura 4.0.0 RC b48
  - OS: Raspbian 9 on RPi3
  - Browser:
    - Name: Firefox
    - Version: 61.0

MMaiero:

Closes #2279

### DIFF
--- a/kura/examples/org.eclipse.kura.demo.heater/OSGI-INF/heater.xml
+++ b/kura/examples/org.eclipse.kura.demo.heater/OSGI-INF/heater.xml
@@ -30,10 +30,9 @@
    </service>
 
    <reference name="CloudPublisher"
-           policy="static"
+           policy="dynamic"
            bind="setCloudPublisher"
            unbind="unsetCloudPublisher"
-           policy-option="greedy"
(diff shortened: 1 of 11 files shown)


MMaiero:
@markoer Could you please have a look if this fixes your issue?

markoer (approved):
Looks OK - I'll try it in the new RC.

PR #2341, opened 2018-12-17
For issue 2240 and OCD localization and others

spnettec:


kura/org.eclipse.kura.broker.artemis.xml/OSGI-INF/metatype/org.eclipse.kura.broker.artemis.xml.BrokerInstance.xml
```diff
@@ -47,7 +47,7 @@
 			description="User configuration in the format: user=password|role1,role2,...|TextArea"
 			type="String"
 			cardinality="1"
-			required="true" />
+			required="false" />
```
MMaiero, line 50: Why the required is set now to false?

kura/org.eclipse.kura.core.cloud/src/main/java/org/eclipse/kura/core/cloud/CloudPayloadJsonDecoder.java
```diff
@@ -88,9 +88,15 @@ public static KuraPayload buildFromByteArray(byte[] array) {
     }
 
     private static void decodeTimestamp(KuraPayload payload, JsonValue timestampValue) {
-        if (timestampValue != null && timestampValue.isNumber()) {
-            long timestamp = timestampValue.asLong();
-            payload.setTimestamp(new Date(timestamp));
+        if (timestampValue != null) {
```
MMaiero, line 91: Please apply the formatter and cleanup profiles to all your code changes to keep consistency.
You'll find them here: https://github.com/eclipse/kura/tree/develop/kura/setups/formatting

kura/org.eclipse.kura.protocol.can/pom.xml
```diff
@@ -74,7 +54,25 @@
 					</filesets>
 				</configuration>
 			</plugin>
-			
+			<plugin>
+                <groupId>org.apache.maven.plugins</groupId>
```
MMaiero, line 58: Apply formatting also here. Why do you need this shift?

kura/org.eclipse.kura.wire.provider/OSGI-INF/metatype/org.eclipse.kura.wire.graph.WireGraphService.xml
```diff
@@ -15,7 +15,7 @@
         <AD id="WireGraph"
             name="WireGraph"
             type="String"
-            cardinality="0"
+            cardinality="65535"
```
MMaiero, line 18: Why this change? You'll get an Array that is now what we originally intended.

target-platform/config/kura.target-platform.build.properties
```diff
@@ -26,6 +26,8 @@ com.h2database.h2.version=1.4.192
 slf4j.api.version=1.7.25
 jcl.over.slf4j.version=1.7.25
 log4j.version=2.8.2
+log4j.log4j.version=1.2.17
```
MMaiero, line 29: To add those dependencies we need CQs. I've not checked but I believe we don't have them for those bundles.
Is there a specific reason to have them?

target-platform/p2-repo-test-deps/pom.xml
```diff
@@ -85,6 +85,16 @@
 									<artifactId>objenesis</artifactId>
 									<version>${objenesis.version}</version>
 								</artifactItem>
+                                <artifactItem>
```
MMaiero, line 88: Again, apply xml formatting to match the existing content

MMaiero (commented):
I would ask you to split this PR into multiple ones and to open issues to relate the PRs to them. Otherwise it's a little bit difficult to understand the changes and why you need them.
Other comments are inline in the contribution

kura/org.eclipse.kura.broker.artemis.xml/OSGI-INF/metatype/org.eclipse.kura.broker.artemis.xml.BrokerInstance.xml
spnettec: Since the ComponentUtil.getObjectClassDefinition has been fixed.The OCD validation can be excited correctly.  If required="true"  the validation would be failed at first running.

kura/org.eclipse.kura.wire.provider/OSGI-INF/metatype/org.eclipse.kura.wire.graph.WireGraphService.xml
spnettec: This is also  due to the the failure of OCD validation.This value must be greater than 1 to pass the OCD validation. I don't know why. I search some documentation, in this case must to set a large positive value. You can boot kura and test this issue so i believe you can fix it.

kura/org.eclipse.kura.core.cloud/src/main/java/org/eclipse/kura/core/cloud/CloudPayloadJsonDecoder.java
spnettec: Ok. This is my problem.

target-platform/config/kura.target-platform.build.properties
spnettec: When i import kura source into eclipse. Some Testing projects have compile errors in eclipse due to lack this bundles. And i can't compile in eclipse. Only maven.

target-platform/p2-repo-test-deps/pom.xml
spnettec: Got it

kura/org.eclipse.kura.protocol.can/pom.xml
spnettec: When i finish maven compile or install the lib would be deleted. So if shift this the plugins execute clean fist then copy dependencies and the lib dictionary still exist.

spnettec:
Hi @MMaiero If all these changes can be passed, I will split it into several PRs and format the file.
Best regards.

kura/org.eclipse.kura.broker.artemis.xml/OSGI-INF/metatype/org.eclipse.kura.broker.artemis.xml.BrokerInstance.xml
spnettec: I think we can set a default users to avoid first running OCD validation

spnettec:
Are you sure to apply this format and cleanup ? 
[https://github.com/eclipse/kura/tree/develop/kura/setups/formatting](url)
 I have many differences when i apply the format and cleanup.
eg:

image

MMaiero:
Yes, that should be the formatter and cleanup profiles that has to be applied to every contribution

spnettec:
Issue 1: Add typed JSON PayloadJsonDecoder
Issue 2:Fix getObjectClassDefinition  always return null ocd
Issue 3:Add localization ocd features
Issue 4:Fix OCD validation error
Issue 5:Fix test lack log4j and slf4j-log4j12 bundle error in eclipse      ------   remove?
Issue 6: Fix can't delete snapshot xml factoryConpomnent instance

Is it ok?

nicolatimeus:
Hi @spnettec , I had a look at this, I agree with @MMaiero that since this PR modifies different parts of the code, it would be better if you split it into smaller PRs that we can review and merge individually. (e.g. a PR with the changes to JSON encoding and decoding and one with the changes to the ConfigurationService that are not mentioned in #2341).

I would start with the one containing the changes for adding support for TypedJson, that directly address #2341. 

About this topic, I've noticed that you implemented the decoding part only.
Are you willing to work on the encoding part as well? In this way we can add full support for TypedJson to Kura.

In my opinion, it would be better to separate the code that handles Simple JSON and Typed JSON. You could for example add a new pair of classes (e.g. `CloudPayloadTypedJsonDecoder`/`CloudPayloadTypedJsonEncoder`) that are specific for the new format, and add another configuration option to the **payload.encoding** property of default CloudEndpoint configuration (e.g, `Typed JSON`). I would prefer this approach for the following reasons:

* Adding another **payload.encoding** type will be necessary for the encoding part, because while it is possible to create a decoder that handles both simple and typed JSON like you did, it is not possible to create a single encoder that handles both types (you have to know in advance which format to use).

* The logic of the decoders for the two formats will be simpler and targeted to the specific protocol.

* If the code that handles Simple JSON is not modified, we can be quite confident that the implementation of Typed JSON does not introduce new regressions in existing code.

spnettec:
Hi @nicolatimeus 

I think typed json is superset of simple json. Typed json can contained simple json. According this article [Payload Typed Json](https://github.com/eclipse/kapua/wiki/K-Payload-JSON-Format#typed-json-encoding) sentOn, position and body are simple json. This is known type. Only metrics is typed json. It is unknown type. 


In my opinion only one decoder and one encoder to handle both simple json and typed json.

For the encoder if  metrics object type is TypedValue we can easily encode as typed json. This is decided by which KuraPayload and which KuraMessage. If we want send typed json we can set TypedValue object in metrics.

For TypedValue we can add a type "INTEGER64" to fit for mqtt payload.

nicolatimeus:
Hi,

For the decoding part I agree, you can create a single decoder that handles both formats, without any limitation.

For the encoding part, using TypedValues as metric value can certainly solve the problem, but that would introduce a convention that applications must be aware of.

I think we should try to enable existing applications to use SimpleJson, TypedJson or Protobuf without code modifications.

Ideally, an application producing KuraPayload and KuraMessages should not care about message encoding, leaving these details to the cloud stack. In this way a user can pick a specific encoding at configuration time without changing the application.

We recently refactored cloud related APIs trying to enable creating applications that work on different cloud platforms using different conventions/message encodings, you can find more information about this at [1]. Since then, we are trying to keep this "requirement" in mind when we make changes to the cloud components.

[1] https://eclipse.github.io/kura/cloud-api/1-overview.html

PR #2349, opened 2018-12-21
Added check for resolv.conf symlink

pierantoniomerlino:

Brief description of the PR. 
If the resolv.conf file is a symlink, Kura fails to write the DNS servers on it. This PR adds a check before writing on that file.


kura/org.eclipse.kura.linux.net/src/main/java/org/eclipse/kura/linux/net/dns/LinuxDns.java
```diff
@@ -367,6 +370,17 @@ private synchronized String getRealPath(File file) throws KuraException {
 
     private synchronized void writeDnsFile(Set<IPAddress> servers) {
         logger.debug("Writing DNS servers to file");
+        // Check if DNS_FILE_NAME is a symlink and it is broken or points to unmanaged location.
+        // In this case, delete it.
+        try {
```
MMaiero, line 383: Maybe move this block to a specific method with a meaningful name, in order to be able to remove the comment.

PR #2374, opened 2019-02-07
[QA] Updated test dependencies

markoer:
Updated dependency fetching. Also unified it with implementation where configuration is now simplified no longer requiring Ant plugin.
Expected end result is Jenkins build passing again.


MMaiero:
Approving as the drop wizard.metrics dependency was approved as a build and test dependency in CQ 14155

MMaiero (approved):

Issue #2372, opened 2019-02-06
Add Timestamp to Lifecycle Messages


cdealti:
**Is your feature request related to a problem? Please describe.**
The lifecycle messages, `KuraBirthPayload` and `KuraDisconnectPayload` include a metric representing the system uptime. This time interval is relative to a (device) timestamp which is **not** set in the `KuraPayload`.

**Describe the solution you'd like**
Set the timestamp in the lifecycle messages. 

**Describe alternatives you've considered**
None.

**Additional context**
None.

PR #2378, opened 2019-02-11
Add a timestamp in lifecycle messages

pierrepitiot:
Resolve issue #2372


MMaiero:
Would it be a possibility to have the timestamp in the LifecyclePayloadBuilder? In this way we could have a timestamp directly in the payload. Even if the result is probably the same

pierrepitiot:
In this case it would be in KuraBirthPayload build and KuraDisconnectPayload right?

MMaiero:
Let's stick with the current solution you designed.

pierantoniomerlino (approved):

PR #2370, opened 2019-02-05
Cleanup Kura folders

markoer:
Deletes most files that Kura brought to the system.

**Related Issue:** This PR fixes #362.

**Description of the solution adopted:** Removes most of the files. It does leave some behind - those that can still be used after (e.g. dhcpcd, wpa-supplicant, named, hostapd, ppp, etc. configuration).

Issue #362, opened 2016-07-22
Uninstaller leaves most of the files


ctron:
Uninstalling Kura 1.4.0 from a Raspberry PI 2 leaves most of the file on the file system:

The following files remain after a `dpkg --purge kura`

```
/opt/eclipse/kura_1.4.0_raspberry-pi-2/
/opt/eclipse/kura_1.4.0_raspberry-pi-2/plugins
/opt/eclipse/kura_1.4.0_raspberry-pi-2/plugins/slf4j.log4j12_1.6.0.jar
/opt/eclipse/kura_1.4.0_raspberry-pi-2/plugins/org.eclipse.equinox.http.servlet_1.1.300.v20120522-1841.jar
/opt/eclipse/kura_1.4.0_raspberry-pi-2/plugins/javax.servlet_3.0.0.v201112011016.jar
/opt/eclipse/kura_1.4.0_raspberry-pi-2/plugins/org.eclipse.equinox.common_3.6.100.v20120522-1841.jar
/opt/eclipse/kura_1.4.0_raspberry-pi-2/plugins/org.eclipse.jetty.http_8.1.3.v20120522.jar
/opt/eclipse/kura_1.4.0_raspberry-pi-2/plugins/osgi.cmpn_4.3.0.201111022214.jar
/opt/eclipse/kura_1.4.0_raspberry-pi-2/plugins/org.apache.felix.deploymentadmin_0.9.5.jar
/opt/eclipse/kura_1.4.0_raspberry-pi-2/plugins/org.eclipse.jetty.server_8.1.3.v20120522.jar
/opt/eclipse/kura_1.4.0_raspberry-pi-2/plugins/org.eclipse.core.runtime_3.8.0.v20120521-2346.jar
/opt/eclipse/kura_1.4.0_raspberry-pi-2/plugins/org.eclipse.equinox.metatype_1.2.0.v20120522-1841.jar
/opt/eclipse/kura_1.4.0_raspberry-pi-2/plugins/com.codeminders.hidapi_1.1.1.jar
/opt/eclipse/kura_1.4.0_raspberry-pi-2/plugins/org.apache.felix.gogo.command_0.8.0.v201108120515.jar
/opt/eclipse/kura_1.4.0_raspberry-pi-2/plugins/usb4java-javax_1.0.0.jar
/opt/eclipse/kura_1.4.0_raspberry-pi-2/plugins/org.eclipse.osgi_3.8.1.v20120830-144521.jar
[...]

PR #2387, opened 2019-02-15
CentOS/RPM uninstallation cleanup

markoer:
Cleanup of Kura's files during RPM uninstallation.

**Related Issue:** Continues work started in #2370, but for RPMs. Related also to #362.

**Description of the solution adopted:** Removes most of the files. It does leave some behind - those that can still be used after (e.g. dhcpcd, wpa-supplicant, named, hostapd, ppp, etc. configuration).
The approach differs from the one for Debian packages in that this one adds removal script directly to RPM.

### DIFF
--- a/kura/distrib/pom.xml
+++ b/kura/distrib/pom.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!--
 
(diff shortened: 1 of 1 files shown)


pierantoniomerlino (approved):
With this approach only Kura-related files will be removed during uninstallation. The network/firewall configs remain as Kura set them up.

PR #2400, opened 2019-03-21
kura.linux.snapctl: adding snap control provider

kubiko:
kura.linux.snapctl: adding snap control provider
    
Adding dns server service provider for snap confinement
It reflests constrains when Kura runs in snap confinement.
In snap confinement Kura runs in own sandbox, so usual writable paths
might not be accessible.
systemd is also not accessible directly, instead kura can control
own services (named) through snapctl interface

Signed-off-by: Ondrej Kubik <ondrej.kubik@canonical.com>

### DIFF
--- a/kura/distrib/config/kura.build.properties
+++ b/kura/distrib/config/kura.build.properties
@@ -45,6 +45,7 @@ org.eclipse.kura.linux.command.version=1.0.300
 org.eclipse.kura.linux.net.version=1.0.500-SNAPSHOT
 org.eclipse.kura.linux.sysv.provider.version=1.0.0
 org.eclipse.kura.linux.systemd.provider.version=1.0.0
+org.eclipse.kura.linux.snapctl.provider.version=1.0.100-SNAPSHOT
 org.eclipse.kura.linux.debian.provider.version=1.0.0
 org.eclipse.kura.linux.redhat.provider.version=1.0.0
 org.eclipse.kura.linux.gpio.version=1.0.400
--- a/kura/distrib/src/main/ant/build_equinox_distrib.xml
+++ b/kura/distrib/src/main/ant/build_equinox_distrib.xml
@@ -1383,7 +1383,14 @@ fi]]>
 		<zip destfile="${project.build.directory}/${build.output.name}.zip" update="true">
 			<zipfileset file="${project.build.directory}/plugins/org.eclipse.kura.linux.snapctl.provider_${org.eclipse.kura.linux.snapctl.provider.version}.jar"
                         prefix="${build.output.name}/${plugins.folder}" />
+			<zipfileset file="${project.build.directory}/plugins/org.eclipse.kura.linux.usb.*_${org.eclipse.kura.linux.usb.version}.jar"
+                        prefix="${build.output.name}/${plugins.folder}" />
 		</zip>
+		<condition property="is.tinyb.snapshot" value="">
+			<contains string="${tinyb.version}" substring="SNAPSHOT"/>
+		</condition>
+			<antcall target="tinyb-snapshot-jars-all" />
+			<antcall target="tinyb-no-snapshot-jars-all" />
 	</target>
 
 	<target name="tinyb-jars">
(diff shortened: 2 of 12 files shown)


kura/distrib/config/kura.build.properties
```diff
@@ -45,6 +45,7 @@ org.eclipse.kura.linux.command.version=1.0.300
 org.eclipse.kura.linux.net.version=1.0.500-SNAPSHOT
 org.eclipse.kura.linux.sysv.provider.version=1.0.0
 org.eclipse.kura.linux.systemd.provider.version=1.0.0
+org.eclipse.kura.linux.snapctl.provider.version=1.0.100-SNAPSHOT
```
MMaiero, line 48: Since it is a new bundle it would be 1.0.0-SNAPSHOT

kura/distrib/src/main/ant/build_equinox_distrib.xml
```diff
@@ -1315,6 +1326,17 @@ fi]]>
 		</propertyfile>
 	</target>
 
+	<target name="linux-snapctl-config" if="is.snapctl" depends="linux-config">
+		<propertyfile file="${project.build.directory}/${build.output.name}/config.ini">
+			<entry key="osgi.bundles" operation="+"
+					value=", reference:file:${kura.install.dir}/${kura.symlink}/${plugins.folder}/org.eclipse.kura.linux.snapctl.provider_${org.eclipse.kura.linux.snapctl.provider.version}.jar@4" />
+			<entry key="osgi.bundles" operation="+"
```
MMaiero, line 1333: Sorry, probably I've missed the point: the emulator bundles should be provided only in NN profiles, but in the is.snapctl you set that has to be not NN. Right?

kura/org.eclipse.kura.linux.snapctl.provider/OSGI-INF/dnsserver.xml
```diff
@@ -0,0 +1,21 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+
+    Copyright (c) 2019 Eurotech and/or its affiliates
```
MMaiero, line 4: > Copyright (c) 2019 Eurotech and/or its affiliates

this should be: Copyright (c) 2019 Canonical

kura/org.eclipse.kura.linux.snapctl.provider/build.properties
```diff
@@ -0,0 +1,20 @@
+#
+# Copyright (c) 2018 Eurotech and/or its affiliates
```
MMaiero, line 2: Same as per the other copyright. But please correct the date

kura/org.eclipse.kura.linux.snapctl.provider/pom.xml
```diff
@@ -0,0 +1,32 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+
+    Copyright (c) 2019 Eurotech and/or its affiliates
```
MMaiero, line 4: Same as before

kura/org.eclipse.kura.linux.snapctl.provider/src/main/java/org/eclipse/kura/internal/linux/snapctl/net/dns/DnsServerServiceImpl.java
```diff
@@ -0,0 +1,333 @@
+/*******************************************************************************
+ * Copyright (c) 2019 Eurotech and others
```
MMaiero, line 2: Same as before

kura/distrib/src/main/ant/build_equinox_distrib.xml
```diff
@@ -1391,6 +1431,28 @@ fi]]>
 	    </zip>
 	</target>
 
+    <target name="tinyb-snapshot-jars-all" if="is.tinyb.snapshot">
```
MMaiero, line 1434: Why this?

kura/distrib/src/main/ant/build_equinox_distrib.xml
kubiko: @MMaiero this is attempt to add only one additional target. Like supper set bundle, so it has emulator and real network providers, and correct one is chosen when snap is built by editing config file.
But I agree that this was created bundle is confusing.
@MMaiero up to you to guide me what's better approach, add one ubuntu-snap target, or add two targets `ubuntu-snap` `ubuntu-snap-nn`

MMaiero: If that is not a problem for you, I'll prefer to have two targets. In this way it keeps also the consistency with the other targets.

kubiko: @MMaiero no problem. Then next question is if you are OK to have folded into one target differentiations architectures? This is related to comment below about tinyb
So `ubuntu-snap/ubuntu-snap-nn` will have libraries for all architectures, this is mostly related to tinyb one.
Otherwise we will need following targets:
```
ubuntu-snap-armhf
ubuntu-snap-arm64
ubuntu-snap-amf64
ubuntu-snap-nn-armhf
ubuntu-snap-nn-arm64
ubuntu-snap-nn-amf64
```
You have to remember that snaps support all different architectures, I have skipped here 32b x86, as you do not seem to have any support for it

kura/distrib/src/main/ant/build_equinox_distrib.xml
kubiko: relates to my comment above, we need to include all the architectures as we want to be able to build snap for armhf/arm64/amd64 from single target. 
Unless you prefer to have separate target for each architecture. Choice is yours :)

MMaiero:
Old PR. Closing for now

PR #2402, opened 2019-03-21
Adding snap layout support

kubiko:
When kura is deployed as snap, it runs in own sandbox and some parts of the system and read only or not accessible at all. Therefore some configuration files and tmp files needs to be relocated when snap environment is detected

/etc/ and  /etc/network/ directories are read only so following files needs to be relocated:
 - /etc/resolv.conf.save 
 - /etc/hostapd-
 - /etc/wpa_supplicant-
 - /etc/hostapd.conf.tmp
 - /etc/wpa_supplicant.conf.tmp
 - /etc/network/interfaces.tmp

Signed-off-by: Ondrej Kubik ondrej.kubik@canonical.com

### DIFF
--- a/kura/org.eclipse.kura.linux.debian.provider/src/main/java/org/eclipse/kura/internal/linux/net/config/NetInterfaceConfigSerializationServiceImpl.java
+++ b/kura/org.eclipse.kura.linux.debian.provider/src/main/java/org/eclipse/kura/internal/linux/net/config/NetInterfaceConfigSerializationServiceImpl.java
@@ -45,7 +45,10 @@ public class NetInterfaceConfigSerializationServiceImpl implements NetInterfaceC
     private static final Logger logger = LoggerFactory.getLogger(NetInterfaceConfigSerializationServiceImpl.class);
 
     private static final String DEBIAN_NET_CONFIGURATION_FILE = "/etc/network/interfaces";
-    private static final String DEBIAN_TMP_NET_CONFIGURATION_FILE = "/etc/network/interfaces.tmp";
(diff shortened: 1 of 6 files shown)


MMaiero:
I was thinking that we may have a more general property that defines a path prefix, instead of relating that to the "snap" concept.
Also, do we need to have it as a System property passed in launch script or, maybe, it is better to have a property in Kura.properties file that is manageable using the Kura System service?

kubiko:
> I was thinking that we may have a more general property that defines a path prefix, instead of relating that to the "snap" concept.
> Also, do we need to have it as a System property passed in launch script or, maybe, it is better to have a property in Kura.properties file that is manageable using the Kura System service?

@MMaiero I'd use your guidance here, what is best approach so it is done in least invasive way. Having generic path prefix sounds like good option here.
Only possible complication I can think of, we have three paths
- read only package deployment, that's where content of snap lives: $SNAP/
- writable path, common to all revisions, usually good for configurations and settings, SNAP_COMMON. 
- writable path, per revision, usually used for temp data, or writable data which are revision exclusive.

what would be best naming for those three options?

While division between SNAP_COMMON, and SNAP_DATA is not mandatory, and only one can be used. But good use of it, will increase chances of success when upgrade fails and we need to rollback to old version. Then having good unmodified snapshot of writable data improves our chances of success. But you will probably know better than me what's best strategy here.
I use SNAP_DATA for the moment only for `logs`, `/run` directory and `/var/lib/dhclient`

MMaiero:
Old PR. Closing for now

PR #2403, opened 2019-03-21
Fixing file attributes for missing executable flag

kubiko:
Set correct executable file flag, so it is deployed correctly

Signed-off-by: Ondrej Kubik ondrej.kubik@canonical.com

### DIFF
--- a/kura/distrib/src/main/ant/build_equinox_distrib.xml
+++ b/kura/distrib/src/main/ant/build_equinox_distrib.xml
@@ -443,27 +443,27 @@ fi]]>
 
 			<zipfileset
 				file="${project.build.directory}/${build.output.name}/kura_install.sh"
-				prefix="${build.output.name}/${install.folder}/" />
+				prefix="${build.output.name}/${install.folder}/" filemode="777" />
 
 			<zipfileset file="src/main/resources/${build.name}/kuranet.conf"
 				prefix="${build.output.name}/${install.folder}/" />
 
 			<zipfileset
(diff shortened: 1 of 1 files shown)


MMaiero (approved):

kura/distrib/src/main/ant/build_equinox_distrib.xml
```diff
@@ -401,25 +401,25 @@ fi]]>
                         prefix="${build.output.name}/${user.config.folder}/" />
 
 			<zipfileset file="${project.build.directory}/${build.output.name}/kura_install.sh"
-                        prefix="${build.output.name}/${install.folder}/" />
+                        prefix="${build.output.name}/${install.folder}/" filemode="777" />
```
pierantoniomerlino, line 447: Maybe 755 or 700 is more appropriate. Could we move this chmod operation to the kura_install.sh?

kubiko: @pierantoniomerlino we can move it to `kura_install.sh`
motivation to change it here was that we have single location for change, rather than editing `kura_install.sh` for each target instead

pierantoniomerlino: Good point. So, leave it in `build_equinox_distrib.xml`, but please change the mode. :-)

MMaiero:
@kubiko Hi, any update on these PRs?
We really would like to merge them and finalise the Snap integration.

Thanks.
  Matteo

MMaiero:
@pierantoniomerlino Could you please check if that is ok for you?

PR #2447, opened 2019-05-15
Updated snapshot 0 with latest changes to kura codebase for cloud con…

MMaiero:
…nections.

This will prevent an empty list in cloud connections when restoring snapshot 0.


MMaiero:
@markoer Any thoughts? Can this be merged?
Thanks

markoer (changes requested):
I guess you'll have to merge with http service changes. It's not active and https doesn't work.

MMaiero:
Yes, this branch was developed before the https merging. In theory, since there are no conflicts the merge will be done automatically closing the pull request. 
In any case, I can rebase the branch, if needed, but the issue should be visible and proven to be fixed even without the https part.

markoer (approved):
True, merge should go through OK.

PR #2489, opened 2019-07-10
Fix variable replacement in topic string

### LINKED ISSUE #2448: Variable replacement in topic string does not work anymore in new Kura 4 Cloud Publisher
In a Wires graph, we use a javascript filter to decode the data received from lora sensors, and we add a string metric named "mote" that represents the name of the channel (the sensor) to add it at the end of the general topic:

the javascript filter =
var outRecord = newWireRecord()
outRecord.assetName = newStringValue(record.assetName.getValue())
**outRecord.mote = newStringValue(prop)**

and in the publisher
Application.topic = Motes/**$mote**

in Kura 3 the result is:
topic :#account-name/#client-id/Lora/Motes/**Adeunis**

but in Kura 4 with the new publisher structure the result is:
topic :#account-name/#client-id/Lora/Motes/**$mote**

The $mote variable is not anymore replaced by the real value. It should be noted that only the $assetName variable is correctly replaced.

pierrepitiot:
In Wires publisher, rather than limiting the variable replacement to only "assetName", all the variables from a record should be available.

**Related Issue:** This PR fixes/closes #2448


kura/org.eclipse.kura.wire.component.provider/src/main/java/org/eclipse/kura/internal/wire/publisher/CloudPublisher.java
```diff
@@ -295,8 +296,10 @@ private void publish(final List<WireRecord> wireRecords) {
         Map<String, TypedValue<?>> wireRecordProps = wireRecord.getProperties();
 
         final Map<String, Object> properties = new HashMap<>();
-        if (wireRecordProps.containsKey(ASSET_NAME_PROPERTY_KEY)) {
-            properties.put(ASSET_NAME_PROPERTY_KEY, wireRecordProps.get(ASSET_NAME_PROPERTY_KEY).getValue());
+        List<String> l = new ArrayList<>(wireRecordProps.keySet());
+        for(String s:l) {
```
MMaiero: please apply formatting

MMaiero (commented):
Could you have a look if it is possible to add unit/integration tests to prove that the substitution works?

pierrepitiot:
@MMaiero I am afraid I don't know how to do that ...

MMaiero:
Have a look here: https://github.com/eclipse/kura/tree/develop/kura/test/org.eclipse.kura.wire.component.provider.test

pierrepitiot:
@MMaiero , sorry but I don't find how I can test the topic in this test. the CloudPublisher that is taking care of that is a mock in this test and I didn't find how to replace it with a real CloudPublisher...

MMaiero:
Can't you just do a unit test verifying that the substitution is performed correctly by your code?

MMaiero (commented):
@pierrepitiot Why didn't you add the test to the wire.component.provider.test project?

pierrepitiot:
Because I made a lot of different tries and it was simpler to use a new project but I can merge it in wire.component.provider.test if you think it's better

MMaiero:
Yes please, it would match with what we already have

PR #2510, opened 2019-07-26
Backoff jitter

pasqualenaclerio:
**Description of the solution adopted:** Added an exponential backoff with jitter

### DIFF
--- a/kura/org.eclipse.kura.core/OSGI-INF/metatype/org.eclipse.kura.data.DataService.xml
+++ b/kura/org.eclipse.kura.core/OSGI-INF/metatype/org.eclipse.kura.data.DataService.xml
@@ -27,14 +27,14 @@
             default="false"
             description="Enable automatic connect of the Data Publishers on startup and after a disconnection."/>
                         
-        <AD id="connect.retry-interval"
-            name="Connect Retry-interval"
+        <AD id="backoff.max.tentatives"
+            name="Backoff max tentatives"
             type="Integer"
             cardinality="0" 
             required="true"
-            default="60"
+            default="10"
             min="1"
-            description="Frequency in seconds to retry a connection of the Data Publishers after a disconnect (Minimum value 1)."/>
+            description="Maximum number of tentatives used to calculate backoff jitter. (Minimum 1)"/>
             
         <AD id="enable.recovery.on.connection.failure"
             name="Enable Recovery On Connection Failure"
--- a/kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/data/DataServiceImpl.java
+++ b/kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/data/DataServiceImpl.java
@@ -562,6 +562,7 @@ public class DataServiceImpl implements DataService, DataTransportListener, Conf
     }
 
     private boolean startConnectionMonitorTask() {
+
         if (this.connectionMonitorFuture != null && !this.connectionMonitorFuture.isDone()) {
             logger.error("Reconnect task already running");
             throw new IllegalStateException("Reconnect task already running");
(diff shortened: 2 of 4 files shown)


pierantoniomerlino:
For the backoff jitter implementation, please take a look [here](https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/) and [here](https://docs.aws.amazon.com/general/latest/gr/api-retries.html).

kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/data/DataServiceImpl.java
```diff
@@ -579,57 +581,71 @@ private boolean startConnectionMonitorTask() {
...
-                                    .getAndIncrement() < DataServiceImpl.this.dataServiceOptions
-                                            .getRecoveryMaximumAllowedFailures()) {
-                                logger.info("Checkin done.");
-                                DataServiceImpl.this.watchdogService.checkin(DataServiceImpl.this);
+                    int attempts = 1;
+                    int backoffMaxTentatives = DataServiceImpl.this.dataServiceOptions.getBackoffMaxTentatives();
+
+                    while (!DataServiceImpl.this.dataTransportService.isConnected() || autoConnect) {
+
+                        int reconnectDelay = (new Random().nextInt((int) Math.pow(2, attempts)) + 1) * 1000;
```
pierantoniomerlino, line 595: According to [this document](https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/), the reconnect delay should be implemented as
`sleep = random_between(0, min(cap, base * 2 ** attempt))`
So, it is assumed that the base is 1 (this is fine), but the cap is missing.

kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/data/DataServiceOptions.java
```diff
@@ -155,6 +155,6 @@ int getRecoveryMaximumAllowedFailures() {
     }
 
     int getCriticalComponentTimeout() {
-        return getConnectDelay() * CONNECT_CRITICAL_COMPONENT_TIMEOUT_MULTIPLIER;
+        return getBackoffMaxTentatives() * CONNECT_CRITICAL_COMPONENT_TIMEOUT_MULTIPLIER;
```
pierantoniomerlino, line 158: The timeout for the critical components was computed based on the connection delay. Now that the connection delay is replaced with the maximum number of attempts, a new formula should be used.

kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/data/DataServiceOptions.java
```diff
@@ -40,7 +40,7 @@
     private static final String RECOVERY_MAX_FAILURES_PROP_NAME = "connection.recovery.max.failures";
 
     private static final boolean AUTOCONNECT_PROP_DEFAULT = false;
-    private static final int CONNECT_DELAY_DEFAULT = 60;
+    private static final int BACKOFF_MAX_TENTATIVES_DEFAULT = 60;
```
pierantoniomerlino, line 43: The default value is 10, according to the metatype.

kura/org.eclipse.kura.core/OSGI-INF/metatype/org.eclipse.kura.data.DataService.xml
```diff
@@ -27,14 +27,14 @@
             default="false"
             description="Enable automatic connect of the Data Publishers on startup and after a disconnection."/>
                         
-        <AD id="connect.retry-interval"
-            name="Connect Retry-interval"
+        <AD id="backoff.max.tentatives"
+            name="Backoff max tentatives"
             type="Integer"
```
pierantoniomerlino, line 32: Please change the word "tentative" with "attempt" or "try". Tentative means temporary, uncertain.

PR #2534, opened 2019-08-26
feat(api)!: Added equals and hashcode methods in Password API.

MMaiero:
`Password` did not override `equals` and `hashCode`, so two instances holding the same value were never equal. Configuration code that compares property maps with `Objects.deepEquals` (for example `ConfigurationServiceImpl` when applying a snapshot) always saw `Password` properties as changed and issued unnecessary configuration updates, and every options class wrapping a password had to unwrap the `char[]` to compare it.

This PR implements value-based `equals` and `hashCode` on `Password` using `Arrays.equals` and `Arrays.hashCode`, which also handle a null password value consistently. A `PasswordTest` is added to the `AllCoreTests` suite of `org.eclipse.kura.core.test`, whose JaCoCo report is the one the API bundle publishes to Sonar, so the new code is covered by the quality gate.


mattdibi:
Since we're now working on a *major* release and we can perform breaking changes can we merge this? @MMaiero 

3 days ago it was this PR's 5th birthday... we need a birthday present 😄

MMaiero:
Not building. It needs to be rechecked

salvatore-coppola:
@mattdibi it's time to merge this, I don't want to assist to the 6th birthday

mattdibi:
@salvatore-coppola build is failing, cannot be merged as-is

salvatore-coppola:
@mattdibi THE TIME HAS COME !

salvatore-coppola:
Cleaned a little bit and updated the description

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/configuration/Password.java
```diff
@@ -42,4 +44,24 @@ public char[] getPassword() {
...
         return new String(this.passwordVal);
     }
+
+    @Override
+    public int hashCode() {
+        return Arrays.hashCode(this.passwordVal);
+    }
+
+    @Override
+    public boolean equals(Object obj) {
```
marcellorinaldo, line 50:
```java
  char[] x1 = new char[] {'a','b','c'};
  char[] x2 = new char[] {'a','b','c'};
  Password p1 = new Password(x1);
  Password p2 = new Password(x2);
  
  System.out.println(p1.equals(p2)); // true
  x1[1] = 'x';
  System.out.println(p1.equals(p2)); // false
```

Some other changes are needed. The two equals should return both true.

marcellorinaldo (changes requested):

marcellorinaldo (changes requested):
Moreover, some problems on `null` values (also prior this PR):

```java
Password p1 = new Password(null);
// produces error: reference to Password is ambiguous
```

```java
Password p1 = new Password((String) null);
p1.toString(); // throws NPE
```

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/configuration/Password.java
salvatore-coppola: solved

salvatore-coppola:
> Moreover, some problems on `null` values (also prior this PR):
> 
> ```java
> Password p1 = new Password(null);
> // produces error: reference to Password is ambiguous
> ```
> 
> ```java
> Password p1 = new Password((String) null);
> p1.toString(); // throws NPE
> ```

Decide to maintain the tolerance for the null value for the password argument for backward compatibility.

marcellorinaldo (approved):

marcellorinaldo:
> > Moreover, some problems on `null` values (also prior this PR):
> > ```java
> > Password p1 = new Password(null);
> > // produces error: reference to Password is ambiguous
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > ```java
> > Password p1 = new Password((String) null);
> > p1.toString(); // throws NPE
> > ```
> 
> Decide to maintain the tolerance for the null value for the password argument for backward compatibility.

This commit is suspicious https://github.com/eclipse-kura/kura/commit/3ed5a1ef7c15ab1324da3186783c5b64be614ab2

salvatore-coppola:
We will merge in order to solve an old bug in a dependent component. Later, we will evaluate whether to remove the ability to allow null arguments.

Issue #2544, opened 2019-09-17
Asset is removed from Drivers view when it is deleted in WireGraph


pintify:
**Describe the bug**
Given an Asset, and a WireGraph which uses it in the flow, when the asset is removed from the graph, then the asset is also removed from the Drivers & Assets view. It also happens when the whole WireGraph is removed.

**Expected behavior**
The asset remains in Drivers & Assets view and still available as a possible asset to include in the graph.

MMaiero:
Hi,
this was an original choice to link the lifetime of an asset added to the graph to the graph itself.
For what I remember the decision was meant to keep the overall code simple and matching the behaviour of all the other wire components.

Best regards,
Matteo

pintify:
I can see how that makes sense from programming point of view, but as user there is a non-reversible step when the asset is first included in the WireGraph:

Step 1: asset created from Drivers & Assets view
Status 1: asset existent but not present in WireGraph
Step 2: asset added to WireGraph
Status 2: asset both existent and present in WireGraph

Logic says that the status 1 is available after doing a reverse operation of step 2 such as removing the asset from WireGraph, but there is no way to get there unless you remove and recreate the asset.

spnettec:
> Hi,
> this was an original choice to link the lifetime of an asset added to the graph to the graph itself.
> For what I remember the decision was meant to keep the overall code simple and matching the behaviour of all the other wire components.
> 
> Best regards,
> Matteo

Hi @MMaiero, The behaviour is wrong. When remove graph the Drivers & Assets can not be removed. It is huge bug for long time. I have fixed this bug for myself. 
[patch.txt](https://github.com/eclipse/kura/files/3666520/patch.txt)

MMaiero:
Hi @spnettec 
thank you for your feedback.
I would suggest to open a pull request. Unfortunately we cannot merge directly a patch file, due to Eclipse Foundation rules and the fact that it is correct that you get the credits for your changes.

We will be very happy to review your PR.

Best regards,
Matteo

markoer:
Related to #1981.

MMaiero:
Closing as this should be fixed now

PR #2549, opened 2019-09-30
Asset is removed from Drivers view when it is deleted in WireGraph

spnettec:
Asset is removed from Drivers view when it is deleted in WireGraph

Describe the bug
Given an Asset, and a WireGraph which uses it in the flow, when the asset is removed from the graph, then the asset is also removed from the Drivers & Assets view. It also happens when the whole WireGraph is removed.

Expected behavior
The asset remains in Drivers & Assets view and still available as a possible asset to include in the graph

see #2544

### DIFF
--- a/kura/org.eclipse.kura.web2/www
+++ b/kura/org.eclipse.kura.web2/www
@@ -1,1 +1,1 @@
-src/main/webapp/
+src/main/webapp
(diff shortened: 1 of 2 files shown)


MMaiero:
Please have a look at this failure: https://accounts.eclipse.org/legal/eca/validation/15251

spnettec:
> Please have a look at this failure: https://accounts.eclipse.org/legal/eca/validation/15251

How to remove this problem?

PR #2550, opened 2019-10-01
Asset is removed from Drivers view when it is deleted in WireGraph

spnettec:

Describe the bug
Given an Asset, and a WireGraph which uses it in the flow, when the asset is removed from the graph, then the asset is also removed from the Drivers & Assets view. It also happens when the whole WireGraph is removed.

Expected behavior
The asset remains in Drivers & Assets view and still available as a possible asset to include in the graph.

see issue #2544


MMaiero:
Please apply the official Kura cleanup and formatting profiles to your changes.

MMaiero (commented):
I believe that the cleanup profile is not applied.
Furthermore, you need to change the version in kura/distrib/config/kura.build.properties to have the updated bundle included in the final distribution.

Finally, I tested that after the changes, the generation of an asset from the wires page does not work anymore. Even if the component is there in the graph, it is not created and accessing again the wires page displays the newly created asset in red, as the corresponding instance does not exist.
This behaviour works in develop.

spnettec:
I‘m sure that i have fixed this bug long time ago in my develop version and it work fine. I don't test this PR. Maybe some other class should fixed because i have changed too many file. I'll build and test soon. Sorry about this PR.

spnettec:
@MMaiero 
Can you try my image to see if it is the correct behavior?
docker run -d -p 1883:1883 -p 1888:1888 -p 8080:80 -p 443:443 -p 8088:8088 --name yofciot heyoulin/kura:v4.2.9

MMaiero:
No, the problem is there as well, for what I can see

spnettec:
@MMaiero Could you send me snapshot.xml file? I can't understand what you see.

spnettec:
By the way i don't think it is good idea that creating asset from graph view.

MMaiero:
Why do you say so? I find it useful to be able to create and update drivers and assets from the wire tab: I'm able to define the whole graph without having to switch around the UI.

spnettec:
> Why do you say so? I find it useful to be able to create and update drivers and assets from the wire tab: I'm able to define the whole graph without having to switch around the UI.

Assets are generally pre-defined. Then use it in the graph. Otherwise, this PR should not be submitted. Assets are created in the Assets view and should not be deleted in the graph. Therefore  it should not be created in the graphics. This reduces the complexity and the logic is clearer. This is why most users are confused about this behavior.

MMaiero:
Obviously I'm reporting my experience. And the times I've done demos or I've talked with customers, they are focused in creating the graph first and they start mocking it in the canvas. They they need an asset, and disliked the fact that they had to save the graph, move to the drivers and assets section to create and configure the asset and then move back to the wires canvas. That is the reason why it was implemented in that way.

I appreciate your PR and would like to get some feedback from the community to understand which way we need to follow.

spnettec:
> Obviously I'm reporting my experience. And the times I've done demos or I've talked with customers, they are focused in creating the graph first and they start mocking it in the canvas. They they need an asset, and disliked the fact that they had to save the graph, move to the drivers and assets section to create and configure the asset and then move back to the wires canvas. That is the reason why it was implemented in that way.
> 
> I appreciate your PR and would like to get some feedback from the community to understand which way we need to follow.

Hi @MMaiero, The latest modified version is suitable for both cases. Can you review this change?

MMaiero (approved):

spnettec:
> So, now the behaviour in wires canvas seems correct. But, from what I see, if I remove the asset from wires, then it is removed also in drivers and assets that is like the current behaviour. Or I'm missing something?


The problem is that when you create an asset in the AssetView in advance, then use it in the WireView, and then want to delete it, finally the pre-created one is also deleted.

PR #2558, opened 2019-10-23
udhcpc pid file parameter fix

kandrej:
The command that starts the **udhcpc** dhcp client is missing the -p parameter before the pid file. The pid file is not created


MMaiero:
Hi @kandrej could you please check your ECA and signature? Otherwise we cannot merge your PR.

kura/org.eclipse.kura.linux.net/src/main/java/org/eclipse/kura/linux/net/dhcp/DhcpClientManager.java
```diff
@@ -159,6 +159,7 @@ private static String formCommand(String interfaceName, boolean useLeasesFile, b
             sb.append(interfaceName);
             sb.append(' ');
             if (usePidFile) {
+            	sb.append("-p ");
```
MMaiero: May I ask you to apply the Kura formatter here? The line added seems misaligned compared to the others. Thanks

MMaiero (approved):

PR #2567, opened 2019-10-30
fix #2565

### LINKED ISSUE #2565: Error appears after changing console username/password
**Describe the bug**
Error message appears after changing console username and password 
error message is `404 ProxyServlet: /admin/denali/xsrf` 

**To Reproduce**
1. Go to WebConsole
2. Change values for username and password
3. Click on Apply button
4. Click "Yes" to confirm changing

**Expected behavior**
Username and password should be changed without error appearing 

**Screenshots**
console_password

 - Board: Raspberry Pi 3 (also able to reproduce on RG10-20)
 - Kura version: 4.2-SNAPSHOT 

**Additional context**
- username and password are changed 
- there is no error in kura.log
- unable to reproduce on Kura 4.1

spnettec:

**Related Issue:** #2565

### DIFF
--- a/kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/Console.java
+++ b/kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/Console.java
@@ -378,7 +378,8 @@ public class Console implements ConfigurableComponent, org.eclipse.kura.web.api.
                 chain(baseHandler, sessionAuthHandler).sendErrorOnFailure(401));
 
         // exception on admin console path, redirect to login page on failure instead of sending 401 status
-        routingHandler.addRouteHandler(consoleOptions.getAppRoot()::equals,
+        // why consoleOptions.getAppRoot()::equals does't work?
+        routingHandler.addRouteHandler(p->consoleOptions.getAppRoot().equals(p),
                 defaultHandler.redirectOnFailure(AUTH_PATH));
 
         return new HttpContextImpl(routingHandler, defaultContext);
--- a/kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/Console.java
+++ b/kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/Console.java
@@ -222,7 +222,7 @@ public class Console implements ConfigurableComponent, org.eclipse.kura.web.api.
         }
 
         ConsoleOptions options = new ConsoleOptions(properties);
-        String oldAppRoot = consoleOptions.getAppRoot();
+        String oldAppRoot = this.appRoot;
         Console.setConsoleOptions(options);
 
         try {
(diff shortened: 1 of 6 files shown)


mstankovic:
@MMaiero @spnettec I checked this PR and i was unable to recreate #2565
i also checked few more scenarios with logging and changing username/password and everything works fine.

spnettec:
@nicolatimeus  I think this bug has been fixed, please review it.

kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/Console.java
```diff
@@ -372,80 +391,82 @@ private HttpContext initResourceContext(final HttpContext defaultContext) {
...
-        this.httpService.registerServlet(AUTH_RESOURCE_PATH, new SendStatusServlet(404), null, resourceContext);
-        this.httpService.registerServlet(CONSOLE_RESOURCE_PATH, new SendStatusServlet(404), null, resourceContext);
+        this.httpService.registerServlet("/", new RedirectServlet("/"::equals, () -> this.appRoot), null,
+                this.resourceContext);
+        this.httpService.registerServlet(AUTH_RESOURCE_PATH, new SendStatusServlet(404), null, this.resourceContext);
+        this.httpService.registerServlet(CONSOLE_RESOURCE_PATH, new SendStatusServlet(404), null, this.resourceContext);
 
         this.httpService.registerServlet(PASSWORD_AUTH_PATH,
-                new GwtPasswordAuthenticationServiceImpl(this.authMgr, CONSOLE_PATH), null, sessionContext);
+                new GwtPasswordAuthenticationServiceImpl(this.authMgr, () -> this.appRoot), null, this.sessionContext);
```
nicolatimeus, line 410: The `appRoot` parameter was originally meant to allow having a 3rd party web ui available on the root. It is not related the console path. Maybe we can improve the parameter description to avoid misunderstanding. If you want to enable the customisation of the console path (which can be desirable) please introduce another configuration parameter.

nicolatimeus (changes requested):
Hi @spnettec, the changes look good to me, just please address the comment below. Also please check why Eclipse CLA check is failing.

kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/Console.java
spnettec: I don't understand what means this appRoot is. If change the appRoot this code
`
this.httpService.registerServlet("/", new RedirectServlet("/"::equals, this.appRoot), null, resourceContext);
`
will redirect console to the changed appRoot. That's why i thought appRoot is console path.

By the way i don't know why Eclipse CLA check is aways failing.

spnettec: Should i change to this?
`this.httpService.registerServlet("/", new RedirectServlet("/"::equals, CONSOLE_PATH), null, resourceContext);`

If not next time the we can't enter console by root path '/'.

nicolatimeus: I will try to explain better the use case for `appRoot`:

Suppose you want to add another web application to the gateway that is completely different from Kura web ui, for example a web application developed by a customer that manages only its application, and the customer wants it to be available on the root path `/`.

You can achieve this in the following way:

1. Create a custom bundle that serves the application on `/mycompany/myapp`.
2. Set the `appRoot` parameter to `/mycompany/myapp`.

In this way the custom application will be available on the root path, Kura console will be still available on `/admin/console` which is currently hardcoded.

In other words `appRoot` is simply the relative path to which the user will be redirected when accessing the root path, which is not necessarily the Kura console path.

spnettec: Thanks. I see.  You can review the change of #2568 about this and i'll close this PR and open other PR later because of Eclipse CLA failing

spnettec:
Close this PR because of Eclipse CLA failing

PR #2568, opened 2019-11-02
Kura4.3

spnettec:
This PR completed the web localization work and fixed many bugs.

**Related Issue:** 

Uncertain

- Some Non OSGI package generated by p2-maven-plugin. 
- Snapshot xml read by stream and can read large xml file .
- Can run above java 8. eg: java11 .

  Kura API have changed

Signed-off-by: heyoulin heyoulin@gmail.com


MMaiero:
Hello @spnettec, thank you for this PR!
It shows that you are really committed to the project and did a lot of work in improving it.

I would suggest to split this PR into multiple ones, in order to have self contained PRs that can be easier to be evaluated, but also analysed and discussed by the community. Each PR should be associated to an issue that describes the problem/enhancement/feature.

As is, the PR is really difficult to understand and evaluate, considering the number of commits and changes applied.
Eventually, we can discuss about this during the next weekly meeting.

Best regards,
Matteo

PR #2569, opened 2019-11-04

spnettec:
see #2565


MMaiero:
@nicolatimeus Can you provide some feedback on this?

kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/Console.java
```diff
@@ -386,7 +386,7 @@ private synchronized void initHTTPService() throws NamespaceException, ServletEx
         this.httpService.registerServlet(CONSOLE_RESOURCE_PATH, new SendStatusServlet(404), null, resourceContext);
 
         this.httpService.registerServlet(PASSWORD_AUTH_PATH,
-                new GwtPasswordAuthenticationServiceImpl(this.authMgr, CONSOLE_PATH), null, sessionContext);
```
nicolatimeus: I think this should be left to CONSOLE_PATH, for the reasons we discussed in the past.

MMaiero:
@spnettec Using your change, the error due to the reload of the web environment seems to be solved. The real issue that I see after this change is that the user is not logged out: the session is not invalidated and you still can perform changes to the web UI. In other environments I've seen, usually, after a password change you are logged out and asked to login again.

PR #2610, opened 2019-12-20

### LINKED ISSUE #2565: Error appears after changing console username/password

spnettec:
Fix #2565

### DIFF
--- a/kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/Console.java
+++ b/kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/Console.java
@@ -160,6 +160,9 @@ public class Console implements ConfigurableComponent, org.eclipse.kura.web.api.
         this.eventAdmin = null;
     }
 
+    private ConsoleOptions preOptions;
+    private Boolean invalidateSession = false;
+
     // ----------------------------------------------------------------
     //
     // Activation APIs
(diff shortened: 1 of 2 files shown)


MMaiero:
Thank you for the change. But now, if I'm not wrong, a change in the session timeout will not be reflected immediately, right? As well as for the app root.

MMaiero:
Other than that, please apply to the updated code the Kura formatter and cleanup profiles, to match the code style.

spnettec:
> Thank you for the change. But now, if I'm not wrong, a change in the session timeout will not be reflected immediately, right? As well as for the app root.

Yes. I think session timeout and app root is not necessary reflected immediately. If you want i can do that.

spnettec:
I just fixed the cleanup and format. Why the travis-ci failed?

MMaiero:
It reports an error with one of the tests. I have tried to restart the build.

MMaiero:
The original issue should be fixed now in develop branch. Closing for now.

PR #2619, opened 2020-01-09
Fixed timer issues on system time changes

### LINKED ISSUE #2605: Wires Timer stop ticking if system time is set in the past
**Describe the bug**
Changing the system time one our in the past causes all wires timers to stop ticking, the problem is not solved by setting the date back in the future. The only way for restoring behaviour as of now is updating Wire Graph configuration

**To Reproduce**
1. Create a Timer and connect it to a Logger.
2. Note current time
3. Set date one hour in the past

**Expected behavior**
The time continues to tick regardless of system time shifts

nicolatimeus:

* Use a `SingleThreadScheduledExecutor` per Timer instead of the Quartz scheduler for `SIMPLE` mode, this should make `SIMPLE` timers not affected by system time changes.
* Implemented rescheduling of Quartz triggers for `CRON` mode on ClockEvent
* Replaced Timer unit tests with integration tests
* Added utility bundle for wire component testing

For `CRON` Timers, the issue is fixed only for system time changes caused by Kura ClockService, system time changes caused by other components or processes will still cause the issue.

**Related Issue:**
Closes #2605


kura/test/org.eclipse.kura.wire.component.provider.test/OSGI-INF/TimerTest.xml
```diff
@@ -0,0 +1,25 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- Copyright (c) 2019 Eurotech and/or its affiliates and others All rights
```
MMaiero: This header is a little bit messed up

kura/test/org.eclipse.kura.util.wire.test/META-INF/MANIFEST.MF
```diff
@@ -0,0 +1,24 @@
+Manifest-Version: 1.0
+Bundle-ManifestVersion: 2
+Bundle-Name: Wire Component test utilities
+Bundle-SymbolicName: org.eclipse.kura.util.wire.test;singleton:=true
+Bundle-Version: 1.0.0.qualifier
```
MMaiero: I believe we have the bundle version matching the kura version for tests

MMaiero (commented):
Since the contribution is made in 2020, maybe we can set that in all the headers.

MMaiero (approved):

Issue #2617, opened 2020-01-09
Pid and ExitCode simplification


pierantoniomerlino:
**Is your feature request related to a problem? Please describe.**
The `Pid` and `ExitCode` objects used by the CommandExecutorService are interfaces that simply wrap the pid of a running process and the exit code of a command. The `LinuxPid` and `LinuxExitCode` classes are concrete implementations of these interfaces specific for Linux. However, they essentially wrap a Integer. Since in most OS both pids and exit codes are integers numbers, the interfaces should be simplified or replaced with classes.

**Describe the solution you'd like**
Refactor the service to use integers instead of `Pid` and `ExitCode` interfaces.

PR #2623, opened 2020-01-14
Enhance CommandExecutor service

### LINKED ISSUE #2615: CommandExecutorService tests improvement
The tests implemented for the new `CommandExecutorService` check the method functionality running simple command on the underlying OS. For example the `ls -all` command is run and the return code is checked.

This approach is problematic when the tests are run on a OS other than Linux (i.e. Mac OS X). Moreover, the `UnprivilegedExecutorService` tests are based on the existence of a specific Linux user (i.e. kura), but this cannot be automatically created in automatic build environment.

The test should be modified to be OS independent.

**Describe alternatives you've considered**
The kura user is created in the `BeforeClass` section of the tests and it works on Travis, but it not so reliable.

**Additional context**

pierantoniomerlino:
This PR improves the CommandExecutor service removing the ExitStatus class, simplifying the Pid class and enhancing the tests for this service. 

**Related Issue:** This PR fixes/closes #2615 and #2617 

**Description of the solution adopted:** Since the tests for the CommandExecutor service are OS dependent, this PR tries to solve this limitation. The number of tests and the code coverage is indeed reduced.
In order to simplify the service, the ExitStatus class is replaced by integer number. Also the Pid class is simplified, since the type returned by the `getPid`method is now a integer.


kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/internal/linux/executor/DefaultExecutorFactory.java
```diff
@@ -0,0 +1,14 @@
+package org.eclipse.kura.core.internal.linux.executor;
```
MMaiero, line 12: Copyright

kura/org.eclipse.kura.api/META-INF/MANIFEST.MF
```diff
@@ -67,6 +67,7 @@ Import-Package: javax.comm;version="1.2.0",
  javax.microedition.io;resolution:=optional,
  javax.net.ssl,
  javax.usb;version="1.0.2",
+ org.apache.commons.exec;version="1.3.0",
```
MMaiero, line 70: Let's check if it is possible to remove this dependency in APIs

kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/linux/executor/privileged/PrivilegedExecutorServiceImpl.java
```diff
@@ -33,20 +33,33 @@
 
     private static final Logger logger = LoggerFactory.getLogger(PrivilegedExecutorServiceImpl.class);
     private static final LinuxSignal DEFAULT_SIGNAL = LinuxSignal.SIGTERM;
+    private ExecutorUtil executorUtil;
 
     @SuppressWarnings("unused")
     private ComponentContext ctx;
 
+    public PrivilegedExecutorServiceImpl() {
```
MMaiero: Why this constructor?

kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/linux/executor/unprivileged/UnprivilegedExecutorServiceImpl.java
```diff
@@ -38,20 +38,33 @@
     @SuppressWarnings("unused")
     private ComponentContext ctx;
     private UnprivilegedExecutorServiceOptions options;
+    private ExecutorUtil executorUtil;
+
+    public UnprivilegedExecutorServiceImpl() {
```
MMaiero: same here

pierantoniomerlino:
We decide to split the solutions to issues #2615 and #2617 and close this PR.

Issue #2013, opened 2018-04-03
Unsaved changes dialog triggers incorrectly


ctron:
While testing Kura 3.2 RC I noticed a few times that the "there are unsaved changes" dialog triggers incorrectly. Making a change, pressing "Apply" applies the changes, the "apply" button even gets disabled. Yet the Web UI asks to discard changes when moving away from the service.

MMaiero:
Could you please provide a replicable set of actions in order to identify the issue?

ctron:
I thought that would be clear by:

> Making a change, pressing "Apply" applies the changes, the "apply" button even gets disabled. Yet the Web UI asks to discard changes when moving away from the service.

As this is an issue which sporadically appears (at least I haven't found a pattern yet) I cannot give you a set of instructions to replicate this with with 100% certainty.

tafitzgerald:
I have seen this occur before as well and it happens irregularly. This is due to the field being marked as "dirty" when it is touched. This seems to be a minor bug in the system and will be corrected. Thanks for your input on this @ctron.

MMaiero:
@tafitzgerald Any progress on this topic?

tafitzgerald:
Should be finished. I'm waiting to test before creating PR.

tafitzgerald:
Tested and submitted PR #2631

Issue #1960, opened 2018-03-12
[Web] Unable to identify component type after creation


ctron:
After creating a new component with the Kura Web UI, the component type is no longer visible and the user has to figure out the component type based on the structure of the service's attributes. Of course this will only work if the services do have different attributes.

IMHO the service type/component type should be visible in the UI alongside the service/component ID.

nicolatimeus:
The component type should be visible in the heading of the configuration section
<img width="493" alt="screen shot 2018-03-15 at 09 19 46" src="https://user-images.githubusercontent.com/8960773/37451548-263af3f8-2832-11e8-8f11-9615c79377ec.png">

ctron:
You are just looking at the Wires UI I assume:

![image](https://user-images.githubusercontent.com/202474/37509605-43d492b8-28f8-11e8-8c67-966fd3138d82.png)

I agree that (if the service has a description) with Kura 3.2 this will be a bit improved. Still you have to deduct it from the description. And if that is empty you are back to the original problem.

tafitzgerald:
Wires components are labeled differently than the services panel. I am working different options to display the component type on the main ui panel for services, but real estate can be tricky to display both component type and name.

PR #2631, opened 2020-01-21
Change the way we set component fields "dirty" so as to not generate dialog incorrectly

tafitzgerald:

Currently, component text fields are being set as "dirty" on loss of focus and key events. This incorrectly causes a dialog to display asking "There are unsaved changes. Do you want to discard and continue?" when no change has been made to the field.

**Related Issue:** This PR fixes/closes issue #2013 

**Description of the solution adopted:** This solution now only displays the dialogue when a change has been made to the field's value after loss of focus. No arbitrary key event or loss of field focus without change will trigger the dialog.

NOTE: 
* This does not apply to password fields.
* The value change handler only registers on modified value, even when the field has been edited but reverted back to original value.

**Screenshots:** Below is a screenshot of dialog in reference:

image


MMaiero (approved):

MMaiero:
@tafitzgerald I had to rebase the branch. Please have a look

tafitzgerald:
@MMaiero I see, my PR for issue #1960 seemed to be up to date.

MMaiero:
Hi, could you check why now all the pre-merge checks fail?

PR #2641, opened 2020-01-24
Wire components ordered alphabeticall, indenting like components

### LINKED ISSUE #1958: [Wires] Use alphabetical ordering for Wires components
IMHO the current ordering of Wire components is not really intuitive. I guess using an alphabetical ordering based on the label would make sense.

tafitzgerald:

The ordering of wires components in the UI right panel can seem confusing to the user.

**Related Issue:** This PR fixes/closes #1958 

**Description of the solution adopted:** Wires components are being ordered alphabetically first and then by number of inputs versus outputs in the wire component right UI panel. Furthermore, components in the same family are here being indented to show similar components to the user. 

**Screenshots:** 
2020-01-24-132320_1824x984_scrot

**Any side note on the changes made:** There are no major side effects to this PR, but review is requested to make sure that this is beneficial to the community.


MMaiero:
Hi @tafitzgerald,
the alphabetical order works for me.
I'm not sure about the grouping and shifting. It does not seem to be so obvious. 
As a user, I would be a little bit confused about this shifting since it is not clear or explained anywhere.

In the original post, I believe, we were supposing to group components in three categories: producers, consumers and the component that act as both.
The idea of the foldable category was to be able to fold or unfold the specific category to reduce the visible number of components and make it easier to interact with the wire graph.
Something like the collapse option shown here: https://gwtbootstrap3.github.io/gwtbootstrap3-demo/#collapse

MMaiero:
Same issue here with commits. Please have a look.

PR #2644, opened 2020-01-28
fix apply changes dialog opening when it shouldn't

tafitzgerald:
Signed-off-by: thomas.fitzgerald thomas.fitzgerald@eurotech.com

Currently, component text fields are being set as "dirty" on loss of focus and key events. This incorrectly causes a dialog to display asking "There are unsaved changes. Do you want to discard and continue?" when no change has been made to the field.

Related Issue: This PR fixes/closes issue #2013

Description of the solution adopted: This solution now only displays the dialogue when a change has been made to the field's value after loss of focus. No arbitrary key event or loss of field focus without change will trigger the dialog.

NOTE:

This does not apply to password fields.
The value change handler only registers on modified value, even when the field has been edited but reverted back to original value.
Screenshots: Below is a screenshot of dialog in reference:

2020-01-21-111136_1824x984_scrot


tafitzgerald:
I created a new branch in my fork and rebased to latest eclipse/kura to fix the ECA issue for the last PR.

MMaiero (approved):

PR #2650, opened 2020-01-30
Applied checkstyle to net.admin bundle

MMaiero:


kura/org.eclipse.kura.net.admin/src/main/java/org/eclipse/kura/net/admin/modem/hspa/HspaModem.java
```diff
@@ -616,19 +616,14 @@ public ModemRegistrationStatus getRegistrationStatus() throws KuraException {
                 String[] regStatusSplit = sRegStatus.split(",");
                 if (regStatusSplit.length >= 2) {
                     int status = Integer.parseInt(regStatusSplit[1]);
-                    switch (status) {
-                    case 0:
+                    if (status == 0) {
```
pierantoniomerlino, line 619: Why you replaced the switch statement with if-else?

pierantoniomerlino (approved):

PR #2660, opened 2020-01-31
Order wire components alphabetically

### LINKED ISSUE #1958: [Wires] Use alphabetical ordering for Wires components

tafitzgerald:

The ordering of wires components in the UI right panel can seem confusing to the user.

**Related Issue:** This PR fixes/closes #1958 

**Description of the solution adopted:** Wires components are being ordered alphabetically and grouped by type of component in their own collapsible panels under the components panel.

The latest screen looks like below:
2020-02-13-131426_1738x905_scrot

**Any side note on the changes made:** Modified the panel headers to match the page stylesheet and appear uniform. Using consumer/producer since that is the terminology used by wire components.


MMaiero:
Hi @tafitzgerald 
that is really great!

Few comments: 
- It's not really intuitive that those are collapsible elements. Maybe a visual hint is needed. Or at, least, the categories should be visually different from the section title ("Wire Components")
- Having no scroll means that the whole wire configuration section at the bottom of the page is moved more downwards. If you have a lot of components there (maybe as a result of new components added from the marketplace) it means that, if you click on a wire component you have to scroll a lot to be able to change the corresponding configuration. It seems a little bit inconvenient.

What do you think?

tafitzgerald:
@MMaiero I can see how the absence of the scroll for the right component panel could cause inconvenience. I will change the components panel to be a scroll panel itself and test it. I didn't like the aesthetics of setting a static size for the collapsible panels so I will try to leave that out as well. 

Do you think adding a tooltip for each collapsible panel and a different color scheme for the panel header would suffice to let the user know that each collapses/expands?

MMaiero:
I believe that something visible like the grouping done in the "Accordion Example using PanelCollapse (whole header)" example can be helpful.

tafitzgerald:
@MMaiero please have a look at the updated screenshots and code. I added icon to indicate to user that the panels collapse/expand and set the components panel to be a scroll panel with defined height.

tafitzgerald:
@MMaiero I tried using buttons for each dropdown wire component panel but that did not seem as intuitive to me that clicking on the button would drop the panel down. I decided to go with uniform sub-panels for each component type in the wire components panel. Emitter/receiver might not be the most intuitive terminology to use here, maybe input/output would be better.

MMaiero:
Hi @tafitzgerald 
a minor bug found: when you access the page the "emitters" section is expanded but the symbol in the left is set like the section is folded.

tafitzgerald:
> Hi @tafitzgerald
> a minor bug found: when you access the page the "emitters" section is expanded but the symbol in the left is set like the section is folded.

@MMaiero thanks, I missed that. I am also thinking it makes more sense to label the panels as Producers and Consumers since the components underneath are labeled as such. I made that change as well.

MMaiero (approved):

MMaiero:
Hi @tafitzgerald 
I've made some minor changes.

PR #2662, opened 2020-01-31
Reporting system uptime in MacOS from kern.boottime

### LINKED ISSUE #1932: SystemAdminService.getUptime() returns SystemAdminService#UNKNOWN on MacOS when locale is not English
Since the code to figure out the uptime on Mac uses string pattern matching to parse the result of uptime (which contains "days", "hours", and things like that), it doesn't work when the local is not 'en'.

The implementation should be replaced to use the kernel boottime as a reference and compute the uptime from there (see `sysctl kern.boottime`)

tafitzgerald:

The current way we read system uptime on MacOS reports "UNKNOWN" when locality is not set to en.

**Related Issue:** This PR fixes/closes #1932 

**Description of the solution adopted:** This proposed change computes the time since the system has booted by reading from the kernel boottime (sysctl kern.boottime). This file returns a string containing seconds and microseconds of the time the system was booted (since the epoch). This solution parses that output and retrieves the second portion, getting the difference from the current system time. 

**Any side note on the changes made:** This solution should work regardless of locality.

### DIFF
--- a/kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/system/SystemAdminServiceImpl.java
+++ b/kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/system/SystemAdminServiceImpl.java
@@ -15,7 +15,6 @@ import java.io.BufferedReader;
 import java.io.File;
 import java.io.FileReader;
 import java.io.IOException;
-import java.io.InputStreamReader;
 import java.text.DateFormat;
 import java.text.SimpleDateFormat;
 import java.util.Date;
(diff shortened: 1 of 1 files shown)


kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/system/SystemAdminServiceImpl.java
```diff
@@ -128,42 +129,33 @@ public String getUptime() {
...
-                        // Only showing mins
-                        mins = Integer.parseInt(uptimePart.split("\\s*mins")[0]);
-                    } else {
-                        logger.error("uptime could not be parsed correctly: {}", uptimeParts[0]);
-                    }
-
-                    uptime = (long) ((days * 24 + hours) * 60 + mins) * 60;
-                    uptimeStr = Long.toString(uptime * 1000);
-                }
+            	p = Runtime.getRuntime().exec("sysctl -n kern.boottime");
```
pierantoniomerlino: Why do you use the `Runtime.exec` here? The class is already bound to the `CommandExecutorService` that was developed to run command on the system. I suggest to replace it with the `runSystemCommand` method that internally uses the `CommandExecutorService` and return a string representing the output of the command.

tafitzgerald: I see your point. Changed to use CommandExecutorService to run system command.

kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/system/SystemAdminServiceImpl.java
```diff
@@ -128,42 +128,20 @@ public String getUptime() {
...
-                        logger.error("uptime could not be parsed correctly: {}", uptimeParts[0]);
-                    }
-
-                    uptime = (long) ((days * 24 + hours) * 60 + mins) * 60;
-                    uptimeStr = Long.toString(uptime * 1000);
-                }
+        	try {
+        		String lastBootupSysCmd = runSystemCommand("sysctl -n kern.boottime", false, this.executorService);
+            	
+            	if (!lastBootupSysCmd.equals("")) {
```
pierantoniomerlino: You could use `!lastBootupSysCmd.isEmpty()` instead of `!lastBootupSysCmd.equals("")`.

pierantoniomerlino (approved):

MMaiero:
@tafitzgerald Can we make the change suggested by @pierantoniomerlino? Then, I believe, we can proceed with merging

kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/system/SystemAdminServiceImpl.java
```diff
@@ -128,42 +128,20 @@ public String getUptime() {
...
-                        logger.error("uptime could not be parsed correctly: {}", uptimeParts[0]);
-                    }
-
-                    uptime = (long) ((days * 24 + hours) * 60 + mins) * 60;
-                    uptimeStr = Long.toString(uptime * 1000);
-                }
+        	try {
+        		String lastBootupSysCmd = runSystemCommand("sysctl -n kern.boottime", false, this.executorService);
+            	
+            	if (!lastBootupSysCmd.isEmpty()) {
```
tafitzgerald: made changes per @pierantoniomerlino suggestion

pierantoniomerlino (approved):

PR #2667, opened 2020-02-04
Fix opcua array to string (#2658)

### LINKED ISSUE #2658: OPC-UA DRIVER - Return a String Array
---
name: OPC-UA DRIVER - Return a String Array
about: I found a simple solution to return a StringArray from OPC UA.

---

My project need to process a StringArray from a OPC UA Server, I found a way to read it.

On the DataTypeMapper class I insert the following code the map function.
image
` if (value instanceof String[]) {` 
 `                    String[] stringArray = (String[]) value;` 
 `                    return TypedValues.newStringValue(String.join(",", stringArray));` 
 `               } else {` 
`                     return TypedValues.newStringValue(value.toString());` 
`                 }`

**Additional context**
I have we this demo server to test:  opc.tcp://uademo.prosysopc.com:53530/OPCUA/SimulationServer

The original value reads using Prosys OPC UA Browser.
image

I have used this asset configuration to test.

image

The result was the values separated with comma as it is represented on the value field of OPC UA browser.

image

It results the same value on the log:
image

alebianchin:
Added a check to convert from arrays to String
**Related Issue:** This PR fixes/closes #2658

### DIFF
--- a/kura/org.eclipse.kura.driver.opcua.provider/src/main/java/org/eclipse/kura/internal/driver/opcua/DataTypeMapper.java
+++ b/kura/org.eclipse.kura.driver.opcua.provider/src/main/java/org/eclipse/kura/internal/driver/opcua/DataTypeMapper.java
@@ -11,6 +11,8 @@
  */
 package org.eclipse.kura.internal.driver.opcua;
 
+import java.util.Arrays;
+
 import org.eclipse.kura.type.DataType;
 import org.eclipse.kura.type.TypedValue;
 import org.eclipse.kura.type.TypedValues;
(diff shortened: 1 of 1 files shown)


nicolatimeus (approved):

nicolatimeus:
Please update copyright header year.

leandrofabior:
I put the changed code on my driver version and builded it.
The result was the following, I saw that your version put [ ] around data.

image

This [] around is the same representation that OPC Browser uses. Then this is correct and the PR is ok.

image

leandrofabior:
Thanks for the help.

PR #2675, opened 2020-02-12
Updated KuraFormatter.xml

pierantoniomerlino:
Updated KuraFormatter.xml file according to the new checkstyle rules.


MMaiero (commented):
Not sure why it added so many new options to the file

MMaiero (approved):

PR #2818, opened 2020-04-28
Added attempts management in modem monitor

pierantoniomerlino:
With this PR, the modem is reset only after 3 attempts to apply the configuration.

**Related Issue:** N/A

**Description of the solution adopted:** Currently, when the modem monitor catches an exception the modem is suddenly reset. With this PR, the monitor tries to apply the configuration 3 times before resetting the modem.


MMaiero:
Can we have this in develop also?

MMaiero:
Let's also open an issue to show how the corresponding issue can be triggered.

MMaiero (approved):

PR #2872, opened 2020-05-13
Pr/watchdog metatype

### LINKED ISSUE #2854: Change WatchdogService refresh interval parameter description
The current description in the metatype for the watchdog service refresh interval parameter is the following:

`Watchdog refresh interval in ms. The value can be set between 1 and 60 seconds.`

This might cause the user to assume that this parameter changes the hardware watchdog refresh interval. If the user sets it by mistake to a value above the hw watchdog expected refresh interval, the gateway will reset as soon as the WatchdogService starts.

We should improve the description to warn about this, another possibility can be using a property in `kura.properties` for this parameter, that is hw/OS dependent, to avoid exposing it to the user.

iacco:
Brief description of the PR: improved WatchdogService's metatype and corrected typo.

**Related Issue:** This PR fixes/closes #2854


kura/org.eclipse.kura.linux.watchdog/OSGI-INF/metatype/org.eclipse.kura.watchdog.WatchdogService.xml
```diff
@@ -34,7 +34,7 @@
             required="true"
             default="10000"
             max="60000"
-            description="Watchdog refresh interval in ms. The value can be set between 1 and 60 seconds."/>
+            description="WatchdogService\'s refresh interval in ms of the Watchdog device. The value can be set between 1 and 60 seconds and should not be set to a value more or equal to the Watchdog device\'s timeout value"/>
```
MMaiero: Maybe better saying "Watchdog device refresh interval in ms."?

iacco: I am unsure about this: IMHO it is not sufficiently clear that the WatchdogService is only resetting the watchdog device's timeout, rather than setting its timeout in the first place.

Another solution I was thinking about would be to leave this to 10000 by default and remove it from metatype's attribute definition.

Let me know your thoughts.

iacco: So, after a bit of digging around, and gathering information, here's some thoughts:

- In the following statements I will mostly refer to hardware watchdogs, rather than software watchdogs which interface with them; to be more specific, I will refer to their Kernel driver implementations;

- It is probably not a good idea to "hardcode" the timeout, since according to the Kernel Documentation, while a lot of watchdog devices and their drivers have default timeouts of 30 or 60 seconds, there are also watchdog devices with default timeouts lower or equal to 10 seconds: even though most of the platforms on top Kura will run will not have this default, it would be unwise to remove it [1]

- On lots (most?) of the devices the watchdog timeout can be modified at userspace level according to the needs: as painful as it looks to imagine watchdogs with really low timeouts, it might be a possibility, also during the framework "life": it would be unwise to hardcode this property, since changes in the timeout will require an update of the bundle;

- Based on the points above however, it is IMHO useful to state clearly that this WatchdogServiceImpl will reset the watchdog timeout, can disable it (where supported) with the Magic Character, but cannot set the refresh rate of a watchdog device.

[1](https://www.kernel.org/doc/html/latest/watchdog/watchdog-parameters.html)

MMaiero: > Based on the points above however, it is IMHO useful to state clearly that this WatchdogServiceImpl will reset the watchdog timeout, can disable it (where supported) with the Magic Character, but cannot set the refresh rate of a watchdog device.

Agreed

MMaiero (approved):

PR #2882, opened 2020-05-19
Enhancement: added Areas to S7 PLC Driver implementation 

LucaDazi:
Added a new parameter to the S7 Channel Descriptor, so that information on S7 Area Code can be forwarded to Moka7.

Channel configuration is backward compatible with previous versions.

### DIFF
--- a/kura/org.eclipse.kura.driver.s7plc.provider/src/main/java/org/eclipse/kura/internal/driver/s7plc/S7PlcArea.java
+++ b/kura/org.eclipse.kura.driver.s7plc.provider/src/main/java/org/eclipse/kura/internal/driver/s7plc/S7PlcArea.java
@@ -11,6 +11,9 @@
  */
 package org.eclipse.kura.internal.driver.s7plc;
 
+import org.eclipse.kura.KuraErrorCode;
+import org.eclipse.kura.KuraException;
+
 import Moka7.S7;
 
 public enum S7PlcArea {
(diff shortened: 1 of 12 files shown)


kura/org.eclipse.kura.driver.s7plc.provider/src/main/java/org/eclipse/kura/internal/driver/s7plc/S7PlcArea.java
```diff
@@ -0,0 +1,59 @@
+package org.eclipse.kura.internal.driver.s7plc;
```
MMaiero, line 12: Copyright banner is required

kura/org.eclipse.kura.driver.s7plc.provider/src/main/java/org/eclipse/kura/internal/driver/s7plc/S7PlcChannelDescriptor.java
```diff
@@ -20,6 +20,8 @@
 import org.eclipse.kura.driver.ChannelDescriptor;
 import org.eclipse.kura.util.collection.CollectionUtil;
 
+import Moka7.S7;
```
MMaiero, line 23: Please update all the copyright banners

MMaiero (commented):
Since this adds new functionality, I would say that we should move its version to 1.1.0-SNAPSHOT

nicolatimeus:
It looks like the `Counters` and `Timers` areas require special treatment, in this case the value of the `Amount` argument of `ReadArea`/`WriteArea` needs to be adjusted (see Moka7 manual).
In this case the library will transfer `Amount * 2` bytes, the length of the supplied array must be at least `Amount * 2`.
For the other areas the current implementation should be correct, the library will transfer `Amount` bytes.
I can work on a fix.

PR #2956, opened 2020-06-19
Changed dns displaying, using a textarea

MMaiero:

### DIFF
--- a/kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/client/ui/network/TabTcpIpUi.java
+++ b/kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/client/ui/network/TabTcpIpUi.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2011, 2018 Eurotech and/or its affiliates
+ * Copyright (c) 2011, 2020 Eurotech and/or its affiliates
  *
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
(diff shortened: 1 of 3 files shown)


LeoNerdoG:
Hey @MMaiero,
I have checked this and it looks fine, I just have 2 comments.
1. Error message should be changed if user has e.g. 15 DNS addresses and one is wrong (Kura reports "Invalid IPv4 Address", which is not correct, it should really say "one of ip addresses not valid, please check" or something similar. But this is just cosmetic
Screenshot 2020-06-19 at 14 49 54

2. The field is not limited - but there should be some sort of limitation.

MMaiero:
What do you mean with 

> but there should be some sort of limitation.

LeoNerdoG:
I meant that there is no upper limit for number of character user can enter into the field. If there is no limitation, one can break the system by entering 10M+ characters into the field and applying the changes. 
So I mean that there should be a limit for e.g. 10000 characters (that is almost 700 full-length DNS addresses, e.g. 255.255.255.255).

MMaiero:
I have to see if there are limits set at linux level
Looking at my Mac there seem to be no limit. But I'll double check

pierantoniomerlino:
Indeed, there aren't any limit on the number of dns servers that can be added to the  resolv.conf file.

pierantoniomerlino (approved):

kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/client/ui/network/TabTcpIpUi.java
```diff
@@ -42,6 +42,7 @@
 import org.gwtbootstrap3.client.ui.ListBox;
 import org.gwtbootstrap3.client.ui.Modal;
 import org.gwtbootstrap3.client.ui.PanelHeader;
+import org.gwtbootstrap3.client.ui.TextArea;
```
LeoNerdoG, line 45: Hey @MMaiero,
Can you please update the header to "2020"?

MMaiero:
Fixed the copyright header and the message, as suggested.

pierantoniomerlino (approved):

PR #2959, opened 2020-06-22
Drag-and-drop of dp files

### LINKED ISSUE #2940: Drag and drop of deployment packages
Currently the drag-and-drop feature for installing deployment packages is limited to the Eclipse marketplace. So, only a link to the marketplace is allowed in to the packages tab.

IMHO, a useful feature could be the upload of .dp files from the local filesystem with drag-and-drop.

pierantoniomerlino:
This PR adds the possibility to drag-and-drop deployment packages directly from the local filesystem.

**Related Issue:** This PR fixes/closes #2940 

Thanks to @nicolatimeus for the contribution!


kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/client/util/DropSupport.java
```diff
@@ -12,7 +12,10 @@
 
 package org.eclipse.kura.web.client.util;
 
+import java.util.List;
```
LeoNerdoG, line 15: Please update header of the file.

LeoNerdoG (changes requested):
Hey @pierantoniomerlino,
I have checked this PR and I have two comments:
1. please fix the header in DropSupport.java file
2. if user selects multiple packages from desktop and drops them in kura, Kura asks user if he wants to install the package (only one package) - if you say yes, nothing happens. Is this something we can ignore or should be fixed. 
All other scenarios are OK.

MMaiero:
I believe it works great!
I have tested the corner case with multiple dps selected. In my case it actually installed one of them.
But yes, we should take care of that case, in order to prevent issues. Can we check the number of files selected? If so, we may want to throw an error due to the fact that we support just one dp at a time.

kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/client/ui/packages/PackagesPanelUi.java
```diff
@@ -606,22 +611,53 @@ private boolean isEclipseMarketplaceUrl(String url) {
...
+                    event.stopPropagation();
                     PackagesPanelUi.this.packagesDropzone.removeStyleName(DROPZONE_ACTIVE_STYLE_NAME);
-                    if (isEclipseMarketplaceUrl(url)) {
-                        PackagesPanelUi.this.confirmDialog.show(MSGS.packagesMarketplaceInstallConfirmMessage(),
-                                () -> eclipseMarketplaceInstall(url));
+
+                    final List<File> droppedFiles = event.getFiles();
+
+                    if (!droppedFiles.isEmpty()) {
+
```
MMaiero: maybe, for readability, it is better to move the content to a specific method.

MMaiero (approved):
I made a small change in the message when multiple dp files are provided. @pierantoniomerlino can you have a look at it? Other than that for me it can be merged.

pierantoniomerlino:
@MMaiero For me is ok!

PR #2960, opened 2020-06-22
Add support for Bluetooth 5.X LE Extended Advertising Reports

### LINKED ISSUE #2705: HCI 5.0 / Extended Advertising support
Bluetooth adapters which support HCI 5.0 and extended advertising do not work with the current BluetoothLEBeacon implementation with newer Linux kernels and Bluez versions. This is because the current implementation is using deprecated command-line tools (hcitool and hcidump). The Bluetooth 5.0 spec states that if a card supports extended advertising this will be used by default which means legacy LE scan commands are rejected. the 'hcitool' is deprecated and doesn't support LE scans with HCI 5.0 adapters.

My suggestion would be to re-use the BluetoothLeAdapterImpl with TransportType set to LE to kick off a beacon scan using Bluez DBus API and then kick off a 'btmon' command-line process which can be used to parse advertising packets similar to 'hcidump'.

 - Board Up2 + Intel 9260 WiFi/BT adapter
 - OS version: Ubuntu 18, Linux Kernel 5.3.x

srware:
This PR adds support for Bluetooth 5.X adapters and LE Extended Advertising Reports.
The deprecated 'hcitool' currently used for beacon scanning is replaced by BlueZ Bluetooth Monitor utility.

This PR closes #2705


kura/org.eclipse.kura.ble.provider/src/main/java/org/eclipse/kura/internal/ble/BluetoothLeAdapterImpl.java
```diff
@@ -144,6 +144,14 @@ public boolean isDiscovering() {
...
     }
 
+    public void startDiscovery() throws KuraBluetoothDiscoveryException {
+        try {
+            this.adapter.startDiscovery();
+        } catch (BluetoothException e) {
+            throw new KuraBluetoothDiscoveryException(e, "Start discovery failed");
+        }
+    }
+
```
pierantoniomerlino, line 185: I think that we can add the signature of this method as API in the `BluetoothLeAdapter`.

kura/org.eclipse.kura.ble.provider/src/main/java/org/eclipse/kura/internal/ble/beacon/BluetoothLeBeaconScannerImpl.java
```diff
@@ -15,15 +15,16 @@
 import org.eclipse.kura.bluetooth.le.beacon.BluetoothLeBeaconDecoder;
 import org.eclipse.kura.bluetooth.le.beacon.BluetoothLeBeaconScanner;
 import org.eclipse.kura.bluetooth.le.beacon.listener.BluetoothLeBeaconListener;
+import org.eclipse.kura.internal.ble.BluetoothLeAdapterImpl;
 
 public class BluetoothLeBeaconScannerImpl<T extends BluetoothLeBeacon> implements BluetoothLeBeaconScanner<T> {
 
-    private final BluetoothLeAdapter adapter;
+    private final BluetoothLeAdapterImpl adapter;
     private final BluetoothLeBeaconDecoder<T> decoder;
```
pierantoniomerlino: Moving the `startDiscovery` in the API, we can leave the `BluetoothLeAdapter`. So, we'll pass the interface, not the implementation.

kura/org.eclipse.kura.ble.provider/src/main/java/org/eclipse/kura/internal/ble/util/BluetoothProcess.java
```diff
@@ -106,17 +110,30 @@ void execSnoop(String[] cmdArray, final BTSnoopListener listener) {
...
         this.executorService.execute(command, callback);
 
         this.futureInputGobbler = streamGobblers.submit(() -> {
             Thread.currentThread().setName("BluetoothProcess BTSnoop Gobbler");
             try {
                 readBTSnoopStreamFully(this.readOutputStream, listener);
             } catch (IOException e) {
-                logger.error("Process snoop input stream failed", e);
+                logger.debug("Process snoop input stream failed", e);
             }
```
pierantoniomerlino, line 137: Why `debug`?

kura/org.eclipse.kura.ble.provider/src/main/java/org/eclipse/kura/internal/ble/util/BluetoothProcess.java
```diff
@@ -126,7 +143,7 @@ void execSnoop(String[] cmdArray, final BTSnoopListener listener) {
             try {
                 readBTErrorStreamFully(this.readErrorStream, listener);
             } catch (IOException e) {
-                logger.error("Process snoop error stream failed", e);
+                logger.debug("Process snoop error stream failed", e);
             }
```
pierantoniomerlino, line 147: Why `debug`?

kura/org.eclipse.kura.ble.provider/src/main/java/org/eclipse/kura/internal/ble/util/BluetoothProcess.java
```diff
@@ -162,11 +179,12 @@ private void readBTSnoopStreamFully(InputStream is, BTSnoopListener listener) th
...
             if (is != null) {
                 byte[] packet = this.parser.readRecord();
-                listener.processBTSnoopRecord(packet);
+                if (packet.length > 0) {
+                    listener.processBTSnoopRecord(packet);
+                }
             }
         }
-        closeStreams();
-        logger.debug("End of stream!");
```
pierantoniomerlino, line 172: Why did you remove the streams close?

pierantoniomerlino:
@srware Thank you very much for your contribution! I took a look at the code and it's seems to be fine.  I added only few comments. BTW, some examples and test are broken with these modifications: could you take a look at that? Probably only the versions of the dependencies should be updated.

kura/org.eclipse.kura.ble.provider/src/main/java/org/eclipse/kura/internal/ble/BluetoothLeAdapterImpl.java
srware: This makes sense, I just wasn't sure if you wanted to expose this functionality in the API directly given the results of starting a discovery would not be consumable through existing API methods.

pierantoniomerlino: Yes, I understand your concern... but I prefer to have it in the API (along with the `stopDiscovery`..)

kura/org.eclipse.kura.ble.provider/src/main/java/org/eclipse/kura/internal/ble/util/BluetoothProcess.java
srware: Since the introduction of the CommandExecutorService the btdump process is killed independent of the btsnoop parsing which means the stream almost always ends with an incomplete btsnoop packet. The btsnoop parser doesn't check for available bytes before attempting to read which is fair enough, but does mean this exception is thrown more often than not when the btdump process ends which is an expected error. This does have the potential to hide more serious issues parsing the btsnoop data but unless the BluetoothProcess handles the btdump process directly like it used to this isn't easily avoided.

kura/org.eclipse.kura.ble.provider/src/main/java/org/eclipse/kura/internal/ble/util/BluetoothProcess.java
srware: The while() loop here doesn't quit unless you call destroyBTSnoop(). This function includes a call to closeStreams() already so calling it again here is unnecessary from what I can see.

srware:
@pierantoniomerlino , I will make the changes to the BluetoothLEAdapter implementation and take a look at the examples & tests. Thanks for taking a look and providing feedback.

pierantoniomerlino:
Testing the functionality with iBeacon/Eddystone scanners, I noticed that the scanner reports only few advertising packets and it seems that is due to the duplicates filtering of the discovery. Indeed, the `hcitool` command is called with the `--duplicates` option to get all the packets. AFAIK, there isn't an option to remove the filter and receive duplicate packets using Tinyb/Bluez.

@srware do you know if there is the possibility to get all the advertisement packets during a discovery?

srware:
@pierantoniomerlino , good catch. I was still getting 20+ packets over a 5 second period from a single MAC in my test service so didn't consider duplicate filtering. TinyB doesn't offer this option and given it hasn't been maintained in a long time I would be inclined to look at an alternative library like [bluez-dbus](https://github.com/hypfvieh/bluez-dbus) which doesn't require the native libraries and does have support for enabling duplicate data as a discovery filter call. Thoughts?

pierantoniomerlino:
@srware Ah! bluez-dbus is interesting... I'll take a look at that.

srware:
@pierantoniomerlino , I have made the changes you mentioned and fix the examples and tests. Before I push the changes to this PR are you happy to merge this without the duplicate fix and I will attempt to replace tinyb with bluez-dbus in a separate PR or would you prefer that I try and do that and add it to this PR in a separate commit maybe?

pierantoniomerlino:
@srware Maybe it's better to contribute a full feature, that is try to integrate bluez-dbus in this PR in a separate commit. Thanks!

srware:
An update on this PR. I have finished the migration to bluez-dbus but several bugs were fixed in the library during the process so I am waiting for a new release of bluez-dbus before finalising this.

kura/org.eclipse.kura.ble.provider/src/main/java/org/eclipse/kura/internal/ble/BluetoothLeAdapterImpl.java
```diff
@@ -144,6 +144,15 @@ public boolean isDiscovering() {
         return uuidList.toArray(uuids);
     }
 
+    @Override
```
LeoNerdoG, line 175: Please update the header to "2017, 2020".

kura/org.eclipse.kura.ble.provider/src/main/java/org/eclipse/kura/internal/ble/beacon/BluetoothLeBeaconManagerImpl.java
```diff
@@ -6,6 +6,8 @@
  * which accompanies this distribution, and is available at
  * http://www.eclipse.org/legal/epl-v10.html
  *
+ * Contributors:
```
LeoNerdoG, line 10: Please update the header to "2017, 2020".

kura/org.eclipse.kura.ble.provider/src/main/java/org/eclipse/kura/internal/ble/beacon/BluetoothLeBeaconScannerImpl.java
```diff
@@ -44,7 +44,7 @@ public BluetoothLeAdapter getAdapter() {
     @Override
     public void startBeaconScan(long timeout) throws KuraBluetoothCommandException {
         try {
-            this.beaconManager.startBeaconScan(this.adapter.getInterfaceName());
+            this.beaconManager.startBeaconScan(this.adapter);
```
LeoNerdoG, line 50: Please update the header to "2017, 2020".

kura/org.eclipse.kura.ble.provider/src/main/java/org/eclipse/kura/internal/ble/util/BTSnoopParser.java
```diff
@@ -37,18 +37,23 @@ public void setInputStream(InputStream is) {
             this.gotHeader = true;
         }
 
-        readInt();
-        int includedLength = readInt();
-        readInt();
-        readInt();
+        readInt(); // Original Length
```
LeoNerdoG, line 43: Please update the header to "2017, 2020".

kura/test/org.eclipse.kura.internal.ble.test/src/test/java/org/eclipse/kura/internal/ble/beacon/BluetoothLeBeaconManagerImplTest.java
```diff
@@ -469,40 +469,6 @@ public void testProcessInputStreamFail14() throws KuraException {
         }
     }
```
LeoNerdoG, line 476: Please update the header to "2018, 2020".

kura/test/org.eclipse.kura.internal.ble.test/src/test/java/org/eclipse/kura/internal/ble/beacon/BluetoothLeBeaconScannerImplTest.java
```diff
@@ -43,7 +43,7 @@ public void testStartBeaconScan() throws KuraBluetoothCommandException, Interrup
                 lock.notifyAll();
             }
             return null;
-        }).when(beaconManagerMock).stopBeaconScan(devName);
+        }).when(beaconManagerMock).stopBeaconScan(adapterMock);
```
LeoNerdoG, line 50: Please update the header to "2018, 2020".

LeoNerdoG (changes requested):
Hey @srware,
can you please update the headers in some of the files? Thanks!

srware:
@pierantoniomerlino , I have pushed another commit replacing TinyB with bluez-dbus. This should now be complete.

Thanks.

MMaiero:
@pierantoniomerlino Can you please have a look at this? If it is ok, we can then see if CQs are needed.

pierantoniomerlino:
@srware I'm sorry for the delay... I'm currently checking the code and try it on my device!

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/bluetooth/le/BluetoothLeAdapter.java
```diff
@@ -195,7 +202,7 @@
      * Sets the discoverable timeout the adapter. A value of 0 disables
      * the timeout.
      */
-    public void setDiscoverableTimout(long value);
+    public void setDiscoverableTimeout(long value);
```
pierantoniomerlino, line 220: For compatibility reasons, we'd keep the old method, deprecate it and add the new one.

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/bluetooth/le/BluetoothLeAdapter.java
```diff
@@ -274,10 +281,12 @@
      *            the pathloss value
      * @param trasportType
      *            the trasportType (LE or BREDR)
+     * @param duplicateData
+     *            whether to filter duplicate data
      *
      * @since 2.0
      */
-    public void setDiscoveryFilter(List<UUID> uuids, int rssi, int pathloss, BluetoothTransportType transportType);
+    public void setDiscoveryFilter(List<UUID> uuids, int rssi, int pathloss, BluetoothTransportType transportType, boolean duplicateData);
```
pierantoniomerlino, line 339: As for the previous method, we'd keep the old one. We can decide to deprecate it on favor of the new one or keep both methods.

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/bluetooth/le/BluetoothLeDevice.java
```diff
@@ -263,10 +264,9 @@
...
      * will not be valid anymore and any operation on it will have no effect.
      *
-     * @return TRUE if the device has been removed
      * @throws BluetKuraBluetoothRemoveExceptionoothException
      *
      * @since 2.0
      */
-    public boolean remove() throws KuraBluetoothRemoveException;
+    public void remove() throws KuraBluetoothRemoveException;
 }
```
pierantoniomerlino, line 277: Here we've two choices:
1. simply modify the method
2. keep the old one, deprecate it and create a new one with a different name

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/bluetooth/le/beacon/AdvertisingReportAddressType.java
```diff
@@ -14,14 +14,18 @@
  * Possible values are:
  * 0x00 : Public Device Address
  * 0x01 : Random Device Address
- * 0x05 - 0xFF : Reserved for future use
+ * 0x02 : Public Identity Address
+ * 0x03 : Random Identity Address
  *
```
pierantoniomerlino, line 23: Could you please add also the UNRESOLVED?

kura/org.eclipse.kura.ble.eddystone.provider/META-INF/MANIFEST.MF
```diff
@@ -5,8 +5,8 @@ Bundle-SymbolicName: org.eclipse.kura.ble.eddystone.provider;singleton:=true
 Bundle-Version: 1.0.400.qualifier
 Bundle-RequiredExecutionEnvironment: JavaSE-1.8
 Import-Package: org.eclipse.kura;version="[1.0,2.0)",
- org.eclipse.kura.bluetooth.le;version="[1.2,1.3)",
- org.eclipse.kura.bluetooth.le.beacon;version="[1.0,1.1)",
+ org.eclipse.kura.bluetooth.le;version="[2.0,3.0)",
+ org.eclipse.kura.bluetooth.le.beacon;version="[1.0,3.0)",
```
pierantoniomerlino: I'm not sure that [1.0,3.0) is the right values... maybe [2.0,3.0)?

kura/org.eclipse.kura.ble.ibeacon.provider/META-INF/MANIFEST.MF
```diff
@@ -5,8 +5,8 @@ Bundle-SymbolicName: org.eclipse.kura.ble.ibeacon.provider;singleton:=true
 Bundle-Version: 1.0.400.qualifier
 Bundle-RequiredExecutionEnvironment: JavaSE-1.8
 Import-Package: org.eclipse.kura;version="[1.0,2.0)",
- org.eclipse.kura.bluetooth.le;version="[1.2,1.3)",
- org.eclipse.kura.bluetooth.le.beacon;version="[1.0,1.1)",
+ org.eclipse.kura.bluetooth.le;version="[2.0,3.0)",
+ org.eclipse.kura.bluetooth.le.beacon;version="[1.0,3.0)",
  org.eclipse.kura.bluetooth.le.beacon.listener;version="[1.0,1.1)",
```
pierantoniomerlino, line 10: The same as above

kura/org.eclipse.kura.ble.provider/src/main/java/org/eclipse/kura/internal/ble/BluetoothLeDeviceImpl.java
```diff
@@ -25,40 +24,41 @@
...
         } else {
             throw new KuraBluetoothResourceNotFoundException("Gatt service not found.");
         }
     }
 
+    @Override
+    public BluetoothLeGattService findService(UUID uuid, long timeout) throws KuraBluetoothResourceNotFoundException {
+        return this.findService(uuid);
+    }
+
```
pierantoniomerlino, line 57: You removed the timeout functionality from this method. Is this because you are sure that the method will always return or throw an exception? Or it can be possible that the device will get stuck forever?

kura/org.eclipse.kura.ble.provider/src/main/java/org/eclipse/kura/internal/ble/BluetoothLeGattCharacteristicImpl.java
```diff
@@ -72,44 +75,65 @@ public BluetoothLeGattDescriptor findDescriptor(UUID uuid, long timeout)
...
+                        if (!props.getPath().contains(BluetoothLeGattCharacteristicImpl.this.characteristic.getDbusPath())) {
+                            return;
+                        }
+
+                        for (Map.Entry<String, Variant<?>> entry : props.getPropertiesChanged().entrySet()) {
+                            if (entry.getKey().equals("Value")) {
+                                Object obj = entry.getValue().getValue();
+                                byte[] val = (byte[]) obj;
+                                callback.accept(val);
+                            }
```
pierantoniomerlino:
```
byte[] value = (byte[]) props.getPropertiesChanged().get("Value").getValue();
if (value != null) {
    callback.accept(value);
}
```

kura/org.eclipse.kura.ble.provider/src/main/java/org/eclipse/kura/internal/ble/BluetoothLeServiceImpl.java
```diff
@@ -21,36 +21,42 @@
...
+
+import com.github.hypfvieh.bluetooth.DeviceManager;
+import com.github.hypfvieh.bluetooth.wrapper.BluetoothAdapter;
 
 public class BluetoothLeServiceImpl implements BluetoothLeService {
 
     private static final Logger logger = LogManager.getLogger(BluetoothLeServiceImpl.class);
 
-    private BluetoothManager bluetoothManager;
+    private DeviceManager bluetoothManager;
```
pierantoniomerlino, line 46: It's worth to change the variable name from `bluetoothManager` to `deviceManager`?

kura/org.eclipse.kura.ble.provider/src/main/java/org/eclipse/kura/internal/ble/BluetoothLeGattCharacteristicImpl.java
```diff
@@ -72,44 +75,65 @@ public BluetoothLeGattDescriptor findDescriptor(UUID uuid, long timeout)
...
+                        }
+                    }
+                }
+            });
+            this.characteristic.startNotify();
+        } catch (DBusException e) {
```
pierantoniomerlino, line 99: Why we need to call `startNotify`?

pierantoniomerlino:
@srware Great work! I finished to check the code. I tested the SensorTag example/driver, iBeacon and Eddystone bundles (only the XDK is missing) on a RPi and all is almost fine. I have few concerns:

- in your opinion is is possible to replace (totally or partially) the use of `hcitool`, `btmon`/`hcidump` with the new library?
- even if the iBeacon/Eddystone scanner works, they detect only few advertising packets as if `DuplicateData` is not set. According to [1], is not possible to set the `DuplicateData` using dbus, but only sending the commands at low level using `hcitool`.

Finally, please rebase your branch with the new modifications on the develop branch.

Thank you!

[1] https://stackoverflow.com/a/64361977/4213620

kura/examples/org.eclipse.kura.example.ble.tisensortag.dbus/OSGI-INF/metatype/org.eclipse.kura.example.ble.tisensortag.dbus.BluetoothLe.xml
```diff
@@ -49,8 +49,8 @@
             default="120"
             description="Period in seconds between 2 publishes. It must be greater than scan_time."/>
             
-        <AD id="enableTermometer"
-            name="enableTermometer"
+        <AD id="enableTemperature"
+            name="enableTemperature"
             type="Boolean"
```
pierantoniomerlino, line 58: The `termometer` is for sure a typo... but in all others properties we use the name of the tool/sensor used to read that physical quantity. So, we can use `thermometer` or we can decide to replace all properties name with the name of the quantity.

kura/examples/org.eclipse.kura.example.ble.tisensortag.dbus/META-INF/MANIFEST.MF
```diff
@@ -1,12 +1,12 @@
 Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
-Bundle-Name: org.eclipse.kura.example.ble.tisensortag.tinyb
-Bundle-SymbolicName: org.eclipse.kura.example.ble.tisensortag.tinyb;singleton:=true
+Bundle-Name: org.eclipse.kura.example.ble.tisensortag.dbus
+Bundle-SymbolicName: org.eclipse.kura.example.ble.tisensortag.dbus;singleton:=true
 Bundle-Version: 1.0.400.qualifier
```
pierantoniomerlino: Since we have changed the name of the bundle, is it a new one? If so, have we to start again from 1.0.0?

kura/org.eclipse.kura.ble.provider/src/main/java/org/eclipse/kura/internal/ble/BluetoothLeGattCharacteristicImpl.java
srware: Once we setup the PropertyHandler you have to set the characteristic notifying on change in order to receive anything on the callback. The previous TinyB enableValueNotifications() method did the same internally.

kura/examples/org.eclipse.kura.example.ble.tisensortag.dbus/META-INF/MANIFEST.MF
srware: Good question. It's technically the same bundle but I figured calling it "tinyb" after a library no longer in use didn't make much sense.

I will let you instruct as to your preference on versioning.

kura/org.eclipse.kura.ble.provider/src/main/java/org/eclipse/kura/internal/ble/BluetoothLeDeviceImpl.java
srware: The bluez-dbus library has no such method with a configurable timeout. There is a default timeout associated with the DBus connection (10 seconds) at which point this method will throw an error.

kura/org.eclipse.kura.ble.provider/src/main/java/org/eclipse/kura/internal/ble/BluetoothLeServiceImpl.java
srware: Probably makes sense :)

srware:
@pierantoniomerlino , I have rebased and made the recommended changes. CI seems to have stalled doing the checks but hopefully it's OK.

Thanks.

pierantoniomerlino:
@srware Great! I restarted the tests and now they work fine. I'm managing the CQs in IPZilla for bluez-dbus and the PR. Then we'll merge the PR on develop.

We have still the issue regarding the duplicate advertisement packets, but we'll fix that later.
Thanks!

pierantoniomerlino:
CQs submitted!

pierantoniomerlino:
The CQs were submitted, but the `dbus-java` library was rejected for a license incompatibility. So, we asked the owner of that project if it was possible to change the license to a more "eclipse-friendly" one. He accepted and [here](https://github.com/hypfvieh/dbus-java/issues/129) is the issue for updating the license to MIT. We're currently waiting for the next release of the `dbus-java` and `bluez-dbus`.

srware:
Thanks for the update @pierantoniomerlino .

pierantoniomerlino:
@srware The new version of [bluez-dbus](https://github.com/hypfvieh/bluez-dbus/tree/bluez-dbus-parent-0.1.4) has been finally released! I'll re-submit the CQs. If you agree, I can update this PR with the last library versions.

srware:
> @srware The new version of [bluez-dbus](https://github.com/hypfvieh/bluez-dbus/tree/bluez-dbus-parent-0.1.4) has been finally released! I'll re-submit the CQs. If you agree, I can update this PR with the last library versions.

@pierantoniomerlino, sounds good from my side :)

pierantoniomerlino:
I added some tests for the ble.provider bundle. I had to recover the `hcitool` command for beacon scanning since it wasn't possible to get all the advertisement packets. I hope that a next release of `bluez-dbus` will fix this...

kura/test/org.eclipse.kura.internal.ble.test/src/test/java/org/eclipse/kura/internal/ble/BluetoothLeGattCharacteristicImplTest.java
```diff
@@ -0,0 +1,117 @@
+package org.eclipse.kura.internal.ble;
```
MMaiero, line 13: Copyright header

srware:
> I added some tests for the ble.provider bundle. I had to recover the `hcitool` command for beacon scanning since it wasn't possible to get all the advertisement packets. I hope that a next release of `bluez-dbus` will fix this...

@pierantoniomerlino , yes, i'm not sure where the "duplicates" issue lies. I would assume either the bluez-dbus library is not invoking the filter capabilities of the BlueZ DBus API correctly or there is an issue with BlueZ itself. If this were the case I would expect the "bluetoothctl" utility to also not work when filters are cleared.

Unfortunately invoking "hcitool" on a BT5.0 adapter puts them into "legacy" mode which means no BLE extended advertisements etc...

pierantoniomerlino:
@srware This is a good point. According to this [site](https://ukbaz.github.io/howto/beacon_scan_cmd_line.html), it seems that Bluez filters the duplicate packets, but even if `DuplicateData` is set, not more than 1 message every 10s is received (I also checked it with `bluectoothctl` and `dbus-monitor`). 

I opened an issue in the Bluez repo to get support. In the meanwhile, I'm going to recover the scan using `bluez-dbus` and make the `hcitool` optional.

MMaiero (commented):
If we are moving from 2.2 to 3.0 for the API, then we need to update all the corresponding annotations.

MMaiero (approved):

kura/distrib/config/kura.build.properties
```diff
@@ -65,7 +65,7 @@ org.eclipse.kura.example.project.version=1.0.400
 org.eclipse.kura.example.publisher.version=1.0.500
 org.eclipse.kura.example.serial.publisher.version=1.0.400
 org.eclipse.kura.example.ble.tisensortag.version=1.0.500
-org.eclipse.kura.example.ble.tisensortag.tinyb.version=1.0.300
+org.eclipse.kura.example.ble.tisensortag.dbus.version=1.0.300
```
MMaiero: maybe this can be a version 1.0.0 since the renaming means a new bundle

pierantoniomerlino (approved):

nicolatimeus (approved):

pierantoniomerlino:
@srware We've merged your PR, but we found that an additional CQ is requested by Eclipse, since your contribution is more than 1000 lines... So, we reverted the merge and filled a new CQ for the code. But, we cannot re-open your PR. Could you please open a new one with all the code modifications?

We're really sorry for this.

srware:
@pierantoniomerlino 

Done: https://github.com/eclipse/kura/pull/3377

PR #2973, opened 2020-07-03
Fixed timer description

### LINKED ISSUE #2703: Wrong Description in "Timer" Wire Component
When creating a timer wire component, there is a description above "simple.interval" field, which says: "Interval of the timer ticks, in the time unit specified by the simple.time.unit property (Needs to be set if SIMPLE type is selected)"
The problem is in part "(Needs to be set if SIMPLE type is selected)" which is clearly not true, since this field has to be set in every case, since it is a mandatory field. 
The same goes for "cron.interval" field - these fields are marked as mandatory (asterix symbol) so the part in the description should be removed or the fields should be modified to be in sync with the "type" field (e.g. if user select "CRON" the field for simple interval gets disabled and vice versa for "SIMPLE" option).

**Testflow:**
2. Go to Wire Graph, create a timer wire component (e.g. timer1)
3. Select timer1 and observe the "simple.interval" and "cron.interval" field descirptions.

**Expected behavior**
See above. 

**Screenshot:**
Screenshot 2020-03-11 at 10 29 12

Board: Raspberry Pi 3
Additional info: /

MMaiero:

Closes #2703

### DIFF
--- a/kura/org.eclipse.kura.wire.component.provider/OSGI-INF/metatype/org.eclipse.kura.wire.Timer.xml
+++ b/kura/org.eclipse.kura.wire.component.provider/OSGI-INF/metatype/org.eclipse.kura.wire.Timer.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!--
-    Copyright (c) 2016, 2017 Eurotech and/or its affiliates and others
+    Copyright (c) 2016, 2020 Eurotech and/or its affiliates and others
 
      All rights reserved. This program and the accompanying materials
      are made available under the terms of the Eclipse Public License v1.0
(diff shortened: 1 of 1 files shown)


kura/org.eclipse.kura.wire.component.provider/OSGI-INF/metatype/org.eclipse.kura.wire.Timer.xml
```diff
@@ -29,7 +29,7 @@
             cardinality="0"
             required="true"
             default="10"
-            description="Interval of the timer ticks, in the time unit specified by the simple.time.unit property (Needs to be set if SIMPLE type is selected)"
+            description="Interval of the timer ticks, in the time unit specified by the simple.time.unit property. This value is used when type is set to SIMPLE."
```
LeoNerdoG, line 32: Change description is OK, please update the header of the document.

LeoNerdoG (commented):
Hey @MMaiero,
I have checked this PR and the description is OK now. 
Please update the header and then it can be merged.

kura/org.eclipse.kura.wire.component.provider/OSGI-INF/metatype/org.eclipse.kura.wire.Timer.xml
MMaiero: Done

LeoNerdoG (approved):

PR #2980, opened 2020-07-09
Added a check in bg scan thresholds in order to prevent wrong interva…

### LINKED ISSUE #2887: "Bgscan Short Interval" Can be Longer Than "Bgscan Long Interval"
If user sets Wireless Mode to Station Mode, two options open: Bgscan short and long interval. The short one specifies the minimum value and the long interval specifies the max scanning interval. However, user is able to set short interaval to be longer then the Long one - is this okay?

**Testflow:**
2. Go to Network -> wlan0 -> Wireless
3. Select Station mode, enter "10" for Bgscan short interval
4. Enter "1" for Bgscan long interval
5. Kura does not return any errors

**Expected behavior**
Bgscan Short Interval should be always shorter than Bgscan Long Interval. 

**Screenshot:**
Screenshot 2020-05-20 at 09 31 46

Board: Raspberry Pi 3
Kura version: 4.2.0_SNAPSHOT ([Link to the specific Kura version -from May 20th 2020](https://ci.eclipse.org/kura/job/kura-develop/lastSuccessfulBuild/artifact/kura/distrib/target/kura_4.2.0-SNAPSHOT_raspberry-pi-2-3_installer.deb)

MMaiero:
…l values setup

Closes #2887

### DIFF
--- a/kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/client/ui/network/TabWirelessUi.ui.xml
+++ b/kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/client/ui/network/TabWirelessUi.ui.xml
@@ -167,8 +167,10 @@
                                 <b:FormLabel for="shortI"
                                     ui:field="labelShortI" />
                                 <util:HelpButton ui:field="shortIHelp"/>
+                                <b:HelpBlock color="red"
+                                    ui:field="helpShortI" />
                                 <b:TextBox b:id="shortI"
-                                    ui:field="shortI" />
+                                    ui:field="shortI" validateOnBlur="true" />
                             </b:FormGroup>
 
                             <b:FormGroup ui:field="groupLongI">
@@ -176,7 +178,7 @@
                                     ui:field="labelLongI" />
                                 <util:HelpButton ui:field="longIHelp"/>
                                 <b:TextBox b:id="longI"
-                                    ui:field="longI" />
+                                    ui:field="longI" validateOnBlur="true" />
                             </b:FormGroup>
 
                             <b:FormGroup>
--- a/kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/client/ui/network/TabWirelessUi.ui.xml
+++ b/kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/client/ui/network/TabWirelessUi.ui.xml
@@ -2,7 +2,7 @@
 
 <!--
 
-    Copyright (c) 2011, 2019 Eurotech and/or its affiliates
+    Copyright (c) 2011, 2020 Eurotech and/or its affiliates
 
      All rights reserved. This program and the accompanying materials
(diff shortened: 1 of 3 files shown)


LeoNerdoG:
Hey @MMaiero,
I have checked this and now Kura reports that the field is in error state, but the changes can be applied anyway. 
Can you please check on your side?

MMaiero:
I've made an improvement. Can you please check if this works?

LeoNerdoG:
Hey @MMaiero,
I have checked this and:

1. the error message is "The bgscan short interval value must be less than the long value", but validation lets user to set long and short interval to the same value. 

Screenshot 2020-07-13 at 08 35 22

Screenshot 2020-07-13 at 08 35 39

2. There is no validation nor error message beyond sixth number in the fields (so validation is done only if user has <= 5 total numbers)

Screenshot 2020-07-13 at 08 39 01

3. Both fields can be set to zero (is this ok?)

4. No validation if user enters a character into Bgscan long interv

Screenshot 2020-07-13 at 09 10 20

MMaiero:
Right. I've made few improvements.

LeoNerdoG:
Hey @MMaiero,
I have rechecked this PR, it is much better now 😄 ,  but...

1. If user enters shorter short interval than long interval, the error messages are the same ("The bgscan short interval value must be less than the long value") - is this intended? shouldn't the error messages be different for the two fields? 
Screenshot 2020-07-13 at 14 32 35

2. If user enters a character or a symbol into on of the two fields, both fields are in error state (which shouldn't be, if one field is not in error state, it should not be labeled red):
Screenshot 2020-07-13 at 14 32 51

3. If you entere erroneous stuff in either of the fields, change views (discard changes!) and return, both fields are still in error state. 

4. if user enters e.g. "               12            " into longer field and "1" in short field, there is no way of saving configuration - apply button does not work.

MMaiero:
1 and 2 are in that way due to the fact that the two fields are related. A change in one field means a validation in both. We could remove this link but that means that if short is 100, long is 10 than short will be in error. But if I change long to a valid field than short will stay in error due to the fact that the validation is done on blur.
For the others, I will take a look.

MMaiero:
I have fixed 4. 3 will stay there for the time being. I'll open an issue for that, but needs more work.

kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/client/ui/network/TabWirelessUi.ui.xml
```diff
@@ -167,16 +167,20 @@
                                 <b:FormLabel for="shortI"
                                     ui:field="labelShortI" />
                                 <util:HelpButton ui:field="shortIHelp"/>
+                                <b:HelpBlock color="red"
```
LeoNerdoG, line 170: Please update the header to "2020".

LeoNerdoG:
hey @MMaiero,
I have checekd this PR and it looks OK, so it can be merged into develop.

LeoNerdoG (approved):

PR #2984, opened 2020-07-15
Changed target selection code using OSGi bundle context to discover i…

### LINKED ISSUE #2983: UI Target selection is not working in all cases
**Describe the bug**
Currently the target filter in UI is based on a Class.forName invocation that works only for the interfaces imported by the Web bundle, as it uses the local bundle class loader.
If the user tries to use the target filter with its own APIs, the filtering will return an empty list

**To Reproduce**
1. Develop a new bundle that implements an interface that is not imported by the web bundle
2. The corresponding target filter in UI will not be able to find the corresponding implementations 

**Expected behavior**
The target filter should work across all the bundles.

**Screenshots**
N/A

 - Board: ALL
 - OS version: ALL
 - Additional info: N/A

**Additional context**
Tested with current develop but applicable also to previous versions of the framework

MMaiero:
…mplementors

Closes #2983

### DIFF
--- a/kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/server/GwtComponentServiceImpl.java
+++ b/kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/server/GwtComponentServiceImpl.java
@@ -873,18 +873,14 @@ public class GwtComponentServiceImpl extends OsgiRemoteServiceServlet implements
 
             referenceInterfaces.forEach(reference -> {
                 try {
-                    Class<?> t = Class.forName(reference);
-                    Collection<?> cloudServiceReferences = ServiceLocator.getInstance().getServiceReferences(t, null);
+                    ServiceReference<?>[] serviceReferences = context.getServiceReferences(reference, null);
 
-                    for (Object cloudServiceReferenceObject : cloudServiceReferences) {
-                        if (cloudServiceReferenceObject instanceof ServiceReference) {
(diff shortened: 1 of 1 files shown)


pierantoniomerlino (approved):

kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/server/GwtComponentServiceImpl.java
```diff
@@ -873,18 +873,14 @@ public boolean updateProperties(GwtXSRFToken xsrfToken, String pid, Map<String,
...
-                    Collection<?> cloudServiceReferences = ServiceLocator.getInstance().getServiceReferences(t, null);
-
-                    for (Object cloudServiceReferenceObject : cloudServiceReferences) {
-                        if (cloudServiceReferenceObject instanceof ServiceReference) {
-                            ServiceReference<?> cloudServiceReference = (ServiceReference<?>) cloudServiceReferenceObject;
-                            String cloudServicePid = (String) cloudServiceReference.getProperty(KURA_SERVICE_PID);
-                            result.add(cloudServicePid);
-                            ServiceLocator.getInstance().ungetService(cloudServiceReference);
-                        }
+                    ServiceReference<?>[] serviceReferences = context.getServiceReferences(reference, null);
```
LeoNerdoG, line 876: Please update the header of the file to 2020.

Issue #2752, opened 2020-04-01
Uploading Channels Not Working Properly?


LeoNerdoG:
The "Replace current channel" description in Wire graph -> Wire Asset -> upload channels is: "If checked, channels will be replaced by the imported ones. Otherwise, they will be appended to the list." 
But if you download the channels, edit one of them and then upload it again, the channel gets updated - is this the desired behaviour or not? This is logical, but is not consistent with the description. 

**Testflow:**
1. Login to Kura as admin/admin
2. Go to Packages, install wires-development-package
3. Go to Wire graph, create an asset with a component containing four channels (channel-1, 2, 3 and 4)
4. Download the channels and change parameters of Channel-1
5. Upload the document and observe the result - channel-1 gets updated

**Expected behavior**
I don't know what is the desired behaviour here, but maybe only description should be tweaked a bit?

**Screenshot:**
![Screenshot 2020-04-01 at 14 24 09](https://user-images.githubusercontent.com/15121999/78137860-472e0900-7426-11ea-8126-cccd779a5583.png)

Target Environment (please complete the following information):
Board: Raspberry Pi 3
OS version: Linux raspberrypi 4.19.75-v7+ # 1270 SMP Tue Sep 24 18:45:11 BST 2019 armv7l GNU/Linux
Additional info: /

LeoNerdoG:
I have verified that this issue has been resolved.

Issue #2755, opened 2020-04-02
Different Behaviour When Adding and Uploading Channels


LeoNerdoG:
If user adds a channel "manually" (by clicking "Add channel" button) to an asset in Wire Graph, the "Download" button gets disabled and Reset and Apply buttons get enabled. This is fine, but when user uploads channels, this does not happen, although user might add several new channels as well. We should unify this behaviour regardless of user adding channels manually or uploading them. 

**Testflow:**
1. Login to Kura as admin/admin
2. Go to Packages, install Wires-development-package from Eclipse Marketplace
3. Go to Wire Graph, create a wireAsset comonent and add a channel to it
4. Try to upload a random .txt file via "Upload channels" button
5. Observe the error message

**Expected behavior**
Same behaviour when adding channels manually or uploading them.

**Screenshots**
/
Target Environment (please complete the following information):
Board: Raspberry Pi 3
OS version: Linux raspberrypi 4.19.75-v7+ # 1270 SMP Tue Sep 24 18:45:11 BST 2019 armv7l GNU/Linux
Additional info: /

MMaiero:
I believe #2996 has addressed also this issue.
@LeoNerdoG can you please check and close this issue if the assumption is correct?

LeoNerdoG:
Hey @MMaiero,
yes, you are right. I have checked this and it has been successfully resolved with PR #2996.

PR #2996, opened 2020-07-22
Improved CSV channel upload

nicolatimeus:

* Uploading Asset channels from CSV no longer alters device configuration
* The only requirement for upload to work now is that the Driver instance associated with the Asset exists on the system, this enables channel uploads in more scenarios:
  * Assets with dirty config
  * Asset existing in the Web ui but not yet created on the device
* Upload channels button should now always be in the enabled state

### DIFF
--- a/kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/client/ui/ConfigurableComponentUi.java
+++ b/kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/client/ui/ConfigurableComponentUi.java
@@ -65,7 +65,9 @@ public class ConfigurableComponentUi extends AbstractServicesUi implements HasCo
             if (isDirtyStateChanged) {
                 this.listener.onDirtyStateChanged(this);
             }
-            this.listener.onConfigurationChanged(this);
+            if (this.dirty) {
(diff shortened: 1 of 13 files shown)


nicolatimeus:
Improves a bit the upload modal description as suggested in #2752.
Download button should be disabled now both when adding and uploading channels, probably fixing #2755. A change similar to this one could be applied to channel download as well, allowing to have the download button always enabled as well.

MMaiero:
I've seen that once instantiated, the download button is available. Is that expected? 
It creates a csv with no channels in it but you cannot upload it, since it is rejected while uploading

LeoNerdoG:
Hey @nicolatimeus,
I have checekd this PR and it fixes issue #2752. I am also confirming that I can reproduce matteo's issue/comment.

nicolatimeus:
I tried to reproduce:
* If I create an empty asset in wire composer, download button is not enabled.
* If I create an empty asset in the drivers an assets section, then the download button is enabled as reported.

Can you confirm this?

In the latter case probably the download button is enabled because the asset is also created on the device, and so its configuration is not "dirty".
Since a CSV file with no channels is considered invalid by the upload validation logic, we could add an additional check that disables channel download if asset has no channels defined.

MMaiero (approved):

PR #2999, opened 2020-07-24
Improved CSV channel download

nicolatimeus:
Adopted a similar approach as #2996 for channel download.
 * Downloading channels is now possible even if an Asset does not exist on the device or if its configuration is dirty.
 * The download channel button should always be enabled, unless the channel table is empty or if Asset configuration is not valid.

### DIFF
--- a/kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/server/servlet/ChannelServlet.java
+++ b/kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/server/servlet/ChannelServlet.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2018 Eurotech and/or its affiliates and others
+ * Copyright (c) 2018, 2020 Eurotech and/or its affiliates and others
  *
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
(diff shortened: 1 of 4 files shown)


nicolatimeus:
Needs to be rebased after #2996 is merged

kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/server/servlet/ChannelServlet.java
```diff
@@ -12,40 +12,19 @@
...
-import java.util.Collection;
-import java.util.List;
-import java.util.concurrent.atomic.AtomicBoolean;
 
 import javax.servlet.ServletException;
 import javax.servlet.http.HttpServlet;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 import javax.servlet.http.HttpSession;
+import javax.xml.ws.http.HTTPException;
```
LeoNerdoG, line 23: Please update header to "2020".

LeoNerdoG (commented):
Hey @nicolatimeus,
I have checked this and the second part is OK (The download channel button should always be enabled).
I would just like to know how to achieve dirty or nonexisting asset and how to download the channels after the asset is deleted...is there a testflow? 
Also you forgot to update one header.

nicolatimeus:
Hi @LeoNerdoG , basically the web ui allows you to perform changes like creating components and updating component configurations without committing the changes until the **Apply** button is pressed.
So by Asset with dirty configuration I mean an asset whose channels have been modified by the user but the changes have not been committed yet by using the **Apply** button. This can be acheived both in Wires composer and in the Drivers and Asset section.
By nonexisting Asset I mean an Asset that has not been created on the device yet, it can be obtained by creating a new Asset in the wire composer without committing the change using the **Apply** button. This can be obtained only in the Wire composer but not in the Drivers and Asset section, since in the latter case Assets are created on the device directly.

LeoNerdoG:
Hey @nicolatimeus,
thank you for the clarification. I am confirming that this PR is OK and can be merged into develop.

MMaiero (approved):

PR #3012, opened 2020-07-30
Added min, max, median and square root to the math singleport components

### LINKED ISSUE #3011: More Singleport Mathematical components
It would be nice to add minimum, maximum, median and square root to the single port wire mathematical components.

iacco:
Signed-off-by: Luca Iaconissi <luca.iaconissi@gmail.com>

**Related Issue:** This PR fixes/closes #3011


kura/examples/org.eclipse.kura.example.wire.math.singleport.provider/META-INF/MANIFEST.MF
```diff
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: Singleport Math Wire Component Provider
 Bundle-SymbolicName: org.eclipse.kura.example.wire.math.singleport.provider;singleton:=true
-Bundle-Version: 1.0.300.qualifier
+Bundle-Version: 1.0.400.qualifier
```
cdealti: Why?

kura/examples/org.eclipse.kura.example.wire.math.singleport.provider/OSGI-INF/MaximumComponent.xml
```diff
@@ -0,0 +1,41 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+    Copyright (c) 2018 Eurotech and/or its affiliates and others
```
cdealti: Fix date. Do the same for the others too

kura/examples/org.eclipse.kura.example.wire.math.singleport.provider/OSGI-INF/metatype/org.eclipse.kura.wire.Minimum.xml
```diff
@@ -0,0 +1,51 @@
...
+
+     All rights reserved. This program and the accompanying materials
+     are made available under the terms of the Eclipse Public License v1.0
+     which accompanies this distribution, and is available at
+     http://www.eclipse.org/legal/epl-v10.html
+-->
+<MetaData xmlns="http://www.osgi.org/xmlns/metatype/v1.2.0" localization="en_us">
+    <OCD id="org.eclipse.kura.wire.Minimum" 
+         name="Minimum" 
+         description="A wire component that finds the Minimum on a running window of samples for a configurable input property.">
```
cdealti: Please get the habit to explain the details, for example

"A wire component that finds the minimum value of a numeric property in a fixed size buffer containing the last received values, emitting the result. For comparison, values are promoted to double which is also the type of the result."

kura/examples/org.eclipse.kura.example.wire.math.singleport.provider/OSGI-INF/metatype/org.eclipse.kura.wire.SquareRoot.xml
```diff
@@ -0,0 +1,43 @@
...
+    Copyright (c) 2018 Eurotech and/or its affiliates and others
+
+     All rights reserved. This program and the accompanying materials
+     are made available under the terms of the Eclipse Public License v1.0
+     which accompanies this distribution, and is available at
+     http://www.eclipse.org/legal/epl-v10.html
+-->
+<MetaData xmlns="http://www.osgi.org/xmlns/metatype/v1.2.0" localization="en_us">
+    <OCD id="org.eclipse.kura.wire.SquareRoot" 
+         name="SquareRoot"
```
cdealti: "Square Root"

kura/examples/org.eclipse.kura.example.wire.math.singleport.provider/OSGI-INF/metatype/org.eclipse.kura.wire.SquareRoot.xml
```diff
@@ -0,0 +1,43 @@
...
+
+     All rights reserved. This program and the accompanying materials
+     are made available under the terms of the Eclipse Public License v1.0
+     which accompanies this distribution, and is available at
+     http://www.eclipse.org/legal/epl-v10.html
+-->
+<MetaData xmlns="http://www.osgi.org/xmlns/metatype/v1.2.0" localization="en_us">
+    <OCD id="org.eclipse.kura.wire.SquareRoot" 
+         name="SquareRoot" 
+         description="A wire component that performs the Square Root of a configurable input property.">
```
cdealti: do not capitalize words in the description

kura/examples/org.eclipse.kura.example.wire.math.singleport.provider/pom.xml
```diff
@@ -18,7 +18,7 @@
     </parent>
 
     <artifactId>org.eclipse.kura.example.wire.math.singleport.provider</artifactId>
-    <version>1.0.300-SNAPSHOT</version>
+    <version>1.0.400-SNAPSHOT</version>
```
cdealti: same as above

kura/examples/org.eclipse.kura.example.wire.math.singleport.provider/src/main/java/org/eclipse/kura/example/wire/math/singleport/maximum/MaximumComponent.java
```diff
@@ -0,0 +1,36 @@
...
+    protected void init() {
+        this.runningExtremes = null;
+    }
+
+    @Override
+    public TypedValue<?> apply(TypedValue<?> t) {
+        if (runningExtremes == null) {
+            this.runningExtremes = new RunningExtremes(this.options.getWindowSize());
+        }
+        final double value = ((Number) t.getValue()).doubleValue();
```
cdealti, line 35: I'd rather prefer checking the `DataType` and returning an `NaN` double value if the input is not numeric. This is to prevent log flooding if the input is not numeric

kura/examples/org.eclipse.kura.example.wire.math.singleport.provider/OSGI-INF/metatype/org.eclipse.kura.wire.Maximum.xml
```diff
@@ -0,0 +1,51 @@
...
+            name="result.name"
+            type="String" cardinality="0"
+            required="true"
+            default="result"
+            description="Specifies the name of the result property in emitted envelope.">
+        </AD>
+        
+        <AD id="window.size"
+            name="window.size"
+            type="Integer" cardinality="0"
```
cdealti, line 39: Add minimum

kura/examples/org.eclipse.kura.example.wire.math.singleport.provider/OSGI-INF/metatype/org.eclipse.kura.wire.Median.xml
```diff
@@ -0,0 +1,51 @@
...
+            required="true"
+            default="result"
+            description="Specifies the name of the result property in emitted envelope.">
+        </AD>
+        
+        <AD id="window.size"
+            name="window.size"
+            type="Integer" cardinality="0"
+            required="true"
+            default="10"
```
cdealti, line 42: Add minimum

kura/examples/org.eclipse.kura.example.wire.math.singleport.provider/OSGI-INF/metatype/org.eclipse.kura.wire.Minimum.xml
```diff
@@ -0,0 +1,51 @@
...
+            required="true"
+            default="result"
+            description="Specifies the name of the result property in emitted envelope.">
+        </AD>
+        
+        <AD id="window.size"
+            name="window.size"
+            type="Integer" cardinality="0"
+            required="true"
+            default="10"
```
cdealti, line 42: Add minimum

kura/examples/org.eclipse.kura.example.wire.math.singleport.provider/src/main/java/org/eclipse/kura/example/wire/math/singleport/RunningExtremes.java
```diff
@@ -0,0 +1,64 @@
...
+import java.util.ArrayDeque;
+
+public class RunningExtremes {
+
+    private final ArrayDeque<Double> window;
+    private final int windowSize;
+
+    private double max;
+    private double min;
+    private double last;
```
cdealti: you don't actually need this field

kura/examples/org.eclipse.kura.example.wire.math.singleport.provider/src/main/java/org/eclipse/kura/example/wire/math/singleport/RunningExtremes.java
```diff
@@ -0,0 +1,64 @@
...
+        if (value > this.max) {
+            this.max = value;
+        }
+        if (value < this.min) {
+            this.min = value;
+        }
+    }
+
+    private void updateExtremes() {
+        double localMax = this.window.getLast();
```
cdealti: why do you need local variables?

kura/examples/org.eclipse.kura.example.wire.math.singleport.provider/src/main/java/org/eclipse/kura/example/wire/math/singleport/RunningExtremes.java
```diff
@@ -0,0 +1,64 @@
...
+    private double min;
+    private double last;
+
+    public RunningExtremes(final int windowSize) {
+        this.window = new ArrayDeque<>(windowSize);
+        this.windowSize = windowSize;
+    }
+
+    private void updateWindow(double value) {
+        if (this.window.size() >= windowSize) {
```
cdealti: This needs to be refactored.

1. The buffer is at maximum capacity. The oldest element is removed and stored in a local variable
2. The input value is added to the buffer
3. if the oldest element was removed from the buffer (i.e. not null) AND is the current min or max, the current min and max are recomputed on the whole buffer (which contains the input value)
4. otherwise the current min and max are updated comparing them with the input value

It is almost the same as your implementation but not exactly

kura/examples/org.eclipse.kura.example.wire.math.singleport.provider/src/main/java/org/eclipse/kura/example/wire/math/singleport/RunningExtremes.java
```diff
@@ -0,0 +1,64 @@
...
+    private double max;
+    private double min;
+    private double last;
+
+    public RunningExtremes(final int windowSize) {
+        this.window = new ArrayDeque<>(windowSize);
+        this.windowSize = windowSize;
+    }
+
+    private void updateWindow(double value) {
```
cdealti: Use `final` consistently; use the last Kura formatter

kura/examples/org.eclipse.kura.example.wire.math.singleport.provider/OSGI-INF/metatype/org.eclipse.kura.wire.GainOffset.xml
```diff
@@ -16,7 +16,7 @@
             name="configuration"
             type="String" cardinality="0"
             required="true"
-            default="toBeMultipliedByTwo | 2&#13;toBeMultipliedBy3AndIncreasedBy1 | 3 | 1"
+            default="toBeMultipliedByTwo | 2&#10;toBeMultipliedBy3AndIncreasedBy1 | 3 | 1"
```
cdealti, line 25: Try avoiding unrelated changes or submit them in a separate PR

kura/examples/org.eclipse.kura.example.wire.math.singleport.provider/src/main/java/org/eclipse/kura/example/wire/math/singleport/AbstractSingleportMathComponent.java
```diff
@@ -116,5 +116,4 @@ public void onWireReceive(WireEnvelope wireEnvelope) {
                     .singletonList(new WireRecord(Collections.singletonMap(this.options.getResultName(), result))));
         }
     }
-
```
cdealti, line 122: Try avoiding unrelated changes or submit them in a separate PR

kura/examples/org.eclipse.kura.example.wire.math.singleport.provider/src/main/java/org/eclipse/kura/example/wire/math/singleport/RunningMedian.java
```diff
@@ -0,0 +1,35 @@
...
+        if (this.window.size() >= windowSize) {
+            this.window.removeFirst();
+        }
+        this.window.addLast(value);
+        double[] values = new double[windowSize];
+        for (Double windowValue : window) {
+            values[index++] = windowValue.doubleValue();
+        }
+        Arrays.sort(values);
+        if (windowSize % 2 == 0) {
```
cdealti: The buffer may not be at full capacity. Did you mean `window.size()`?

kura/examples/org.eclipse.kura.example.wire.math.singleport.provider/src/main/java/org/eclipse/kura/example/wire/math/singleport/RunningMedian.java
```diff
@@ -0,0 +1,35 @@
...
+    }
+
+    public double updateAndGetMedian(double value) {
+        int index = 0;
+        if (this.window.size() >= windowSize) {
+            this.window.removeFirst();
+        }
+        this.window.addLast(value);
+        double[] values = new double[windowSize];
+        for (Double windowValue : window) {
```
cdealti: If the size of the queue is large, it would be more efficient to maintain a sorted collection field like a `TreeSet` as the cost of insertion and removal is log(n) instead of n log(n) of `Arrays.sort`. 

In general, `Collection`S provide `toArray` methods to avoid copying the element yourself
Use https://docs.oracle.com/javase/8/docs/api/java/util/ArrayDeque.html#toArray(T[])

cdealti (changes requested):
I've stopped the review before the test.
Before requesting another review make also sure you
1. Add Copyright headers with correct date and without mention to contributors in all new files
2. Run latest version of Kura formatted and cleanup
3. Apply requested changes to all parts where they apply (I did not repeat the same comment everyware)

kura/examples/org.eclipse.kura.example.wire.math.singleport.provider/src/main/java/org/eclipse/kura/example/wire/math/singleport/RunningExtremes.java
iacco: The usage of "last" has been thought in order to reduce computation for this component.

The value being extracted from the array is checked in order to understand if it was either the maximum or the minimum; if it is not, there is no need to update these two values, which can then be checked against the new value being inserted in the array.
Assuming my maximum in the window is 5 and my minimum is 1, if I extract a value of 2, I can still keep those two variables without re-check the whole window. Then I can check the new value against reliable local maximums and minimums. I reckon with a window of quite a lot of elements, and a quick running Wire Graph application, this can help.

It is of course feasible to implement this without this check, but at each iteration the component will have to check the whole array, which is pretty much a waste of resources.

kura/examples/org.eclipse.kura.example.wire.math.singleport.provider/OSGI-INF/metatype/org.eclipse.kura.wire.GainOffset.xml
iacco: Without this change, tests can either be ineffective/useless or will not work for this component.
To me it does not make sense to either:

- push tests only for the new components
- push tests also for the old components, but which fail because of the way tests work
- push tests also for the old components, which do not actually perform useful tests

kura/examples/org.eclipse.kura.example.wire.math.singleport.provider/src/main/java/org/eclipse/kura/example/wire/math/singleport/RunningExtremes.java
iacco: I am already using the formatter and cleanup profiles contained in kura/kura/setups/formatting/

Can you please explain what do you mean with "use final consistently"?

kura/examples/org.eclipse.kura.example.wire.math.singleport.provider/META-INF/MANIFEST.MF
iacco: noted.

kura/examples/org.eclipse.kura.example.wire.math.singleport.provider/OSGI-INF/MaximumComponent.xml
iacco: noted.

kura/examples/org.eclipse.kura.example.wire.math.singleport.provider/OSGI-INF/metatype/org.eclipse.kura.wire.Minimum.xml
iacco: noted.

kura/examples/org.eclipse.kura.example.wire.math.singleport.provider/OSGI-INF/metatype/org.eclipse.kura.wire.SquareRoot.xml
iacco: noted.

kura/examples/org.eclipse.kura.example.wire.math.singleport.provider/OSGI-INF/metatype/org.eclipse.kura.wire.SquareRoot.xml
iacco: noted.

kura/examples/org.eclipse.kura.example.wire.math.singleport.provider/pom.xml
iacco: noted.

kura/examples/org.eclipse.kura.example.wire.math.singleport.provider/OSGI-INF/metatype/org.eclipse.kura.wire.Maximum.xml
iacco: noted.

kura/examples/org.eclipse.kura.example.wire.math.singleport.provider/OSGI-INF/metatype/org.eclipse.kura.wire.Median.xml
iacco: noted.

kura/examples/org.eclipse.kura.example.wire.math.singleport.provider/OSGI-INF/metatype/org.eclipse.kura.wire.Minimum.xml
iacco: noted.

kura/examples/org.eclipse.kura.example.wire.math.singleport.provider/src/main/java/org/eclipse/kura/example/wire/math/singleport/RunningExtremes.java
iacco: noted.

kura/examples/org.eclipse.kura.example.wire.math.singleport.provider/src/main/java/org/eclipse/kura/example/wire/math/singleport/RunningMedian.java
iacco: noted.

kura/examples/org.eclipse.kura.example.wire.math.singleport.provider/src/main/java/org/eclipse/kura/example/wire/math/singleport/maximum/MaximumComponent.java
iacco: noted.

kura/examples/org.eclipse.kura.example.wire.math.singleport.provider/src/main/java/org/eclipse/kura/example/wire/math/singleport/RunningExtremes.java
iacco: noted?

kura/examples/org.eclipse.kura.example.wire.math.singleport.provider/src/main/java/org/eclipse/kura/example/wire/math/singleport/RunningMedian.java
iacco: Set cannot be used because for the nature of it, you cannot store multiple entries with the same value.

iacco:
Ready for the next review.

kura/examples/org.eclipse.kura.example.wire.math.singleport.provider/src/main/java/org/eclipse/kura/example/wire/math/singleport/RunningExtremes.java
cdealti: `final double value`

kura/examples/org.eclipse.kura.example.wire.math.singleport.provider/src/main/java/org/eclipse/kura/example/wire/math/singleport/RunningMedian.java
cdealti: It can be done with a TreeSet, provided the element contains the value and a monotonic sequence number attached to the value. This would allow for elements with the same value (but a different sequence number).

The complexity of this solution is O(n) as, to get the "central" elements in the TreeSet requires using an iterator (insertion and removal have complexity O(log n))

iacco:
attempted modification to travis configuration in order to solve build issues

cdealti (approved):

cdealti:
@MMaiero this contribution exceeds 1kLOC

LeoNerdoG:
Hey @iacco,
I have started checking this PR and I have several questions about Min and Max component (Sqrt is OK). 
The problem is the same on both components, so I will onyl describe it for Min component. 

Becasue this component has only one input (probably becasue you can have a lot of input values in array style I guess), I have a problem with creating such an array/buffer with Wire components. 
Try 1: Using JS filter with code:

> var outRecord = newWireRecord()
> var byteArray = newByteArray(2) 
> byteArray[0] = 1
> byteArray[1] = 2
> outRecord.byteArrayTest = newByteArrayValue(byteArray)
> output.add(outRecord)
In this case I get "Not a number" error (see screenshot bellow):
Screenshot 2020-09-25 at 09 57 14

Try 2: using "Join" component does not work, becasue I still have to set the input parameter of one of the inputs in "Join" component. I also get "Missing operand error in this case: 
Screenshot 2020-09-25 at 13 50 54
The code used for JS components in this case is:
> var outRecord = newWireRecord()
outRecord.myValue = newIntegerValue(38)
output.add(outRecord)

I was looking at [this document which describes TypedValues](https://download.eclipse.org/kura/docs/api/3.0.0/apidocs/org/eclipse/kura/type/TypedValues.html#newByteArrayValue-byte:A-) but found nothing "usefull", because there is no way to create array of integers. If I create array of byte values, I get error as described above. 

Do you have any ideas or suggestions how to tackle this?

iacco:
Hey @LeoNerdoG,

unfortunately this components keeps by itself a running buffer, so it processes single numeric property in a sampling window which progresses every time an envelope is received. When the sampling window is full, it will discard the oldest value.

Because of this, the scenario you are describing is not the scenario this component was developed for.

You could however implement a solution like:
script filter A, connected to both Min/Max component and another script filter B.
both the Min/Max and script filter B then could be connected to a Join.

```
A -> Min/Max -> Join --> etc.
  |          |
   ->B  -----      

```

Implementation wise:
**Script Filter A**: take the array in input most likely as multiple properties with different names (e.g. prop1, prop2) or in a bytearray, iterate in the array and emit every property as a single property (e.g. **value**) in an envelope.
At the end, emit an envelope with a different property (e.g. **done**) with value true. 

**Script Filter B**: ignore messages containing property value. If a message with value **done** is received emit.

**Min/Max**: specify **exactly** the array size

**Join**: use barrier.

What will happen is that:
- Script filter A will emit all the messages to a MIN/MAX and perform the operation. Each message before the last one is discarded by the barriered Join, only the last will be kept
- Script filter B will wait for script A to be done to publish, enabling the Join (and therefore providing the actual MIN/MAX of the array)

Let me know your thoughts.

With Best Regards,
iacco

LeoNerdoG:
Heu @iacco,
thanks for all explanations!
I have checked this PR and it it looks OK, so from my standpoint it can be merged.

MMaiero:
The CQ has passed the first check: https://dev.eclipse.org/ipzilla/show_bug.cgi?id=22765
As reported, the code can be now merged but the contribution will be further checked later on.

PR #3038, opened 2020-08-19
Restoring the previous regex validation.

### LINKED ISSUE #2954: Wireless Password Not Parsed Correctly
If your Raspebrry is set as Access Point and your wireless password contains symbols such as ⁄‹›€ı–°·‚·±_‚ˇ¿ˇÈ~¿}{|Ææ etc. Kura reports an error "Password must be 8-63 characters" although the password has the required length (it looks like that very special symbols are not properly parsed). 

**Testflow:**
2. Go to network -> wlan0 -> wireless
3. Set Wireless Mode to Access Point
4. Enter a valid SSID and a password that contains one fo the upper symbols
5. Observe the error.

**Expected behavior**
Either the symbols mentioned should not be allowed (changed error message) or the field should be properly parsed. 

**Screenshots**
Screenshot 2020-06-18 at 15 31 45

**Target Environment:**
Board: Raspberry Pi 3
Kura version: 4.2.0_SNAPSHOT ([Link to the specific Kura version -from June 16th, 2020](https://ci.eclipse.org/kura/job/kura-develop/lastSuccessfulBuild/artifact/kura/distrib/target/kura_4.2.0-SNAPSHOT_raspberry-pi-2-3_installer.deb)

MMaiero:
The GWT implementation seems to be incompatible with the regexes defined for wifi and
have some known limitations.

Closes #2954


MMaiero:
@mstankovic Can you give this a look? I believe we should have some network tests to validate different type of passwords. In this way we can prevent regressions in the wireless part.

mstankovic:
@MMaiero sure, i will run tests now and let you know if there are some problems

mstankovic:
@MMaiero nope, we dont have tests for different type of passwords..
I run all networking tests (wifi) and all have passed.
also i checked #2954 and iam still reproducing issue described: field is invalid for password: "testKEYS€"

MMaiero:
The regex defined is the following: `^[ -~]{8,63}$`
and means that all the characters from space to ~ are accepted and there should be a minimum of 8 and a maximum of 63 chars.
Looking here: http://www.asciitable.com the € character is not in that set. That is the reason why it is rejected.

I don't know if at linux level the '€' character is allowed but at least it is not for our current implementation.
I believe we can improve the message. Maybe with something like: "Incorrect password. Password must be 8-63 characters and with ASCII characters from code 32 to 126."

mstankovic:
ok. in that case you can merge this PR.

MMaiero:
I ended up with something like this:

Screenshot 2020-08-21 at 14 20 56

nicolatimeus (approved):

PR #3042, opened 2020-08-21
Fixes case where a read for all channels fails and reports empty value.

### LINKED ISSUE #3044: Incorrect Asset status report from cloud call
**Describe the bug**
When a read for all the channels is requested by the cloud, if an error happens, for example the driver is not able to reach the target PLC, the catcher error is not reported, resulting in an empty list of channels:

```
[
  {
    "name": "modbus_asset_1",
    "channels": [
      {
        "name": "Channel-1",
        "timestamp": 1598004632308,
        "value": "4",
        "type": "INTEGER"
      },
      {
        "name": "Channel-2",
        "timestamp": 1598004632309,
        "value": "6",
        "type": "INTEGER"
      },
      {
        "name": "Channel-3",
        "timestamp": 1598004632310,
        "error": "Unknown error"
      }
    ]
  },
  { "name": "modbus_asset_2", "channels": [] },
  {
    "name": "modbus_asset_3",
    "channels": [
      {
        "name": "Channel-1",
        "timestamp": 1598004632348,
        "value": "4",
        "type": "INTEGER"
      }
    ]
  }
]
```

Asset 1 and 3 correctly report even if there is an error in a specific channel.
For asset 2, that has connection issues, what is reported is an empty list

**To Reproduce**
1. Define an asset that has a driver, for example Modbus, with a wrong target IP configured. It will return usually, also via local ESF REST call a connection error 
2. Using the cloud platform request a read for all the channels of the given asset
3. See error

**Expected behavior**
The result reported to the cloud platform should state that an error happened.

**Screenshots**
N/A

 - Board [e.g. Raspberry Pi 3]
 - OS version: [e.g. provide the result of `uname -a`]
 - Additional info: 

**Additional context**
N/A

MMaiero:

### DIFF
--- a/kura/org.eclipse.kura.asset.cloudlet.provider/src/main/java/org/eclipse/kura/internal/asset/cloudlet/AssetCloudlet.java
+++ b/kura/org.eclipse.kura.asset.cloudlet.provider/src/main/java/org/eclipse/kura/internal/asset/cloudlet/AssetCloudlet.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2016, 2018 Eurotech and/or its affiliates and others
+ * Copyright (c) 2016, 2020 Eurotech and/or its affiliates and others
  *
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
--- a/kura/org.eclipse.kura.asset.cloudlet.provider/src/main/java/org/eclipse/kura/internal/asset/cloudlet/AssetCloudlet.java
+++ b/kura/org.eclipse.kura.asset.cloudlet.provider/src/main/java/org/eclipse/kura/internal/asset/cloudlet/AssetCloudlet.java
@@ -22,12 +22,15 @@ import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Optional;
 import java.util.Set;
+import java.util.stream.Collectors;
 
 import org.eclipse.kura.KuraErrorCode;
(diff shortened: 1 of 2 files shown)


nicolatimeus (approved):

MMaiero:
Closes #3044

MMaiero:
@mstankovic Can you please validate this PR?

kura/org.eclipse.kura.asset.cloudlet.provider/src/main/java/org/eclipse/kura/internal/asset/cloudlet/AssetCloudlet.java
```diff
@@ -22,12 +22,15 @@
 import java.util.Map.Entry;
 import java.util.Optional;
 import java.util.Set;
+import java.util.stream.Collectors;
```
LeoNerdoG, line 25: Please change header to "2020".

LeoNerdoG (changes requested):
hey @MMaiero, please update the header in one of the files.

kura/org.eclipse.kura.asset.cloudlet.provider/src/main/java/org/eclipse/kura/internal/asset/cloudlet/AssetCloudlet.java
MMaiero: Done

LeoNerdoG (approved):

PR #3049, opened 2020-08-27
Updated kura exception messages

### LINKED ISSUE #3021: KuraException with some KuraErrorCode are not printed properly.
Hi Kura Team,

**Describe the bug**
While trying to log some `KuraException`, a simple `System.out.println` or a logging of that exception are failing to retrieve the associated message:

`
WARN  o.e.k.KuraException - Could not load Exception Messages Bundle for Locale en_US
org.eclipse.kura.KuraException: Generic Error - INVALID_PARAMETER: {1}
	at org.eclipse.kura.core.configuration.ConfigurationServiceImpl.createFactoryConfiguration(ConfigurationServiceImpl.java:409)
...
`

The error code here is `INVALID_PARAMETER`.

After investigation, it seams that properties into `KuraExceptionMessagesBundle.properties` from bundle `org.eclipse.kura.api` are not in line with the enums defined into `enum KuraErrorCode` of that same bundle.

**Expected behavior**
log a pre-defined error message despite the "GenericError"'s one.

**Target Environment :**
 - KuraEmulator (for now)

Thank you all for your work!

pierantoniomerlino:
Closes #3021 


MMaiero:
Is it worth having some arguments for those added messages? For example for INVALID_PARAMETER maybe we can post what the parameters is? Just thinking out loud...

kura/org.eclipse.kura.protocol.can/src/main/java/org/eclipse/kura/protocol/can/CanConnectionServiceImpl.java
```diff
@@ -64,7 +64,7 @@ public void disconnectCanSocket() throws IOException {
     @Override
     public void sendCanMessage(String ifName, int canId, byte[] message) throws KuraException, IOException {
         if (message.length > 8) {
-            throw new KuraException(KuraErrorCode.OPERATION_NOT_SUPPORTED, "CAN send : Incorrect frame length");
+            throw new KuraException(KuraErrorCode.INVALID_PARAMETER, "CAN send : Incorrect frame length");
```
MMaiero, line 67: Is this a sort of breaking change?

pierantoniomerlino: Mhhh... why? The method `sendCanMessage` is still throwing a Kura Exception...

MMaiero: But if someone has done some logic on top of the error code, he could be in trouble, I think

pierantoniomerlino: Yes, this is true. I can revert the modification if you want, but the invalid parameter exception seems to be more appropriate for me.

MMaiero: the other way around is to note the breaking change updating versions

MMaiero (approved):

PR #3037, opened 2020-08-19
Added reset confirmation in settings tabs

MMaiero:
Signed-off-by: Maiero <matteo.maiero@eurotech.com>

Closes #3002

PR #3050, opened 2020-08-28
Fix missing validation in device certs tab

### LINKED ISSUE #2719: SSL Fields Stay in Error State Although Fields Reset
Although user resets all the fields or changes the views without saving changes, all the fields are marked red. 

**Testflow:**
2. Go to Settings -> try to change anyt of the SSL fields
3. After you enter something in the fields, erase it and press Reset
4. Observe the fields that are marked in red

**Expected behavior**
After the Reset button is pressed, all the fields should be stripped of the red makration. 

**Screenshots**
Screenshot 2020-03-25 at 13 42 31

Board: Raspberry Pi 3
Additional info: /

pierantoniomerlino:
This PR adds the validation for the private key field in the SSL Device certificate tab.

**Related Issue:** This PR fixes/closes #2719 

### DIFF
--- a/kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/client/ui/settings/CommandUserTabUi.java
+++ b/kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/client/ui/settings/CommandUserTabUi.java
@@ -151,13 +151,15 @@ public class CommandUserTabUi extends AbstractServicesUi implements Tab {
 
     @Override
     public void clear() {
-        reset();
+        restoreConfiguration(originalConfig);
+        renderForm();
+        setButtonsEnabled(false);
+        setDirty(false);
     }
 
     private void apply() {
         if (isValid()) {
             if (isDirty()) {
-                // TODO ask for confirmation first
                 Modal modal = new Modal();
                 modal.setClosable(false);
                 modal.setFade(true);
@@ -219,8 +221,7 @@ public class CommandUserTabUi extends AbstractServicesUi implements Tab {
                                         public void onSuccess(Void result) {
                                             modal.hide();
                                             logger.info(MSGS.info() + ": " + MSGS.deviceConfigApplied());
-                                            CommandUserTabUi.this.apply.setEnabled(false);
-                                            CommandUserTabUi.this.reset.setEnabled(false);
+                                            setButtonsEnabled(false);
                                             setDirty(false);
                                             CommandUserTabUi.this.originalConfig = CommandUserTabUi.this.configurableComponent;
                                             EntryClassUi.hideWaitModal();
--- a/kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/client/ui/settings/DeviceCertsTabUi.java
+++ b/kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/client/ui/settings/DeviceCertsTabUi.java
@@ -36,8 +36,10 @@ import org.gwtbootstrap3.client.ui.constants.ValidationState;
 import org.gwtbootstrap3.client.ui.html.Span;
 
 import com.google.gwt.core.client.GWT;
+import com.google.gwt.event.dom.client.BlurEvent;
 import com.google.gwt.uibinder.client.UiBinder;
 import com.google.gwt.uibinder.client.UiField;
+import com.google.gwt.uibinder.client.UiHandler;
 import com.google.gwt.user.client.rpc.AsyncCallback;
 import com.google.gwt.user.client.ui.Composite;
 import com.google.gwt.user.client.ui.HTMLPanel;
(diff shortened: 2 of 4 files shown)


pierantoniomerlino:
@LeoNerdoG Could you please check this?

mstankovic:
I'm confirming that issue is fixed. Also tried to reproduce issues from #3037 and was unable to.
qa validation passed.

LeoNerdoG:
Hey @pierantoniomerlino,
the problem persists. If you change focus to one of the mandatory fields, enter nothing and swtich views, there is no prompt for inclomplete fields - can you check again please?

pierantoniomerlino:
Hi @LeoNerdoG, 
this is my testflow:
1. click on the Settings -> Device SSL Certificate tab
2. click on the Storage Alias field but not write nothing
3. click outside the field (that will be colored in red)
4. click on another tab (i.e. Command User)

Do you expect that the user will be prompted with an alert for incomplete fields?

LeoNerdoG:
Hey @pierantoniomerlino,
Yes, if any field is in error state, user should receive at least a notification that there is a dirty field (only because of focus change?)? 
if this is not the case, then this PR is OK, becasue I don't know exactly what is the desired behaviour here.

pierantoniomerlino:
Yes, IMHO the behavior is correct as is now. The user has only clicked on a field without any data, so he shouldn't be notified for some missing data. 

Thanks for the report!

MMaiero:
@LeoNerdoG Can you have a look at the latest commit? I think it should improve a bit the experience

MMaiero (approved):

LeoNerdoG:
Hey @pierantoniomerlino,
I have checekd this PR and found one more issue, everything else is fine. 
This happens on all cert tabs. 

if you anter something in some mandatory fields (e.g. in Storage alias), click outside, then deleting the contents and click Reset, the "Field cannot be blank" text does not disappear, but is instead grayed out (see screenshot bellow). Can you please check? 

After this user can change views without any prompt, so the problem is that there is an grayed out error message in the GUI which can be confusing for end user. 

Screenshot 2020-09-03 at 08 52 04

pierantoniomerlino:
Yeah, I'm aware of that issue... but I've no idea how to fix it! :-(

LeoNerdoG:
One wise programmer once said to me: "if you wait long enough, the problem will solve itself!"
Jokes aside, maybe we should ask others for some new ideas...?

pierantoniomerlino:
> One wise programmer once said to me: "if you wait long enough, the problem will solve itself!"

It's my current approach! :-D
Yes, I'll ask to others for fixing this...

pierantoniomerlino:
@LeoNerdoG I waited long enough... now the issue seems solved :-D

LeoNerdoG:
Hey @pierantoniomerlino,
I have checked this again and it does not seem to be resolved, sorry 😸 
Testflow:
1. Click on a free entry field (e.g. "Storage Alias")
2. Click outside the field (observe error)
3. Try to change tabs - no error message is shown (see GIF bellow). 


 
Can you pleae check if you can reproduce it?

pierantoniomerlino:
@LeoNerdoG Yes, the behavior is exactly that. I thought that, indeed, the user didn't fill any data, so why the webUi has to alert that something is changed? 
If user added some data and deleted it, instead, the webUI shows the alert because the user has modified the fields. What do you think?

LeoNerdoG:
Hey @pierantoniomerlino,
I would expect that this behaviour would be unified in a way, that if user clicks on a mandatory field (change of focus) and then wants to change views, Kura will show an error (so this behaviour is exaclty the same as if user fills the field with data and afterwards deletes everything, without saving it first); If you think current behaviour is OK, cool. I just think that it is not logical from end-user point of view to have an erroneous field and to be able to change views...it is strange. 
Nonetheless, this is not a critical issue and can be closed as far as I am concerned, becasue if you wanted to achieve this, then I guess the fix is ok! 😄

pierantoniomerlino:
Hi @LeoNerdoG, I understand your point about the user-experience. So, I changed a bit the fields behavior. Please take a look on it.

LeoNerdoG:
Hey @pierantoniomerlino,
I have checked this and now it seems OK!

MMaiero (approved):

MMaiero:
I have done a small commit to clean a bit the implementation. Could you please check @pierantoniomerlino @LeoNerdoG?

pierantoniomerlino:
@MMaiero Yes, I had promised myself to leave only the `.reset()` methods, but I forgot it! So, the modification is fine  for me.

LeoNerdoG:
Hey @MMaiero,
I have checked this again and found a bug in "Command User" tab (I don't know if it was there before or not, I just noticed it now). 
Testflow:
1. Delete the default command username from the fields(kura)
2. Leave the field blank, click on other tabs without saving 
3. Observe how confirmation box opens twice


can you please check?

MMaiero:
@LeoNerdoG I have made the change. Please have a look. If ok, I'm going to merge this

LeoNerdoG:
Hey @MMaiero,
it seems Ok, it can be merged.

PR #3059, opened 2020-09-03
Pr/trigonometry

### LINKED ISSUE #3055: Trigonometry components
Is your feature request related to a problem? Please describe.
This feature would add to the Kura examples a singleport component which allows to perform trigonometric operations (sin, cos, tan, etc.) onto the Wire Graph.

iacco:
PR to implement a Wire Component which can handle trigonometric calculations

**Related Issue:** This PR fixes/closes #3055


kura/examples/org.eclipse.kura.example.wire.trigonometric.functions.provider/META-INF/MANIFEST.MF
```diff
@@ -0,0 +1,19 @@
+Manifest-Version: 1.0
+Bundle-ManifestVersion: 2
+Bundle-Name: Trigonometric Operations for Wires
```
cdealti: "Trigonometric Operations" -> "Trigonometric functions"

kura/examples/org.eclipse.kura.example.wire.trigonometric.functions.provider/META-INF/MANIFEST.MF
```diff
@@ -0,0 +1,19 @@
...
+Bundle-SymbolicName: org.eclipse.kura.example.wire.trigonometric.functions.provider;singleton:=true
```
cdealti: org.eclipse.kura.example.wire.trigonometric.functions.provider -> org.eclipse.kura.example.wire.math.trig

Please rename the project directory and the Java packages too

kura/examples/org.eclipse.kura.example.wire.trigonometric.functions.provider/OSGI-INF/metatype/org.eclipse.kura.wire.TrigonometricOperators.xml
```diff
@@ -0,0 +1,44 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- Copyright (c) 2020 Eurotech and/or its affiliates and others All rights 
+	reserved. This program and the accompanying materials are made available 
+	under the terms of the Eclipse Public License v1.0 which accompanies this 
+	distribution, and is available at http://www.eclipse.org/legal/epl-v10.html -->
+<MetaData xmlns="http://www.osgi.org/xmlns/metatype/v1.2.0"
+	localization="en_us">
+	<OCD id="org.eclipse.kura.wire.TrigonometricOperators"
+		name="Trigonometric Operations for Wires"
```
cdealti: "Trigonometric Operations for Wires" -> "Trigonometric Functions for Wires"

kura/examples/org.eclipse.kura.example.wire.trigonometric.functions.provider/OSGI-INF/metatype/org.eclipse.kura.wire.TrigonometricOperators.xml
```diff
@@ -0,0 +1,44 @@
...
+		description="A component that applies a trigonometric function to a numeric property received in input in the component. The result will be emitted as a new property of type double.">
+
+		<AD id="operand.name" name="Operand Name" type="String"
+			cardinality="0" required="true" default="operand"
+			description="Specifies the name of the operand property in the received envelope.">
+		</AD>
+
+		<AD id="trigonometric.operator" name="Operator" type="String"
```
cdealti: `id`: "trigonometric.operator" -> "trigonometric.function".
`name`: Function

kura/examples/org.eclipse.kura.example.wire.trigonometric.functions.provider/OSGI-INF/metatype/org.eclipse.kura.wire.TrigonometricOperators.xml
```diff
@@ -0,0 +1,44 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- Copyright (c) 2020 Eurotech and/or its affiliates and others All rights 
+	reserved. This program and the accompanying materials are made available 
+	under the terms of the Eclipse Public License v1.0 which accompanies this 
+	distribution, and is available at http://www.eclipse.org/legal/epl-v10.html -->
+<MetaData xmlns="http://www.osgi.org/xmlns/metatype/v1.2.0"
+	localization="en_us">
+	<OCD id="org.eclipse.kura.wire.TrigonometricOperators"
+		name="Trigonometric Operations for Wires"
+		description="A component that applies a trigonometric function to a numeric property received in input in the component. The result will be emitted as a new property of type double.">
```
cdealti: Please reword as

"A component that applies a trigonometric function to a numeric property converted to double and involving rounding. The result will be emitted as a new property of type double."

kura/examples/org.eclipse.kura.example.wire.trigonometric.functions.provider/OSGI-INF/metatype/org.eclipse.kura.wire.TrigonometricOperators.xml
```diff
@@ -0,0 +1,44 @@
...
+
+		<AD id="operand.name" name="Operand Name" type="String"
```
cdealti: A "function" accepts a "parameter". Please rename
`id`: parameter.name
`name`: Parameter Name

kura/examples/org.eclipse.kura.example.wire.trigonometric.functions.provider/src/main/java/org/eclipse/kura/example/wire/trigonometric/functions/provider/TrigonometricComponent.java
```diff
@@ -0,0 +1,128 @@
...
+
+    @Override
+    public void producersConnected(Wire[] wires) {
+        wireSupport.producersConnected(wires);
+    }
+
+    @Override
+    public void onWireReceive(WireEnvelope wireEnvelope) {
+        final Map<String, TypedValue<?>> properties = wireEnvelope.getRecords().get(0).getProperties();
+        final Double operand = Double.valueOf(properties.get(this.options.getOperandName()).getValue().toString());
```
cdealti: Please take a look to
https://docs.oracle.com/javase/8/docs/api/java/lang/Number.html#doubleValue--

kura/examples/org.eclipse.kura.example.wire.trigonometric.functions.provider/src/main/java/org/eclipse/kura/example/wire/trigonometric/functions/provider/TrigonometricComponent.java
```diff
@@ -0,0 +1,128 @@
...
+        case TAN:
+            return Math.tan(operand.doubleValue());
+        case ASIN:
+            return Math.asin(operand.doubleValue());
+        case ACOS:
+            return Math.acos(operand.doubleValue());
+        case ATAN:
+            return Math.atan(operand.doubleValue());
+        default:
+            return operand;
```
cdealti: 1. To avoid the switch, cannot you use the functional approach also used in your PR for the logical operators?
2. for the same reason, you should apply a default function

kura/examples/org.eclipse.kura.example.wire.trigonometric.functions.provider/OSGI-INF/TrigonometricComponents.xml
```diff
@@ -0,0 +1,42 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- Copyright (c) 2020 Eurotech and/or its affiliates and others All rights 
+	reserved. This program and the accompanying materials are made available 
+	under the terms of the Eclipse Public License v1.0 which accompanies this 
+	distribution, and is available at http://www.eclipse.org/legal/epl-v10.html -->
+<scr:component
+	xmlns:scr="http://www.osgi.org/xmlns/scr/v1.1.0"
+	name="org.eclipse.kura.wire.TrigonometricOperators" activate="activate"
```
cdealti: Rename according to the above request

kura/examples/test/org.eclipse.kura.example.wire.trigonometric.functions.provider.test/src/main/java/org/eclipse/kura/example/wire/trigonometric/functions/provider/test/TrigonometricComponentTest.java
```diff
@@ -0,0 +1,348 @@
...
+    private static final String EMIT_RECEIVED_PROPERTIES = "emit.received.properties";
+    private static final String TRIGONOMETRIC_OPERATION = "trigonometric.operator";
+
+    private static final String OPERAND_NAME_DEFAULT = "operand";
+    private static final String RESULT_NAME_DEFAULT = "result";
+    private static final boolean EMIT_RECEIVED_PROPERTIES_DEFAULT = false;
+    private static final String TRIGONOMETRIC_OPERATION_DEFAULT = "SIN";
+
+    // tolerance for errors with inverse trigonometric functions
+    private static final double ACCEPTABLE_ERROR = 0.0000000000000004;
```
cdealti: Can you please explain this?

cdealti (changes requested):

LeoNerdoG:
Hey @iacco,
I have build this branch but were unable to find the wire asset with trigonometric functions. 
Is this package/wire asset bundled into this or do I have to install a custom package from Eclipse marketplace in order to get this working?
When building the project I have selected "core-dp" and "raspebrry-pi-2-3" - should I select anything alse while building this?

kura/examples/test/org.eclipse.kura.example.wire.trigonometric.functions.provider.test/src/main/java/org/eclipse/kura/example/wire/trigonometric/functions/provider/test/TrigonometricComponentTest.java
iacco: Because of how floating point values are treated, and because of approximations, results of Math functions can slightly differ from the real values.
Different functions lead to slightly different errors, all less than the ACCEPTABLE_ERROR value.

If you take SIN of PI, expected result should be 0, while result is off slightly (1.2246467991473532E-16).

For tests' sake, instead of rounding, or performing the same mathematical operation with java.lang.Math and checking the value (IMHO pretty much useless), I preferred matching the values to the expected values of plus/minus one, zero or Pi fractions (using Math.PI), taking in account of this error.

iacco:
Hello @LeoNerdoG! 
same as per boolean and mathematical (and all of the examples), I think that they are not included in the main build.
In order to build them, you have to invoke maven using the pom in the "kura/examples" directory.

Also, if you check in the Travis configuration, you'll see that the examples are built separately after the main build.
Let me know if you need anything else.

iacco:
Applied changes, rebased repo.

Waiting for new review.

LeoNerdoG:
Hey @iacco,
thanks for the info - I wasn't paying attention when building. Nonetheless, I have successfully build the packages and am alreay testing this. I will try to finish this as soon as possible and will post issues (if I find any) into the comments section here.

kura/examples/test/org.eclipse.kura.example.wire.trigonometric.functions.provider.test/src/main/java/org/eclipse/kura/example/wire/trigonometric/functions/provider/test/TrigonometricComponentTest.java
cdealti: The accuracy of the java.lang.Math methods is documented (see https://docs.oracle.com/javase/8/docs/api/java/lang/Math.html#sin-double-)

> or performing the same mathematical operation with java.lang.Math and checking the value (IMHO pretty much useless)

why useless?

kura/examples/org.eclipse.kura.example.wire.trigonometric.functions.provider/src/main/java/org/eclipse/kura/example/wire/trigonometric/functions/provider/TrigonometricComponent.java
cdealti: Now the behavior no longer matches the description of the component. Other components in the code base convert a `Number` to `Double` while now this component requires a `Double`.

kura/examples/org.eclipse.kura.example.wire.math.trig/src/main/java/org/eclipse/kura/example/wire/math/trig/TrigonometricComponentOptions.java
```diff
@@ -0,0 +1,109 @@
...
+            return parameter -> Math.cos(parameter.doubleValue());
+        case TAN:
+            return parameter -> Math.tan(parameter.doubleValue());
+        case ASIN:
+            return parameter -> Math.asin(parameter.doubleValue());
+        case ACOS:
+            return parameter -> Math.acos(parameter.doubleValue());
+        case ATAN:
+            return parameter -> Math.atan(parameter.doubleValue());
+        default:
```
cdealti: Please reorder the last two cases so that `default` is the last one

cdealti:
Some review comments, even if outdated, must be resolved.

kura/examples/org.eclipse.kura.example.wire.trigonometric.functions.provider/src/main/java/org/eclipse/kura/example/wire/trigonometric/functions/provider/TrigonometricComponent.java
iacco: Tried the following:

final Double parameter = new Double(properties.get(this.options.getParameterName()).getValue().toString()).doubleValue();
--> Lint states *Use "Double.parseDouble" for this string-to-double conversion.*

final Double parameter = Double.parseDouble(properties.get(this.options.getParameterName()).getValue().toString());
--> Lint OK

final Double parameter = new Double(properties.get(this.options.getParameterName()).getValue().toString());
--> Lint OK

final Double parameter = (Double) properties.get(this.options.getParameterName()).getValue();
--> Lint OK

final Double parameter = ((Double) properties.get(this.options.getParameterName()).getValue()).doubleValue();
-->Lint states *Remove the unboxing from "Double".*

Please select the one that most suits you , and I will apply. If none of the above is good, suggest how it should be implemented.

kura/examples/test/org.eclipse.kura.example.wire.trigonometric.functions.provider.test/src/main/java/org/eclipse/kura/example/wire/trigonometric/functions/provider/test/TrigonometricComponentTest.java
iacco: First, to use the output values (like the 1.2246467991473532E-16 "zero" from my previous comment) is really not human friendly. 
If we consider that tests most of the time have the side effect of showing how the associated bundle works, human readibility in my opinion is important.

Second, by my taste to have something like the following is quite ugly:

> Double result = <output of the component, e.g. SIN of Pi)>
assertTrue(Math.sin(Math.PI) == result);

Let me know how you want it done.

cdealti (approved):

cdealti:
@MMaiero according to Github this contribution slightly exceeds 1kLOC. However, in the following comment Eclipse clarifies that

>FWIW, committers have some discretion with regard to what 1KLOC means. What we're really looking for is contributions of significant new intellectual property. These significant contributions need to be reviewed (and tracked) by the IP team for licensing, provenance, and general shenanigans. I tend to think of it as 1KLOC of meaningful content and "really close" counts (e.g., a contribution of 1,100 lines that includes file headers and blank lines)."

On top of that, this contribution includes OSGi tests that are not redistributed with the Kura runtime. If we require that a contributor develop tests it is likely that many contributions will exceed the limit.

LeoNerdoG:
Hey @iacco,
I have a question regarding this trigonometric element in wires. 
On screenshot bellow you can see my wire graph - I would like to know if "timer" is best option here for generating pulses? In my opinion it is not, but I could not find any other components that would produce these signals (not in generated examples nor on Eclipse marketplace). 
If this is the correct element for generating pulse, please let me know. I am not getting any output on the logger side. 

Screenshot 2020-09-22 at 13 51 31

iacco:
Dear @LeoNerdoG,

timer can be used, but of course since it only emits an epoch called TIMER, you would need to set the input parameter's name to TIMER. You can check the name of the property by putting in parallel a verbose logger out of a TIMER.

Another (probably simpler) way would be to put in place a script filter in between this component and the timer.
In its code create something like (out of memory with the JS code)

> var outRecord = newWireRecord()
> outRecord.myValue = newIntegerValue(<some value>)
> output.add(outRecord)

Then of course put as parameter name "myValue".

Let me know if you need anything else.

iacco

LeoNerdoG:
Hey @iacco,
this helped a lot, thanks. I was able to get the result out of the trigonometric component. 
I will post possible issues here, it will take some time to check all the scenarios.

LeoNerdoG:
Hey @iacco,
I have checked this PR and there are (in my opinion) some changes to be made. 
1. Result Name field is not limited, so we can get something like this in the log:
Screenshot 2020-09-23 at 11 40 26

2. Parameter Value field is also not limited (there is no such thing in logs as is for "Result Name", but should be limited anyways). I know this will also affect other components, but I think it should be done. 

3. Everything after dot symbol is ignored (e.g. if you want to calculate sin1.234, you will actually get result of sin1).
Screenshot 2020-09-23 at 11 49 22

4. If you have (for example) two JS filters with same output name (e.g. myValue1) and bot values are different (in one case "12" and in other "24"), you will get two results on the output side of Trigonometric component - is this ok? 

The calculactions seem to be ok, so there are only problems that I described above. 
Can you please confirm this?

iacco:
Hey @LeoNerdoG,

1. as far as I know property name is not limited in any wire component. Also, maybe it would be not the best to limit it, since you would add complexity and corner-cases (what happens when the name is longer, do you "cut" it to the max size? do you simpy not work?).

2. same as above

3. I performed some simple tests, and could not reproduce this: used a small JSFilter with the following code:

```
var outRecord = newWireRecord()
outRecord.test = newDoubleValue(1.3)
output.add(outRecord)

```

by changing from 1.2 to 1.3 or 1.5 I could see different values with SIN function.

4. I believe it should be correct: 2 inputs lead to 2 outputs.
the component does simply process an incoming wire envelope, look for the parameter, perform the action and emit the result with a given name, does not perform buffering, or any kind of logic in regards to which component it got the envelope from.

Let me know your thoughts.

With Best Regards,
iacco

LeoNerdoG:
Hey @iacco,
for the first two I don't expect to be major blockers, I just wanted to state that I think this should be fixed in the future (not in this PR) and wanted to know you thoughts. 

For the third one: I am an idiot. I must have had Integer value copy-pasted from previous tests instead having a double and...yeah. I tried with double value and it worsk fine. Luca:Leo 1:0 😄 

For the last one: OK, It makes sense. 
I think this can be merged, I didn't find anything else that would seem like an issue. Good job! 😄

MMaiero:
The CQ has passed the first check: https://dev.eclipse.org/ipzilla/show_bug.cgi?id=22766
As reported, the code can be now merged but the contribution will be further checked later on.

MMaiero:
The code needs to be EPL-v2 compatible. @iacco anything against?

iacco:
fine for me!

PR #3066, opened 2020-09-07
Backport services target selection

MMaiero:
Backport of https://github.com/eclipse/kura/issues/2983 and https://github.com/eclipse/kura/pull/3023


kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/server/GwtComponentServiceImpl.java
```diff
@@ -732,18 +734,18 @@ public boolean updateProperties(GwtXSRFToken xsrfToken, String pid, Map<String,
 
             referenceInterfaces.forEach(reference -> {
```
nicolatimeus, line 743: The code that creates the `referenceInterfaces` should only consider the references included in the definition of the component with `kura.service.pid=pid`. The `getPidsFromTarget()` method seem to ignore the `pid` argument completely.
This might cause incorrect interfaces to be added to the set and included later in the search.

MMaiero:
I have committed a change. Please have a look. If it works, I'll also update develop

LeoNerdoG:
Hey @MMaiero,
I have tried several times now to build and deply this on my Raspberry but when after a successfull build, Kura is not accessible - can you please check if there is an error on the branch? Becasue all other branches work ok, I tried with several of them.

MMaiero:
How are you accessing this? Remember that this is Kura 4 so the access is not via https

LeoNerdoG:
Hey Matteo, yeah, I totally forgot...i was now able to access it. Such a rookie mistake...

nicolatimeus (approved):

LeoNerdoG:
Hey, I have finally checked this and it looks ok! 
Sorry for taking so long.

PR #3095, opened 2020-10-06
Issue 2598 quectel modem

sterwen:
This pull request add the support of several Quectel Modems: EC25/EC21/BG96

**Related Issue:** This PR fixes/closes issue 2598 (that has been prematurally closed)

**Description of the solution adopted:** Support added canonically versus the existing ones.


pierantoniomerlino:
@sterwen This PR has conflicts with some file already present in the develop branch. In particular, the Quectel EC/EG25 modem is already in that files. Could you fix it?
However, there are some differences in the way the modem is configured. In the **develop** branch the modem is as follows:

- AT Port: 2
- Data Port: 3
- GPS Port: -1

that means that the modem uses port 2 for AT commands, 3 for the connection and is hasn't a gps device.
In this PR, instead:

- AT Port: 3
- Data Port: 3
- GPS Port: 1

So, the same port is used for AT command and connection. This can cause some issues during normal operation, since Kura will try to query the modem using the same port used for connection, causing failures.
Are some Quectel EC/EG25 modem equipped with GPS?

sterwen:
Will re-submit

PR #3096, opened 2020-10-09
Added support for Telefonica US version

javadakhlaghinia:
Added support for Telefonica US version


MMaiero:
Hello,
you have to rebase your branch and be even to develop before creating a new PR.
Otherwise there will be conflicts, as you can see from GitHub.

MMaiero:
As you can see there are still issues. The commit needs only your changes.
What I can suggest is to start with a branch in your fork that is even with the Kura/develop and start committing from there

PR #3108, opened 2020-10-16
Changed directory check during unzip of compressed files

### LINKED ISSUE #3107: The upload of zip archive with folders fails
**Describe the bug**
The CommandService can be used to upload compressed files to the gateway. If the zip file contains directories, the procedure fails with the following error:

```
2020-10-16T10:33:57,518 [pool-25-thread-1] ERROR o.e.k.c.a.c.CommandCloudApp - Error unzipping command zip bytes
java.io.FileNotFoundException: /tmp/hcitrace/hcitrace.txt (Not a directory)
```

**To Reproduce**
1. Create a zip file with a folder inside
2. Upload to the gateway with Device->Command tab
3. Check that there is an error in `kura.log` and in the /tmp folder there is a file with the same name of the folder in the zip archive.

**Expected behavior**
No errors on the logs and the content of the zip file should be decompressed on the /tmp folder

pierantoniomerlino:
This PR changes the directory check that is performed during the unzip of compressed files.

**Related Issue:** This PR fixes/closes #3107 

The `file.isDirectory()` method checks whether the file is a regular one or a directory. However, it returns false if the file doesn't exist. So, during the decompression of a zip archive, the method cannot be used since the resource is not copied to the filesystem yet. 
According to [this](https://stackoverflow.com/questions/23920684/java-extracting-zip-file-with-multiple-subdirectories) a safe way to check if the resource is a file is a folder or not, is to test the last character of the path: a folder path always ends with '/'.

### DIFF
--- a/kura/org.eclipse.kura.core.cloud/src/main/java/org/eclipse/kura/cloud/app/command/UnZip.java
+++ b/kura/org.eclipse.kura.core.cloud/src/main/java/org/eclipse/kura/cloud/app/command/UnZip.java
@@ -95,7 +95,7 @@ public class UnZip {
                 String fileName = validateFileName(expectedFilePath, folder.getPath());
                 File newFile = new File(fileName);
 
-                if (newFile.isDirectory()) {
+                if (expectedFilePath.endsWith("/")) {
                     newFile.mkdirs();
                     ze = zis.getNextEntry();
                     continue;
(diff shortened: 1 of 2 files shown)


MMaiero (approved):

MMaiero:
Does this need a back port for 4.1.2? Or we just fix for 5.0.0 for now?

pierantoniomerlino:
We agreed to port the fix to the 4.1.2

PR #3114, opened 2020-10-19
Modified iBeacon packet parser

### LINKED ISSUE #3113: The iBeacon Scanner sometimes fails to detect beacons
**Describe the bug**
The iBeacon Scanner example [1] sometimes fails to detect the incoming beacons. It seems that if the beacons are generated by the iBeacon advertiser [2] running in a RPi, it works as expected. When the iBeacons are emitted by an Android smartphone with some app, the iBeacon scanner cannot detect any packets.

**To Reproduce**
1. Install the iBeaconScanner example and enable it
2. Install on a different device the iBeaconAdvertiser and enable it
3. See on the log of the first device that the iBeacons are detected
4. Disable the iBeaconAdvertiser and use an Android app to emit iBeacon packets
5. See on the logs of the first device that any packet is detected.

**Expected behavior**
All packets should be detected regardless the device has emitted them.

[1] https://github.com/eclipse/kura/tree/develop/kura/examples/org.eclipse.kura.example.ibeacon.scanner
[2] https://github.com/eclipse/kura/tree/develop/kura/examples/org.eclipse.kura.example.ibeacon.advertiser

pierantoniomerlino:
This PR modifies the parser for detecting iBeacon advertising packets.

**Related Issue:** This PR fixes/closes #3113 

As briefly presented here [1], the advertising packet for iBeacon is composed by two sections:

- AD0: it contains the Flags Advertising Data Type
- AD1: it contains the iBeacon Data Type

The current implementation of the iBeacon packet parser searches for the two section and if the first one is missing, the parser fails and ignore the incoming packet. It seems that many Android apps don't send both AD0, causing the failure on the parser. 

To avoid that, a check is added before reading the flag section. If the AD0 is missing the iBeacon packet object is filled with default flag values.

[1] https://www.argenox.com/library/bluetooth-low-energy/ble-advertising-primer/


MMaiero (approved):

MMaiero:
Don't we need a develop PR?

MMaiero:
Tested on my machine since Travis seems to struggle. It builds fine.

PR #3141, opened 2020-11-13
Add support for multiple cellular modems

pierantoniomerlino:
This PR adds the support for multiple cellular modems.

**Description of the solution adopted:** Kura natively supports multiple cellular modems except for the reset procedure. So, the `GatewayModemDriver` is updated to use the `ModemDevice` object instead of the vid/pid values. In this way, the code can link the specific modem with the usb port where it is attached to and toggle the correct reset gpio.
The `SupportedUsbModems` class is updated as well to use the `UsbDevice` instead the pid/vid where possible


kura/org.eclipse.kura.linux.net/src/main/java/org/eclipse/kura/linux/net/modem/SupportedUsbModems.java
```diff
@@ -87,63 +92,54 @@ public static void installModemDrivers(CommandExecutorService executorService) {
         }
     }
 
-    public static SupportedUsbModemInfo getModem(String vendorId, String productId) {
-        if (vendorId == null || productId == null) {
+    public static SupportedUsbModemInfo getModem(UsbDevice usbDevice) {
```
MMaiero, line 113: I believe those changes here are a sort of breaking change. Is the manifest updated to export a new major? Or, if we can still be retro-compatible, can we have method overloading with deprecation?

kura/org.eclipse.kura.linux.net/src/main/java/org/eclipse/kura/linux/net/modem/SupportedUsbModemsInfo.java
```diff
@@ -41,7 +41,7 @@ public static SupportedUsbModemInfo getModem(String vendorId, String productId,
         return null;
     }
 
-    public static boolean isSupported(String vendorId, String productId, String productName) {
-        return SupportedUsbModems.isSupported(vendorId, productId, productName);
+    public static boolean isSupported(UsbDevice usbDevice) {
```
MMaiero, line 50: Same here

kura/org.eclipse.kura.linux.net/src/main/java/org/eclipse/kura/linux/net/modem/UsbModemDriver.java
```diff
@@ -58,43 +59,43 @@ public String getProduct() {
         return this.product;
     }
 
-    public void enable() throws KuraException {
+    public void enable(UsbModemDevice device) throws KuraException {
```
MMaiero, line 77: Same here

kura/org.eclipse.kura.linux.net/src/main/java/org/eclipse/kura/linux/net/modem/SupportedUsbModems.java
pierantoniomerlino: I can move org.eclipse.kura.linux.net.modem from 1.1.1 to 2.0.0

MMaiero: Is it possible to deprecate the old and add the new method?

MMaiero (approved):

PR #3142, opened 2020-11-13
Restricted file upload type in Command ui

### LINKED ISSUE #3132: User Gets Logged Out if PDF File Uploaded in Device -> Command
If user selects a .PDF fiel as file to be uploaded and executed in Device -> Command and clicks "Execute" several times (without adding a command to be executed, e.g. "pwd"), Kura logs user out and saying "Session has expired". 

**Testflow:**
2. Go to Device -> Command and select a .PDF document in "File" section
3. Without entering a command to execution, press "Execute" button several times - observe the error and how you will get logged out. 

**Expected behavior**
Kura should not log user out in this case. 

**Screenshots**
Screenshot 2020-11-09 at 09 16 40

 - Board: Raspberry Pi 3
 - OS version: Linux raspberrypi 4.19.118-v7+ 1311 SMP Mon Apr 27 14:21:24 BST 2020 armv7l GNU/Linux
 - Additional info: latest develop brach

**Additional context**
/

nicolatimeus:
Related to #3132 

Signed-off-by: Timeus <nicola.timeus@eurotech.com>

### DIFF
--- a/kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/client/ui/device/CommandTabUi.java
+++ b/kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/client/ui/device/CommandTabUi.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2011, 2019 Eurotech and/or its affiliates
+ * Copyright (c) 2011, 2020 Eurotech and/or its affiliates
  *
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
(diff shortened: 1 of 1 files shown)


MMaiero:
@LeoNerdoG Can you please check if the corresponding issue is solved?

LeoNerdoG:
Hey @nicolatimeus,
I have checked this on a clean browser and if user does not click "Execute" twice, it works fine. But if user click "Execute" twice (without waiting for the first command to execute properly), user gets logged out. 
This happnes with .sh and .zip files. 

Is there a way that screen gets disabled with rotating cog so user cannot click Execute multiple times? 
@MMaiero what do you think? Is disabling the screen after the "Execute" is pressed a valid option?

MMaiero:
I believe we have already done something similar for other views.

MMaiero (approved):
To be noted that in Chrome also .jar files can be uploaded. Probably because all of them are considered forms of compressed files.

PR #3143, opened 2020-11-13
Added support for HTTPS client side authentication

nicolatimeus:
* Adds support for HTTPS client side authentication to Web UI
* Updates Jetty to 9.4.34.v20201102


MMaiero:
@mstankovic can you have a check on this? I believe most of our tests will be broken by this.
We also need to understand if it is possible to add new tests for the hardware key or just simply the SSL mutual authentication implemented.

mstankovic:
@MMaiero merging this will not broke our tests since the "password" is default **Authentication Method**

Capture


i will wait for instructions how to test "certificate" **Authentication Method** and let you now if tests can be automated.

nicolatimeus:
Hi @mstankovic, yes you will be probably able to login in the same way as before with username and password.
This PR also adds a new "Security" tab, some elements from other sections (e.g. the WebConsole configuration) have been moved there, please have a look at that new section as well to verify if some tests need to be adjusted.

mstankovic:
@nicolatimeus thank you for that info. Yes I need to adjust tests where we adding server certificates. Also I found few issues so far:

1. once I added server certificate i don't see it on the certificates list table (even after restarting kura)
2. by default Uninstall button is enabled although no one item is selected
3. I cannot see that item from the table is selected although its selected
3. if I uninstall both default certs - im not able to access kura anymore
4. Im able to disable both auth methods (password ad certificate) from WebConsole service and after that im able to enable them back only from EC if device is connected to the cloud.
5. If I try to add some invalid string as certificate for any type of certs im getting error behind Add Certificate modal dialog - I cannot see error until I close modal

@LeoNerdoG 
it would be great if you can take a look at this PR im pretty sure that there are more issues

kura/org.eclipse.kura.http.server.manager/OSGI-INF/metatype/org.eclipse.kura.http.server.manager.HttpService.xml
```diff
@@ -45,34 +45,87 @@
...
+            default=""
+            description="Specifies the URI for OCSP responders."/>
+            
+        <AD id="https.client.crl.path"
+            name="Https Client CRL path"
+            type="String"
+            cardinality="0"
+            required="false"
+            default=""
+            description="Specifies the path for the CRL for client certificate revcation check."/>
```
MMaiero: revocation

kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/SelfConfiguringComponentProperty.java
```diff
@@ -0,0 +1,190 @@
...
+import java.util.stream.Collectors;
+
+import org.eclipse.kura.KuraException;
+import org.eclipse.kura.configuration.Password;
+import org.eclipse.kura.configuration.metatype.AD;
+import org.eclipse.kura.configuration.metatype.Scalar;
+import org.eclipse.kura.core.configuration.metatype.Tad;
+import org.eclipse.kura.crypto.CryptoService;
+
+/*******************************************************************************
```
MMaiero: needs to be reworked in the proper order

MMaiero (approved):

MMaiero:
Few minor comments made.
We will merge the PR with those fixed. On the Issues reported, we will cover them in another PR that is going to be posted shortly

MMaiero:
Since Travis seems not functional, I have tested the build in my pc and verified it works on a raspberry. I'm going to merge it in order to let the team continue on the QA and development process.

PR #3196, opened 2020-11-26
Modified database interaction code to prevent that an interrupted exc…

MMaiero:
…eption can corrupt the db

It adds also some simplifications in the data service.

### DIFF
--- a/kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/db/H2DbServiceImpl.java
+++ b/kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/db/H2DbServiceImpl.java
@@ -20,7 +20,10 @@ import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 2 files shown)


kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/db/H2DbServiceImpl.java
```diff
@@ -279,6 +303,12 @@ private void updateInternal(final Map<String, Object> properties) {
                 restartDefragTask(newConfiguration);
             }
 
+            if (this.executorService != null) {
+                this.executorService.shutdown();
```
nicolatimeus: If I understood the documentation correctly, after `shutdown()` is called, the current executor will not accept new tasks and `ExecutorService.submit()` should fail throwing the `RejectedExecutionException` unchecked exception.
We should check that `withConnection()` is not affected by this.

kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/db/H2DbServiceImpl.java
```diff
@@ -279,6 +303,12 @@ private void updateInternal(final Map<String, Object> properties) {
...
+                awaitExecutorServiceTermination();
```
nicolatimeus: The current thread is holding `this.rwLock.writeLock`, there might be tasks in the executor that are waiting to acquire a `this.rwLock.readLock`. If this happens, `awaitExecutorServiceTermination()` will not return before the 1 minute timeout and these tasks will run after `awaitExecutorServiceTermination()` returns.

kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/db/H2DbServiceImpl.java
```diff
@@ -154,20 +162,36 @@ public Connection getConnection() throws SQLException {
             syncWithExecutor();
         }
 
-        final Lock lock = this.rwLock.readLock();
-        lock.lock();
-        Connection connection = null;
+        Future<T> result = this.executorService.submit(() -> {
```
nicolatimeus: Should we submit the task with the `readLock` held? Can this solve the issues in the comments above?

PR #3255, opened 2021-01-07
Virtual network interfaces management

pierantoniomerlino:
Currently, Kura treats virtual interfaces as standard ones. This PR improves the virtual interfaces management implementing the `isVirtual` method in the NetworkService and adding a (configurable) default configuration for them.

**Description of the solution adopted:** The `isVirtual` method in the NetworkService is implemented, checking if the path on the filesystem contains the virtual keyword. Moreover, the `kura.net.virtual.devices.config` property is added to the `kura.properties` file to set the default configuration for the virtual interfaces: this can be `unmanaged` or `disabled`. The configuration can be overridden using the usual `kuranet.conf` file.


MMaiero:
Do you believe some unit tests can be added to support this new feature?

pierantoniomerlino:
> Do you believe some unit tests can be added to support this new feature?

I was thinking about that when I submitted the PR...

MMaiero (approved):

PR #3274, opened 2021-01-18
Allow restricting REST APIs and Console access to dedicated ports

nicolatimeus:
* Adds support for multiple equivalent HTTP ports (e.g. more than one unencrypted HTTP port)
* Added support for restricting access to REST APIs and console to dedicated ports
* Includes a (breaking?) change in `ConfigurationService` that enables serializing empty arrays in xml snapshot


MMaiero:
I can confirm the issue with Java8. With Java11 it builds

mstankovic:
@nicolatimeus i found two minor issues but important for qa automation:

- rest service: once you enter desired port -> apply button should be enabled automatically (currently need to remove focus)
- https service: unsaved warning appears after applying changes (add port and apply changes and change tab for example)

MMaiero:
@mstankovic Regarding the first point, are we changing the behaviour compared to other services?
Regarding the second point, we will investigate.

mstankovic:
@MMaiero I used "Command Password Value" (at CommandService) for the reference but i see now that other inputs has the same behavior. ok, lets ignore this i will try to find solution for automation.

nicolatimeus:
Hi @mstankovic, I've checked the 

> unsaved warning appears after applying changes (add port and apply changes and change tab for example)

part.
I could reproduce it, basically the following happens:

1. The `HttpService` update will cause the Jetty instance to be restarted.
2. The restart of the Jetty instance causes the configuration update RPC call performed by the browser to fail with a `StatusCodeException` with `code == 0`, since an exception is returned, the HttpService configuration is still marked as "dirty" (there are unsaved changes).
3. `StatusCodeException`s with `code == 0` are ignored by the code that reports errors to the user, so no error dialog will be shown.
4. When the Jetty instance restarts the Web UI will detect that the session expired and a page refresh will be triggered. At this point the browser will report that there are unsaved changes due to point 2.

Would it be better if we just clear the dirty state of the HttpService and force a page refresh just after the configuration update RPC returns either success or failure?

mstankovic:
@nicolatimeus Yes, I think that would be better option.

mstankovic:
@nicolatimeus its much better now!

MMaiero:
I believe we have lost in moving the http service configuration from one section to the other the note that stated that a change on the service will determine a reload of the http/s service.
At least in that way the user is warned.

Is there a way in the service configuration to have only the rest service on a certain port while the UI in a different one?

PR #3278, opened 2021-01-22
Adapt log download feature to journald

pierantoniomerlino:
This PR modifies the log download feature to include also journald log in the compressed file.

**Description of the solution adopted:** The `LogServlet` uses the ´journalctl´ command to generate a log file dumping the journald logs related to Kura and the whole system. If the command succeeds, the files are included in the compressed file sent to the user.
The fields included in the log can be customized using the `kura.log.download.journal.fields` property in the `kura.properties`. The default configuration is `SYSLOG_IDENTIFIER,PRIORITY,MESSAGE,STACKTRACE`.

The files kura_journal.log and system_journal.log are added to the compressed file.


kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/server/servlet/LogServlet.java
```diff
@@ -63,6 +68,17 @@ protected void doGet(HttpServletRequest httpServletRequest, HttpServletResponse
             return;
         }
 
+        PrivilegedExecutorService pes = null;
```
MMaiero, line 72: Can't the pes be obtained in the write method, since it seems the only usage point?

pierantoniomerlino: @MMaiero Yes, but then I've to pass to the `writeJournaldLog` also the locator and the session...

MMaiero (approved):

PR #3281, opened 2021-01-26
Restored stream close instruction

pierantoniomerlino:


MMaiero:
Reporting the analysis done by @pierantoniomerlino: "Even if removing the call to close instruction seems correct, the resulting compressed file cannot be opened in some OSes (Windows, Raspbian, ecc.). So, keep this call."

MMaiero (approved):

PR #3285, opened 2021-02-01
Attempt to fix CommandService always uploading file in Java temp directory

alebianchin:
Signed-off-by: alebianchin <alessandrobianchin311@gmail.com>

This PR is a draft because it still breaks when getting a reference to the CommandCloudApp class.


code-incubus:
There is a problem when you try to set a non-existent working directory of commands on the CMD service, more precisely, non-existent directories are created with the name of the previously set path, there are steps to reproduce the issue:

1. Go to Command Service and enable it
2. On the same page set Command Working Directory with the value "null" and apply changes.
3. Go to Device
4. Select the Command tab
5. Select file for the upload in an appropriate format
6. Set Execute command e.g. "ls -la" and click on execute button
7. Output field will return all files/directories which uploaded file contains
8. Go back to Execute command and check the current path of a directory with "pwd"
9. Observe the path and you will see that a new "null" directory will be created to "/opt/eclipse/kura_5.0.0-SNAPSHOT_raspberry-pi-2/" path

kura/org.eclipse.kura.core.cloud/META-INF/MANIFEST.MF
```diff
@@ -43,3 +43,4 @@ Import-Package: com.eclipsesource.json;version="0.9.5",
  org.osgi.service.event;version="1.3.0",
  org.osgi.util.tracker;version="1.5.1",
  org.slf4j;version="1.6.4"
+Export-Package: org.eclipse.kura.cloud.app.command
```
MMaiero: Is this change needed since we are getting the properties needed from OSGi?

kura/org.eclipse.kura.core.cloud/src/main/java/org/eclipse/kura/cloud/app/command/CommandCloudApp.java
```diff
@@ -269,7 +269,7 @@ public String execute(String cmd, String password) throws KuraException {
     }
 
     // command service defaults getters
-    private String getDefaultWorkDir() {
+    public String getDefaultWorkDir() {
```
MMaiero: Is this change needed since we are getting the properties needed from OSGi?

kura/org.eclipse.kura.web2/META-INF/MANIFEST.MF
```diff
@@ -31,6 +31,7 @@ Import-Package: com.eclipsesource.json;version="0.9.5",
  org.eclipse.kura.certificate;version="[2.1,3.0)",
  org.eclipse.kura.channel;version="[1.0,2.0)",
  org.eclipse.kura.cloud;version="[1.0,2.0)",
+ org.eclipse.kura.cloud.app.command,
```
MMaiero: Is this change needed since we are getting the properties needed from OSGi?

MMaiero:
Forgot to mention: please update the copyright year in the files changed

code-incubus:
Hey @MMaiero, I've checked this PR after your last commit and it's fine from my side.

PR #3307, opened 2021-03-02
Fixes for Unprivileged Command Service

pierantoniomerlino:
This PR fixes the Unprivileged Command Service after the merge of #3301.

**Description of the solution adopted:** The #3301 removes the `kurad` user from the `sudoers`, so the `sudo` command cannot be used from Kura. This PR replaces the `sudo` command used by the Unprivileged Command Service with `su`. A new configuration is added to `/etc/pam.d/su` to grant the needed privileged to the user. Morever, the selection of the user that the Unprivileged Command Service uses for running commands is removed from the webUI and moved to the `kura.properties` file. Finally, a radio button is added to the CommandService for selecting the Privileged/Unprivileged executor.


pierantoniomerlino:
👍

MMaiero (approved):

PR #3309, opened 2021-03-03
Added tamper detection APIs

nicolatimeus:
* Adds new APIs that allow to implement tamper detection mechanisms.
* A tamper detection event can cause republishing of birth certificate.
* Adds an example that provides a simulated implementation of the APIs.

The following new web ui elements will be available if a tamper detection API implementation is available:

* The **Status** section will report tamper status.
* The new **Security** -> **Tamper Detection** section allows to view and reset tamper status.


MMaiero:
@nicolatimeus Please have a look at the sonar reports

MMaiero (approved):

Issue #3323, opened 2021-03-10
Firewall settings not saved in the file system after network configuration change


sterwen:
**Describe the bug**
When changing the network configuration, the firewall rules are not permanently saved to /etc/sysconfig/iptables. Therefore upon the next reboot the gateway is not accessible anymore.

**To Reproduce**
1. Start from a default network configuration Eth0 => WAN DHCP client, wlan0 LAN DHCP server.
2. Access to the gateway works (https/ssh/...). Routing LAN to WAN is working
3. Configure ppp0 as WAN, Eth0 LAN DHCP client, Wlan0 LAN DHCP server
4. Access to the gateway works (https/ssh/...). Routing LAN to WAN is working
5. Here we verify /etc/sysconfig/iptables and we can see that the firewall configuration is not saved
6. reboot
7. After reboot, the gateway starts normally and we can verify that from the console but no remote access possible

*Applying /usr/sbin/iptables-save before the reboot solve the issue*

**Expected behavior**
Same behavior after reboot.

**Screenshots**
N/A

**Target Environment (please complete the following information):**
 - Board NXP i.MX6
 - OS version: Linux BS184300123 5.4.47-imx6-sr #1 SMP Mon Feb 15 22:07:42 UTC 2021 armv7l armv7l armv7l GNU/Linux
 - Additional info: 

**Additional context**

pierantoniomerlino:
I follow the sequence of commands reported in the issue and I found that indeed `/etc/sysconfig/iptables` is not updated with all the needed rules. After a reboot, it is possible to access to the web console but the ip forwarding rules between wlan and ppp is gone. **Maybe** this is related to #1195.

pierantoniomerlino:
This issue is not directly related to #1195, but it is due to a bug on the `LinuxFirewall`.

pierantoniomerlino:
@sterwen I tried to fix this issue in #3334. Could you please test it on your device?

sterwen:
Hello,

Tested and the correction is OK.

PR #3334, opened 2021-03-22
New firewall chains for IP/Port forwarding

### LINKED ISSUE #1195: [Net] Firewall IP Forwarding rules entered in the Web UI lost on reboot

pierantoniomerlino:
This PR tries to fix some issues found in the firewall.

**Related Issue:** This PR fixes/closes #1195 #3323

**Description of the solution adopted:** The #1195 reports that the IP forwarding rules are lost after a reboot. This is caused by the impossibility for the LinuxFirewall to parse the iptables rules and recognize them as IP or Port forwarding rules. In #2119 the comment module is used to tag the rules, but unfortunately that module is not present on all the devices. This PR tries to overcome the issue using more iptables chain to group the IP/Port forwarding rules together. 
The `forward-kura-ip` and `forward-kura-ipf` chains are added to the filter tables, while the `postrouting-kura-pf`, `postrouting-kura-ipf` and `prerouting-kura-pf` ones are used in the nat table.

Moreover, this PR fixes a bug in the `LinuxFirewall` that prevents the saving of the forwarding rules.


kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/net/firewall/FirewallNatConfig.java
```diff
@@ -49,6 +62,35 @@ public FirewallNatConfig(String srcIface, String dstIface, String protocol, Stri
...
+     *            the source network/host address in CIDR notation
+     * @param dst
+     *            the destination network/host address in CIDR notation
+     * @param masquerade
+     *            whether or not MASQUERADE should be enabled
+     * @param type
+     *            the type of the rule (IP forwarding, Port forwarding or generic)
+     */
+    public FirewallNatConfig(String srcIface, String dstIface, String protocol, String src, String dst,
+            boolean masquerade, RuleType type) {
```
MMaiero, line 90: Missing since annotation

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/net/firewall/FirewallNatConfig.java
```diff
@@ -75,6 +117,10 @@ public boolean isMasquerade() {
         return this.masquerade;
     }
 
+    public RuleType getRuleType() {
```
MMaiero, line 128: Missing since annotation

MMaiero (commented):
Seems really great! It works fine on my raspberry. Prior to merging, please have a look at Sonar, since it is failing.

MMaiero (approved):

PR #3336, opened 2021-03-23
Partial rollback on a previous PR to restore empty string support

MMaiero:

This PR fixes an error in the local UI that prevented to set a value to "". For example, once the client id was set, from the local UI it was not possible to remove the defined value, setting the entry as empty.


MMaiero:
@mstankovic Can you please check if this works?

mstankovic:
@MMaiero verified as fixed

nicolatimeus (approved):

PR #3337, opened 2021-03-24
Reworked the SSL Manager keystore code

MMaiero:
Added a check to prevent the password update if the password has not changed.
This prevents the case, for example during boot, where the SSL SslManagerServiceImpl
triggers a keystore and a security manager password update even if the password was
not actually changed.


MMaiero:
@mstankovic Can you please check if this works?

nicolatimeus (approved):

PR #3349, opened 2021-04-20
Enhanced key pair entry

pierantoniomerlino:
This PR do the following:

- some of the methods in `KeystoreServiceRemoteKeystore` are made public;
- the method for inserting new certificates/keypair from the web console is modified to accept DSA/RSA/EC keys;
- the list of certificates can be ordered by keystoreName/type/alias;
- tests added.

**Any side note on the changes made:** Even if EC certificates are accepted, It isn't possible to add EC keypairs.


MMaiero (approved):

MMaiero:
Merging. The build is giving errors on tamper resistance tests that we are managing in another PR

PR #3375, opened 2021-05-13
Certificate list error dialog fix

marcellorinaldo:

Brief description of the PR. Caught keystore exceptions raised when populating the certificates list in the "Security" UI with wrong-configured certificates. In this way, the user is able to see at least the correct certificates, whereas the faulty ones are hidden.

**Related Issue:** N/A.

**Description of the solution adopted:** A "KuraException" is caught when populating the certificates list. The exception and the relative stack trace are logged.

**Screenshots:** N/A.


mstankovic:
Verified as fixed!

PR #3381, opened 2021-05-18
Fixed broken dev-env

salvatore-coppola:
**Fixed broken dev-env**

Modified dependencies version, launch configurations (for both OSX and linux), initial emulator snapshot and bundles composition on emulator target platform.

Open points:

- An `java.lang.IllegalStateException` is still thrown because the default clientId (primal network interface mac address) in MqttDataTransport could not be read if the primal network interface is not eth0. Not blocking but we could set a default string as clientId.
- A `java.lang.IllegalArgumentException` is thrown at equinox startup because standard property `osgi.configuration.area` is overridden by `org.osgi.framework.storage` (not a real issue).
- A `javax.management.InstanceAlreadyExistsException` is thrown by `log4j2` mbean (not an issue).


salvatore-coppola:
- Add a random generated clientId to the emulator's snapshot_0.

MMaiero:
Can you check the sonar results?

salvatore-coppola:
@MMaiero currently there isn't test coverage for the emulator bundle, should I add it ?

kura/emulator/org.eclipse.kura.emulator/src/main/java/org/eclipse/kura/emulator/Emulator.java
```diff
@@ -66,23 +74,23 @@ protected void deactivate(ComponentContext componentContext) {
...
-            if (fileOutput != null) {
-                fileOutput.close();
-            }
-            if (fileInput != null) {
-                fileInput.close();
-            }
+        URL internalSnapshotURL = this.componentContext.getBundleContext().getBundle().getResource(SNAPSHOT_0_NAME);
+
+        try (InputStream fileInput = internalSnapshotURL.openStream();
+                OutputStream fileOutput = new FileOutputStream(snapshotFolderPath + File.separator + SNAPSHOT_0_NAME)) {
```
MMaiero, line 80: Formatting seems not applied properly

target-platform/config/kura.target-platform.build.properties
```diff
@@ -7,7 +7,7 @@ org.usb4java.version=1.2.500
 usb4java-javax.version=1.2.500
 com.gwt.user.version=1.1.100
 jdk.dio.version=1.0.500
-org.eclipse.kura.camel.sun.misc.version=1.0.400
+org.eclipse.kura.camel.sun.misc.version=1.0.500-SNAPSHOT
```
MMaiero, line 10: Is this change required?

MMaiero:
In my setup the dev-env is starting but the system packages section is not loading. Probably because it is searching for RPMs?

target-platform/config/kura.target-platform.build.properties
salvatore-coppola: I checked and yes we need it

MMaiero (approved):

PR #3393, opened 2021-06-01
Fix snapshots size

salvatore-coppola:
Now snapshots are serialized without OCD component definition.

### DIFF
--- a/build-all.bat
+++ b/build-all.bat
@@ -1,4 +1,0 @@
-mvn -f target-platform/pom.xml clean install "-Dmaven.test.skip=true" &&^
-mvn -Pcan -f kura/pom.xml clean install "-Dmaven.test.skip=true" &&^
-mvn -Pcan -f kura/examples/pom.xml clean install "-Dmaven.test.skip=true" 
-rem && mvn -f kura/distrib/pom.xml clean install "-Dmaven.test.skip=true"
--- a/build-all.bat
+++ b/build-all.bat
@@ -1,0 +1,4 @@
+mvn -f target-platform/pom.xml clean install "-Dmaven.test.skip=true" &&^
+mvn -Pcan -f kura/pom.xml clean install "-Dmaven.test.skip=true" &&^
+mvn -Pcan -f kura/examples/pom.xml clean install "-Dmaven.test.skip=true" 
+rem && mvn -f kura/distrib/pom.xml clean install "-Dmaven.test.skip=true"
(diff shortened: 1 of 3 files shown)


kura/test/org.eclipse.kura.core.configuration.test/src/test/java/org/eclipse/kura/core/configuration/ConfigurationServiceTest.java
```diff
@@ -2171,6 +2162,7 @@ protected String marshal(Object object) {
     }
 
     @Test
+    @Ignore
```
nicolatimeus: Is this intentional?

salvatore-coppola: no, this is a typo!

kura/org.eclipse.kura.core.configuration/src/main/java/org/eclipse/kura/core/configuration/ConfigurationServiceImpl.java
```diff
@@ -1034,10 +1034,21 @@ private void encryptPlainSnapshots() throws KuraException, IOException {
         }
     }
 
-    private synchronized long saveSnapshot(List<ComponentConfiguration> configs) throws KuraException {
+    private synchronized long saveSnapshot(List<ComponentConfiguration> configs, boolean includeDefinition)
```
MMaiero: Do we need to keep the "includeDefinition" parameter? All the usages we have right now set the value to false.
is there any other usage that I've missed?

salvatore-coppola: I'd keep it, at the moment no method set the flag to true but maybe in the future some methods will do it, also a configuration normally includes the definition so an invoker could expect to have the configuration definition serialized.

MMaiero: Correct.
But that is a private method. So, users will not be able to leverage it in any way, right?

salvatore-coppola: that's right, it's more a concern about someone who could use the method in the future, but however we can remove the flag

MMaiero: Thanks.

MMaiero (approved):

build-all.bat
```diff
@@ -0,0 +1,4 @@
+mvn -f target-platform/pom.xml clean install "-Dmaven.test.skip=true" &&^
```
MMaiero: Is this file new? Is it related to this PR or not?

salvatore-coppola: no this is a script file of mine.

PR #3406, opened 2021-06-13
feat: allow dynamic Topic (using message properties) with rawMQTT

GizmoIchitzo:
The "raw mqtt" publisher topic can now add wildcase values ($ value) to be dynamically enhanced by the kuramessage


GizmoIchitzo:
I update the raw mqtt cloud connection, using the same operation than CloudPublisherImpl, so I duplicated Code.
I don't find existing tests with publish MQTT :s

MMaiero:
Thanks. I'll have a look if it is possible to create some tests. I'll keep you posted.

mattdibi:
Closing for inactivity

Issue #3160, opened 2020-11-17
Dialogue Box for Deleting Certificate Contains Certificate Alias


LeoNerdoG:
When deleting a certificate from the list, a dialogue box opens for confirming the decision. The problem here is that dialogue box contains certificate alias, which can lead to a very big dialogue box. 

**Testflow:**
1. Login to Kura as admin/admin
2. Go to Security -> Add
3. Selecte Server SSL Certificate
4. Enter a very long name for certificate alias
5. Add a valid certificate, click Apply
6. Select the newly created ceritifacet and click "Uninstall"
7. Observe the dialogue box test

**Expected Behaviour:**
I recommend that this gets changed to "Are you sure you want to remove the selected certificate? Removing a certificate might have unexpected effects on the stability of the framework. If in doubt, answer "No"."

**Screenshot:**
![Screenshot 2020-11-17 at 10 54 04](https://user-images.githubusercontent.com/15121999/99375734-57ff9200-28c4-11eb-845f-923ea8d8b9e7.png)

**Target Environment (please complete the following information):**
 - Board: Raspberry Pi 3
 - OS version: Linux raspberrypi 5.4.72-v7+ 1356 SMP Thu Oct 22 13:56:54 BST 2020 armv7l GNU/Linux
 - _**Tested branch: PR #3143**_ 

**Additional context**
As stated above, this issue has been found on branch PR 3143 and not on latest develop branch.

salvatore-coppola:
Already patched.
Won't fix.

PR #3408, opened 2021-06-14
Fix UI security section

salvatore-coppola:
Several UI fixies that close the issue mentioned in the comments.


kura/org.eclipse.kura.http.server.manager/OSGI-INF/metatype/org.eclipse.kura.http.server.manager.HttpService.xml
```diff
@@ -85,6 +85,12 @@
             default="false"
             description="Specifies whether the revocation soft fail is enabled or not. If it is enabled and the gateway is not able verify the revocation status of a client certificate (for example due to a connectivity problem), the certificate will be rejected. This parameter is ignored if Revocation Check Enabled is set to false."/>
 
+            type="Boolean"
```
MMaiero: I believe this is a leftover from the rebase

kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/client/ui/NotInListValidator.java
```diff
@@ -0,0 +1,46 @@
+/*******************************************************************************
+ * Copyright (c) 2011, 2021 Eurotech and/or its affiliates and others
```
MMaiero: New file so this is only 2021

kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/client/ui/PEMValidator.java
```diff
@@ -0,0 +1,23 @@
+/*******************************************************************************
+ * Copyright (c) 2011, 2021 Eurotech and/or its affiliates and others
```
MMaiero: New file so this is only 2021

kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/client/ui/PKCS8Validator.java
```diff
@@ -0,0 +1,22 @@
+/*******************************************************************************
+ * Copyright (c) 2011, 2021 Eurotech and/or its affiliates and others
```
MMaiero: New file so this is only 2021

kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/client/ui/ServicesUi.ui.xml
```diff
@@ -50,6 +50,9 @@
     </ui:style>
 
     <b:Container fluid="true">
+		<b:Row>
```
MMaiero: The formatting seems odd here

kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/client/ui/security/CertificateListTabUi.java
```diff
@@ -237,70 +238,73 @@ public String getValue(GwtKeystoreEntry object) {
...
-					return -1;
-				if(item2 == null)
-					return 1;
-				
-				return item1.compareTo(item2);
-			}
-    	};
+    private <U extends Comparable<U>> Comparator<GwtKeystoreEntry> getComparator(
+            Function<GwtKeystoreEntry, U> comparableElementSupplier) {
+        return (o1, o2) -> {
```
MMaiero, line 243: Formatting seems strange

kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/client/ui/security/CertificateListTabUi.ui.xml
```diff
@@ -20,6 +20,8 @@
     xmlns:g="urn:import:com.google.gwt.user.client.ui" xmlns:gwt="urn:import:org.gwtbootstrap3.client.ui.gwt"
     xmlns:kura="urn:import:org.eclipse.kura.web.client.ui">
 
+	<ui:with field="msgs" type="org.eclipse.kura.web.client.messages.Messages" />
```
MMaiero: Also here

kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/client/ui/security/CertificateListTabUi.ui.xml
```diff
@@ -32,6 +34,11 @@
 
     <b:Container fluid="true">
         <b:Column size="MD_12">
+        	<b:Row>
+				<b.html:Paragraph alignment="LEFT">
```
MMaiero: And here. At least looking from GitHub

kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/client/ui/ServicesUi.ui.xml
salvatore-coppola: I see it properly in my eclipse

kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/client/ui/security/CertificateListTabUi.java
salvatore-coppola: I see it properly in my eclipse

kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/client/ui/security/CertificateListTabUi.ui.xml
salvatore-coppola: I see it properly in my eclipse

kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/client/ui/security/CertificateListTabUi.ui.xml
salvatore-coppola: I see it properly in my eclipse

kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/client/ui/ServicesUi.ui.xml
MMaiero: Maybe a configuration problem related to spaces and tabs?

salvatore-coppola: yes, i replaced tabs with whitespaces

kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/client/ui/security/CertificateListTabUi.java
salvatore-coppola: i have checked formatting and indentention and i tseems ok to me, re-applying the formatter doesn't change anything in the file

MMaiero:
HI @salvatore-coppola I'm noticing an issue with the certificates page in this PR. When following the procedure to add a new certificate or key pair, the view is like attached:

Screenshot 2021-06-16 at 08 53 32

All 4 buttons are available and enabled. Clicking on apply or reset simply closes the modal.

Furthermore, at least for me, having all 4 buttons available kind of makes difficult to understand what to do. 

At this point I'm kind not sure if the reset button is indeed needed. Or if the order of those buttons is the most useful for the end user. I believe that usually the apply button in our UI is the last one on the right, for example

salvatore-coppola:
@MMaiero this behaviour is a a bug, i'm going to fix it

MMaiero:
Another small comment on the description. Is it possible to have a bit of more space between the tab and the description? I don't know for the others, but it seems to me that the description is maybe too close to the top tab.

Screenshot 2021-06-16 at 09 00 50

Also, I have not checked if it is a problem also in the develop branch and/or in the release branch, but I'm having duplicate descriptions in all the services.
Screenshot 2021-06-16 at 09 01 37

kura/org.eclipse.kura.web2/src/main/resources/org/eclipse/kura/web/client/messages/Messages.properties
```diff
@@ -144,6 +144,7 @@ fileDownloadSuccess=File was successfully downloaded
 fileDownloadFailure=File download failed
 
 securityIntro=Review and update the available security settings.
+securityCertificateListWarning=Review and update the available certificate list.<br />After you add all the necessary certificates, a device reboot could be required in order for changes to be applied, in case the selected Keystore is not an SSLKeystore or an HttpsKeystore.<br />Update either SSLKeystore or HttpsKeystore could trigger a web console/mqtt data connection restart.
```
MMaiero: > available certificate list

available certificates list

> After you add all the necessary certificates, a device reboot could be required in order for changes to be applied, in case the selected Keystore is not an SSLKeystore or an HttpsKeystore.

For some keystore types, a device reboot might be required after a certificate update to make the change effective. For certain services like SSL and Web UI, the corresponding service will automatically restart after a keystore change. 

Something like this?

salvatore-coppola:
I have updated the warning message.

code-incubus:
Hi, @salvatore-coppola  I found a couple of typos in the following descriptions on the Http Service page:

2. Go to Security -> HTTP Service
3. Observe the syntax errors (described below)

* Instead "If enabled, the revocation status of client certificates will be **_ckeched_** during TLS handshake. If a revoked certificate is detected, handshake will fail. The revocation status will be checked using OCSP, CRLDP or the CRLs cached by the attached KeystoreService instance, depending on the value of the Revocation Check Mode parameter. If not enabled, revocation **_ckeck_** will not be performed." for **_"Revocation Check Enabled"_** should be _"If enabled, the revocation status of client certificates will be checked during TLS handshake. If a revoked certificate is detected, a handshake will fail. The revocation status will be checked using OCSP, CRLDP, or the CRLs cached by the attached KeystoreService instance, depending on the value of the Revocation Check Mode parameter. If not enabled, a revocation check will not be performed."_

* Instead "Specifies whether the revocation soft fail is enabled or not. If it is enabled and the gateway is not able **_verify_** the revocation status of a client certificate (for example due to a connectivity problem), the certificate will be rejected. This parameter is ignored if Revocation Check Enabled is set to false." for **_"Revocation Soft-fail Enabled"_** should be _"Specifies whether the revocation soft fail is enabled or not. If it is enabled and the gateway is not able to verify the revocation status of a client certificate (for example due to a connectivity problem), the certificate will be rejected. This parameter is ignored if Revocation Check Enabled is set to false."_

code-incubus:
Also, the user is still able to create visual certificates with the same "Alias".

Steps to reproduce:

2. Go to Security -> Add
3. Select Certificate, add a valid certificate with storage alias "ssl-1"
4. Try to add another certificate with the same alias but put one space before the name (e.g " ssl-1")
5. repeat step 4. adding a different number of spaces before/after the alias name

issue#3162

code-incubus:
Related to #3160 there is still one issue. When deleting a certificate from the list, a dialogue box opens for confirming the decision. The problem here is that the dialogue box contains a certificate alias, which can lead out of the dialog box.

issue#3160

code-incubus:
The footer text is now overlapped with the left menu.

footer_issue

code-incubus:
I'll agree with @MMaiero for the possibility to have a bit more space between the tab and the description and also buttons on the page. Take as an example the Certificate List tab... same space between tab above and buttons below.  Now, the SSL Configuration description is too close to the top bar but the spacing between buttons is OK.  e.g:

image


image

This should be the same for all descriptions.

kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/client/ui/security/CertificateListTabUi.ui.xml
```diff
@@ -50,17 +58,19 @@
...
-  					<b:Button dataDismiss="MODAL" b:id="closeModalButton" ui:field="closeModalButton">Close</b:Button>
-  					<b:Button type="PRIMARY" b:id="nextStepButton" ui:field="nextStepButton">Do something</b:Button>
-  				</b:ModalFooter>
-  			</b:Modal>
+                <b:ModalBody b:id="certAddModalBody">
+                    <g:ScrollPanel ui:field="certAddModalBody" addStyleNames="{style.modal-scroll-panel}"/>
+                </b:ModalBody>
+                <b:ModalFooter>
+                    <b:Button dataDismiss="MODAL" b:id="resetModalButton" ui:field="resetModalButton">Reset</b:Button>
+                    <b:Button type="PRIMARY" b:id="nextStepButton" ui:field="nextStepButton">Do something</b:Button>
```
MMaiero: I kind of find more natural the old ordering of buttons: in the first view the next button should be on the right like in the following view where the confirmation with the submit button is in the right.
I find odd that in the first view we have, in order, next and close and in the following view reset, close, apply.

salvatore-coppola: 6975c4d0efab4662ff7787a6afd856b100cdc0d2 revert the order as was before.

kura/org.eclipse.kura.http.server.manager/OSGI-INF/metatype/org.eclipse.kura.http.server.manager.HttpService.xml
```diff
@@ -56,14 +56,14 @@
...
-
+      
         <AD id="https.revocation.check.enabled"
             name="Revocation Check Enabled"
             type="Boolean"
             cardinality="0"
             required="false"
             default="false"
-            description="If enabled, the revocation status of client certificates will be ckeched during TLS handshake. If a revoked certificate is detected, handshake will fail. The revocation status will be checked using OCSP, CRLDP or the CRLs cached by the attached KeystoreService instance, depending on the value of the Revocation Check Mode parameter. If not enabled, revocation ckeck will not be performed."/>
+            description="If enabled, the revocation status of client certificates will be checked during TLS handshake. If a revoked certificate is detected, a handshake will fail. The revocation status will be checked using OCSP, CRLDP, or the CRLs cached by the attached KeystoreService instance, depending on the value of the Revocation Check Mode parameter. If not enabled, a revocation check will not be performed."/>
```
MMaiero: > a handshake will fail

the following handshakes will fail.

> If not enabled, a revocation check will not be performed.

the revocation check

MMaiero (approved):

MMaiero:
Can you also create the back port for Release 5.0.0? Thanks!

PR #3414, opened 2021-06-15
[Backport] Updated README.MD with the default exposed port.

salvatore-coppola:
Updated README.MD with the default exposed port.


MMaiero:
I believe there is an issue with this branch. It should only contain the back ported commit

PR #3432, opened 2021-06-22
Tentative fix for Alpine.

MMaiero:
Modified the way we get the keystore password.

This Pr closes an issue with Alpine Docker but more generally with every distribution. In brief, a concurrency issue could lead the web server to not have the proper keystore in time and could lead to not have Jetty running.


MMaiero:
@mstankovic verified that it is working. Merging.

PR #3455, opened 2021-06-29
Restrict pkcs8 accepted key and fix error message

salvatore-coppola:
Fixed private key regex
Fixed apply button event-handling duplication.
Now more runtime exceptions are handled by GwtCertificatesServiceImpl.

### DIFF
--- a/kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/client/ui/security/KeyPairTabUi.java
+++ b/kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/client/ui/security/KeyPairTabUi.java
@@ -21,7 +21,6 @@ import org.eclipse.kura.web.client.ui.validator.PEMValidator;
 import org.eclipse.kura.web.client.ui.validator.PKCS8Validator;
 import org.eclipse.kura.web.client.ui.validator.StringLengthValidator;
 import org.eclipse.kura.web.client.ui.validator.StringNotInListValidator;
-import org.eclipse.kura.web.client.util.FailureHandler;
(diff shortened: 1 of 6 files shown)


salvatore-coppola:
@nicolatimeus if you want to check

kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/client/ui/security/KeyPairTabUi.java
```diff
@@ -219,21 +225,42 @@ private void storeKeyPair() {
...
+                            public void onSuccess(Void result) {
+                                reset();
+                                setDirty(false);
+                                KeyPairTabUi.this.listener.onKeystoreChanged();
+                            }
+
+                            @Override
+                            public void onFailure(Throwable caught) {
+                                caught.printStackTrace();
+                                FailureHandler.handle(caught);
```
nicolatimeus: Probably this is not needed

kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/client/ui/security/KeyPairTabUi.java
```diff
@@ -211,6 +216,7 @@ private void initForm(final List<String> keyStorePids, List<String> usedAliases)
                     storeCertificate();
                 }
                 this.listener.onApply(isValid);
+                this.applyHandler.removeHandler();
```
nicolatimeus: We should make sure that no handlers are leaked

mstankovic:
verified as implemented.

nicolatimeus (approved):

PR #3470, opened 2021-07-05
Enable receive timeout for modem serial communication

nicolatimeus:

Enables serial receive timeout for communication with modems. The receive timeout is set to 5 seconds by default and can be customized using the newly introduced `net.admin.modem.receive.timeout.ms` property in `kura.properties`.
Given the current implementation of modem communication that relies heavily on the `InputStream.available()` method, the timeout should never trigger, but this PR enables it as a safety measure to avoid potential reads without a timeout.
The behavior of the previous implementation can be recovered by setting the`net.admin.modem.receive.timeout.ms` property to a value lesser or equal than zero.


MMaiero:
Merging as it is a refactoring of the existing code. No tests are available right now. We will add them in the next sprints

MMaiero (approved):

PR #3489, opened 2021-07-13
[Backport] Deprecated SafeProcess and related classes

pierantoniomerlino:
Backport of #3488 


kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/linux/util/LinuxProcessUtil.java
```diff
@@ -34,6 +34,10 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+/**
+ * @deprecated since 2.2 in favor of {@link org.eclipse.kura.executor.CommandExecutorService}
```
MMaiero: I'm not sure if it is correct in the core bundle to mention the API version. In theory those packages are separate and not related.

pierantoniomerlino: Indeed these are not "official" APIs, so they aren't related to the version...

MMaiero (approved):

PR #3503, opened 2021-07-21
Disable hwclock update with chrony.

salvatore-coppola:
Now onClockUpdate accepts a parameter to handle hardware clock update.


nicolatimeus (approved):

nicolatimeus:
If not done already, I think we should update the documentation mentioning that if chrony is used, Kura will not change system time and/or update hardware clock, so that the user is aware of this.

salvatore-coppola:
> 
> 
> If not done already, I think we should update the documentation mentioning that if chrony is used, Kura will not change system time and/or update hardware clock, so that the user is aware of this.

I've updated the documentation in this pr: #3481

PR #3525, opened 2021-08-03
Changed ownership and permission for named files

pierantoniomerlino:

This PR fixes the following error in the named logs:

```
named[16289]: open: /etc/bind/rndc.key: permission denied
```


MMaiero:
Is it needed just in those 2 platforms?

pierantoniomerlino:
@MMaiero The other profiles are -nn and strangely in the `intel-up2-centos-7` we don't copy any named files...

MMaiero:
Is that due to a bug?

MMaiero (approved):

PR #3528, opened 2021-08-04
Removed string trims in input and password validation

marcellorinaldo:

Brief description of the PR. Removed string trim before validation.

**Related Issue:** N/A.

**Description of the solution adopted:** Trimming the value before validation was not safe in certain conditions. An error is now shown if a user inserts spaces before and after the values.
Removed trim also when checking if password is empty: a password consisting of only whitespaces is no more identified as an empty one.

**Screenshots:** N/A.


MMaiero:
I believe this was changed due to other issues. Can you check? Also, probably a password of spaces makes no sense

PR #3531, opened 2021-08-05
Clock service refactor

MMaiero:
Brief description of the PR. This PR tries to refactor the clock service to have a single executor running. Plus a more aligned config management with identification of changes

**Related Issue:** N/A

**Description of the solution adopted:** N/A

**Screenshots:** N/A

**Any side note on the changes made:** N/A


kura/org.eclipse.kura.linux.clock/src/main/java/org/eclipse/kura/linux/clock/AbstractNtpClockSyncProvider.java
```diff
@@ -42,12 +39,19 @@
...
     @Override
-    public void init(Map<String, Object> properties, ClockSyncListener listener) throws KuraException {
-        this.properties = properties;
+    public void init(ClockServiceConfig clockServiceConfig, ScheduledExecutorService scheduler,
+            ClockSyncListener listener) throws KuraException {
+        this.scheduler = scheduler;
         this.listener = listener;
-
-        readProperties();
+        this.ntpHost = clockServiceConfig.getNtpHost();
```
nicolatimeus, line 49: Since `ClockServiceConfig` is immutable, we could also define a `clockServiceConfig` field for storing it instead of the multiple per property fields, but it is fine also as is.

MMaiero: I don't think we gain so much if we keep a single field. Maybe it is more clear, being an abstract class to have the fields separate. Don't know, I think that both solutions work fine.

nicolatimeus (approved):

PR #3534, opened 2021-08-19
Some improvements in the command executor service

pierantoniomerlino:
Some improvements on the Command Executor Service:

- the `stopStreamHandler`method is added to explicitly close the streams
- the Command Executor Service tests are replaced


kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/internal/linux/executor/ExecutorUtil.java
```diff
@@ -278,15 +295,19 @@ private static DefaultExecutor configureExecutor(Command command) {
         return executor;
     }
 
-    private static void executeAsync(Command command, CommandLine commandLine, Consumer<CommandStatus> callback) {
+    protected Executor getExecutor() {
```
MMaiero, line 298: Does this need to be protected or maybe better private?

kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/internal/linux/executor/ExecutorUtil.java
```diff
@@ -358,4 +380,8 @@ private static CommandLine buildPrivilegedCommand(Command command) {
         }
         return commandLine;
     }
+
+    protected ByteArrayOutputStream createStream() {
```
MMaiero, line 384: Also for this one

kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/internal/linux/executor/ExecutorUtil.java
pierantoniomerlino: I used it only for injecting a mock for testing. I wasn't able to make it work when configured as private.

MMaiero (approved):

PR #3558, opened 2021-09-01
Wifi 5Ghz support

salvatore-coppola:
Removed channel filter in Wifi hotspot list.


kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/server/GwtNetworkServiceImpl.java
```diff
@@ -1018,8 +1016,7 @@ public void updateNetInterfaceConfigurations(GwtXSRFToken xsrfToken, GwtNetInter
                     // if(!ssid.matches("[0-9A-Za-z/.@*#:\\ \\_\\-]+")){
                     // ssid= null;
                     // }
-                    if (wifiHotspotInfo.getChannel() <= systemService.getKuraWifiTopChannel() && ssid != null
-                            && !ssid.equals(wirelessSsid)) {
+                    if (ssid != null && !ssid.equals(wirelessSsid)) {
```
MMaiero, line 1019: Without this check are we still compatible with the regulations, for example, in Japan or US?

salvatore-coppola: I believe yes because in station mode we only work with frenquncies that hotspot use, so it's their responsability to complain with the regulations (or not ? ). Obviously the max channel parameter become useless for the station mode,

PR #3561, opened 2021-09-02
MQTT birth certificate extended properties providing via API

### LINKED ISSUE #3553: API to externally provide extended properties for the MQTT birth certificate
Currently, extended properties for the MQTT birth certificate can be provided by implementing the `SystemServiceImpl.getExtendedProperties` method. This requires modifying a Kura class, and there is no architectural way to provide properties from external bundles. It would be great if external bundles could provide properties for the birth certificate in an easy way not involving Kura manipulation.

Create a new interface in the Kura API bundle which any service in any bundle can implement to provide its own properties. The interface should have a single method that returns the properties. A registry should globally track all services implementing this new interface, providing a method that returns their instances. The `SystemServiceImpl.getExtendedProperties` method, which currently returns no property, should instead be implemented to use the registry to get all the properties holding services, and collect the properties from them when required, passing them to the underlying methods.

**Additional context**
We already implemented the solution and can provide a pull request as soon as possible.

robertodamiani:
Signed-off-by: Roberto Damiani <ing.roberto.damiani@gmail.com>

This PR provides a new API interface to provide MQTT birth certificate extended properties from third-party external bundles. Any time a birth certificate has to be created, Kura will collect all properties from all services implementing the new interface and will add them to the birth certificate.

**Related Issue:** This PR fixes/closes #3553

* Added the `ExtendedPropertiesHolder` interface to the API bundle, which can be implemented by third-party bundles to provide extended properties that will be added to the MQTT birth certificate;
* Added the `ExtendedPropertiesService` service to obtain the current properties, and added the default implementation `ExtendedPropertiesServiceImpl` that tracks the `ExtendedPropertiesHolder` services and collects all properties from them, passing the properties to the calling methods;
* Implemented the `SystemServiceImpl.getExtendedProperties` method to obtain the extended properties from the `ExtendedPropertiesService`, handling the case in which the service is not registered;
* Added a new event, implemented by the `ExtendedPropertiesEvent`, which notifies that properties have changed;
* Added the new event handling in the cloud service, so that birth certificate is retransmitted when properties change.


nicolatimeus:
Hi @robertodamiani , thank you for your contribution!
Have you already considered using `ServiceTracker` [1], or a dynamic Declarative Services reference with 0..n cardinality [2] for implementing the tracking of the `ExtendedPropertiesHolder` instances?

[1] http://docs.osgi.org/javadoc/r4v42/org/osgi/util/tracker/ServiceTracker.html
[2] http://docs.osgi.org/specification/osgi.cmpn/7.0.0/service.component.html

robertodamiani:
Hi @nicolatimeus, thank you for your feedback.

The registry architecture was surely too complex and unnecessary, so I agree that the solutions you suggest are more appropriate.
For simplicity, I used the solution [2], and I dropped the registry service and used a simple 0..n reference to track the `ExtendedPropertiesHolder` services.

I pushed the new solution for you to review again and I updated the PR description.
 
Thank you.

robertodamiani:
Hi @nicolatimeus,
I just updated the PR fixing the develop branch conflict.

Thank you.

robertodamiani:
hi @nicolatimeus,
I added a unit test for the feature to the PR.

Thank you.

nicolatimeus:
Sorry for the late reply, the changes look good, I have only the following two concerns: 

1. The modifications do not allow to change the version of the returned properties.
2. Since the process of tracking the `ExtendedPropertiesHolder` instances is dynamic, the system service might not have tracked all of the expected `ExtendedPropertiesHolder` instances when a birth certificate is created.

Regarding the first point, one possible approach could be allowing the SystemService to delegate the implementation of the`getExtendedProperties()` method to a class that should be included in a fragment that can be attached to the system service provider bundle.
We could then move your current implementation to a "default" fragment that we can include in Kura distribution. In this way a user could provide a custom fragment implementation that could allow to completely control the version and also eventually. having "static" extended properties solving the second problem.
Regarding the second point, should we define an OSGi event that notifies that extended properties have changed that can be received by cloud connector implementations triggering a birth republish?

robertodamiani:
Hi @nicolatimeus,
I understood the problems and I agree with the solutions.

I just would like to ask you what do you mean by "default fragment".
Are you talking about the `SystemService` interface `getExtendedProperties` method, which should have a default implementation in the interface itself?


Thanks.

robertodamiani:
I think you mean this:
https://www.ibm.com/docs/en/radfws/9.6?topic=SSRTLW_9.6.0/com.ibm.aries.osgi.doc/topics/cbundlefragment.html

I have a look and try to design a solution as a fragment.

I'll update the PR as soon as possible.

robertodamiani:
Hi @nicolatimeus,

I just updated the pull request implementing the additional features:
* the logics to collect properties from the `ExtendedPropertiesHolder` have been moved to the `ExtendedPropertiesService` service;
* the concrete implementation of the new service has been added to the new module `org.eclipse.kura.core.extendedproperties`, which can be eventually replaced;
* an OSGI event has been added to notify that the properties have changed;
* the new event is handled by the cloud service and the birth certificate is retransmitted.

For the solution I did not use OSGI fragments as you suggested as I'm not experienced enough, but if you think it is a better solution I can rework the code again.

Let me know any thoughts.

Thank you.

nicolatimeus:
Hi @robertodamiani , the approach looks fine to me, a couple of comments:

* Changing the `org.eclipse.kura.core.system` package in a backwards compatible way requires to increase the minor version of that package in the manifest file.
* Please make sure that all of the new types inside the API bundle are annotated either with the ProviderType or ConsumerType annotation, according to this guide https://github.com/eclipse/kura/wiki/Kura-Semantic-Versioning
  * We usually prefer to place provider and consumer types in different api packages
* I saw that there were test failures, can you please have a look at that (I tried to restart the build).

PR #3569, opened 2021-09-13
Removed logging of properties on CommandCloudApp update

MMaiero:


kura/org.eclipse.kura.core.cloud/src/main/java/org/eclipse/kura/cloud/app/command/CommandCloudApp.java
```diff
@@ -140,7 +140,7 @@ protected void activate(Map<String, Object> properties) {
     }
 
     public void updated(Map<String, Object> properties) {
-        logger.info("updated...: {}", properties);
+        logger.info("Request handler {} updated...");
```
pierantoniomerlino: Why the {}?

pierantoniomerlino (approved):

PR #3573, opened 2021-09-14
When running into a timeout, report an error

ctron:
When the installer fails with a timeout, it still reports "success" to the caller.

Now, the installer scripts returns "2", which indicates an error.

### DIFF
--- a/kura/distrib/src/main/sh/extract.sh
+++ b/kura/distrib/src/main/sh/extract.sh
@@ -57,6 +57,7 @@ function run_kura_install {
             if [ "$DELTA" -ge "$TIMEOUT_TIME" ]; then
                 echo "The installation process is not responding. It'll be stopped."
                 kill -9 $PID >> /dev/null 2>&1
+                exit 2
             fi
             sleep $REFRESH_TIME
         done
@@ -71,6 +72,7 @@ function run_kura_install {
             if [ "$DELTA" -ge "$TIMEOUT_TIME" ]; then
                 echo "The installation process is not responding. It'll be stopped."
                 kill -9 $PID >> /dev/null 2>&1
+                exit 2
             fi
             sleep $REFRESH_TIME
         done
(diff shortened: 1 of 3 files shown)


kura/distrib/src/main/sh/extract.sh
```diff
@@ -57,6 +57,7 @@ function run_kura_install {
             if [ "$DELTA" -ge "$TIMEOUT_TIME" ]; then
                 echo "The installation process is not responding. It'll be stopped."
                 kill -9 $PID >> /dev/null 2>&1
+                exit 2
```
pierantoniomerlino, line 60: Is it better to add the `stop_watchdog` call also before the `exit 2` command?

pierantoniomerlino:
I totally agree that the installer has to return an error code when timeout. According to [this](https://tldp.org/LDP/abs/html/exitcodes.html), why use exit code 2? I admit that I don't know which value can be used. Maybe 128+9 since we call `kill -9`?

ctron:
> I totally agree that the installer has to return an error code when timeout. According to [this](https://tldp.org/LDP/abs/html/exitcodes.html), why use exit code 2? I admit that I don't know which value can be used. Maybe 128+9 since we call `kill -9`?

Well, `exit 1` is being used for missing dependencies. The next free one seemed `2`.

pierantoniomerlino:
@ctron Ah ok. Now I understand the logic. What  do you think about the comment on the `stop_watchdog` call?

kura/distrib/src/main/sh/extract.sh
ctron: I am not sure. I guess when the script terminates, the watchdog would end too.

pierantoniomerlino: I don't think so. If not stopped or continuously kicked, the watchdog eventually reboot the device.

ctron: Isn't that a good idea in this case?

pierantoniomerlino: Not sure... a reboot would not solve the error.

PR #3574, opened 2021-09-15
Added password decryption in NetworkConfigurationService

marcellorinaldo:

Brief description of the PR. On password-related properties, the value is checked to verify that the content is decrypted. If so, the value will be decrypted before creating the network configuration.

**Related Issue:** N/A.

**Description of the solution adopted:** Before this PR, the properties that represent passwords might be received encrypted in the NetworkConfigurationService. In that case, the property is simply applied without further checks creating a non-functioning network configuration. With this PR, such properties are decrypted before application.

**Screenshots:** N/A.


MMaiero:
@marcellorinaldo Can we try to satisfy sonar?

marcellorinaldo:
> @marcellorinaldo Can we try to satisfy sonar?

Sure.

MMaiero (approved):

PR #3601, opened 2021-09-24
Added scale factor parameter in BaseAsset

### LINKED ISSUE #2312: Asset configuration: Unit factor field
We've recently started using Kura to read industrial devices on some plant we are monitoring to evaluate a full implementation in our systems but found a difficulty in variables processing. Most of the industrial devices we read on our systems requires some scale on the value read to convert to the proper unit:

- Modbus register: 42045 (cV)
- Value we intend to read: 420,45 V

We have solved the issue with an intermediate specific component that process this scaling, but we think the real solution would be to include it in the Asset configuration so that any driver implemented by third party developers have to take this in consideration.

The final solution we are proposing would be a field on _BaseChannelDescriptor_ named similar to value.scale or value.factor with default value to 1 that is multiplied by the value read from the industrial device right before sending the value to the next wire component in case the value is numeric

pintify:
Added scale factor parameters in BaseAsset

**Related Issue:** This PR fixes/closes #2312

**Description of the solution adopted:** Added a scale factor for the numeric variables so that the compatible drivers include by default support to scale the value.

### DIFF
--- a/kura/org.eclipse.kura.asset.provider/src/main/java/org/eclipse/kura/asset/provider/AssetConstants.java
+++ b/kura/org.eclipse.kura.asset.provider/src/main/java/org/eclipse/kura/asset/provider/AssetConstants.java
@@ -43,7 +43,10 @@ public enum AssetConstants {
     TYPE(CHANNEL_DEFAULT_PROPERTY_PREFIX.value() + "type"),
 
     /** Value type Property to be used in the configuration. */
-    VALUE_TYPE(CHANNEL_DEFAULT_PROPERTY_PREFIX.value() + "value.type");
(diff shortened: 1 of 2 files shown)


pintify:
Sorry, I really don't know what I must do to overcome the given error:

imagen

MMaiero:
Hello, as you can see there are some errors with the ECA that needs to be signed, then also some tests related to the asset are failing.
Since there are new code additions, I believe you will need to add few tests.

MMaiero:
Scaling factor has been added to Kura recently. Please have a look if that satisfies your needs. Best regards

pintify:
As far as I see on source, yes, it should be perfect. Thank you!

PR #3606, opened 2021-09-24
Now SslContext cache is cleaned when KeystoreService in SSLManager is…

### LINKED ISSUE #3605: Updating KeystoreService in SSLManagerService doesn't trigger the sslContext update.
Updating KeystoreService in SSLManagerService doesn't trigger the sslContext update., this results in a wrong sslfactory retrieved when another service try to get it.

salvatore-coppola:
Now SslContext cache is cleaned when KeystoreService in SSLManager is updated.

This close #3605

### DIFF
--- a/kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/ssl/SslManagerServiceImpl.java
+++ b/kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/ssl/SslManagerServiceImpl.java
@@ -99,6 +99,8 @@ public class SslManagerServiceImpl implements SslManagerService, ConfigurableCom
         this.keystoreService = keystoreService;
         this.keystoreServicePid = Optional.of((String) properties.get(ConfigurationService.KURA_SERVICE_PID));
 
+        this.clearSslContexCache();
+
         if (this.sslServiceListeners != null) {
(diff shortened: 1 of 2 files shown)


kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/ssl/SslManagerServiceImpl.java
```diff
@@ -107,6 +109,14 @@ public void setKeystoreService(KeystoreService keystoreService, final Map<String
...
         if (this.keystoreService == keystoreService) {
+
+            this.clearSslContexCache();
+
+            if (this.sslServiceListeners != null) {
+                // Notify listeners that service has been updated
+                this.sslServiceListeners.onConfigurationUpdated();
+            }
+
             this.keystoreService = null;
```
nicolatimeus: I think the listeners should be called after setting the KeystoreService to `null`, otherwise the old keystore will be still be available when they are called.

salvatore-coppola: you are right

nicolatimeus (approved):

PR #3623, opened 2021-09-29
Allow suppressing the start of the bluetooth daemon

ctron:
This change allows suppressing the start of the bluetooth daemon.

This functionality already exists in the context of snap packages. However, this PR makes this a more generic functionality, allowing container to use this too.

### DIFF
--- a/kura/org.eclipse.kura.ble.provider/src/main/java/org/eclipse/kura/internal/ble/BluetoothLeServiceImpl.java
+++ b/kura/org.eclipse.kura.ble.provider/src/main/java/org/eclipse/kura/internal/ble/BluetoothLeServiceImpl.java
@@ -46,7 +46,7 @@ public class BluetoothLeServiceImpl implements BluetoothLeService {
 
     protected void activate(ComponentContext context) {
         logger.info("Activating Bluetooth Le Service...");
-        if (!startBluetoothUbuntuSnap() && !startBluetoothSystemd() && !startBluetoothInitd()) {
+        if (!startBluetoothSuppressed() && !startBluetoothUbuntuSnap() && !startBluetoothSystemd() && !startBluetoothInitd()) {
             startBluetoothDaemon();
         }
         try {
@@ -108,6 +108,11 @@ public class BluetoothLeServiceImpl implements BluetoothLeService {
         } else {
             return false;
         }
+    }
+
+    private boolean startBluetoothSuppressed() {
+        // Allow to disable the bluetooth service start, e.g. when running inside a container
+        return Boolean.getBoolean("kura.ble.suppressBluetoothDaemonStart");
     }
 
     private boolean execute(String commandLine) {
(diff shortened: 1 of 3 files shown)


pierantoniomerlino:
@ctron I agree that this is a good functionality and should be added to the code. I've a couple of thoughts:

1. if the bluetooth service is already started, Kura will not stop it.
2. why not use the SystemService and eventually put the property in the kura.properties file? This is the way Kura has managed properties until now.

ctron:
> * if the bluetooth service is already started, Kura will not stop it.
> 

If it didn't start it, it also shouldn't stop it.

> * why not use the SystemService and eventually put the property in the kura.properties file? This is the way Kura has managed properties until now.

I did align that with the snap variant, which also uses `System.getProperty`. I assumed there was a reason behind that approach (like startup order). So is it ok to add a reference to the system service for this?

pierantoniomerlino:
@ctron It's true that the snap variant uses this approach, but probably it was an error reviewing the PR :-)
So, please, use the SystemService if possible.

ctron:
I updated the PR to use the system service.

kura/org.eclipse.kura.ble.provider/src/main/java/org/eclipse/kura/internal/ble/BluetoothLeServiceImpl.java
```diff
@@ -110,6 +122,12 @@ private boolean startBluetoothUbuntuSnap() {
         }
     }
 
+    private boolean startBluetoothSuppressed() {
+        // Allow to disable the bluetooth service start, e.g. when running inside a container
+        String value = this.systemService.getProperties().getProperty("kura.ble.suppressBluetoothDaemonStart");
+        return Boolean.parseBoolean(value);
```
pierantoniomerlino, line 128: What happens if the property is not defined? Is it better to add a default value instead of return null?

ctron: I don't think making this more complex makes any sense: https://docs.oracle.com/javase/7/docs/api/java/lang/Boolean.html#parseBoolean(java.lang.String)

> The boolean returned represents the value true if the string argument is not null and is equal, ignoring case, to the string "true".

So "null" is already handled, as "false".

pierantoniomerlino: That's fine. Could you please take a look at the failing tests?

ctron:
I updated the tests. And can so no test failures with the CI. But it still shows up red?!

MMaiero:
From the latest build I see this error:

```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.3:testCompile (compiletests) on project org.eclipse.kura.internal.ble.test: Compilation failure
[ERROR] /home/jenkins/agent/workspace/multibranch_PR-3623/kura/kura/test/org.eclipse.kura.internal.ble.test/src/test/java/org/eclipse/kura/internal/ble/BluetoothLeServiceImplTest.java:[60,57] cannot find symbol
[ERROR]   symbol:   class Properties
[ERROR]   location: class org.eclipse.kura.internal.ble.BluetoothLeServiceImplTest
[ERROR] -> [Help 1]
[ERROR] 
[ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch.
[ERROR] Re-run Maven using the -X switch to enable full debug logging.
[ERROR] 
[ERROR] For more information about the errors and possible solutions, please read the following articles:
[ERROR] [Help 1] http://cwiki.apache.org/confluence/display/MAVEN/MojoFailureException
[ERROR] 
[ERROR] After correcting the problems, you can resume the build with the command
[ERROR]   mvn <args> -rf :org.eclipse.kura.internal.ble.test
```

pierantoniomerlino (approved):

PR #3637, opened 2021-10-08
Fix wpa encoded password save

salvatore-coppola:
Added a check before the password validation.
Now the validation is skipped if the password is encoded.


salvatore-coppola:
@marcellorinaldo could you give a look ?

marcellorinaldo:
> @marcellorinaldo could you give a look ?

Yes

marcellorinaldo (approved):
Tested on RaspberryPi 4, everything works as expected.

kura/org.eclipse.kura.net.admin/src/main/java/org/eclipse/kura/net/admin/visitor/linux/WpaSupplicantConfigWriter.java
```diff
@@ -325,28 +325,30 @@ private String updateWPA(WifiConfig wifiConfig, String fileAsString) throws Kura
         String result = fileAsString;
         String passKey = new String(wifiConfig.getPasskey().getPassword());
 
-        boolean checkAndEnconde = true;
+        boolean doCheckAndEnconde = true;
```
MMaiero: Encode, right?

salvatore-coppola: yep

kura/org.eclipse.kura.util/src/main/java/org/eclipse/kura/util/base/TypeUtil.java
```diff
@@ -63,4 +63,51 @@ private TypeUtil() {
         return b.toByteArray();
     }
 
+    /**
```
MMaiero, line 66: Do we have to update the exported package version in the OSGi manifest since we are adding a new public method?

salvatore-coppola: yes

kura/org.eclipse.kura.net.admin/META-INF/MANIFEST.MF
```diff
@@ -42,7 +43,7 @@ Import-Package: javax.microedition.io,
  org.eclipse.kura.security;version="[1.1,2.0)",
  org.eclipse.kura.system;version="[1.0,2.0)",
  org.eclipse.kura.usb;version="[1.0,2.0)",
- org.eclipse.kura.crypto;version="[1.1,2.0)",
+ org.eclipse.kura.util.base;version="[1.0,2.0)",
```
MMaiero: I believe we need the parseHexBinary right? So the correct import should be [1.1,2.0)?

salvatore-coppola: you are right

MMaiero (approved):

PR #3643, opened 2021-10-12
fixed channel parsing.

salvatore-coppola:
replaced the channel in ht capabilites with the one in DS parameter set


salvatore-coppola:
@MMaiero this fix must be backported in kura 5

pierantoniomerlino (approved):

PR #3647, opened 2021-10-13
Enabled "Apply" button in identities UI when changing from password enabled to disabled

marcellorinaldo:

Brief description of the PR. Enabled "Apply" button in identities UI when changing from password enabled to disabled. Previously, if an identity configuration was modified by disabling the password, the "Apply" button won't be enabled.
In order to enable it the user had to alter another configuration parameter (e.g. assigning and removing a permission).

**Related Issue:** N/A.

**Description of the solution adopted:** Change listener is triggered when switching from password enabled to disabled.

**Screenshots:** N/A.


MMaiero:
Do we plan to back port to 5?

nicolatimeus (approved):

PR #3667, opened 2021-10-29
Raspberry 3 & 4 ubuntu20 profile fixes

salvatore-coppola:
Added gpio required dependency and fixed pins name.


salvatore-coppola:
@marcellorinaldo has tested the pr on its own device.

MMaiero (approved):

PR #3704, opened 2021-12-09
Added UI support for JSON snapshot format

nicolatimeus:

* Adds a new dialog that is shown in the Settings -> Snapshots and in the Wires UI that is shown when a snapshot download is requested, it allows to choose between the existing XML format and the new CONF-V2 JSON format.
* Snapshots downloaded in JSON format can be applied using the web ui in Settings -> Snapshots like XML snapshots, or using `CONF-V2/PUT/configurableComponents/configurations/_update`.
* Changed CONF-V2 methods to use `true` as default value for the `takeSnapshot` parameter if not specified


MMaiero:
@marcellorinaldo @mstankovic  Can you please have a review?

marcellorinaldo (approved):

marcellorinaldo:
Tested on Raspberry Pi4, everything seems ok.

PR #3711, opened 2021-12-13
Fix deb dependencies during multiple profile building

salvatore-coppola:
This changes fix the problem with the build when multiple profiles are used.
Before these changes all deb files used the same nvidia-jetson-nano dependencies.


mstankovic:
seems good to me. There are no more errors related to iptables and named services. All networking tests are passing at least for Rpi3

salvatore-coppola:
@MMaiero this pr could be reviewed

kura/distrib/src/main/deb/control-nvidia-jetson-nano/prerm
```diff
@@ -0,0 +1,114 @@
+#!/bin/bash
```
MMaiero: Would it be better if we have those files here in the specific target device folder like in here https://github.com/eclipse/kura/tree/develop/kura/distrib/src/main/resources/nvidia-jetson-nano?
What do you think?

salvatore-coppola: ok, but I need to do another round of tests to be sure there aren't broken paths to those files.

MMaiero (approved):

MMaiero:
The changes go over 1000 lines, but they are mainly due to the duplication of files needed to keep up with different dependencies for different profiles.

PR #3715, opened 2021-12-15
Add log UI new window button

marcellorinaldo:
Brief description of the PR. Added a new button in the "Device Logs" section to open a new Window in browser.

**Related Issue:** N/A.

**Description of the solution adopted:** This button can be handy when trying to debug some functionalities, having in separate windows the logs in real-time and the Kura UI. The button allows to open a new browser window and login with the current user with just one click.

**Screenshots:** N/A.

### DIFF
--- a/kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/client/ui/device/LogTabUi.java
+++ b/kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/client/ui/device/LogTabUi.java
@@ -77,7 +77,7 @@ public class LogTabUi extends Composite {
     @UiField
     CheckBox showMoreInfoCheckbox;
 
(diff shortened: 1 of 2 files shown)


MMaiero:
Is it possible to open the new window in the system logs tab? Or to have in the new window just the text area with the logs?

marcellorinaldo:
> Is it possible to open the new window in the system logs tab?

No, as of now we cannot know which page we have visited using GWT.

> Or to have in the new window just the text area with the logs?

It **might** be possible to embed custom javascript code into GWT. But that requires further investigation and discussion (is it possible to make Java RPC's using native javascript? If not, I think we need to use secure WebSockets).

The PR provided here is the simplest solution with what we have at our disposal right now.

MMaiero (approved):

PR #3740, opened 2022-01-11
Added Support for Gpsd

salvatore-coppola:
Added Support for Gpsd through `de.taimos:gpsd4java:1.10`


kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/position/PositionService.java
```diff
@@ -79,8 +79,10 @@
     public boolean isLocked();
 
     /**
-     * Returns the last sentence received from the gps.
+     * @deprecated
+     *             Returns the last sentence received from the gps.
```
MMaiero, line 108: Do we have an API method to use instead? If so, maybe let's reference it here.

MMaiero: Le's add also from when the API has been deprecated. It will help also with removals

kura/org.eclipse.kura.linux.position/OSGI-INF/gpsdPositionPositionProvider.xml
```diff
@@ -0,0 +1,24 @@
...
+   available under the terms of the Eclipse Public License 2.0
+   which is available at https://www.eclipse.org/legal/epl-2.0/
+ 
+    SPDX-License-Identifier: EPL-2.0
+    
+    Contributors:
+     Eurotech
+
+-->
+<scr:component xmlns:scr="http://www.osgi.org/xmlns/scr/v1.1.0" enabled="true" immediate="false" name="GpsdPositionProvider">
```
MMaiero: The name of the component definition file has position 2 times. Is that expected?

kura/org.eclipse.kura.linux.position/OSGI-INF/serialDevicePositionProvider.xml
```diff
@@ -0,0 +1,30 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+
+   Copyright (c) 2011, 2021 Eurotech and/or its affiliates and others
```
MMaiero: It's a new file so the copyright header should have only 2022

kura/org.eclipse.kura.linux.position/src/main/java/org/eclipse/kura/linux/position/GpsdPositionProvider.java
```diff
@@ -0,0 +1,286 @@
+/*******************************************************************************
+ * Copyright (c) 2011, 2021 Eurotech and/or its affiliates and others
```
MMaiero: Only 2022 since it is a new file

kura/org.eclipse.kura.linux.position/src/main/java/org/eclipse/kura/linux/position/PositionProvider.java
```diff
@@ -0,0 +1,58 @@
+package org.eclipse.kura.linux.position;
```
MMaiero, line 14: copyright header missing

kura/org.eclipse.kura.linux.position/src/main/java/org/eclipse/kura/linux/position/PositionProvider.java
```diff
@@ -0,0 +1,58 @@
...
+
+public interface PositionProvider {
+
+    public void start();
+
+    public void stop();
+
+    public Position getPosition();
+
+    /**
```
MMaiero: If this is a new interface, not clear the deprecations.

kura/org.eclipse.kura.linux.position/src/main/java/org/eclipse/kura/linux/position/PositionProviderType.java
```diff
@@ -0,0 +1,33 @@
+package org.eclipse.kura.linux.position;
```
MMaiero, line 14: copyright header missing

kura/org.eclipse.kura.linux.position/src/main/java/org/eclipse/kura/linux/position/SerialDevicePositionProvider.java
```diff
@@ -0,0 +1,246 @@
+/*******************************************************************************
+ * Copyright (c) 2011, 2021 Eurotech and/or its affiliates and others
```
MMaiero: 2022 if this is a new file

kura/test/org.eclipse.kura.linux.position.test/src/test/resources/gpsd-raw-json-boltgate-10-12-el27.1.txt
```diff
@@ -0,0 +1,141 @@
+{"class":"VERSION","release":"3.22","rev":"3.22","proto_major":3,"proto_minor":14}
```
MMaiero: I would anonymise the name of such files

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/position/PositionService.java
salvatore-coppola: > Do we have an API method to use instead? If so, maybe let's reference it here.

No, the idea is to now allow to access directly to the underlying position provider.

salvatore-coppola: > Le's add also from when the API has been deprecated. It will help also with removals

I've added since 2.2.0

kura/org.eclipse.kura.linux.position/src/main/java/org/eclipse/kura/linux/position/PositionProvider.java
salvatore-coppola: to maintain the  compatibility with SerialDevicePositionProvider  than now implements Position Provider.

nicolatimeus: Maybe we can split this in two internal interfaces? For example PositionProvider, which only allows to get the position and time, and NMEAPositionProvider that provides the NMEA-related informations.

kura/org.eclipse.kura.linux.position/src/main/java/org/eclipse/kura/linux/position/GpsdPositionProvider.java
```diff
@@ -0,0 +1,290 @@
...
+    private AtomicReference<Double> longitude = new AtomicReference<>();
+    private AtomicReference<Double> longitudeError = new AtomicReference<>();
+    private AtomicReference<Double> altitude = new AtomicReference<>();
+    private AtomicReference<Double> altitudeError = new AtomicReference<>();
+    private AtomicReference<Double> speed = new AtomicReference<>();
+    private AtomicReference<Double> speedError = new AtomicReference<>();
+    private AtomicReference<Double> course = new AtomicReference<>();
+    private AtomicReference<Double> courseError = new AtomicReference<>();
+    private AtomicReference<Double> timestamp = new AtomicReference<>();
+    private AtomicReference<ENMEAMode> mode = new AtomicReference<>();
```
nicolatimeus: Could it make sense to define an internal "State" class containing all of these fields? In this way we can keep a single `AtomicReference<State>` here and have the whole state updated atomically. Otherwise it is possible that different fields come from different `TPVObject` instances due to race conditions.

kura/org.eclipse.kura.linux.position/src/main/java/org/eclipse/kura/linux/position/PositionProvider.java
salvatore-coppola: > Maybe we can split this in two internal interfaces? For example PositionProvider, which only allows to get the position and time, and NMEAPositionProvider that provides the NMEA-related informations.

I can do that, but also GpsdPositionProvider has to implement NMEAPositionProvider interface because PositionServiceImpl invoke the getNMEAPosition, unless i do a check on runtime class that I find an ugly solution.

kura/org.eclipse.kura.linux.position/src/main/java/org/eclipse/kura/linux/position/GpsdPositionProvider.java
salvatore-coppola: Looking at the current Gpsd4Java implementation it seems hard the race condition will happen, however it is a good idea. In the future we have to notice if a field is related to the others before add it in the internal state.

kura/org.eclipse.kura.linux.position/src/main/java/org/eclipse/kura/linux/position/GpsdPositionProvider.java
```diff
@@ -0,0 +1,300 @@
...
+    @Override
+    public void stop() {
+        if (this.gpsEndpoint != null) {
+            this.gpsEndpoint.stop();
+        }
+    }
+
+    @Override
+    public Position getPosition() {
+        return new Position(
```
nicolatimeus: Should we create a local variable such as
```
final GpsdInternalState currentState = this.internalState.get();
```
and then get the fields out of it? in this way we are sure that all measurements come from the same `TPVObject`.

kura/org.eclipse.kura.linux.position/src/main/java/org/eclipse/kura/linux/position/GpsdPositionProvider.java
```diff
@@ -0,0 +1,300 @@
...
+    }
+
+    @Override
+    public void handleSUBFRAME(SUBFRAMEObject subframe) {
+        // Noting to do.
+    }
+
+    @Override
+    public void handleTPV(TPVObject tpv) {
+        internalState.get().setLatitude(tpv.getLatitude());
```
nicolatimeus: Probably we should create a new `GpsdInternalState` instance here, fill it from the `TPVObject` and then "publish" it atomically with `internalState.set()`.

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/position/PositionService.java
salvatore-coppola: deprecated annotation removed

kura/org.eclipse.kura.linux.position/src/main/java/org/eclipse/kura/linux/position/GpsdPositionProvider.java
salvatore-coppola: done!

MMaiero:
Build is failing but not due to tests introduced by this PR

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/position/PositionService.java
```diff
@@ -60,27 +62,52 @@
...
+    @Deprecated
     public String getNmeaTime();
 
     /**
      * Returns the current NMEA date from ZDA sentence
+     * 
+     * @throws UnsupportedOperetaionException
+     *             {@link PositionServiceProvider} could not support this method.
+     * 
+     * @deprecated Since 2.2 use {@link #getDateTime()}. Access to underlying gps data layer is discouraged.
```
MMaiero: Is it deprecated since 2.2 or 2.3?

target-platform/p2-repo-common/pom.xml
```diff
@@ -613,6 +618,7 @@
                                 <move file="target/source/plugins/org.apache.felix.useradmin.jar" tofile="target/source/plugins/org.apache.felix.useradmin_${org.apache.felix.useradmin.version}.jar"/>
                                 <move file="target/source/plugins/bcpkix-jdk15on.jar" tofile="target/source/plugins/org.bouncycastle.bcpkix-jdk15on_${org.bouncycastle.version}.jar"/>
                                 <move file="target/source/plugins/bcprov-jdk15on.jar" tofile="target/source/plugins/org.bouncycastle.bcprov-jdk15on_${org.bouncycastle.version}.jar"/>
+                                <move file="target/source/plugins/gpsd4java.jar" tofile="target/source/plugins/de.taimos.gpsd4java_${de.taimos.gpsd4java.version}.jar"/>
```
nicolatimeus: Is this correct?

MMaiero (approved):
The contribution is above 1000 lines. But contains also gps test values useful for feature testing and that have raised the size of the PR.

nicolatimeus (approved):

MMaiero:
The CQ for the gpsd dependency is still open https://dev.eclipse.org/ipzilla/show_bug.cgi?id=23923 but we have the PMC approval and, due to the new Eclipse rules, we will merge the contribution, keeping it on hold for releases until full approval by the Eclipse IP Team

PR #3744, opened 2022-01-13
Adding property to disable ppp logging

MMaiero:

### DIFF
--- a/kura/org.eclipse.kura.core.system/src/main/java/org/eclipse/kura/core/system/SystemServiceImpl.java
+++ b/kura/org.eclipse.kura.core.system/src/main/java/org/eclipse/kura/core/system/SystemServiceImpl.java
@@ -1483,5 +1483,15 @@ public class SystemServiceImpl extends SuperSystemService implements SystemServi
 
         return false;
     }
+    
+    @Override
+    public boolean isLegacyPPPLoggingEnabled() {
+        final Optional<String> override = getProperty(KEY_LEGACY_PPP_LOGGING);
+        if (override.isPresent()) {
+            return Boolean.parseBoolean(override.get());
+        }
+
+        return false;
+    }
 
 }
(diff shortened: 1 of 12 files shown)


kura/org.eclipse.kura.core.system/src/main/java/org/eclipse/kura/core/system/SystemServiceImpl.java
```diff
@@ -1483,5 +1483,15 @@ public boolean isLegacyBluetoothBeaconScan() {
...
     }
+    
+    @Override
+    public boolean isLegacyPPPLoggingEnabled() {
+        final Optional<String> override = getProperty(KEY_LEGACY_PPP_LOGGING);
+        if (override.isPresent()) {
+            return Boolean.parseBoolean(override.get());
+        }
+
+        return false;
```
pierantoniomerlino, line 1494: Does it make sense to return true (the default value) if the property is not present?

kura/org.eclipse.kura.net.admin/src/main/java/org/eclipse/kura/net/admin/visitor/linux/PppConfigWriter.java
```diff
@@ -192,6 +218,10 @@ public String formPeerFilename(UsbDevice usbDevice) {
     }
 
     public String formPppLogFilename(UsbDevice usbDevice) {
+        if (!this.logEnabled) {
+            return null;
+        }
+
```
pierantoniomerlino: If the logEnabled is false, the filename is not written in the peers file. But, where will be written the logs?

MMaiero: PPP can write directly in the journal for systemd systems

kura/org.eclipse.kura.core.system/src/main/java/org/eclipse/kura/core/system/SystemServiceImpl.java
MMaiero: I thought about that. But considering that our targets from now on are systemd systems where the PPP daemon can directly write to the journal, I thought that maybe this property is better to be false by default and for legacy systems it can be set to true. 
But I'm open to change the default if we believe otherwise.

pierantoniomerlino: So, the default value is false (a.k.a. log in the journal) and in our profiles we override it and set to true (a.k.a. log in a separate file). It makes sense.

pierantoniomerlino (approved):

PR #3749, opened 2022-01-17
Network configuration clean up

pierantoniomerlino:
This PR tries to clean the network properties and the related snapshot.
The properties for the status of the networking have been removed from the snapshot. Now, it contains only the desidered configuration. Moreover, the monitors have been modified for logging the complete Network configuration (with current values) when an event for status changed is fired.

### DIFF
--- a/kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/net/wifi/WifiConfig.java
+++ b/kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/net/wifi/WifiConfig.java
@@ -174,8 +174,10 @@ public class WifiConfig implements NetConfig {
     }
 
     public void setPasskey(String key) {
-        Password psswd = new Password(key);
-        this.passkey = psswd;
+        if (key != null) {
+            Password psswd = new Password(key);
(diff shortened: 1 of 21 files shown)


MMaiero:
Tests are failing

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/net/wifi/WifiConfig.java
```diff
@@ -91,7 +93,21 @@ public WifiConfig(WifiMode mode, String ssid, int[] channels, WifiSecurity secur
         this.security = security;
         this.passkey = new Password(passkey);
         this.hwMode = hwMode;
-        this.broadcast = broadcast;
+        this.bgscan = bgscan;
+        this.channelFrequencies = null;
+        this.wifiCountryCode = null;
+    }
+
+    public WifiConfig(WifiMode mode, String ssid, int[] channels, WifiSecurity security, String passkey, String hwMode,
```
MMaiero, line 109: Missing since annotation

MMaiero (approved):

PR #3760, opened 2022-01-26
Added support for password change on login 

nicolatimeus:

Allows the admin to enforce password change for an identity at next login:
* A new dialog for password change will be shown after login if password change is requested
* Web ui access, most GWT RPCs and REST APIs will return unauthorized status until the password is changed
* Certificate based access is still permitted


MMaiero:
@marcellorinaldo Can you please have a review of this PR? Thanks.

kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/UserManager.java
```diff
@@ -106,6 +107,19 @@ public void requirePermissions(final String username, final String... permission
...
+        final String userRoleName = getUserRoleName(username);
+        final Role role = userAdmin.getRole(userRoleName);
+
+        if (!(role instanceof User)) {
+            throw new KuraException(KuraErrorCode.SECURITY_EXCEPTION);
+        }
+
+        final User asUser = (User) role;
+
+        return "true".equals(asUser.getProperties().get("kura.need.password.change"));
```
marcellorinaldo: I would use the constant KURA_NEED_PASSWORD_CHANGE here instead of the hard-coded value

kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/shared/model/GwtUserConfig.java
```diff
@@ -21,14 +21,17 @@
...
 
-    public GwtUserConfig(final String userName, final Set<String> permissions, final boolean isPasswordAuthEnabled) {
+    public GwtUserConfig(final String userName, final Set<String> permissions, final boolean isPasswordAuthEnabled,
+            final boolean isPasswordChangeNeeded) {
         super(userName, permissions);
         setPasswordAuthEnabled(isPasswordAuthEnabled);
+        setPasswordChangeNeeded(isPasswordChangeNeeded);
     }
 
     public boolean isPasswordAuthEnabled() {
```
marcellorinaldo, line 38: I would also add a constant PASSWORD_ENABLED_KEY for string "password.enabled"

marcellorinaldo (changes requested):

marcellorinaldo:
If minimum password length is 0 (disabled) and digits, special chars, and uppercase and lowercase letters are required, then the password change on next login is not validating such restrictions. Only the first one seems to be validated (at least one digit) and the length requirement.

MMaiero:
Just to double check, can also the admin user be forced to change its password during the first login?

nicolatimeus:
Yes, the feature works the same regardless of assigned permissions

MMaiero:
Great! Thanks!

MMaiero:
I have updated snapshots 0 to require password change by default.
If the change is ok, we can definitely merge

MMaiero (approved):

marcellorinaldo (approved):

PR #3766, opened 2022-01-28
Fix typo in kura gpio trigger

marcomatteo:
Fixed a typo in the KuraGPIOTrigger enum.

**Related Issue:** N/A

**Description of the solution adopted:** Fixed the returned value RAISING_EDGE of KuraGPIOTrigger enum with RISING_EDGE. Kept the older value in order to backward compatibility. 

**Screenshots:** N/A

**Any side note on the changes made:** Correct the directory path to Cleanup and Formatter from kura/setup to kura/setups in the CONTRIBUTING.md documentation.


kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/gpio/KuraGPIOTrigger.java
```diff
@@ -13,7 +13,7 @@
 package org.eclipse.kura.gpio;
 
 public enum KuraGPIOTrigger {
-    RAISING_EDGE,
+    RISING_EDGE,
```
MMaiero, line 21: Hello and thanks for the contribution!
This is a braking change to the APIs. 
It would be better if you could deprecate the old option and add the new one.
Implementation-wise you should try to keep the compatibility with the old option since it is deprecated but still there.

Thanks
Best regards,
Matteo

marcomatteo: Hi Matteo and thank you for the feedback!
Ok then I will add the deprecate annotation for the old option.

Kind regards,
Marco

nicolatimeus:
I think that returning the new `RISING_EDGE` variant instead of the old one in API methods is a breaking change since it is not known by existing clients.

marcomatteo:
@nicolatimeus Thanks Nicola. 
A question: can the org.eclipse.kura.gpio package version still be 1.2.0 since the old variant is still in the enum but flagged as deprecated?

nicolatimeus:
Hi, adding an enum variant requires to increase the minor version of the java package. In this way applications that use the new variant can depend on a package version that provides it.

PR #3769, opened 2022-02-02
Added validation to DHCP/NAT tab

pierantoniomerlino:
This PR adds validations on DHCP IP addresses and Lease times fields in the dhcp/nat tab.

**Description of the solution adopted:** The following validations are added:

- the default and max lease time cannot be 0 or negative
- the range of the dhcp ip address are checked (end must be greater than begin)


MMaiero:
By default the fields are not checked and the form allows you to apply incorrect values.

Screenshot 2022-02-03 at 13 40 49

We should define the required fields as such in the configuration, like in the firewall part
Screenshot 2022-02-03 at 13 42 07

pierantoniomerlino:
This happens when a "fresh" form is initiated. Currently the validation is performed when the user fills a field. This behavior must be definitely fixed.

MMaiero:
Yes, agreed

MMaiero (approved):

PR #3773, opened 2022-02-04
Divided payload column into small and large

marcellorinaldo:

Brief description of the PR. Given the new [H2 restrictions](https://h2database.com/html/advanced.html#limits_limitations), the maximum length of the `VARBINARY` type is now reduced (1048576 bytes). This PR tries to maintain retrocompatibility in the Data Store service for payloads that previously could have exceeded this new limit.

**Related Issue:** N/A.

**Description of the solution adopted:** Previously, the payload column was changed to be of type `BLOB`. According to the [H2 documentation](https://h2database.com/html/advanced.html#large_objects), this type is inefficient when the size is small < 200 bytes. To keep high efficiency, the Data Store tables now have two columns for the payload:
- `smallPayload`: of type `VARBINARY` that is meant to store payloads < 200 bytes
- `largePayload`: of type `BLOB` to store payloads >= 200 bytes

For each row, one between the two column is always populated with `null`. The separation logic is entirely handled in the `DbDataStore.storeInternal` and `DbDataStore.buildDataMessage` methods.

**Screenshots:** N/A.


kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/data/store/DbDataStore.java
```diff
@@ -312,25 +316,29 @@ private synchronized DataMessage storeInternal(String topic, byte[] payload, int
...
+                pstmt.setString(1, topic);                                          // topic
+                pstmt.setInt(2, qos);                                               // qos
+                pstmt.setBoolean(3, retain);                                        // retain
+                pstmt.setTimestamp(4, now, this.utcCalendar);                       // createdOn
+                pstmt.setTimestamp(5, null);                                        // publishedOn
+                pstmt.setInt(6, -1);                                                // publishedMessageId
+                pstmt.setTimestamp(7, null);                                        // confirmedOn
+
+                if (payload.length > PAYLOAD_BYTE_SIZE_THRESHOLD) {
+                    pstmt.setBytes(8, payload);                                     // smallPayload
```
MMaiero, line 346: the comment is broken, right?

marcellorinaldo: Code is broken. Fixed and added tests to capture this behavior.

MMaiero (commented):
Can we have a look if we are able to solve this issue that is causing the sonar failure? https://sonarcloud.io/project/issues?branch=PR-3773&id=org.eclipse.kura%3Akura&issues=AX7j-HyD7Jb3vuVPTV2W&open=AX7j-HyD7Jb3vuVPTV2W

nicolatimeus:
In the previous version there was a limit in the payload size defined by the `payload VARBINARY(16777216)` column type.. is that limit still enforced?

marcellorinaldo:
> In the previous version there was a limit in the payload size defined by the `payload VARBINARY(16777216)` column type.. is that limit still enforced?

No, now we do not have limits

marcellorinaldo:
@nicolatimeus @MMaiero Maximum allowed payload size restored to 16777216 to maintain same behavior as in previous version.

MMaiero (approved):

MMaiero:
If @nicolatimeus agrees, I believe we can merge

nicolatimeus (approved):

Issue #3733, opened 2021-12-30
Jetty errors while working with Wire Graph


mstankovic:
**Describe the bug**
From time to time Im getting jetty errors while Im working with Wire Graphs. Find them bellow. P.S. I dont have exact steps to reproduce those errors.

error1:

```
2021-12-27T10:21:38,165 [qtp30344791-139] INFO  o.e.k.w.s.s.WiresBlinkServlet - Session ended: 1640600197877
2021-12-27T10:22:12,172 [qtp30344791-325967] INFO  o.e.k.w.s.s.WiresBlinkServlet - Session ended: 1640599929279
2021-12-27T10:22:15,197 [qtp30344791-323670] INFO  o.e.k.w.s.s.WiresBlinkServlet - Session started: 1640599929279
2021-12-27T10:22:30,175 [qtp30344791-325923] INFO  o.e.k.w.s.s.WiresBlinkServlet - Session ended: 1640600249916
2021-12-27T10:22:30,181 [qtp30344791-325923] WARN  o.e.k.w.s.s.WiresBlinkServlet - failed to close stream
org.eclipse.jetty.io.EofException
        at org.eclipse.jetty.io.ChannelEndPoint.flush(ChannelEndPoint.java:279)
        at org.eclipse.jetty.io.ssl.SslConnection.networkFlush(SslConnection.java:474)
        at org.eclipse.jetty.io.ssl.SslConnection$DecryptedEndPoint.flush(SslConnection.java:1080)
        at org.eclipse.jetty.io.WriteFlusher.flush(WriteFlusher.java:422)
        at org.eclipse.jetty.io.WriteFlusher.write(WriteFlusher.java:277)
        at org.eclipse.jetty.io.AbstractEndPoint.write(AbstractEndPoint.java:381)
        at org.eclipse.jetty.server.HttpConnection$SendCallback.process(HttpConnection.java:829)
        at org.eclipse.jetty.util.IteratingCallback.processing(IteratingCallback.java:241)
        at org.eclipse.jetty.util.IteratingCallback.iterate(IteratingCallback.java:223)
        at org.eclipse.jetty.server.HttpConnection.send(HttpConnection.java:550)
        at org.eclipse.jetty.server.HttpChannel.sendResponse(HttpChannel.java:915)
        at org.eclipse.jetty.server.HttpChannel.write(HttpChannel.java:987)
[...]

Issue #3308, opened 2021-03-03
Camel: java.lang.OutOfMemoryError after a long time of execution


mariosanroman:
Hello everyone.

I don't know if this is a bug or if it is that I'm doing something wrong. I hope someone could help me.

I'm basically following the example in https://dentrassi.de/2018/09/14/the-power-of-apache-camel-in-eclipse-kura/

I have kura installed in a raspberry pi 3 with this O.S.:
Linux raspberrypi 4.19.97-v7+ #1294 SMP Thu Jan 30 13:15:58 GMT 2020 armv7l GNU/Linux

I set up a wire with a timer with 1 second interval linked to a Camel Producer with this configuration:

![Captura de pantalla 2021-03-03 a las 13 04 49](https://user-images.githubusercontent.com/5113316/109803200-0c79b580-7c21-11eb-8a94-eac62c5d8589.png)

The ID of the Camel Context: camel1
Endpoint URI: seda:input1

![Captura de pantalla 2021-03-03 a las 13 07 23](https://user-images.githubusercontent.com/5113316/109803502-6a0e0200-7c21-11eb-93c3-60d78fd3c434.png)

camel1 is a Configurable Camel-based XML router with this configuration:

Router XML:
```xml
<routes xmlns="http://camel.apache.org/schema/spring">
    <route>
        <from uri="seda:input1"/>
        <to uri="stream:out"/>
    </route>
</routes>
```
Required Camel Components:
stream

![Captura de pantalla 2021-03-03 a las 13 08 06](https://user-images.githubusercontent.com/5113316/109803580-80b45900-7c21-11eb-8cb2-602f2906576e.png)

Everything seems to work fine. I see the expected outputs in the kura console just as the previous post says I should.

The problem is that I get this in the kura-console.log if I leave it executing for a long time (maybe an hour or more)

```shell
Mar 03, 2021 12:23:02 PM org.apache.camel.com.github.benmanes.caffeine.cache.BoundedLocalCache scheduleDrainBuffers
WARNING: Exception thrown when submitting maintenance task
java.lang.OutOfMemoryError: unable to create new native thread
	at java.lang.Thread.start0(Native Method)
	at java.lang.Thread.start(Thread.java:717)
	at java.util.concurrent.ForkJoinPool.createWorker(ForkJoinPool.java:1486)
	at java.util.concurrent.ForkJoinPool.tryAddWorker(ForkJoinPool.java:1517)
[...]

PR #3775, opened 2022-02-06
fix #3444  #3308  #3733

### LINKED ISSUE #3444: There are errors in JS console after opening Status/Network system item
**Describe the bug**
There are errors in JS console after opening Status/Network system item

**To Reproduce**
1. OPen WEB UI
2. Open developer JS console
3. Click on Status/Network system item
4. See errors

**Expected behavior**
There should not be errors

**Screenshots**
Capture
Capture1

 - Board Raspberry Pi 3
 - OS version: Linux raspberrypi 5.10.17-v7+ # 1403 SMP Mon Feb 22 11:29:51 GMT 2021 armv7l GNU/Linux
 - Additional info: build 5.0 - RC2

spnettec:


spnettec:
Sorry. It is very hard to set up a development environment in China. The modifications in these places are done through the github webpage. The version I modified is too different from the deveop version. There is no way to submit it from my current version. I have upgrade camel to v3.14.1 and artemis to v2.20.0.  And add Chinese translation.  I use p2-maven-plugin for the p2-site generation.

nicolatimeus:
Hi @spnettec , the changes look good for me.. Can you please update all copyright headers of modified files? You may also want to add yourself to the contributor list in the copyright headers.
@ctron are the Camel related changes ok for you?

spnettec:
> Hi @spnettec , the changes look good for me.. Can you please update all copyright headers of modified files? You may also want to add yourself to the contributor list in the copyright headers. @ctron are the Camel related changes ok for you?

Sure. Thanks. I will update tonigt.

kura/org.eclipse.kura.wire.camel/src/main/java/org/eclipse/kura/wire/camel/CamelProduce.java
```diff
@@ -15,18 +15,46 @@
...
+    private static final Logger logger = LoggerFactory.getLogger(CamelProduce.class);
+
+    private FluentProducerTemplate template = null;
+
     @Override
     protected void processReceive(final CamelContext context, final String endpointUri, final WireEnvelope envelope)
             throws Exception {
-
-        on(context) //
+        if (template == null) {
```
ctron: I think the logic itself is good. However, I also think there can be a race condition when multiple threads are active.

Assuming that the `if == null` check will most often be false, and setting the template will be much less frequent than getting it, a "read/write lock" makes sense. Just beware, you can't upgrade it from a read lock to a write lock. So you need to unlock the read lock first, then write lock, and check again (if some other thread already created the template).

spnettec: Add readWrite lock for template

kura/org.eclipse.kura.wire.camel/src/main/java/org/eclipse/kura/wire/camel/CamelProduce.java
```diff
@@ -46,11 +49,41 @@ protected void processReceive(final CamelContext context, final String endpointU
...
+        } finally {
+            wlock.unlock();
+        }
+    }
+
+    private void checkTemplate(final CamelContext context) {
+        final Lock rlock = this.rwLock.readLock();
+        rlock.lock();
+        try {
+            templateAvailable = template != null && template.getCamelContext() == context;
```
ctron: I wouldn't use a global/class variable here. I would just return the value. Or maybe I just don't understand the reason for this pattern. Then a bit of source doc would help ;-)

kura/org.eclipse.kura.wire.camel/src/main/java/org/eclipse/kura/wire/camel/CamelProduce.java
```diff
@@ -46,11 +49,41 @@ protected void processReceive(final CamelContext context, final String endpointU
...
+        } finally {
+            rlock.unlock();
+        }
+    }
+
+    private void createTemplate(final CamelContext context) {
+        final Lock wlock = this.rwLock.writeLock();
+        wlock.lock();
+        try {
+            closeTemplate();
```
ctron: At the point it might be, that some other thread already processes this. So what would happen here, is that the existing template would get destroyed. I think there should be a second check here, and only then a new template should be created.

ctron (changes requested):
I think both points can be solves but refactoring the `checkTemplate` method to return a `boolean` and perfom the read lock outside of it (or wrap it in another method). And call the same method in `createTemplate` again.

The rest looks good!

kura/org.eclipse.kura.wire.camel/src/main/java/org/eclipse/kura/wire/camel/CamelProduce.java
spnettec: When readLock unlock there have a race condition the other threads change templateAvailable. So i declare a volatile boolean global/class variable

spnettec: I know what you mean. This is my first version
```
public class CamelProduce extends AbstractReceiverWireComponent {

    private final ReadWriteLock rwLock = new ReentrantReadWriteLock(true);

    private FluentProducerTemplate template = null;

    @Override
    protected void processReceive(final CamelContext context, final String endpointUri, final WireEnvelope envelope)
            throws Exception {
        
        boolean templateAvailable = false;
        final Lock rlock = this.rwLock.readLock();
        rlock.lock();
        try {
            templateAvailable = isTemplateAvailable(context);
        } finally {
            rlock.unlock();
        }
        if (!templateAvailable) {
            createTemplate(context);
        }
        template //
                .withBody(envelope) //
                .to(endpointUri) //
                .asyncSend();
    }

    @Override
    protected void deactivate() {
        super.deactivate();
        final Lock wlock = this.rwLock.writeLock();
        wlock.lock();
        try {
            closeTemplate();
        } finally {
            wlock.unlock();
        }
    }

    private boolean isTemplateAvailable(final CamelContext context) {
        return template != null && template.getCamelContext() == context;
    }

    private void createTemplate(final CamelContext context) {
        final Lock wlock = this.rwLock.writeLock();
        wlock.lock();
        try {
            closeTemplate();
            template = on(context);
        } finally {
            wlock.unlock();
        }
    }

    private void closeTemplate() {
        if (template != null) {
            try {
                template.stop();
            } catch (Exception ignored) {
            } finally {
                template = null;
            }
        }
    }

}


```

kura/org.eclipse.kura.wire.camel/src/main/java/org/eclipse/kura/wire/camel/CamelProduce.java
spnettec: > 

Only need to ensure that the template is destroyed to guaranteed not to cause memory leaks. No need to check again.

spnettec: closeTemplate method have second check and if template created by other thread then stop and recreate

kura/org.eclipse.kura.wire.camel/src/main/java/org/eclipse/kura/wire/camel/CamelProduce.java
ctron: True, there can be a race condition. However, I don't see how the volatile variable will fix this. As the state can still change.

Destroying and re-creating is actually not required. So I would avoid it, if possible.

I think that actually the locking must be a bit stricter. As what could happen (especially with destroying and re-creating the template), is that the template was available when checking. However, it gets destroyed afterwards. So the code would run into a NullPointerException or call into a closed template.

I think what needs to be done is:

```
while true {
  with-read-lock {
    if template-can-be-used {
      call template
      return
    }
  }
  with-write-lock {
    if template-can-be-used {
      continue:
    } else {
      create template
    }
  }
}
```

spnettec: I just don't want 'call template' to be within the read lock. It will reduce the efficiency.

kura/org.eclipse.kura.wire.camel/src/main/java/org/eclipse/kura/wire/camel/CamelProduce.java
```diff
@@ -9,24 +9,97 @@
...
+        rlock.lock();
+        try {
+            templateAvailable = isTemplateAvailable(context);
+        } finally {
+            rlock.unlock();
+        }
+        if (!templateAvailable) {
+            createTemplate(context);
+        }
+        if (template != null) {
```
ctron: As there is no sychronization, it could be that the template is "non null" in this millisecond, but null in the next one.

kura/org.eclipse.kura.wire.camel/src/main/java/org/eclipse/kura/wire/camel/CamelProduce.java
spnettec: I don't think it's the best way to do this. Template re-create to discard messages is acceptable.
Sending messages is not efficient within a read lock.

What do you think? @ctron 

```
public class CamelProduce extends AbstractReceiverWireComponent {

    private final ReadWriteLock rwLock = new ReentrantReadWriteLock(true);

    private FluentProducerTemplate template = null;

    @Override
    protected void processReceive(final CamelContext context, final String endpointUri, final WireEnvelope envelope)
            throws Exception {

        while (true) {
            final Lock rlock = this.rwLock.readLock();
            rlock.lock();
            try {
                if (isTemplateAvailable(context)) {
                    template //
                            .withBody(envelope) //
                            .to(endpointUri) //
                            .asyncSend();
                    return;
                }
            } finally {
                rlock.unlock();
            }

            final Lock wlock = this.rwLock.writeLock();
            wlock.lock();
            try {
                if (!isTemplateAvailable(context)) {
                    createTemplate(context);
                }
            } finally {
                wlock.unlock();
            }
        }
    }

    @Override
    protected void deactivate() {
        super.deactivate();
        final Lock wlock = this.rwLock.writeLock();
        wlock.lock();
        try {
            closeTemplate();
        } finally {
            wlock.unlock();
        }
    }

    private boolean isTemplateAvailable(final CamelContext context) {
        return template != null && template.getCamelContext() == context;
    }

    private void createTemplate(final CamelContext context) {
        closeTemplate();
        template = on(context);
    }

    private void closeTemplate() {
        if (template != null) {
            try {
                template.stop();
            } catch (Exception ignored) {
            } finally {
                template = null;
            }
        }
    }

}

```

kura/org.eclipse.kura.wire.camel/src/main/java/org/eclipse/kura/wire/camel/CamelProduce.java
spnettec: I know that. Because i don't want sending messages within a read lock. 

If no sychronization we can discard this messages. It is acceptable.

If switch template the messages is unuseful. We can discard it and wait next message.

This PR is mainly to solve the thread leak problem instead of modifying the settings causing message loss

spnettec: In AbstractCamelWireComponent and other componet there have many same problem.

For runtime efficiency these errors are acceptable and can automatically recoverable at next message comming. And these error are also justified. We can just debug or info.
But a memory leak is unacceptable.

spnettec: Lots of locks maybe cause system dead lock or slow down the overall system efficiency

spnettec: In my opinion the code should still be the same as the original. just add try catch to skip processReceive massage

Only createTemplate and deactivate add synchronized keyword. These two write methods are rarely used.

How the code is refactored is up to you.

'''

public class CamelProduce extends AbstractReceiverWireComponent {

    private static final Logger logger = LoggerFactory.getLogger(CamelProduce.class);

    private FluentProducerTemplate template = null;

    @Override
    protected void processReceive(final CamelContext context, final String endpointUri, final WireEnvelope envelope)
            throws Exception {

        if (!isTemplateAvailable(context)) {
            createTemplate(context);
        }
        if (template != null) {
            try {
                template //
                        .withBody(envelope) //
                        .to(endpointUri) //
                        .asyncSend();
            } catch (Exception e) {
                logger.error("asyncSend error", e);
            }
        } else {
            logger.debug("FluentProducerTemplate is changing. Skip send massage and wait next massage");
        }
    }

    @Override
    protected synchronized void deactivate() {
        super.deactivate();
        closeTemplate();
    }

    private boolean isTemplateAvailable(final CamelContext context) {
        return template != null && template.getCamelContext() == context;
    }

    private synchronized void createTemplate(final CamelContext context) {
        closeTemplate();
        template = on(context);
    }

    private void closeTemplate() {
        if (template != null) {
            try {
                template.close();
            } catch (Exception ignored) {
            } finally {
                template = null;
            }
        }
    }

}

'''

ctron (approved):
I think using the `bindContext` method for this is the best solution!

nicolatimeus (approved):
Sonar build is failing due to insufficient test coverage.  But currently the Camel part is not covered by any test, merging

PR #3777, opened 2022-02-06
Allowed to submit password change modal with enter key

nicolatimeus:

Should allow to submit the password change on login form with the enter key.


mstankovic:
checked and confirmed that is submitting on enter key.

MMaiero (approved):

PR #3799, opened 2022-02-15
Fix intermittent wifi scan failures.

salvatore-coppola:
Now a temporary network interface is used to scan wifi networks in MASTER mode.

### DIFF
--- a/kura/org.eclipse.kura.linux.net/src/main/java/org/eclipse/kura/linux/net/NetworkServiceImpl.java
+++ b/kura/org.eclipse.kura.linux.net/src/main/java/org/eclipse/kura/linux/net/NetworkServiceImpl.java
@@ -102,7 +102,6 @@ public class NetworkServiceImpl implements NetworkService, EventHandler {
     public static final String PPP_PEERS_DIR = "/etc/ppp/peers/";
     private static final String PPP = "ppp";
     private static final Integer MAX_PPP_NUMBER = 100;
-    private static final String ACCESS_POINT_INTERFACE_SUFFIX = "_ap";
 
     private static final String[] EVENT_TOPICS = new String[] { UsbDeviceAddedEvent.USB_EVENT_DEVICE_ADDED_TOPIC,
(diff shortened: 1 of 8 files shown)


kura/org.eclipse.kura.linux.net/src/main/java/org/eclipse/kura/linux/net/NetworkServiceImpl.java
```diff
@@ -319,6 +321,10 @@ public NetInterfaceState getState(String interfaceName) throws KuraException {
             }
         }
 
+        // remove virtual interface for Access Point
+        interfaceNames = interfaceNames.stream().filter(name -> !name.contains(ACCESS_POINT_INTERFACE_SUFFIX))
+                .collect(Collectors.toList());
+
```
pierantoniomerlino, line 327: Is it better to use .endsWith instead of contains?

kura/org.eclipse.kura.net.admin/src/main/java/org/eclipse/kura/net/admin/NetworkAdminServiceImpl.java
```diff
@@ -88,6 +89,8 @@
 
 public class NetworkAdminServiceImpl implements NetworkAdminService, EventHandler {
 
+    private static final String ACCESS_POINT_INTERFACE_SUFFIX = "_ap";
+
```
pierantoniomerlino: Since this constant is also in the NetworkServiceImpl, maybe it's better to define it only once.

pierantoniomerlino:
@salvatore-coppola We create a virtual wireless interface for performing a scan. Once created, the interface remains active. Is it possible to disable it (not remove) the interface when it is not used?

salvatore-coppola:
> @salvatore-coppola We create a virtual wireless interface for performing a scan. Once created, the interface remains active. Is it possible to disable it (not remove) the interface when it is not used?

obvisously we can but this will slow the scan process, unless we don't care ...

pierantoniomerlino (approved):

PR #3809, opened 2022-02-21
Attempt to remove jakarta.annotation-api

nicolatimeus:

The packages provided by `jakarta.annotation-api` (`javax.annotation`, `javax.annotation.security` and `javax.annotation.sql`) are already provided by `com.eclipsesource.jaxrs.jersey-min`.


MMaiero (approved):

spnettec:
com.eclipsesource.jaxrs.jersey-min is too old

nicolatimeus:
We have removed this because having both `com.eclipsesource.jaxrs.jersey-min` and the Jakarta bundle leads to this issue: https://github.com/hstaudacher/osgi-jax-rs-connector/issues/194#issuecomment-340499769.
Are you having problems with the version of the APIs in the jersey-min bundle? 
Removing the Jakarta bundle is not an ideal solution but since we are close to 5.1 release we applied this change with the idea of finding a better solution in the next release cycle.
If you have a better solution ready please open a PR!

PR #3811, opened 2022-02-22
Modem related fixes

nicolatimeus:

* Fixed GPS tab not enabled even if modem has GPS.
* Fixed ppp not working after Kura restart
* Added partial support for network configurations with name in the `pppX` form
  * These configurations will be migrated to the new usb port naming scheme during `NetworkConfigurationServiceImpl` update
* Ppp interface numbers are now assigned eagerly as soon as the modem is detected, and also when the user requests the name for a given port, even if there is not (yet) a USB device modem tracked on that port.
  * Should fix NPE on `NetworkConfigurationServiceImpl` activate that prevented the visitors from being run to align system configuration with snapshot


MMaiero:
@mstankovic Can you have a check on this PR please?

MMaiero (approved):

Issue #3103, opened 2020-10-15
How can I solve this problem? (about SenseHat Diver)


97Albino:
**Is your feature request related to a problem? Please describe.**
hi
I installed sensehat driver and set configurations like this.
https://marketplace.eclipse.org/content/sensehat-example-driver-eclipse-kura-4xy
![image](https://user-images.githubusercontent.com/58206274/96074670-35084980-0ee4-11eb-97a9-74853fbc093e.png)

I referred to the link below.
http://eclipse.github.io/kura/devices/sensehat-driver.html

When I clicked [data], the following error guide occurred.
![image](https://user-images.githubusercontent.com/58206274/96075123-54ec3d00-0ee5-11eb-8b23-bb216e838bcb.png)
Does this error depend on kura wire or something else?
Does this error become a fatal error in connecting Kura and Hono?
Is this error 

**Describe the solution you'd like**
I want to transfer sensor data(pressure, temp, hummidity) to Hono(MQTT adaptor) by using Kura wire.

![image](https://user-images.githubusercontent.com/58206274/96075722-9df0c100-0ee6-11eb-9919-c1c0038d5a45.png)

**Additional context**
Kura 4.1.1
Rasberry Pi 3

MMaiero:
Hi, 
I believe the error is due to the fact that the Kura Driver is not able to communicate with the sensehat.

The connection to Hono is not affected as is a different thing but for sure, since the data cannot be fetched from the sensehat, that it cannot also be published.

Unfortunately I don't have at the moment a SenseHat to perform some tests
@pierantoniomerlino Do you have an idea on why is not working?

pierantoniomerlino:
IMHO, the Kura Driver is not able to communicate with the SenseHat. Have you tried to communicate with the SenseHat from Linux? Maybe there is something that is not properly configured...
@nicolatimeus Any idea?

97Albino:
Thanks for your replies. :)

Below is the information of [Show Stack Trace].

org.eclipse.kura.example.driver.sensehat.SenseHatInterface.fetch(SenseHatInterface.java:129)
org.eclipse.kura.example.driver.sensehat.SenseHatInterface.runReadRequest(SenseHatInterface.java:206)
org.eclipse.kura.example.driver.sensehat.SenseHatDriver$SenseHatPreparedRead.execute(SenseHatDriver.java:182)
org.eclipse.kura.asset.provider.BaseAsset.lambda$2(BaseAsset.java:313)
[...]

PR #3813, opened 2022-02-23
Applied quick fix to jdk.dio to fix Sensehat driver issue

nicolatimeus:

Applies the quick fix described in #3103 to jdk.dio to solve the issue with SenseHat on recent JVMs, the proper fix would be adding an actual implementation to `Java_com_oracle_dio_impl_Handle_tryLock`.

Also added imports to prevent implicit function declaration warnings during build.


marcellorinaldo (approved):
Works on raspberry pi4

PR #3821, opened 2022-02-25
web2 code contribution for container orchestration 

GregoryIvo:
This PR adds a container management dashboard under the devices tab

image


MMaiero:
The code throws an exception if the service is registered but not connected to a docker instance.
I believe we should return just an empty list or even not register the service if it is not enabled. So that the web UI will not display the tab if the corresponding service is not registered.
Maybe have a look at what @nicolatimeus has done for the tamper detection service and corresponding tab in the web UI.

Screenshot 2022-02-28 at 10 13 56

MMaiero:
Furthermore I believe there is a misspelling in the table as at the last column mentions "ESF managed" but should be "Framework Managed"

GregoryIvo:
> Furthermore I believe there is a misspelling in the table as the last column mentions "ESF managed" but should be "Framework Managed"

- Fixed the Misspelling.  
- UI will not display the tab if the corresponding service is not registered
- Returns just an empty list if API is not enabled. & Shows why in the console.

MMaiero (approved):

PR #3830, opened 2022-03-06
Add listeners in batches

spnettec:

I am using plc4x recently. I found that when adding driver listeners there is no way to add them one by one. We need to add the feature for batch add-driver listeners.


kura/org.eclipse.kura.asset.provider/src/main/java/org/eclipse/kura/asset/provider/DriverState.java
```diff
@@ -84,46 +88,50 @@ public void syncChannelListeners(final Set<ChannelListenerRegistration> targetSt
...
         }
+
     }
 
-    private void attach(final ChannelListenerRegistration reg, final Channel channel) {
+    private void attach(final Map<Channel, ChannelListenerRegistration> regChannels) {
         try {
             logger.debug("Registering Channel Listener for monitoring...");
-            this.driver.registerChannelListener(channel.getConfiguration(), reg.getChannelListener());
-            this.attachedListeners.add(reg);
```
nicolatimeus, line 115: Hi @spnettec, thank you for the contribution. The current implementation of the Asset handles the case of a failure in registering a single channel listener, and if this happens the Asset will attempt to register other listen channels in its configuration (if any). Is this behavior preserved in the new implementation?

spnettec: I would like to confirm that based on the results of the unit test is it possible to register multiple listener instances for one channel Instead of a channel only one instance of the listener can be registered?  At present what I have modified is that a channel can only listen to one listener instance.

PR #3841, opened 2022-03-09
Fixed port forward behaviour with single source port

MMaiero:

### DIFF
--- a/kura/org.eclipse.kura.linux.net/src/main/java/org/eclipse/kura/linux/net/iptables/PortForwardRule.java
+++ b/kura/org.eclipse.kura.linux.net/src/main/java/org/eclipse/kura/linux/net/iptables/PortForwardRule.java
@@ -1,12 +1,12 @@
 /*******************************************************************************
  * Copyright (c) 2011, 2021 Eurotech and/or its affiliates and others
- * 
+ *
  * This program and the accompanying materials are made
  * available under the terms of the Eclipse Public License 2.0
  * which is available at https://www.eclipse.org/legal/epl-2.0/
- * 
+ *
  * SPDX-License-Identifier: EPL-2.0
- * 
+ *
  * Contributors:
  *  Eurotech
  *******************************************************************************/
@@ -18,7 +18,7 @@ import org.eclipse.kura.net.firewall.RuleType;
 
 /**
  * Creates an iptables command for a Port Forward Rule, allowing an incoming port to be forwarded to destinationIP/port.
- * 
+ *
  */
 public class PortForwardRule {
 
@@ -35,7 +35,8 @@ public class PortForwardRule {
     private String permittedNetwork;
     private int permittedNetworkMask;
     private String permittedMAC;
-    private String sourcePortRange;
+    private int sourcePortStart;
+    private int sourcePortEnd;
 
     /**
      * Constructor of <code>PortForwardRule</code> object.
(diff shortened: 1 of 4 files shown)


kura/org.eclipse.kura.linux.net/src/main/java/org/eclipse/kura/linux/net/iptables/PortForwardRule.java
```diff
@@ -128,10 +130,27 @@ public PortForwardRule permittedMAC(String permittedMAC) {
     }
 
     public PortForwardRule sourcePortRange(String sourcePortRange) {
-        this.sourcePortRange = sourcePortRange;
+        parsePortRange(sourcePortRange);
         return this;
     }
 
+    private void parsePortRange(String sourcePortRange) {
```
pierantoniomerlino, line 137: Do we need some check on this field? For example that startPort <= endPort?

kura/org.eclipse.kura.linux.net/src/main/java/org/eclipse/kura/linux/net/iptables/PortForwardRule.java
```diff
@@ -384,7 +391,10 @@ public void setPermittedMAC(String permittedMAC) {
      * @return the sourcePortRange
      */
     public String getSourcePortRange() {
-        return this.sourcePortRange;
+        if (this.sourcePortStart == this.sourcePortEnd && this.sourcePortStart == 0) {
+            return null;
+        }
+        return this.sourcePortStart + ":" + this.sourcePortEnd;
     }
```
pierantoniomerlino, line 398: So, if there is only one source port, the result is port:port (i.e. 10100:10100). Is it correct?

kura/org.eclipse.kura.linux.net/src/main/java/org/eclipse/kura/linux/net/iptables/PortForwardRule.java
MMaiero: We could. But we don't have a way to report the exception. We could act safely by fixing something. The question is, what?

kura/org.eclipse.kura.linux.net/src/main/java/org/eclipse/kura/linux/net/iptables/PortForwardRule.java
MMaiero: yes

kura/org.eclipse.kura.linux.net/src/main/java/org/eclipse/kura/linux/net/iptables/PortForwardRule.java
pierantoniomerlino: Let's see it from another pov: what happens if we set a wrong range in the current implementation? Is there an exception when the rules is applied?

pierantoniomerlino: I verified that in the form in the web console there is a check for this...

MMaiero: True. But the code does not protect itself. For example in the case someone updates the configuration outside of the Web UI.
I think to do so we will need some breaking changes. But probably this is outside the scope of this PR.

pierantoniomerlino: @MMaiero I totally agree. So, leave it as is and let's open an issue if needed.

pierantoniomerlino (approved):

PR #3843, opened 2022-03-09
Remove the recover config script

pierantoniomerlino:
This PR removes the recovery script.

**Description of the solution adopted:** The recovery scripts are removed from the distribution for all profiles.


nicolatimeus:
There are still references to recoverdefaultconfig here: https://github.com/eclipse/kura/blob/39b3e168fcd4414327b5235b30afde56c4b61957/kura/distrib/src/main/resources/intel-up2-ubuntu-20-nn/kura_install.sh#L67

nicolatimeus:
@pierantoniomerlino  can you please solve the conflict?

MMaiero (approved):

PR #3847, opened 2022-03-09
Fixed Invalid returned GPS Position and event firing missing.

salvatore-coppola:
Now GpsdPositionProvider return 0 in every field if no data are available.
Now GpsdPositionProvider fire the onLockStatusChanged correctly.

### DIFF
--- a/kura/org.eclipse.kura.linux.position/src/main/java/org/eclipse/kura/linux/position/GpsdPositionProvider.java
+++ b/kura/org.eclipse.kura.linux.position/src/main/java/org/eclipse/kura/linux/position/GpsdPositionProvider.java
@@ -184,9 +184,10 @@ public class GpsdPositionProvider implements PositionProvider, IObjectListener {
         internalState.setTime(tpv.getTimestamp());
         internalState.setMode(tpv.getMode());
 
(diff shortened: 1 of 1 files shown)


nicolatimeus:
Zero is now shown instead of NaN before getting the first lock, I got a PositionLockedEvent on fix but I'm not getting any event if I disconnect the GPS antenna, the old position is still reported and no event is fired.

salvatore-coppola:
I have tested again and i got the event, the antenna takes a while to lost the fix, can you do another test waiting more and monitoring when gpsd actually loose the fix with `gpsd -w` ?

nicolatimeus (approved):

nicolatimeus:
This change restores the PositionLocked/LostEvents, the reported behavior is probably due to https://lists.gnu.org/archive/html/gpsd-dev/2017-09/msg00027.html, we might want to address that in a follow-up PR

salvatore-coppola:
@MMaiero why the project for this pr is Kura 5.2.0 ?

MMaiero:
Because it is merged in develop that will be 5.2.0. Once merged, the corresponding back port will be merged in 5.1.0 and that PR will be associated to the 5.1.0 project

PR #3855, opened 2022-03-10
H2DB store wire component checks on user input

marcellorinaldo:
Added some checks to the user-definable properties `maximum.table.size` and `cleanup.records.keep`.

**Related Issue:** N/A.

**Description of the solution adopted:** Prior to this PR, if the user set `cleanup.records.keep > maximum.table.size` the cleanup to the DB tables won't happen. In particular, the table truncation was executed by selecting the oldest `maximum.table.size - cleanup.records.keep` entries, which could assume negative values causing an error. Several checks are now implemented and the minimum value for `maximum.table.size` property is set to 1.

Unit tests have been refactored according to Gherkin style and updated with new corner-case checks.

**Screenshots:** N/A.

### DIFF
--- a/kura/org.eclipse.kura.wire.h2db.component.provider/src/main/java/org/eclipse/kura/internal/wire/h2db/store/H2DbWireRecordStore.java
+++ b/kura/org.eclipse.kura.wire.h2db.component.provider/src/main/java/org/eclipse/kura/internal/wire/h2db/store/H2DbWireRecordStore.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2017, 2021 Eurotech and/or its affiliates and others
+ * Copyright (c) 2017, 2022 Eurotech and/or its affiliates and others
  *
  * This program and the accompanying materials are made
  * available under the terms of the Eclipse Public License 2.0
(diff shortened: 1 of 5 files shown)


MMaiero:
@mattdibi Can you please have a review of this PR? Thanks

kura/org.eclipse.kura.wire.h2db.component.provider/src/main/java/org/eclipse/kura/internal/wire/h2db/store/H2DbWireRecordStore.java
```diff
@@ -89,6 +89,9 @@
 
     private static final String[] TABLE_TYPE = new String[] { "TABLE" };
 
+    private static final String NULL_TABLE_NAME_ERROR_MSG = "Table name cannot be null";
+    private static final String NULL_WIRE_RECORD_ERROR_MSG = "";
```
mattdibi: Is this supposed to be empty?

marcellorinaldo: Forgot to readd the error message :)

salvatore-coppola (approved):

mattdibi (approved):
LGTM

PR #3865, opened 2022-03-11
Little change chronyprovider log message

salvatore-coppola:
Now the message is more accurate


pierantoniomerlino (approved):

nicolatimeus:
Do we need to backport this?

PR #3869, opened 2022-03-11
Add container start retry

MMaiero:


MMaiero:
@nicolatimeus @GregoryIvo Initial work on retries and API review.
APIs will need better javadocs. Tests completely need adaptation
Please let me know if the approach works for you and if you already see some needed improvements

GregoryIvo:
@MMaiero I like the direction you are going.  From what I am gathering, you are trying to move out all the 'registration logic' to its own separate layer? API will only be concerned about performing actions on containers, and the DockerServiceListener handles life-cycle tracking?

kura/org.eclipse.kura.container.orchestration.provider/src/main/java/org/eclipse/kura/container/orchestration/provider/DockerService.java
```diff
@@ -22,34 +22,44 @@
...
-     */
-    public String[] listContainers();
-
     /**
      * Lists all available containers by id only, like running 'Docker ps -a'
      *
      * @return
      */
-    public String[] listContainersByID();
+    public List<String> listContainersIds();
```
GregoryIvo: There are many Legacy API functions like this that were created before ContainerDescriptor was realized. I think that going forward we can remove all of these, and rely only on the ContainerDescror API methods.

kura/org.eclipse.kura.container.orchestration.provider/src/main/java/org/eclipse/kura/container/orchestration/provider/DockerService.java
```diff
@@ -22,34 +22,44 @@
...
+    */
+   public void pullImage(String imageName, String imageTag) throws KuraException;
+
+   /**
+    *
+    * @param imageName
+    * @param timeOutSecconds
+    * @throws KuraException
+    */
+   public void pullImage(String imageName, String imageTag, int timeOutSecconds) throws KuraException;
```
GregoryIvo: Going with the theme of my previous comment above, perhaps we add all of the timeOut Parameters into the ContainerDescriptor?

MMaiero:
@GregoryIvo Can you help me with the tests review?
In particular, can you please update the tests in DockerServiceImplTest? I don't want to mess with the logic you used there. Thanks.

kura/org.eclipse.kura.container.orchestration.provider/src/main/java/org/eclipse/kura/container/orchestration/provider/DockerService.java
MMaiero: Don't know. We were discussing with Nicola and maybe it makes sense to have those generic commands that allow to list and run also containers that are not managed by the framework.

kura/org.eclipse.kura.container.orchestration.provider/src/main/java/org/eclipse/kura/container/orchestration/provider/DockerService.java
MMaiero: Agreed. Can you do that?

kura/org.eclipse.kura.container.orchestration.provider/src/main/java/org/eclipse/kura/container/orchestration/provider/ContainerDescriptor.java
```diff
@@ -37,9 +37,9 @@
     private List<String> containerEnvVars;
     private List<String> containerDevices;
     private Map<String, String> containerVolumes;
-    private Boolean containerPrivilaged;
+    private Boolean containerPrivileged;
     private ContainerStates containerState = ContainerStates.STOPPING;
```
nicolatimeus: Keeping the containerState field mutable can be error prone, it requires a lot of internal housekeeping and it can become inconsistent with system state.
What if we just remove it from here and add a method in DockerService that requests the container state to the container daemon given a container id?
Making ContainerDescriptor immutable would also be a nice imporvement.

kura/org.eclipse.kura.container.orchestration.provider/src/main/java/org/eclipse/kura/container/orchestration/provider/impl/DockerServiceImpl.java
```diff
@@ -120,93 +118,70 @@ public void updated(Map<String, Object> properties) {
...
         }
-        return avaliableContainers;
+        return version;
+    }
 
+    private String getContainerTag(Container container) {
+        return container.getImage().split(":")[0];
     }
 
     private int[] parseExternalPortsFromDockerPs(ContainerPort[] ports) {
```
nicolatimeus: This method will always return an empty array, the same applies to parseInternalPortsFromDockerPs, ArrayUtils.add does not modify the passed array argument, which is immutable.

nicolatimeus: The `ipFormatTest.equals("::")` check fails on my device, the reported id in that case is `0.0.0.0`.
Should we expose the IP in our APIs?

kura/org.eclipse.kura.container.orchestration.provider/src/main/java/org/eclipse/kura/container/orchestration/provider/impl/DockerServiceImpl.java
```diff
@@ -271,97 +245,141 @@ public String getContainerIDbyName(String name) {
...
             throw new IllegalStateException(UNABLE_TO_CONNECT_TO_DOCKER_CLI);
         }
-        this.dockerClient.removeContainerCmd(id).exec();
+        this.dockerClient.stopContainerCmd(id).exec();
     }
 
     @Override
-    public void pullImage(String imageName) {
-        pullImage(imageName, 120);
+    public void stopContainer(ContainerDescriptor container) {
```
nicolatimeus: What about just removing this method? The only field that it consider in the `ContainerDescriptor` is the container name (previously the container id). 
Probably the `stopContainer(String id)` is the minimum required functionality here, the clients can figure out the id given a name/container descriptor using the other methods (this is what the current implementation does). The same applies to `deleteContainer(ContainerDescriptor containerDescriptor)`.

MMaiero:
Still not sure if we should change DockerService name to something different. Maybe ContainerOrchestrationService is better? Even though the library we use calls everything Docker...

GregoryIvo:
> 

I agree with the renaming.  In the future, we can add support for other tools that are similar to docker like podman

nicolatimeus (approved):

PR #3870, opened 2022-03-11
Added workaround for GPSD not reporting fix lost

nicolatimeus:
* Added parameter that allows to report the fix as lost if no position updates are received from GPSD for some time
* GPSD now reports the last known position in case of no fix as the Serial provider does


nicolatimeus:
@salvatore-coppola can you please have a look at this?

salvatore-coppola (approved):

PR #3902, opened 2022-03-16
Fixed little typo in ConfigurableGenericDockerService.xml

salvatore-coppola:


MMaiero:
The file is completely changed. Maybe we had bad formatting before?
But it is not clear the small change

salvatore-coppola:
@MMaiero wow you are right, let me check the formatting

PR #3916, opened 2022-03-21
Fix Kura log rotation

mattdibi:
This PR fixes the Kura log rotation in all profiles.

The Kura logrotate configuration was updated for in-place log rotation and following test expectations. The rule now applies to all logs named `kura*.log`. Additionally a cron job is created at Kura installation for running `logrotate` every 5 minutes.

To achieve this we:
- Removed the old global logrotate configuration `logrotate.conf`
- Moved the common `kura.logrotate` configuration to the profiles directories
- Added the cronjob to perform the logrotate call every 5 minutes with the `kura.logrotate` configuration as parameter

Please note that the `raspberry-pi` Raspbian profiles do not need the `su` directive inside the `kura.logrotate` configuration since the `/var/log` directory is already owned by `root:root` which is the default user for logrotate


nicolatimeus:
It seems to work as expected on RPi, just few more notes:
*  the installer currently updates `/etc/cron.d/logrotate-kura` only if the file doesn't already exists, should we update it in any case on installation?
* do we want to remove `/etc/cron.d/logrotate-kura` on uninstallation?

mattdibi:
@nicolatimeus 

> the installer currently updates /etc/cron.d/logrotate-kura only if the file doesn't already exists, should we update it in any case on installation?

This is aligned with what happens on ESF. I can update it to alway overwrite it on installation or something along those lines. What do you think?

> do we want to remove /etc/cron.d/logrotate-kura on uninstallation?

We don't have an uninstall script for Kura though. How should we handle it?

nicolatimeus:
The cron file is now removed as expected, the uninstall script currently removes `/var/log/kura*.log` but leaves the rotated compressed files in place, should we remove these files as well for consistency?

mattdibi:
@nicolatimeus I missed that, sorry. My bad. Fixing it rn.

mattdibi:
@nicolatimeus Tested on Jetson Nano & working as expected.

nicolatimeus (approved):

PR #3936, opened 2022-03-25
feat(web2): password change feature

marcellorinaldo:
This PR introduces a new feature that allows the logged user to change its password.

**Related Issue:** N/A.

**Description of the solution adopted:** The header has been modified to include a dropdown settings menuthat can be triggered by clicking on the cog button next to the username. The menu contains the usual "Logout" button and the new "Change password".

When clicking on "Change password", the password change dialog appears.

The `updatePassword` RPC of the GWT session service has been modified to be more secure by requiring the old password.

**Screenshots:**

Screenshot 1: *Dropdown menu.*

Screenshot 2022-04-22 at 10 57 03

Screenshot 2: *Password change modal.*

Screenshot 2022-04-22 at 11 12 03

### DIFF
--- a/kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/shared/service/GwtSessionService.java
+++ b/kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/shared/service/GwtSessionService.java
@@ -31,5 +31,5 @@ public interface GwtSessionService extends RemoteService {
 
     public GwtUserConfig getUserConfig(GwtXSRFToken token) throws GwtKuraException;
 
-    public void updatePassword(final String newPassword) throws GwtKuraException;
+    public void updatePassword(GwtXSRFToken token, final String newPassword) throws GwtKuraException;
 }
--- a/kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/shared/service/GwtSessionService.java
+++ b/kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/shared/service/GwtSessionService.java
@@ -31,5 +31,6 @@ public interface GwtSessionService extends RemoteService {
 
     public GwtUserConfig getUserConfig(GwtXSRFToken token) throws GwtKuraException;
 
-    public void updatePassword(GwtXSRFToken token, final String newPassword) throws GwtKuraException;
+    public void updatePassword(GwtXSRFToken token, final String oldPassword, final String newPassword)
+            throws GwtKuraException;
 }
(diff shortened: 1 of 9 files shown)


marcellorinaldo:
`TODO1`: refactoring to not have duplicate methods in `EntryClassUi` and `LoginUi`, if possible. Ask for current password in the change modal (this change will reflect also in the first login).

marcellorinaldo:
`TODO2`: add a button with a gear icon near to the username so that it is more intuitive that the user can click in that part. This can be the icon to use: https://fontawesome.com/icons/user-gear?s=solid

kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/shared/service/GwtSessionService.java
```diff
@@ -31,5 +31,6 @@ public interface GwtSessionService extends RemoteService {
 
     public GwtUserConfig getUserConfig(GwtXSRFToken token) throws GwtKuraException;
 
-    public void updatePassword(final String newPassword) throws GwtKuraException;
+    public void updatePassword(GwtXSRFToken token, final String oldPassword, final String newPassword)
```
nicolatimeus, line 35: Please add an audit annotation to this method

nicolatimeus, line 35: Please add an audit annotation to this method

marcellorinaldo: Done.

mstankovic:
feature works as expected. I have found one trivial issue:
we should close dropdown menu if user clicks outside of the menu (anywhere on the app)
question: why we have different behaviour for changing password (for current user) from dropdown menu and from identities service?

marcellorinaldo:
> feature works as expected. I have found one trivial issue:
> we should close dropdown menu if user clicks outside of the menu (anywhere on the app)

It's not trivial in GWT, I will try to do it.

> question: why we have different behaviour for changing password (for current user) from dropdown menu and from identities service?

So that if a user forgets its password the admin can change it.

mstankovic:
> > feature works as expected. I have found one trivial issue:
> > we should close dropdown menu if user clicks outside of the menu (anywhere on the app)
> 
> It's not easy in GWT, I will try to do it.
 _okey_
> > question: why we have different behaviour for changing password (for current user) from dropdown menu and from identities service?
> 
> So that if a user forgets its password the admin can change it.
 _in that case yes, but if logged user wants to change his password. That behaves differently _

marcellorinaldo:
> _in that case yes, but if logged user wants to change his password. That behaves differently _

Only admin users can change identities.

MMaiero:
Is it expected that, once opened, the window in the top right stays open even if you switch sections?

Screenshot 2022-05-02 at 14 09 31
Screenshot 2022-05-02 at 14 09 39

marcellorinaldo:
> Is it expected that, once opened, the window in the top right stays open even if you switch sections?
> 
> Screenshot 2022-05-02 at 14 09 31 Screenshot 2022-05-02 at 14 09 39

yes

MMaiero (approved):

PR #3938, opened 2022-03-28
Draft Branch for Staging: added API's for image management support

GregoryIvo:

this pr will add image management support to the container orchestrator.

MicrosoftTeams-image


GregoryIvo:
@MMaiero Kindly ask for you to review this PR when you have the chance.

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/container/orchestration/ContainerOrchestrationService.java
```diff
@@ -41,6 +41,21 @@
      * @return a list of {@link ContainerInstanceDescriptor}s representing the available containers
      */
     public List<ContainerInstanceDescriptor> listContainerDescriptors();
+    
+    /**
+     * Lists all available images, returning the corresponding {@link ImageInstanceDescriptor}s
+     *
+     * @return a list of {@link ContainerInstanceDescriptor}s representing the available containers
+     */
+    public List<ImageInstanceDescriptor> listImageInstanceDescriptor();
```
MMaiero: needs a since due to the fact that was added after the initial contribution

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/container/orchestration/ContainerOrchestrationService.java
```diff
@@ -41,6 +41,21 @@
...
+    
+    /**
+     * Deletes Image from container engine. Takes ImageID String as parameter. Will throw an error if image is being used by a container.
+     *
+     * @return void
+     * @throws KuraException 
+     */
+    public void deleteImage(String imageId) throws KuraException;
```
MMaiero, line 63: same here

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/container/orchestration/ContainerOrchestrationService.java
```diff
@@ -60,6 +75,24 @@
...
+     *            a string representing the image tag. Must not be null.
+     * @param timeOutSeconds
+     *            a non negative integer representing the image download timeout in seconds
+     * @param registryCredentials
+     *            an optional that can contain the registry URL and credentials for authentication
+     * @throws KuraException
+     *             if the pull operation fails
+     * @throws InterruptedException
+     */
+    public void pullImage(ImageConfiguration imageConfig) throws KuraException, InterruptedException;
```
MMaiero: Why do we need this overloading?

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/container/orchestration/ImageInstanceDescriptor.java
```diff
@@ -0,0 +1,200 @@
...
+import java.util.Objects;
+
+import org.osgi.annotation.versioning.ProviderType;
+
+/**
+ * Object which represents a image. Used to track created images, and images
+ * that exist in the container engine.
+ *
+ * @noimplement This interface is not intended to be implemented by clients.
+ * @since 2.3
```
MMaiero: this is added after 2.3, I think

kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/client/ui/device/DockerContainersTabUi.ui.xml
```diff
@@ -1,45 +1,79 @@
 <!DOCTYPE ui:UiBinder SYSTEM "http://dl.google.com/gwt/DTD/xhtml.ent">
 
-<!--
+<!-- Copyright (c) 2011, 2020 Eurotech and/or its affiliates and others This
```
MMaiero: The formatting messed up the header

MMaiero:
Not sure what this Ubuntu@10 file is

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/container/orchestration/ContainerOrchestrationService.java
GregoryIvo: No need. Decided to do this to not change the API. Will remove it if that is not a problem.

GregoryIvo:
Looks like some un-related tests are causing this to fail 
image

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/container/orchestration/ContainerConfiguration.java
```diff
@@ -72,24 +70,6 @@ public String getContainerName() {
         return this.containerName;
     }
 
-    /**
-     * Returns the base image for the associated container
-     *
-     * @return
-     */
```
MMaiero, line 79: Since this is API, we cannot remove API methods. We need to deprecate them first and then in 1 or 2 major releases, remove them. We should also suggest, for the time being, the alternative to use instead to allow users making the switch before removal

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/container/orchestration/ContainerConfiguration.java
```diff
@@ -72,24 +70,6 @@ public String getContainerName() {
...
-    public String getContainerImage() {
-        return this.containerImage;
-    }
-
-    /**
-     * Returns the image tag for the associated container
-     *
-     * @return
-     */
```
MMaiero, line 88: same here

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/container/orchestration/ContainerConfiguration.java
```diff
@@ -163,21 +145,12 @@ public String getContainerLoggingType() {
     }
 
     /**
-     * Returns the Registry credentials
+     * @since 2.4 Returns the {@link ImageConfiguration} object
      *
      * @return
      */
```
MMaiero, line 154: same here

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/container/orchestration/ContainerConfiguration.java
```diff
@@ -163,21 +145,12 @@ public String getContainerLoggingType() {
...
-    public Optional<RegistryCredentials> getRegistryCredentials() {
-        return this.registryCredentials;
-    }
-
-    /**
-     * Returns the image download timeout (in seconds)
-     *
-     * @return
-     */
```
MMaiero, line 178: and here

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/container/orchestration/ContainerOrchestrationService.java
```diff
@@ -41,25 +41,36 @@
...
-     * @param timeOutSeconds
-     *            a non negative integer representing the image download timeout in seconds
-     * @param registryCredentials
-     *            an optional that can contain the registry URL and credentials for authentication
+     * @param imageConfig
+     *            a ImageConfiguration object which contains info such as image name, tag, pull timeout in secconds and registry credentials.
      * @throws KuraException
      *             if the pull operation fails
      * @throws InterruptedException
      */
```
MMaiero, line 77: same here

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/container/orchestration/ContainerConfiguration.java
```diff
@@ -163,21 +146,52 @@ public String getContainerLoggingType() {
...
+     * @since 2.4 Returns the {@link ImageConfiguration} object
+     *
+     * @return
+     */
+    public ImageConfiguration getImageConfiguration() {
+        return this.imageConfig;
+    }
+    
+    /**
+     * @deprecated Since 2.4 use {@link #getImageConfiguration()}. Image related information has been moved to the {@link ImageConfiguration} class.
```
MMaiero: Let's follow the pattern linked to deprecate things. Usually we add the deprecation at the end

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/container/orchestration/ContainerConfiguration.java
MMaiero: Also for since annotations, please follow the schema used for the other APIs

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/container/orchestration/ContainerConfiguration.java
```diff
@@ -163,21 +146,52 @@ public String getContainerLoggingType() {
...
+     * Returns the base image for the associated container
+     *
+     * @return
+     */
+    public String getContainerImage() {
+        return this.imageConfig.getImageName();
+    }
+
+    /**
+     * @deprecated Since 2.4 use {@link #getImageConfiguration()}. Image related information has been moved to it's own class.
```
MMaiero: Do we really need to deprecate those methods here?

GregoryIvo:
@MMaiero  Based on our previous discussion, rather than depreciating the methods, I thought it would be best to add more information/description encouraging the user to check out the ImageConfig class for more image-related information.

MMaiero:
@GregoryIvo Please give a check on sonar. We need that passing to move on. Thanks

MMaiero:
Be careful that there are still API breakage due to the fact that API methods in ContainerOrchestrationService and ContainerConfiguration have been removed

GregoryIvo:
I bought back the mentioned methods. Using them only to call the new methods so that nothing is effective. Results in one Override for the method pullImage(). Now it can accept either ImageConfig or the variables that make up an ImageConfig (imageName, imageTag... ). 
Hopefully, this is an okay resolution?

MMaiero:
Please have a look at this result: https://ci.eclipse.org/kura/job/multibranch/job/PR-3938/21/console

nicolatimeus:
We have restored the `bnd-baseline-maven-plugin` during `org.eclipse.kura.api` bundle build. The plugin compares the exported APIs of the bundle that is being build with the latest available released version.
This plugin will cause the build to fail if it detects breaking changes not paired with a major version uptick in the corresponding provider packages versions (that signals a backwards incompatible change). 
In this case the build is failing because there are still some methods that have been removed since previous release, can you please restore them? (The major version of the exported packages should not be changed)

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/container/orchestration/ContainerConfiguration.java
```diff
@@ -325,12 +346,22 @@ public ContainerConfiguration build() {
             result.isFrameworkManaged = this.isFrameworkManaged;
             result.containerLoggerParameters = this.containerLoggerParameters;
             result.containerLoggingType = this.containerLoggingType;
-            result.registryCredentials = requireNonNull(this.registryCredentials,
-                    "Request Registry Credentials object cannot be null");
-            result.imageDownloadTimeoutSeconds = this.imageDownloadTimeoutSeconds;
+
```
GregoryIvo: Added this switch to fall back to old builder logic when an ImageConfig is not provided.  Is this an Anti-Pattern?

nicolatimeus: Is it possible/difficult to remove the additional private fields and change the old methods so that they modify the imageConfig field instead? Maybe changing the type of the imageConfig to ImageConfigurationBuilder can help.. in this way the old methods could just call the corresponding ones of the builder.. To make things easier, setImageConfiguration() could also be changed to received a ImageConfigurationBuilder instead of an ImageConfig.. then the build() method could invoke ImageConfigurationBuilder.build() to get the actual ImageConfiguration.. what do you think?

GregoryIvo: I like that a lot. Reduces complexity by getting rid of the IF and allows the end-user to "edit" an imported image config. Making these changes now.

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/container/orchestration/ContainerOrchestrationService.java
```diff
@@ -41,7 +41,37 @@ public interface ContainerOrchestrationService {
...
     public List<ContainerInstanceDescriptor> listContainerDescriptors();
-
+    
+    /**
+     * @since 2.4
+     * Lists all available images, returning the corresponding {@link ImageInstanceDescriptor}s
+     *
+     * @return a list of {@link ContainerInstanceDescriptor}s representing the available containers
+     */
+    public List<ImageInstanceDescriptor> listImageInstanceDescriptor();
```
MMaiero: I think the javadoc should look like this:
/**
     * Lists all available images, returning the corresponding {@link ImageInstanceDescriptor}s
     *
     * @return a list of {@link ContainerInstanceDescriptor}s representing the available containers
     * @ since 2.4
     */

MMaiero: Furthermore, I think we should align to the convention we have that if we return a list, the method should be listImageInstanceDescriptors

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/container/orchestration/ContainerOrchestrationService.java
```diff
@@ -41,7 +41,37 @@ public interface ContainerOrchestrationService {
...
+    
+    /**
+     * @since 2.4
+     * Deletes Image from container engine. Takes ImageID String as parameter. Will throw an error if image is being used by a container.
+     *
+     * @return void
+     * @throws KuraException 
+     */
+    public void deleteImage(String imageId) throws KuraException;
```
MMaiero, line 63: /**
     * Deletes the specified image from the container engine. Takes ImageID String as parameter. Will throw an error if
     * image is being used by a container.
     *
     * @param imageId
     *            string parameter that identifies the image to be deleted
     * @throws KuraException
     *             if the image is used by a container or if the image deletion process fails
     * `@since 2.4`
     */

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/container/orchestration/ContainerOrchestrationService.java
```diff
@@ -41,7 +41,37 @@ public interface ContainerOrchestrationService {
...
+     * @since 2.4
+     * Allows to pull the required image, using the specified tag and credentials. The image will be downloaded
+     * respecting the configured timeout in seconds.
+     *
+     * @param imageConfig
+     *            a ImageConfiguration object which contains info such as image name, tag, pull timeout in secconds and registry credentials.
+     * @throws KuraException
+     *             if the pull operation fails
+     * @throws InterruptedException
+     */
```
MMaiero, line 77: /**
     * Allows to pull the required image, using the specified tag and credentials. The image will be downloaded
     * respecting the configured timeout in seconds.
     *
     * @param imageConfig
     *            a ImageConfiguration object which contains info such as image name, tag, pull timeout in seconds and
     *            registry credentials.
     * @throws KuraException
     *             if the pull operation fails
     * @throws InterruptedException
     * `@since 2.4`
     */

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/container/orchestration/ContainerConfiguration.java
MMaiero: /**
      * Returns the {@link ImageConfiguration} object
      *
      * @return
      * `@since 2.4` 
      */

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/container/orchestration/ContainerConfiguration.java
```diff
@@ -300,22 +301,42 @@ public ContainerConfigurationBuilder setPrivilegedMode(Boolean containerPrivileg
...
-            this.imageDownloadTimeoutSeconds = imageDownloadTimeoutSeconds;
+            this.imageConfigBuilder.setImageDownloadTimeoutSeconds(imageDownloadTimeoutSeconds);
+            return this;
+        }
+
+        /**
+         * @since 2.4 builds image based on {@link ImageConfiguration}
+         *
+         */
+        public ContainerConfigurationBuilder setImageConfiguration(ImageConfiguration imageConfig) {
```
MMaiero, line 330: /**
          *  Sets {@link ImageConfiguration}
          * `@since 2.4`
          */

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/container/orchestration/ImageConfiguration.java
```diff
@@ -0,0 +1,145 @@
...
+    
+    private ImageConfiguration() {
+    }
+
+    /***
+     * Returns a Image's name as a String.
+     * 
+     * @return
+     */
+    public String getImageName() {
```
MMaiero, line 47: /**
      * Returns a Image's name as a String.
      * 
      * @return
      */

MMaiero: Same for all the other methods

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/container/orchestration/ImageConfiguration.java
```diff
@@ -0,0 +1,145 @@
...
+            this.imageDownloadTimeoutSeconds = imageDownloadTimeoutSeconds;
+            return this;
+        }
+
+        public ImageConfiguration build() {
+            ImageConfiguration result = new ImageConfiguration();
+
+            result.imageName = requireNonNull(this.imageName, "Request Image Name cannot be null");
+            result.imageTag = this.imageTag;
+            result.registryCredentials = requireNonNull(this.registryCredentials,
```
MMaiero, line 137: Do we need explicit credentials? If we have the case where no credentials are needed, can't we set this as empty by default and the the user overrides, when needed, with explicit credentials?
In other words, can we initialise the this.registryCredentials with empty?

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/container/orchestration/ImageConfiguration.java
```diff
@@ -0,0 +1,145 @@
...
+        }
+        ImageConfiguration other = (ImageConfiguration) obj;
+        return Objects.equals(this.imageName, other.imageName) && Objects.equals(this.imageTag, other.imageTag)
+                && Objects.equals(this.registryCredentials, other.registryCredentials);
+    }
+
+    public static final class ImageConfigurationBuilder {
+
+        private String imageName;
+        private String imageTag;
```
MMaiero: Can we initialise this by default as latest? So that we don't have to deal with nulls in the rest of the code if the user does not explicitly initialise the value in the builder?

MMaiero: Otherwise, we may use optional also here

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/container/orchestration/ImageInstanceDescriptor.java
```diff
@@ -0,0 +1,200 @@
...
+
+import static java.util.Objects.requireNonNull;
+
+import java.util.Map;
+import java.util.Objects;
+
+import org.osgi.annotation.versioning.ProviderType;
+
+/**
+ * Object which represents a image. Used to track created images, and images
```
MMaiero: Object which represents an image

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/container/orchestration/ImageInstanceDescriptor.java
```diff
@@ -0,0 +1,200 @@
...
+    private String imageId = "";
+    private String imageAuthor = "";
+    private String imageArch = "";
+    private long imageSize = 0;
+    private Map<String, String> imageLabels;
+
+    private ImageInstanceDescriptor() {
+    }
+
+    /***
```
MMaiero: there is a * that is not needed in the javadoc tag

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/container/orchestration/ImageInstanceDescriptor.java
```diff
@@ -0,0 +1,200 @@
...
+@ProviderType
+public class ImageInstanceDescriptor {
+
+    private String imageName = "";
+    private String imageTag = "";
+    private String imageId = "";
+    private String imageAuthor = "";
+    private String imageArch = "";
+    private long imageSize = 0;
+    private Map<String, String> imageLabels;
```
MMaiero, line 40: This is called imageLabels but then it is returned with the method imageTags. What is the correct naming?
Furthermore, why the previous fields are initialised as empty and imageLabels is not initialised? Can we align?

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/container/orchestration/ImageInstanceDescriptor.java
```diff
@@ -0,0 +1,200 @@
...
+            return this;
+        }
+
+        public ImageInstanceDescriptorBuilder setimageSize(long imageSize) {
+            this.imageSize = imageSize;
+            return this;
+        }
+
+        public ImageInstanceDescriptorBuilder setImageLabels(Map<String, String> imageLabels) {
+            this.imageLabels = imageLabels;
```
MMaiero, line 180: what about if this field is null?

kura/org.eclipse.kura.container.orchestration.provider/src/main/java/org/eclipse/kura/container/orchestration/provider/impl/ContainerOrchestrationServiceImpl.java
```diff
@@ -430,13 +413,14 @@ private String createContainer(ContainerConfiguration containerDescription) thro
             throw new IllegalStateException("failed to reach docker engine");
         }
 
-        if (containerDescription == null || containerDescription.getContainerImage() == null
-                || containerDescription.getContainerImageTag() == null) {
+        if (containerDescription == null || containerDescription.getImageConfiguration().getImageName() == null
```
MMaiero: Can the getImageName or imageTag be null now from the imageConfiguration object?

kura/org.eclipse.kura.container.orchestration.provider/src/main/java/org/eclipse/kura/container/orchestration/provider/impl/ContainerOrchestrationServiceImpl.java
```diff
@@ -733,4 +717,82 @@ public boolean equals(Object obj) {
...
+    }
+
+    @Override
+    public void deleteImage(String imageId) throws KuraException {
+        checkRequestEnv(imageId);
+        try {
+            this.dockerClient.removeImageCmd(imageId).exec();
+        } catch (Exception e) {
+            logger.error("Could not remove image {}.", imageId);
+            throw new KuraException(KuraErrorCode.OS_COMMAND_ERROR,
```
MMaiero: This error requires two parameters: https://github.com/eclipse/kura/blob/d8ce302a04c930dc534c7d4f1af68ada82e8d43b/kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/KuraErrorCode.java#L148
So be aware that the error managing will format the message as described here: https://github.com/eclipse/kura/blob/d8ce302a04c930dc534c7d4f1af68ada82e8d43b/kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/KuraErrorCode.java#L144.

kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/shared/service/GwtDeviceService.java
```diff
@@ -60,5 +62,8 @@ public interface GwtDeviceService extends RemoteService {
 
     @Audit(componentName = "UI Device", description = "check if container orchestrator is active")
     public boolean checkIfContainerOrchestratorIsActive(GwtXSRFToken token) throws GwtKuraException;
+    
+    @Audit(componentName = "UI Device", description = "Delete Image")
```
MMaiero: Maybe better "Delete Container Image" in the audit message

MMaiero:
In terms of visual representation, maybe this alignment is better (with the two tables aligned):

Screenshot 2022-04-27 at 09 11 03

I would also rename the tab "Docker Containers" to "Containers" matching what done in Kapua

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/container/orchestration/ImageInstanceDescriptor.java
GregoryIvo: The only field that cannot be null from my research is the Image Id. I have modified it accordingly.

GregoryIvo:
image

MMaiero:
Confirming the tests failures reported in Jenkins. It happens also in my setup:
```
[ERROR] Tests run: 6, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 0.028 s <<< FAILURE! - in org.eclipse.kura.container.orchestration.provider.ContainerConfigurationTest
[ERROR] testRequiredContainerCredentials(org.eclipse.kura.container.orchestration.provider.ContainerConfigurationTest)  Time elapsed: 0.028 s  <<< FAILURE!
java.lang.AssertionError: Expected exception: java.lang.NullPointerException
```

PR #3952, opened 2022-04-04
Kura 5.1.1 release notes

MMaiero:


nicolatimeus:
Should we remove the breaking changes section?

MMaiero:
Not sure. the same principles apply as for 5.1.0, right? So maybe it makes sense to keep them?

nicolatimeus (approved):

PR #3957, opened 2022-04-21
Fix TabWirelessUI

salvatore-coppola:
Now if the selected channel is not available in the regdom the wireless form fields are not filled.


MMaiero:
@mattdibi Can you please have a review? Thanks

mattdibi:
Tested working on RPi4 using my phone in hotspot mode with a 5GHz band. When connected and the error message comes up the fields of the form are not filled, as expected.

mattdibi (approved):

PR #3961, opened 2022-04-26
Restored baseline plugin analysis

nicolatimeus:

Restores the API baseline analysis by setting the previous api bundle version in the `org.eclipse.kura.api` package instead of the no longer maintained `distrib/RELEASE-INFO` folder


MMaiero (approved):

nicolatimeus:
The `bnd-baseline-maven-plugin` should be capable of figuring out the previous version scanning the maven repos without additional configuration.

PR #3975, opened 2022-05-06
feat: Switched to libsocket-can-java with support for all Kura architectures

nicolatimeus:

Switches to `com.eurotech:libsocket-can-java:1.0.0` that includes x86_64, arm32 and aarch64 support.
Pending CQ: https://dev.eclipse.org/ipzilla/show_bug.cgi?id=24095


nicolatimeus:
CQ approved

PR #3977, opened 2022-05-06
feat: Add support for Japanese

MMaiero:


nicolatimeus:
This Japanese messages are shown if the `?locale=jp` query string is added to console URL.
This step must be done manually, do we want to investigate auto-detection of the locale or insert some UI elements to allow a user to configure its locale (or both)?

MMaiero:
Yes, I was thinking on creating few items for that. Also because this translation is only part of the overall Ui visible to the user.

nicolatimeus (approved):

PR #3981, opened 2022-05-12
fix(network): DnsMonitorServiceImpl fixes

pierantoniomerlino:
This PR fixes some bugs in the `DnsMonitorServiceImpl`.

**Description of the solution adopted:** The following fixes are added:

- fixed NPE during configuration retrieval that caused the monitor thread to die;
- fixed wrong configuration comparation check
- the `NetworkConfiguration` now is retrieved by the `NetworkConfigurationService` at each monitor run. This fixes a wrong behavior in case of multiple modems.

### DIFF
--- a/kura/org.eclipse.kura.net.admin/src/main/java/org/eclipse/kura/net/admin/monitor/DnsMonitorServiceImpl.java
+++ b/kura/org.eclipse.kura.net.admin/src/main/java/org/eclipse/kura/net/admin/monitor/DnsMonitorServiceImpl.java
@@ -215,18 +215,21 @@ public class DnsMonitorServiceImpl implements DnsMonitorService, EventHandler {
         logger.debug("handleEvent - topic: {}", event.getTopic());
         String topic = event.getTopic();
 
-        if (topic.equals(NetworkConfigurationChangeEvent.NETWORK_EVENT_CONFIG_CHANGE_TOPIC)) {
-            try {
-                this.networkConfiguration = this.netConfigService.getNetworkConfiguration();
-            } catch (KuraException e) {
-                logger.error("Could not get network configuration", e);
+        Thread thread = new Thread(() -> {
+            if (topic.equals(NetworkConfigurationChangeEvent.NETWORK_EVENT_CONFIG_CHANGE_TOPIC)) {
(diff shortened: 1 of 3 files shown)


kura/org.eclipse.kura.net.admin/src/main/java/org/eclipse/kura/net/admin/monitor/DnsMonitorServiceImpl.java
```diff
@@ -80,23 +78,29 @@ public void setNetworkConfigurationService(NetworkConfigurationService netConfig
...
+        }
     }
 
     public void setDnsServerService(DnsServerService dnsServer) {
         this.dnsServer = dnsServer;
     }
 
     public void unsetDnsServerService(DnsServerService dnsServer) {
-        this.dnsServer = null;
+        if (dnsServer.equals(this.dnsServer)) {
```
nicolatimeus: There should be no need for this if the reference is static with 1..1 cardinality. I think you can even remove the unset methods (also from the component definition xml).

kura/org.eclipse.kura.net.admin/src/main/java/org/eclipse/kura/net/admin/monitor/DnsMonitorServiceImpl.java
```diff
@@ -217,25 +211,15 @@ protected void reconfigureDNSProxy(DnsServerConfigIP4 dnsServerConfigIP4) {
...
-                        props.put(propName, prop);
-                    }
-                }
-                try {
-                    this.networkConfiguration = new NetworkConfiguration(props);
-                } catch (Exception e) {
-                    logger.warn(e.getMessage(), e);
-                }
+            try {
+                this.networkConfiguration = this.netConfigService.getNetworkConfiguration();
```
nicolatimeus: As an enhancement we can consider moving this to a different thread to avoid performing blocking operations during an `EventAdmin` call.

nicolatimeus (approved):
Apart from the previous enhancement comment, changes look ok

MMaiero:
tests are failing

PR #3984, opened 2022-05-16
feat(api): Added new EnrollmentService API.

salvatore-coppola:


kura/org.eclipse.kura.api/META-INF/MANIFEST.MF
```diff
@@ -65,7 +65,7 @@ Export-Package: org.eclipse.kura;version="1.6.0",
  org.eclipse.kura.status;version="1.0.2",
  org.eclipse.kura.system;version="1.6.0",
  org.eclipse.kura.type;version="1.1.0",
- org.eclipse.kura.usb;version="1.3.0",
+ org.eclipse.kura.usb;version="2.0.0",
```
MMaiero: Why these changes if we are just adding a class in a package?

MMaiero: The new package should also be exported, I think.

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/KuraErrorCode.java
```diff
@@ -247,5 +247,11 @@ public enum KuraErrorCode {
      * 
      * @since 2.3
      */
-    UNSUPPORTED_MODEM;
+    UNSUPPORTED_MODEM,
+    /**
+     * Modem unsupported.
```
MMaiero: Probably this is not correct

kura/org.eclipse.kura.api/META-INF/MANIFEST.MF
```diff
@@ -2,10 +2,10 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: org.eclipse.kura.api
 Bundle-SymbolicName: org.eclipse.kura.api;singleton:=true
-Bundle-Version: 2.4.0.qualifier
```
nicolatimeus, line 5: There should be no need for a major version bump

kura/org.eclipse.kura.api/META-INF/MANIFEST.MF
salvatore-coppola: because the baseline check plugin gave me an error preventing the build.

kura/org.eclipse.kura.api/META-INF/MANIFEST.MF
salvatore-coppola: baseline check plugin said me it is required.

kura/org.eclipse.kura.api/META-INF/MANIFEST.MF
MMaiero: Do you have the same error with develop?

salvatore-coppola: in develop i haven't the same problem, i try to revert and see if the error persist.

kura/org.eclipse.kura.api/META-INF/MANIFEST.MF
salvatore-coppola: Solved, the problem appears only building in Eclipse, but after a full build it went away.

kura/org.eclipse.kura.api/META-INF/MANIFEST.MF
salvatore-coppola: Solved, the problem appears only building in Eclipse, but after a full build it went away.

nicolatimeus (approved):

PR #3986, opened 2022-05-16
feat(ai.triton.server): Add Model Encryption support for Triton Server Service

mattdibi:
Add Model Encryption support for Triton Server Service

If a password was specified during configuration **we assume all models in the path are encrypted with the same algorithm and can be decrypted with said password**.

Therefore Kura, upon applying the TritonServerImpl configuration, will perform the following:

-   Create a new folder in /tmp with the required permission to avoid allowing access from users that have no root privileges or need that (i.e. Triton, Kura). For this example we'll call this folder /tmp/decypted_models
    
-   Run the Inference Server using **this new folder as its Model Repository**
    

```
tritonserver --model-repository=/tmp/decrypted-models
```

-   For each model specified in the "Model" property:
    
    -   Will look for it in the "Model Repository Path" passed in the configuration
        
    -   Decrypt it and store it in the /tmp/decrypted_models folder (for example /tmp/decrypted_models/autoencoder_fp32)
        
    -   Tell Triton to load it from the /tmp/decrypted_models folder
        
    -   Wait for Triton to signal correct loading of the decrypted model
        
    -   Wipe the decrypted model folder /tmp/decrypted_models/autoencoder_fp32 from the filesystem as soon as Triton signals model load completion (see "[Model Ready](https://github.com/kserve/kserve/blob/master/docs/predict-api/v2/required_api.md#model-ready)" and [here](https://github.com/kserve/kserve/blob/master/docs/predict-api/v2/required_api.md#model-ready-1))
        

### Encryption procedure

Given a trained model inside the folder tf_autoencoder_fp32

```
tf_autoencoder_fp32
├── 1
│   └── model.savedmodel
│       ├── assets
│       ├── keras_metadata.pb
│       ├── saved_model.pb
│       └── variables
│           ├── variables.data-00000-of-00001
│           └── variables.index
└── config.pbtxt

4 directories, 5 files
```

We'll need to archive it with:

```
zip -vr tf_autoencoder_fp32.zip tf_autoencoder_fp32/
```

and encrypt the archive with:

```
gpg --armor --symmetric --cipher-algo AES256 tf_autoencoder_fp32.zip
```

The resulting archive tf_autoencoder_fp32.zip.asc can be decrypted by [this](https://github.com/bcgit/bc-java/blob/88b6c7c6fff096cc69ebfa5361448c928a67187f/pg/src/main/java/org/bouncycastle/openpgp/examples/PBEFileProcessor.java#L59) method.


kura/org.eclipse.kura.ai.triton.server/META-INF/MANIFEST.MF
```diff
@@ -7,6 +7,8 @@ Bundle-Vendor: Eclipse Kura
 Bundle-RequiredExecutionEnvironment: JavaSE-1.8
 Service-Component: OSGI-INF/*.xml
 Bundle-ClassPath: .,
+ lib/bcpg-jdk15on-1.68.jar,
+ lib/bcprov-jdk15on-1.68.jar,
```
MMaiero: These are in the Kura target platform. In theory you should just import them without having in the lib folder

mattdibi: I'm sure there's one for the provider (the `bcprov`) but I looked for the PGP one (`bcpg`) and couldn't find it. I'll double check.

MMaiero: I think we have never added it. But maybe it makes sense to have it in the target platform for other applications. @nicolatimeus What do you think?

nicolatimeus: Yes, it can make sense to have it in the target platform if it is not that big in size.

MMaiero:
https://sonarcloud.io/summary/new_code?id=org.eclipse.kura%3Akura&branch=PR-3986

kura/org.eclipse.kura.ai.triton.server/META-INF/MANIFEST.MF
```diff
@@ -23,9 +25,12 @@ Import-Package: com.google.common.base;version="25.0.0",
  com.google.gson;version="2.7.0",
  javax.annotation;version="1.2.0",
  org.apache.commons.io;version="2.4.0",
+ org.apache.commons.io.filefilter;version="2.11.0",
  org.eclipse.kura;version="[1.0,2.0)",
  org.eclipse.kura.ai.inference;version="[1.0,1.1)",
+ org.eclipse.kura.cloud.app.command;version="[1.0,2.0)",
  org.eclipse.kura.configuration;version="[1.1,2.0)",
+ org.eclipse.kura.crypto;version="1.2.2",
```
MMaiero: range needed

mattdibi:
> https://sonarcloud.io/summary/new_code?id=org.eclipse.kura%3Akura&branch=PR-3986

On this topic: Sonar highlights [this Security Hotspot](https://sonarcloud.io/project/security_hotspots?id=org.eclipse.kura%3Akura&branch=PR-3986&hotspots=AYDOCWt4PqzFYd-zKJEM) related to the use of a folder inside `/tmp`.

This is not an issue because we're using a sub-folder [with strict access permissions](https://github.com/mattdibi/kura/blob/d09fbc912be396c9c27813aa2a5e0f9d252cf468/kura/org.eclipse.kura.ai.triton.server/src/main/java/org/eclipse/kura/ai/triton/server/TritonServerEncryptionUtils.java#L81) which is what Sonar suggests:

> Recommended Secure Coding Practices
> - Use a dedicated sub-folder with tightly controlled permissions
> - Use ...

Unfortunately I don't have the permission to change the status of the Security Hotspot

MMaiero:
Do you have an idea on why this was not picked up by Sonar?

mattdibi:
> Do you have an idea on why this was not picked up by Sonar?

Mh...I don't think it's smart enough to understand that the folder is created with the correct permissions in another method...

I mean, the issue is highlighted [here](https://github.com/mattdibi/kura/blob/d09fbc912be396c9c27813aa2a5e0f9d252cf468/kura/org.eclipse.kura.ai.triton.server/src/main/java/org/eclipse/kura/ai/triton/server/TritonServerServiceImpl.java#L102) while the folder is created [here](https://github.com/mattdibi/kura/blob/d09fbc912be396c9c27813aa2a5e0f9d252cf468/kura/org.eclipse.kura.ai.triton.server/src/main/java/org/eclipse/kura/ai/triton/server/TritonServerEncryptionUtils.java#L79). Can Sonar "remember" what were the permission when the folder was created? Does it understand them in the first place? I don't think so as it is a little out of scope for Sonar... but I might be wrong. I think this is the reason why it asks us for a review...

What do you think?

kura/org.eclipse.kura.ai.triton.server/src/main/java/org/eclipse/kura/ai/triton/server/TritonServerEncryptionUtils.java
```diff
@@ -0,0 +1,202 @@
...
+    protected static void cleanRepository(String modelRootPath) {
+        if (!Files.isDirectory(Paths.get(modelRootPath))) {
+            logger.warn("Model root folder {} does not exist", modelRootPath);
+            return;
+        }
+
+        try {
+            FileUtils.cleanDirectory(new File(modelRootPath));
+        } catch (IOException e) {
+            logger.warn("Cannot clean directory at path {}", modelRootPath);
```
nicolatimeus: Please log also exception stacktraces on failure.

nicolatimeus: Please log also exception stacktraces on failure.

nicolatimeus (approved):

nicolatimeus:
This PR adds the org.bouncycastle/bcpg-jdk15on/1.68 dependency:
https://gitlab.eclipse.org/eclipsefdn/emo-team/iplab/-/issues/52
MicrosoftTeams-image (5)

PR #3987, opened 2022-05-18
fix(core.keystore): Fixed ability to create a keystore without password.

salvatore-coppola:
Now the default keystore password is immediatly set in the component creation.

Also the default KeystoreService configuration file location is changed from /tmp to /tmp/keystore.ks

### DIFF
--- a/kura/org.eclipse.kura.core.keystore/src/main/java/org/eclipse/kura/core/keystore/FilesystemKeystoreServiceImpl.java
+++ b/kura/org.eclipse.kura.core.keystore/src/main/java/org/eclipse/kura/core/keystore/FilesystemKeystoreServiceImpl.java
@@ -259,7 +259,6 @@ public class FilesystemKeystoreServiceImpl implements KeystoreService, Configura
         }
 
         setKeystorePassword(this.loadKeystore(options), passwordChar);
-
     }
 
     private void updateKeystorePath(KeystoreServiceOptions newOptions) {
--- a/kura/org.eclipse.kura.core.keystore/src/main/java/org/eclipse/kura/core/keystore/FilesystemKeystoreServiceImpl.java
+++ b/kura/org.eclipse.kura.core.keystore/src/main/java/org/eclipse/kura/core/keystore/FilesystemKeystoreServiceImpl.java
@@ -249,12 +249,13 @@ public class FilesystemKeystoreServiceImpl implements KeystoreService, Configura
         }
 
         // Immediately save the keystore with the default password to allow to be loaded with the default password.
-        OutputStream os = new FileOutputStream(fKeyStore);
-        KeyStore newKeystore = KeyStore.getInstance(KeyStore.getDefaultType());
-        newKeystore.load(null, passwordChar);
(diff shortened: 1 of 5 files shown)


kura/org.eclipse.kura.core.keystore/src/main/java/org/eclipse/kura/core/keystore/FilesystemKeystoreServiceImpl.java
```diff
@@ -233,21 +234,38 @@ private boolean keystoreExists(String keystorePath) {
         return keystorePath != null && new File(keystorePath).isFile();
     }
 
-    private void createKeystore(String keystorePath) throws IOException {
+    private void createKeystore(KeystoreServiceOptions options)
+            throws IOException, KuraException, KeyStoreException, NoSuchAlgorithmException, CertificateException {
```
MMaiero: Do we need to throw so many exception to log them immediately on the caller?
Does it make sense to catch and distinguish the different exceptions/causes?

salvatore-coppola: I've captured all exception in one generic and thrown it.

kura/org.eclipse.kura.core.keystore/src/main/java/org/eclipse/kura/core/keystore/FilesystemKeystoreServiceImpl.java
```diff
@@ -233,21 +234,40 @@ private boolean keystoreExists(String keystorePath) {
...
         File fKeyStore = new File(keystorePath);
         if (!fKeyStore.createNewFile()) {
             logger.error("Keystore file already exists at location {}", keystorePath);
+            throw new KuraException(KuraErrorCode.CONFIGURATION_ATTRIBUTE_INVALID, "keystore.path", keystorePath,
+                    "file already exists");
         }
+
+        // Immediately save the keystore with the default password to allow to be loaded with the default password.
+        try (OutputStream os = new FileOutputStream(fKeyStore)) {
+
```
mattdibi: Extra space here

```suggestion
```

kura/org.eclipse.kura.core.keystore/src/main/java/org/eclipse/kura/core/keystore/FilesystemKeystoreServiceImpl.java
```diff
@@ -233,21 +234,40 @@ private boolean keystoreExists(String keystorePath) {
...
+            newKeystore.load(null, passwordChar);
+            newKeystore.store(os, passwordChar);
+            os.flush();
+        } catch (Exception e) {
+            logger.error("Unable to load and store the keystore", e);
+            throw e;
+        }
+
+        setKeystorePassword(this.loadKeystore(options), passwordChar);
+
```
mattdibi: Extra space here too maybe?

```suggestion
```

mattdibi (approved):
LGTM

PR #4008, opened 2022-06-01
fix(ai.triton.server): resolved failed startup on configuration update with loaded models

mattdibi:
Fixed failed Triton Server startup when there were models loaded in the server.

**Description of the solution adopted:** If there are models loaded in Triton, it takes longer to spin down than when there's no model loaded. Since there were no checks in place for verifying Triton proper shutdown on the configuration update it was possible for the `TritonServerServiceImpl` to update the configuration without correctly restarting the Triton Server local instance.

With this PR:
- we introduce a wait on `update` and `deactivate` to check for correct shutdown of the server
- we refactored the sources so that we no longer have magic numbers in the code(`N_ATTEMPT` and `WAIT_ATTEMPT_MS` constants) 
- fixed the log entry on Triton Server shutdown
- we increased the wait time for the various checks (i.e. increased `WAIT_ATTEMPT_MS`).

### DIFF
--- a/kura/org.eclipse.kura.ai.triton.server/src/main/java/org/eclipse/kura/ai/triton/server/TritonServerLocalManager.java
+++ b/kura/org.eclipse.kura.ai.triton.server/src/main/java/org/eclipse/kura/ai/triton/server/TritonServerLocalManager.java
@@ -119,7 +119,7 @@ public class TritonServerLocalManager {
         }
     }
 
-    private boolean isLocalServerRunning() {
+    public boolean isLocalServerRunning() {
         boolean isRunning = false;
         if (this.commandExecutorService.isRunning(TRITONSERVER)) {
             isRunning = true;
--- a/kura/org.eclipse.kura.ai.triton.server/src/main/java/org/eclipse/kura/ai/triton/server/TritonServerLocalManager.java
+++ b/kura/org.eclipse.kura.ai.triton.server/src/main/java/org/eclipse/kura/ai/triton/server/TritonServerLocalManager.java
@@ -119,7 +119,7 @@ public class TritonServerLocalManager {
         }
     }
 
-    public boolean isLocalServerRunning() {
+    protected boolean isLocalServerRunning() {
         boolean isRunning = false;
         if (this.commandExecutorService.isRunning(TRITONSERVER)) {
             isRunning = true;
--- a/kura/org.eclipse.kura.ai.triton.server/src/main/java/org/eclipse/kura/ai/triton/server/TritonServerLocalManager.java
+++ b/kura/org.eclipse.kura.ai.triton.server/src/main/java/org/eclipse/kura/ai/triton/server/TritonServerLocalManager.java
@@ -74,7 +74,7 @@ public class TritonServerLocalManager {
     private void startLocalServer() {
         this.commandExecutorService.execute(this.serverCommand, status -> {
             if (status.getExitStatus().isSuccessful()) {
-                logger.info("Nvidia Triton Server started");
+                logger.info("Nvidia Triton Server gracefully shutdown");
             } else {
                 logger.info("Nvidia Triton Server not started. Exit value: {}", status.getExitStatus().getExitCode());
             }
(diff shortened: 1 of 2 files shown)


kura/org.eclipse.kura.ai.triton.server/src/main/java/org/eclipse/kura/ai/triton/server/TritonServerLocalManager.java
```diff
@@ -74,7 +74,7 @@ private void startLocalServerMonitor() {
     private void startLocalServer() {
         this.commandExecutorService.execute(this.serverCommand, status -> {
             if (status.getExitStatus().isSuccessful()) {
-                logger.info("Nvidia Triton Server started");
+                logger.info("Nvidia Triton Server gracefully shutdown");
```
pierantoniomerlino, line 77: Why here the server is shutdown?

kura/org.eclipse.kura.ai.triton.server/src/main/java/org/eclipse/kura/ai/triton/server/TritonServerLocalManager.java
```diff
@@ -119,7 +119,7 @@ private void stopScheduledExecutor() {
         }
     }
 
-    private boolean isLocalServerRunning() {
+    public boolean isLocalServerRunning() {
```
pierantoniomerlino: Is this method used somewhere else? If no, please keep it private.

kura/org.eclipse.kura.ai.triton.server/src/main/java/org/eclipse/kura/ai/triton/server/TritonServerServiceImpl.java
```diff
@@ -151,6 +153,16 @@ private void startLocalInstance() {
...
+
+        int counter = 0;
+        while (this.tritonServerLocalManager.isLocalServerRunning()) {
+            if (counter++ >= N_ATTEMPT) {
+                logger.warn("Cannot stop local server instance.");
+                return;
+            }
+            TritonServerLocalManager.sleepFor(WAIT_ATTEMPT_MS);
+        }
+
```
pierantoniomerlino, line 165: Indeed the N_ATTEMPTS and WAIT_ATTEMPTS_MS seems to be a bit arbitrary... even if this is an improvement compared with the previous behavior. Is there a way to improve this further? Maybe spawing a thread for checking the server status and calling a callback upon termination?

kura/org.eclipse.kura.ai.triton.server/src/main/java/org/eclipse/kura/ai/triton/server/TritonServerLocalManager.java
mattdibi: Being this a lambda it get executed only once the `this.commandExecutorService.execute` method returns i.e. when the Triton Server is shut down.

kura/org.eclipse.kura.ai.triton.server/src/main/java/org/eclipse/kura/ai/triton/server/TritonServerLocalManager.java
mattdibi: I think the correct way to handle this would be to set it as below since we use it in `TritonServerServiceImpl`:

```suggestion
    protected boolean isLocalServerRunning() {
```

what do you think?

kura/org.eclipse.kura.ai.triton.server/src/main/java/org/eclipse/kura/ai/triton/server/TritonServerServiceImpl.java
mattdibi: I agree. I already had a discussion about this with @MMaiero and we were opting for exposing these two parameters in the metatype for the user to choose.

I also like the callback solution but its scope may require a new PR.

Which of the two would you prefer?

kura/org.eclipse.kura.ai.triton.server/src/main/java/org/eclipse/kura/ai/triton/server/TritonServerLocalManager.java
pierantoniomerlino: That's true. The process is in foreground, so the command exits when the server is shutdown.

kura/org.eclipse.kura.ai.triton.server/src/main/java/org/eclipse/kura/ai/triton/server/TritonServerServiceImpl.java
pierantoniomerlino: I think that for now we can expose to the user a value representing a timeout. After that period of time, the server is killed. For the server start, consider the 0 value that means "try forever".

mattdibi: Good for me. If you agree I would open a new PR for that since it's a meaningful change in the UX that exceeds the scope of this PR (startup bugfix).

PR #4012, opened 2022-06-08
feat(net.admin): Added a cache for the network configuration

pierantoniomerlino:

This PR adds a cache for the network configuration in order to improve the webUI responsiveness.

**Description of the solution adopted:** A `NetworkConfiguration` object is now cached by the `NetworkConfigurationServiceImpl`. In this way, when the webUI request the configuration for showing the values on the screen, a cached object is returned resulting in a fast reply. The internal `NetworkConfiguration` is re-created at every call of the `updated` method.
Since the `NetworkConfiguration` contains also dynamic values (i.e. rssi or ip addresses when in dhcp mode), the request can be performed with a `recompute` option that forces the object recreation. In this way, the webUI will show updated values.
In the network tab of the webUI, a new "Refresh" button is added: it forces the webUI to retrieve updated values from the backend, regenerating the `NetworkConfiguration` object and reading new values from the cellular modem.


kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/net/wifi/WifiClientMonitorService.java
```diff
@@ -27,5 +27,32 @@ public interface WifiClientMonitorService {
...
+     * @param interfaceName
+     *            the name of the wireless interface
+     * @param ssid
+     *            the name of the ssid the interface is attached to
+     * @param recompute
+     *            if set to true, the rssi is recomputed. Otherwise a cached value is returned
+     * @return an integer number representing the rssi
+     * @throws KuraException
+     */
+    public int getSignalLevel(String interfaceName, String ssid, boolean recompute) throws KuraException;
```
MMaiero, line 59: Missing since here?

kura/org.eclipse.kura.net.admin/src/main/java/org/eclipse/kura/net/admin/NetworkAdminServiceImpl.java
```diff
@@ -242,43 +244,47 @@ protected List<WifiAccessPoint> getWifiAccessPoints(String ifaceName) throws Kur
...
-    // FIXME: This api should be deprecated in favor of the following signature:
-    // List<? extends NetInterfaceConfig<? extends NetInterfaceAddressConfig>> getNetworkInterfaceConfigs()
+    public List<? extends NetInterfaceConfig<? extends NetInterfaceAddressConfig>> getNetworkInterfaceConfigs(
+            boolean recompute) throws KuraException {
+
+        logger.debug("Getting all networkInterfaceConfigs");
+        long started = System.currentTimeMillis();
+        List<? extends NetInterfaceConfig<? extends NetInterfaceAddressConfig>> nc = getNetworkConfigurationInternal(
+                recompute).getNetInterfaceConfigs();
+        logger.info("GetNetworkConfiguration {}", (System.currentTimeMillis() - started));
```
MMaiero: Probably better to look at the logging before merging since it logs massively now

kura/org.eclipse.kura.net.admin/src/main/java/org/eclipse/kura/net/admin/NetworkAdminServiceImpl.java
```diff
@@ -1334,6 +1364,12 @@ public void handleEvent(Event event) {
         String topic = event.getTopic();
         if (topic.equals(NetworkConfigurationChangeEvent.NETWORK_EVENT_CONFIG_CHANGE_TOPIC)) {
             this.pendingNetworkConfigurationChange = false;
+            // } else if (topic.equals(NetworkStatusChangeEvent.NETWORK_EVENT_STATUS_CHANGE_TOPIC)) {
```
MMaiero: Does this commented code need to stay?

kura/org.eclipse.kura.net.admin/src/main/java/org/eclipse/kura/net/admin/NetworkConfigurationService.java
```diff
@@ -24,10 +26,36 @@ public interface NetworkConfigurationService {
...
+     * @since 1.5
+     */
+    public Optional<NetworkConfiguration> getNetworkConfiguration(boolean recompute) throws KuraException;
+
+    /**
+     * Update the cached configuration.
+     * 
+     * @throws KuraException
+     */
+    public void updateCurrentNetworkConfiguration() throws KuraException;
```
MMaiero, line 61: Since missing.

kura/org.eclipse.kura.net.admin/src/main/java/org/eclipse/kura/net/admin/NetworkConfigurationServiceImpl.java
```diff
@@ -341,12 +339,36 @@ public synchronized ComponentConfiguration getConfiguration() throws KuraExcepti
         // This method returns the network configuration properties without the current values.
         // i.e. the ip address that should be applied to the system, but not the actual one.
         logger.debug("getConfiguration()");
-        return new ComponentConfigurationImpl(PID, getDefinition(),
-                getNetworkConfiguration().getConfigurationProperties());
+        Optional<NetworkConfiguration> networkConfiguration = getNetworkConfiguration(false);
+        if (networkConfiguration.isPresent()) {
+            return new ComponentConfigurationImpl(PID, getDefinition(),
+                    networkConfiguration.get().getConfigurationProperties());
+        } else {
```
MMaiero: is the else required here?

kura/org.eclipse.kura.net.admin/src/main/java/org/eclipse/kura/net/admin/NetworkConfigurationServiceImpl.java
```diff
@@ -341,12 +339,36 @@ public synchronized ComponentConfiguration getConfiguration() throws KuraExcepti
...
+            throw new KuraRuntimeException(KuraErrorCode.CONFIGURATION_ERROR,
+                    "The network component configuration cannot be retrieved");
+        }
     }
 
     @Override
     public synchronized NetworkConfiguration getNetworkConfiguration() throws KuraException {
+        if (this.currentNetworkConfiguration.isPresent()) {
+            return this.currentNetworkConfiguration.get();
+        } else {
```
MMaiero: is the else required here?

kura/org.eclipse.kura.net.admin/src/main/java/org/eclipse/kura/net/admin/modem/hspa/HspaModem.java
```diff
@@ -83,6 +87,37 @@ public HspaModem(ModemDevice device, String platform, ConnectionFactory connecti
         this.connectionFactory = connectionFactory;
     }
 
+    public void initModemParameters() {
+        if (device != null) {
```
MMaiero: Maybe if it is null we return immediately?

kura/org.eclipse.kura.net.admin/src/main/java/org/eclipse/kura/net/admin/modem/sierra/mc87xx/SierraMc87xx.java
```diff
@@ -146,14 +148,26 @@ public String getSerialNumber() throws KuraException {
         return this.serialNumber;
     }
 
+    @Override
+    public String getMobileSubscriberIdentity(boolean recompute) throws KuraException {
+        // not implemented
+        return null;
```
MMaiero: should we throw here an exception?

MMaiero (approved):

kura/org.eclipse.kura.net.admin/src/main/java/org/eclipse/kura/net/admin/NetworkConfigurationServiceImpl.java
pierantoniomerlino: This is the "old" method, now deprecated. I added the else branch to maintain the old behavior.

MMaiero: True, but since we have the return in the if, we don't need the else, right?

pierantoniomerlino: My concern is that now the `currentNetworkConfiguration` is an Optional. So, if it is empty, we cannot get a value from it. I could return a null, but this is not the same behavior of the previous code.

kura/org.eclipse.kura.net.admin/src/main/java/org/eclipse/kura/net/admin/NetworkConfigurationServiceImpl.java
MMaiero: Still pending here

kura/test/org.eclipse.kura.net.admin.test/src/test/java/org/eclipse/kura/net/admin/modem/QuectelGenericTest.java
```diff
@@ -0,0 +1,236 @@
+package org.eclipse.kura.net.admin.modem;
```
MMaiero: Copyright header missing

MMaiero (approved):

PR #4018, opened 2022-06-10
refactor(wire.h2db.component.provider): Extracted common code from H2DB wire components

pierantoniomerlino:

This PR extracts the common code from the `org.eclipse.kura.wire.h2db.component.provider`

**Description of the solution adopted:** Since the code in the `org.eclipse.kura.wire.h2db.component.provider` bundle can be re-used by bundles that implements `BaseDbService`, this PR extracts the code from there and creates a new `org.eclipse.kura.wire.basedb.component.provider`. Other bundles can extends the common classes inside the bundle anc override the methods if needed.


pierantoniomerlino:
After some investigation, we prefer to have a single wire component (store or filter) that manage multiple type of databases instead of having multiple components for specific dbs.

So, we close this item.

PR #4046, opened 2022-07-11
feat(wire.db.component.provider): Add generic db wire components

pierantoniomerlino:

This PR adds generic DB filter and store wire components that can be used with all the services that implement DBService and H2DBService.

**Description of the solution adopted:** A new `org.eclipse.kura.wire.db.component.provider` has been added. It contains the DB store and filter wire components. They can be uses with the H2DB as well with database services that implement `DBService`.

**Any side note on the changes made:** The old H2 wire components have been deprecated and replaced by the generic ones.


kura/org.eclipse.kura.wire.db.component.provider/OSGI-INF/DbWireRecordFilterComponent.xml
```diff
@@ -0,0 +1,46 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+    
+    Copyright (c) 2022 Eurotech and/or its affiliates and others
+  
+    This program and the accompanying materials are made
+    available under the terms of the Eclipse Public License 2.0
+    which is available at https://www.eclipse.org/legal/epl-2.0/
+ 
+	SPDX-License-Identifier: EPL-2.0
```
MMaiero, line 10: Formatting seems odd

kura/org.eclipse.kura.wire.db.component.provider/src/main/java/org/eclipse/kura/internal/wire/db/common/CommonDbServiceProvider.java
```diff
@@ -0,0 +1,243 @@
+package org.eclipse.kura.internal.wire.db.common;
```
MMaiero, line 13: Missing copyright header

pierantoniomerlino:
The build succeeded, but sonat complains about code duplication. However the duplicated code is with the deprecated classes that have to be removed in a next release.

PR #4055, opened 2022-07-22
fix: Config update could skip wifi file creation

MMaiero:


marcellorinaldo (approved):
Tested and it is working. Files `/etc/wpa-supplicant-{wlan-ifname}.conf` and `/etc/hostapd-{wlan-ifname}.conf` are correctly written when in station or AP mode, even if not existent at installation.

PR #4056, opened 2022-07-22
fix: added missing since annotations in keystore api

MMaiero:


salvatore-coppola:
@MMaiero we are thinking about to remove these two methods from api @nicolatimeus to avoid a bind with bc classes

MMaiero:
Ok for me.

PR #4065, opened 2022-07-27
fix: support topic names using "//" or wildcards

### LINKED ISSUE #3972: Topic name validation: issue with names containing "//" (Cloud Subscriber)
**Describe the bug**
Let Kura establish an MQTT connection as a Cloud Subscriber to a Topic having this kind of name:

```bash
a/b//d
```

Received Payload will not trigger a wiregraph component. No msg will be available on the log.

**To Reproduce**

1. Create a Cloud Connection with a Cloud Subscriber pointing to an MQTT Broker
2. Define the topic inside Kura as a valid one on your broker. Be sure that a valid topic name contains two subsequent "/" as reported in this bug description
3.  Establish a connection to the broker
4. Add a couple of nodes to a Wiregraph: Cloud Subscriber Listener triggered by the connection at step 1; its output goes into a Logger node
5. Start sending an MQTT packet to the topic (simulate it with mosquitto if needed)
6.  The logger component will not be triggered

**Expected behavior**

You should see the payload in the log file (Logger node has been triggered in the wiregraph)

 - Kura 4.1.3 (emulated or on a Raspberry) and 5.x

**Additional context**

Could it be connected to this bug in Paho MQTT?

https://github.com/eclipse/paho.mqtt.java/issues/718

spulci:
Support topic names using "//" or wildcards

**Related Issue:** This PR closes #3972

**Description of the solution adopted:** Embedded Paho Client Mqttv3 jar version 1.2.1 contains a bug (https://github.com/eclipse/paho.mqtt.java/issues/718) that brakes Kura Cloud Connections to topics with names cointaining wildcards or "//".  The bug has been fixed since version 1.2.3; this contrib updates the embedded version to the lastest Paho stable release 1.2.5


spulci:
@MMaiero I need your help troubleshooting this issue. The PR has an issue on MQTT tests using SSL. You may have a look at the CI log

```bash
Caused by: org.eclipse.paho.client.mqttv3.MqttException: MqttException
Caused by: javax.net.ssl.SSLHandshakeException: No name matching localhost found
Caused by: java.security.cert.CertificateException: No name matching localhost found
```

MMaiero:
Hi, did you have a look if this is a known issue of version 1.2.5?
If the tests were working before with the previous version of Paho, maybe they have added a change that is a breaking change.

spulci:
> Hi, did you have a look if this is a known issue of version 1.2.5? If the tests were working before with the previous version of Paho, maybe they have added a change that is a breaking change.

Nope. Versione 1.2.1-k2 is a fork:

https://github.com/eclipse/kura/pull/2794

Kura is actually using a modified version of Paho as reported in that PR. We do not encount any kind of issue in our test on Kura 4.1.3 due to the fact we're not using SSL yet. Hopefully I will port the same fixes on 1.2.5, releasing a forked version for you if the code base is not so different from 1.2.1

I was not aware of this fork, sorry

MMaiero:
Yes, we have created that fork due to some unresolved issues in paho and their threading model that would leave dangling threads over time. I think there is still an issue in their repo that I have open-end and probably also a PR. But those were never addressed by the pho committers.

But the issue seems related to tests failing with 1.2.5 that were previously passing with 1.2.1 (independently from the fork). So that's why I have asked the question.
It may be that there is an issue in 1.2.5 or a breaking change that requires changes to the tests.

spulci:
A real bad day for me.. I have deleted my forked kura repo inadvertently :( I will come back with a new PR when I will have a fixed version of embedded Paho too....so sorry :(

PR #4068, opened 2022-07-28
refactor(ai.triton.server): remove app logic from TritonServerServiceOptions

mattdibi:
This PR removes a part of the application logic from the `TritonServerServiceOptions` which now becomes a simple wrapper aroung the TritonServerService configuration.

This changes are propedeutics to the addition of the future `TritonServerNativeService` and `TritonServerRemoteService` components since they won't rely on the `local.enabled` configuration property.

### DIFF
--- a/kura/org.eclipse.kura.ai.triton.server/src/main/java/org/eclipse/kura/ai/triton/server/TritonServerServiceOptions.java
+++ b/kura/org.eclipse.kura.ai.triton.server/src/main/java/org/eclipse/kura/ai/triton/server/TritonServerServiceOptions.java
@@ -99,16 +99,7 @@ public class TritonServerServiceOptions {
     }
 
     public String getAddress() {
-        String address = "";
-        if (this.isLocal) {
-            address = "localhost";
-        } else {
-            final Object propertyAddress = this.properties.get(PROPERTY_ADDRESS);
-            if (propertyAddress instanceof String) {
-                address = ((String) propertyAddress).trim();
-            }
-        }
-        return address;
+        return getStringProperty(PROPERTY_ADDRESS);
     }
 
     public boolean isLocalEnabled() {
--- a/kura/org.eclipse.kura.ai.triton.server/src/main/java/org/eclipse/kura/ai/triton/server/TritonServerServiceOptions.java
+++ b/kura/org.eclipse.kura.ai.triton.server/src/main/java/org/eclipse/kura/ai/triton/server/TritonServerServiceOptions.java
@@ -99,7 +99,12 @@ public class TritonServerServiceOptions {
     }
 
     public String getAddress() {
-        return getStringProperty(PROPERTY_ADDRESS);
+        String address = "";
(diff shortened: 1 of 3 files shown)


mattdibi:
Tested working on Nvidia Jetson Nano with Triton 2.19.0

```
jetson@jetson-nano:~$ uname -a
Linux jetson-nano 4.9.253-tegra #1 SMP PREEMPT Wed Apr 20 14:25:12 PDT 2022 aarch64 aarch64 aarch64 GNU/Linux
```
```
jetson@jetson-nano:~$ lsb_release -a
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 18.04.6 LTS
Release:	18.04
Codename:	bionic
```
Tested multiple configuration combinations:
- Remote with working server @`localhost` -> Triton correctly loads the desired model
- Remote with no server @`192.168.1.82` -> Reports correctly that it's not able to load the desired model
- Local with server set at bogus address -> Triton correctly starts and load the desired model
- Local with server set at bogus address w/ model encryption -> Triton correctly starts and load the desired model

The fact that Triton loads the model is evidence of the fact that the connection was established correctly.

kura/org.eclipse.kura.ai.triton.server/src/main/java/org/eclipse/kura/ai/triton/server/TritonServerServiceOptions.java
```diff
@@ -99,16 +99,7 @@ public TritonServerServiceOptions(final Map<String, Object> properties) {
...
-        if (this.isLocal) {
-            address = "localhost";
-        } else {
-            final Object propertyAddress = this.properties.get(PROPERTY_ADDRESS);
-            if (propertyAddress instanceof String) {
-                address = ((String) propertyAddress).trim();
-            }
-        }
-        return address;
+        return getStringProperty(PROPERTY_ADDRESS);
```
pierantoniomerlino: Do we need a trim?

mattdibi: Not necessarily but I don't think it hurts. Do you think it can have some negative side effects? I used `getStringProperty` here mainly to avoid code duplication...

pierantoniomerlino: The `getStringProperty` is fine. I mean that I expect that the address has no spaces, so the trimming can be useful in this case. Also to preserve the old behavior.

mattdibi: Sorry, I was answering under the assumption that `getStringProperty` had the `trim`inside. Fixing

pierantoniomerlino (approved):

PR #4070, opened 2022-07-29
feat(ai.triton.server): add Remote Triton Server Service component

mattdibi:
This is the second step towards the deprecation of the current `TritonServerServiceImpl` class following #4064 

The purpose of this PR is to add the `TritonServerRemoteService` Factory Components which exposes the same functionalities as the `TritonServerService` with `local.enabled` turned off (i.e. "Remote" mode) but with a smaller configuration interface.

The final goal is to have 2 different UI pages for configuring the 2 new factory components.

See UML for reference:
image
Highlighted green the component available with this PR
Highlighted yellow and red the upcoming components

Coming soon™ (in the upcoming PRs):
- [X] `TritonServerRemoteServiceImpl` which will expose the `TritonServerRemoteService` factory component with relative metatype
- [ ] `TritonServerNativeServiceImpl` which will expose the `TritonServerNativeService` factory component with relative metatype
- [ ] `TritonServerServiceOrigImpl` deprecation in favour of the new factory components
- [ ] `TritonServerContainerServiceImpl` which will expose the `TritonServerContainerService` factory component with relative metatype

### DIFF
--- a/kura/org.eclipse.kura.ai.triton.server/OSGI-INF/metatype/org.eclipse.kura.ai.triton.server.TritonServerRemoteService.xml
+++ b/kura/org.eclipse.kura.ai.triton.server/OSGI-INF/metatype/org.eclipse.kura.ai.triton.server.TritonServerRemoteService.xml
@@ -16,7 +16,7 @@
 <MetaData xmlns="http://www.osgi.org/xmlns/metatype/v1.2.0" localization="en_us">
     <OCD id="org.eclipse.kura.ai.triton.server.TritonServerRemoteService"
          name="Nvidia Triton Remote Server"
-         description="Configuration for the Remote-mode Nvidia Triton Server">
+         description="Configuration for the Remote Nvidia Triton Server">
 
         <AD id="server.address"
             name="Nvidia Triton Server address"
@@ -24,7 +24,7 @@
             cardinality="0"
             required="true"
             default="localhost"
-            description="The address of the Nvidia Triton Server.">
+            description="The address of the remote Nvidia Triton Server.">
         </AD>
 
         <AD id="server.ports"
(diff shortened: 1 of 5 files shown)


pierantoniomerlino (approved):

pierantoniomerlino:
@mattdibi Only a couple of observations:

- is it better to move the internal apis to a different package inside the same bundle?
- in the `TritonServerRemoteManager` the `isRunning` variable is managed internally. Is it make sense to try to ask to the Server if it actually running? Does exist such type of call?

mattdibi:
@pierantoniomerlino

> is it better to move the internal apis to a different package inside the same bundle?

Yeah, I don't like having these public interfaces laying around like that. With that change we should be able to change the accessibility as we see fit. Could this be addresses in a future PR though? I would like to merge this and the one regarding the `TritonServerNativeImpl` before doing so, so that we have all the components.

> in the TritonServerRemoteManager the isRunning variable is managed internally. Is it make sense to try to ask to the Server if it actually running? Does exist such type of call?

So yeah, this is a weird one.

Think about the `TritonServerRemoteManager` as a compatibility layer between the inner workings of the `TritonServerServiceAbs` and the Remote Triton mode.

In the `TritonServerServiceAbs` we actually check if the server is running through the GRPC protocol. `isRunning` has a real meaning only for the "managed" instances (like the Native and Containerized ones). I was debating to have it simply return `true` since we have no control on the remote instance, but that would brake the `TritonServerServiceAbs` logic.

TLDR: We *do* ask if the server is running but only in the `TritonServerServiceAbs`. This `isRunning` is only meant to check if we ran the executable which doesn't make sense for the Remote instance, hence this weird thing to be able to use these object in an uniform way.

pierantoniomerlino:
@mattdibi 

> is it better to move the internal apis to a different package inside the same bundle?
Yeah, I don't like having these public interfaces laying around like that. With that change we should be able to change the accessibility as we see fit. Could this be addresses in a future PR though? I would like to merge this and the one regarding the TritonServerNativeImpl before doing so, so that we have all the components.

Of course a new PR is required. I forgot to mention it.

> TLDR: We do ask if the server is running but only in the TritonServerServiceAbs. This isRunning is only meant to check if we ran the executable which doesn't make sense for the Remote instance, hence this weird thing.

Yes... we don't have any control on a remote instance of Triton. But maybe it make sense to check is the remote server is running before trying to communicate with it. Leave it as is for now, but let's think if this fits our needs.

PR #4076, opened 2022-08-01
feat(ai.triton.server): add Native Triton Server Service component

mattdibi:
This is the third step towards the deprecation of the current `TritonServerServiceImpl` class following #4064 

The purpose of this PR is to add the `TritonServerNativeService` Factory Components which exposes the same functionalities as the `TritonServerService` with `local.enabled` turned on (i.e. "Local" mode) but with a smaller configuration interface.

The final goal is to have 2 different UI pages for configuring the 2 new factory components.

See UML for reference:
image
Highlighted green the component available with this PR
Highlighted red the upcoming components

Coming soon™ (in the upcoming PRs):
- [X] `TritonServerRemoteServiceImpl` which will expose the `TritonServerRemoteService` factory component with relative metatype
- [X] `TritonServerNativeServiceImpl` which will expose the `TritonServerNativeService` factory component with relative metatype
- [X] `TritonServerServiceOrigImpl` deprecation in favour of the new factory components
- [ ] `TritonServerContainerServiceImpl` which will expose the `TritonServerContainerService` factory component with relative metatype

### DIFF
--- a/kura/org.eclipse.kura.ai.triton.server/src/main/java/org/eclipse/kura/ai/triton/server/TritonServerServiceNativeImpl.java
+++ b/kura/org.eclipse.kura.ai.triton.server/src/main/java/org/eclipse/kura/ai/triton/server/TritonServerServiceNativeImpl.java
@@ -1,0 +1,27 @@
+package org.eclipse.kura.ai.triton.server;
+
+import org.eclipse.kura.executor.CommandExecutorService;
+
+public class TritonServerServiceNativeImpl extends TritonServerServiceAbs {
+
+    @Override
+    TritonServerInstanceManager createInstanceManager(TritonServerServiceOptions options,
+            CommandExecutorService executorService, String decryptionFolderPath) {
+        return new TritonServerNativeManager(options, executorService, decryptionFolderPath);
+    }
+
+    @Override
+    boolean isConfigurationValid() {
+        return !isNullOrEmpty(this.options.getBackendsPath()) && !isNullOrEmpty(this.options.getModelRepositoryPath());
+    }
+
+    @Override
+    boolean isModelEncryptionEnabled() {
+        return this.options.isModelEncryptionPasswordSet();
+    }
+
+    @Override
+    String getServerAddress() {
+        return "localhost";
+    }
+}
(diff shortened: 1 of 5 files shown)


kura/org.eclipse.kura.ai.triton.server/src/main/java/org/eclipse/kura/ai/triton/server/TritonServerServiceNativeImpl.java
```diff
@@ -0,0 +1,39 @@
...
+    @Override
+    TritonServerInstanceManager createInstanceManager(TritonServerServiceOptions options,
+            CommandExecutorService executorService, String decryptionFolderPath) {
+        return new TritonServerNativeManager(options, executorService, decryptionFolderPath);
+    }
+
+    @Override
+    boolean isConfigurationValid() {
+        return !isNullOrEmpty(this.options.getBackendsPath()) && !isNullOrEmpty(this.options.getModelRepositoryPath());
+    }
```
pierantoniomerlino, line 28: Does it make sense to check also the ports and grpc call size (both required parameters)?

mattdibi: Actually `TritonServerServiceOptions` is already performing the required checks on the ports [here](https://github.com/eclipse/kura/blob/66b343825e9a9f7c92a6353016e684cf2e669f38/kura/org.eclipse.kura.ai.triton.server/src/main/java/org/eclipse/kura/ai/triton/server/TritonServerServiceOptions.java#L58-L76) and the GRPC max message dims [here](https://github.com/eclipse/kura/blob/66b343825e9a9f7c92a6353016e684cf2e669f38/kura/org.eclipse.kura.ai.triton.server/src/main/java/org/eclipse/kura/ai/triton/server/TritonServerServiceOptions.java#L93-L98). Shoudn't this already cover what we need?

pierantoniomerlino: Yes, indeed it is covered by that checks.

PR #4082, opened 2022-08-02
refactor(ai.triton.server): add Triton internal package

mattdibi:
Following what was discussed [here](https://github.com/eclipse/kura/pull/4070) about the [creation of a different package inside the same bundle](https://github.com/eclipse/kura/pull/4070#issuecomment-1200860848) this PR creates an `internal` package for the internal APIs we don't want to expose.

Changes:
- Moved `TritonServerInstanceManager` internal interface in the `internal` package
- Moved the implementers of `TritonServerInstanceManager` internal interface in the `internal` package (i.e. `TritonServerNativeManager` and `TritonServerRemoteManager`)
- Moved `TritonServerEncryptionUtils` utilities in the `internal` package
- Moved `TritonServerServiceAbs` abstract class in the `internal` package

With these changes the only components that will be exported will be:
- `TritonServerService*Impl` which represent the factory components
- `DataType`
- `TritonServerServiceOptions`


pierantoniomerlino:
After some discussion with @mattdibi we decided to close this PR and leave the code as is.

PR #4089, opened 2022-08-03
fix(container.orchestrator): blank gpu option, ports fix, and added new RestartOnFailure into comparator and Hashcode

GregoryIvo:
Pr includes a bunch of different fixes as listed in the title.

### DIFF
--- a/kura/org.eclipse.kura.container.orchestration.provider/src/main/java/org/eclipse/kura/container/orchestration/provider/impl/ContainerOrchestrationServiceImpl.java
+++ b/kura/org.eclipse.kura.container.orchestration.provider/src/main/java/org/eclipse/kura/container/orchestration/provider/impl/ContainerOrchestrationServiceImpl.java
@@ -586,6 +586,11 @@ public class ContainerOrchestrationServiceImpl implements ConfigurableComponent,
             HostConfig configuration) {
 
         Optional<String> gpus = containerDescription.getGpus();
+        
+        if (gpus.isPresent() && gpus.get().equals("")) {
+            return configuration;
+        }
+
         gpus.ifPresent(gpu -> configuration.withDeviceRequests(ImmutableList
                 .of(new DeviceRequest().withDriver("nvidia").withCount(gpu.equals("all") ? -1 : Integer.parseInt(gpu))
                         .withCapabilities(ImmutableList.of(ImmutableList.of("gpu"))))));
@@ -612,7 +617,7 @@ public class ContainerOrchestrationServiceImpl implements ConfigurableComponent,
                         Binding.bindPort(containerDescription.getContainerPortsExternal().get(index)));
             }
 
-            commandBuilder.withPortBindings(portbindings);
+            commandBuilder = commandBuilder.withPortBindings(portbindings);
 
         } else {
             logger.error("portsExternal and portsInternal must be int[] of the same size or they do not exist: {}",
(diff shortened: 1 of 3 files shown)


kura/org.eclipse.kura.container.orchestration.provider/src/main/java/org/eclipse/kura/container/orchestration/provider/impl/ContainerOrchestrationServiceImpl.java
```diff
@@ -612,7 +617,7 @@ private HostConfig containerPortManagementHandler(ContainerConfiguration contain
                         Binding.bindPort(containerDescription.getContainerPortsExternal().get(index)));
             }
 
-            commandBuilder.withPortBindings(portbindings);
+            commandBuilder = commandBuilder.withPortBindings(portbindings);
```
mattdibi, line 620: So yeah... I'm sorry but this doesn't change anything.

I still can't see the ports through the CLI

```bash
jetson@jetson-nano:~$ sudo docker ps
CONTAINER ID   IMAGE                 COMMAND                  CREATED         STATUS         PORTS     NAMES
b921d6e2fa1a   tritonserver:latest   "tritonserver --mode…"   8 seconds ago   Up 7 seconds             TRIT
```

Furthermore [if you look at the code of the library](https://github.com/docker-java/docker-java/blob/61f4e559583970fb088a365c7ea16c509e637f97/docker-java-api/src/main/java/com/github/dockerjava/api/model/HostConfig.java#L896-L899) you'll see that the method changes the instance and then returns it so your change shouldn't make any difference.

EDIT: Superseded by #4092

kura/org.eclipse.kura.container.orchestration.provider/src/main/java/org/eclipse/kura/container/orchestration/provider/impl/ContainerOrchestrationServiceImpl.java
```diff
@@ -586,6 +586,11 @@ private HostConfig containerGpusConfigurationHandler(ContainerConfiguration cont
             HostConfig configuration) {
 
         Optional<String> gpus = containerDescription.getGpus();
+        
+        if (gpus.isPresent() && gpus.get().equals("")) {
+            return configuration;
+        }
```
mattdibi, line 592: As I said yesterday: @pierantoniomerlino was already working on a fix and opened a PR [here](https://github.com/eclipse/kura/pull/4090). I would prefer his solution since it goes more at the root of the problem.

mattdibi (changes requested):
In addition to my other comments I would suggest you to split these kind of PRs in multiple PRs in the future.

You're fixing 3 different issues in a single PR. It can get quite hard to review if there's multiple changes in a single PR and by splitting you reduce the number of affected lines, making it easier for reviewers. Furthermore, keeping everything in a single PR, goes against the Conventional Commit principles we recently adopted.

So... for this time around I would propose the following:
- Remove the changes that are not related to the RestartOnFailure stuff
- Rename the PR to only address the RestartOnFailure issues
- For the empty GPU string fix we'll use the PR from @pierantoniomerlino

kura/org.eclipse.kura.container.orchestration.provider/src/main/java/org/eclipse/kura/container/orchestration/provider/impl/ContainerOrchestrationServiceImpl.java
GregoryIvo: I wanted to include his fix here so I could test things before that PR get merged.

Will close this PR now.

PR #4105, opened 2022-08-09
fix: Updated misleading Password Rest Provider message

MMaiero:

### DIFF
--- a/kura/org.eclipse.kura.rest.provider/src/main/java/org/eclipse/kura/internal/rest/provider/PasswordAuthenticationProvider.java
+++ b/kura/org.eclipse.kura.rest.provider/src/main/java/org/eclipse/kura/internal/rest/provider/PasswordAuthenticationProvider.java
@@ -21,7 +21,6 @@ import java.util.Base64;
 import java.util.Base64.Decoder;
 import java.util.Optional;
 import java.util.StringTokenizer;
-import java.util.regex.Pattern;
 
 import javax.annotation.Priority;
 import javax.servlet.http.HttpServletRequest;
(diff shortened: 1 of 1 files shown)


kura/org.eclipse.kura.rest.provider/src/main/java/org/eclipse/kura/internal/rest/provider/PasswordAuthenticationProvider.java
```diff
@@ -45,14 +44,13 @@
 @Priority(200)
 public class PasswordAuthenticationProvider implements AuthenticationProvider {
 
-    private static final String PASSWORD_AUTH_FAILED_MSG = "{} Rest - Failure - Password Authentication failed";
+    private static final String PASSWORD_AUTH_FAILED_MSG = "{} Rest - Failure - Authentication failed as username or password not matching";
```
pierantoniomerlino, line 48: Or simply "Authentication failed"?

MMaiero: There is already a message for authentication failed. I believe this message here is used in audit log to report that the user has failed a login using the password provider.
I think this can be helpful, while inspecting the logs, to understand when and where the user or attacker tried to access.

MMaiero: We have a similar message in the Certificate auth.

pierantoniomerlino (approved):

Issue #4133, opened 2022-08-29
Error parsing Kura log timestamp


mattdibi:
From time to time Im able to see following error from kura.log file:

```
2022-08-25T12:13:06,068 [Thread-7] ERROR o.e.k.l.f.p.KuraLogLineParser - Error parsing Kura log timestamp.
java.text.ParseException: Unparseable date: "[qtp13694860-266585]"
        at java.text.DateFormat.parse(DateFormat.java:366)
        at org.eclipse.kura.log.filesystem.provider.KuraLogLineParser.parseStringToEpoch(KuraLogLineParser.java:117)
        at org.eclipse.kura.log.filesystem.provider.KuraLogLineParser.parseKuraAuditLog(KuraLogLineParser.java:134)
        at org.eclipse.kura.log.filesystem.provider.KuraLogLineParser.createLogEntry(KuraLogLineParser.java:68)
        at org.eclipse.kura.log.filesystem.provider.FilesystemLogProvider$FileLogReader.notifyListeners(FilesystemLogProvider.java:148)
        at org.eclipse.kura.log.filesystem.provider.FilesystemLogProvider$FileLogReader.readLinesAndNotifyListeners(FilesystemLogProvider.java:132)
        at org.eclipse.kura.log.filesystem.provider.FilesystemLogProvider$FileLogReader.run(FilesystemLogProvider.java:98)
```

full kura.log file attached also

[kura.log](https://github.com/eclipse/kura/files/9446038/kura.log)

PR #4144, opened 2022-09-07
fix(log.filesystem.provider): prevent EOF ending incomplete lines and refactor of parser

marcellorinaldo:

The log files that we are reading can be written character by character. This could mislead the reader thread if the line has not been completely written, since once EOF is detected the [RandomAccessFile.readLine()](https://docs.oracle.com/javase/7/docs/api/java/io/RandomAccessFile.html#readLine()) returns the line read up to that point. On the next iteration, the read resumes from that point resulting in an unexpected start of line, thus causing a parse exception. As an example, consider the following case:

*First read, line not completely written by log4j but `readLine` called:*
```
2022-09-07T11:47:23,901 [qtp24668432-131] INFO  o.e.k.w.s.s.WiresBlinkServlet - unregiEOF
```
*Second read, resumes from the last read character and causes exception:*
```
stered
```

Sometimes parsing exceptions arise from the `log.filesystem.provider` component (https://github.com/eclipse/kura/issues/4133). This is due to the `KuraLogLineParser` which is implemented as a singleton but is concurrently accessed by all the instances of `FilesystemLogProvider`s.

**Related Issue:** [Error parsing Kura log timestamp](https://github.com/eclipse/kura/issues/4133).

**Description of the solution adopted:** Added a utility method to read lines until a new line is encountered. It stores a pointer to the last character read that is not an `EOF` so that we an `EOF` is encountered it can resume the reading and finish the line.

`KuraLogLineParser` has been refactored to not be a singleton.

**Screenshots:** N/A.

### DIFF
--- a/kura/org.eclipse.kura.log.filesystem.provider/src/main/java/org/eclipse/kura/log/filesystem/provider/FilesystemLogProvider.java
+++ b/kura/org.eclipse.kura.log.filesystem.provider/src/main/java/org/eclipse/kura/log/filesystem/provider/FilesystemLogProvider.java
@@ -12,6 +12,7 @@
  *******************************************************************************/
 package org.eclipse.kura.log.filesystem.provider;
 
+import java.io.EOFException;
 import java.io.File;
 import java.io.FileNotFoundException;
 import java.io.IOException;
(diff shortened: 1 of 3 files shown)


salvatore-coppola (approved):

marcellorinaldo:
Issue don't fixed yet, observed a `ParseException` while changing firewall settings.

salvatore-coppola:
> Issue don't fixed yet, observed a `ParseException` while changing firewall settings.

try to add a debub log when the line is read so we can se the real read line

salvatore-coppola (approved):
I've some doubt on the UTF-8 characters over the first 256 but let's keept as it is for the moment.

mstankovic:
Im not able to see errors anymore. Issue seems to be fixed.

PR #4172, opened 2022-09-19
fix(linux.net): Fix dns server configuration

pierantoniomerlino:

This PR fixes the dns server configuration files

**Description of the solution adopted:** The `dnssec` parameters were missing in the `named.conf` file. This caused an issue when the `named` service was started. This PR updates the code in order to correctly generate the configuration file. 
Since the `named.conf` file that is copied in the system by Kura during installation is wrong (the forwarders are not consistent), it is better to remove that copy. Kura, indeed, will create the correct file when all the informations will be available.


marcellorinaldo (changes requested):
Please update copyright header in `LinuxDnsServer`

marcellorinaldo (approved):

pierantoniomerlino:
Do not merge this PR! I found an issue that must be fixed!

PR #4177, opened 2022-09-21
fix: Improved handling of floating point values in KuraPayload JSON encoding

nicolatimeus:
…load JSON serialization

* Infinite and NaN metrics and position fields are now ignored instead of causing the entire message to be dropped


salvatore-coppola (approved):

MMaiero:
It seems that all 4 builds by CI terminated because the 2hrs limit has been reached. Can anyone have a look here to see what is going on? Thanks

PR #4187, opened 2022-09-26
feat: support diffrent internet protocols when configuring ports

GregoryIvo:


GregoryIvo:
Trying to do this in a way that doesn't mess up the old API too much. While not being messy code.

1) Create an object to define ports (contains a public implementation for a port string parser as-well)
2) if ports are set via the old way, when the containerDescriptor object is build, it will convert it to the new type.
3) get internal/external ports method should be deprecated, but are still functional - may need to be discussed

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/container/orchestration/ContainerInstanceDescriptor.java
```diff
@@ -99,9 +99,10 @@ public String getContainerId() {
      * Returns the list of external ports that will be mapped to the given container
      *
      * @return
+     * @since 2.5
```
nicolatimeus, line 103: Is the since annotation correct?

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/container/orchestration/ContainerInstanceDescriptor.java
```diff
@@ -33,8 +34,7 @@ public class ContainerInstanceDescriptor {
     private String containerImage;
     private String containerImageTag;
     private String containerID;
-    private List<Integer> containerPortsExternal;
-    private List<Integer> containerPortsInternal;
+    private List<ContainerPort> containerPorts = new ArrayList<>();
```
nicolatimeus, line 38: Should we add a getter for this field that returns a `List<ContainerPort>`? Otherwise the protocol will not be visible to clients.

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/container/orchestration/ContainerPort.java
```diff
@@ -0,0 +1,100 @@
...
+ *******************************************************************************/
+package org.eclipse.kura.container.orchestration;
+
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Objects;
+
+/**
+ * 
+ * @author eurotech
```
nicolatimeus: Please add a description here

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/container/orchestration/ContainerPort.java
```diff
@@ -0,0 +1,100 @@
...
+
+    public int getExternalPort() {
+        return externalPort;
+    }
+
+    public PortInternetProtocol getInternetProtocol() {
+        return internetProtocol;
+    }
+
+    public static List<Integer> continerPortsListInternal(List<ContainerPort> ports) {
```
nicolatimeus: Is it really necessary to have this method in the APIs?

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/container/orchestration/ContainerPort.java
```diff
@@ -0,0 +1,100 @@
...
+        List<Integer> portList = new LinkedList<>();
+        for (ContainerPort port : ports) {
+            portList.add(port.internalPort);
+        }
+        return portList;
+    }
+
+    public static List<Integer> continerPortsListExternal(List<ContainerPort> ports) {
```
nicolatimeus: Is it really necessary to have this method in the APIs?

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/container/orchestration/PortInternetProtocol.java
```diff
@@ -0,0 +1,26 @@
...
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *  Eurotech
+ *******************************************************************************/
+package org.eclipse.kura.container.orchestration;
+
+/**
+ * 
+ * @author eurotech
```
nicolatimeus: Please add a description here

kura/org.eclipse.kura.container.provider/OSGI-INF/metatype/org.eclipse.kura.container.provider.ContainerInstance.xml
```diff
@@ -73,12 +73,12 @@
             min="1"
             default="500" />
 
-        <AD id="container.ports.internal" name="Internal ports"
-            description="A comma separated list of ports. Note, the number of internal ports must be equal to the number of external ports. Example: 80, 443." 
+        <AD id="container.ports.internal" name="Internal Ports"
+            description="A comma-separated list of ports. Note, the number of internal ports must be equal to the number of external ports. A port internet protocol can also be specified with a colon and text after the port number. Example: 80, 443:udp, 8080:tcp."
```
nicolatimeus: Maybe we can mention that if no protocol is specified, the default will be TCP.

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/container/orchestration/ContainerConfiguration.java
```diff
@@ -79,13 +79,22 @@ public String getContainerName() {
...
+    public List<ContainerPort> getContainerPorts() {
+        return this.containerPorts;
+    }
+
     /**
      * Returns the list of external ports that will be mapped to the given container
      *
      * @return
      */
     public List<Integer> getContainerPortsExternal() {
```
nicolatimeus, line 97: Maybe we can consider deprecating (without removing) the API methods that allow to get/set the port mappings as a list of integer suggesting to use the new `ContainerPort` APIs, since the former provide incomplete information (you don't know the protocol).

The deprecation should be done in a follow-up PR to adhere with conventional commits.

GregoryIvo: Resolving this for now, but have added to my to-do list.

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/container/orchestration/ContainerConfiguration.java
```diff
@@ -79,13 +79,26 @@ public String getContainerName() {
         return this.containerName;
     }
 
+    /**
+     * Returns a list of {@link ContainerPort} mapped to the container.
+     *
```
nicolatimeus, line 84: Please add since annotation if it is a new method

nicolatimeus:
Would it be possible to move the configuration parsing error handling logic to the updated() method? Something like this:
* Try to crate the option objects
* If an exception occurs, perform a transition to the "Disabled" state and log an error message
In this way we can remove the error handling logic from the options object making it more clean, and we will be able to react to failures while parsing the other parameters as well.

GregoryIvo:
> Would it be possible to move the configuration parsing error handling logic to the updated() method? Something like this:
> 
> * Try to crate the option objects
> * If an exception occurs, perform a transition to the "Disabled" state and log an error message
>   In this way we can remove the error handling logic from the options object making it more clean, and we will be able to react to failures while parsing the other parameters as well.

Made these changes, appears to work out well on my end and pass tests. Fingers crossed this builds.

nicolatimeus (approved):

PR #4189, opened 2022-09-27
fix: Removed internal usage of ConnectionInfo

nicolatimeus:

Kura currently installs the `/etc/network/if-up.d/ifup-local` script that populates the `/tmp/.kura` with some `coninfo` files that contain some information about network interface state (DNS, GATEWAY, IP address).
Kura pareses these files using the `ConnectionInfo` APIs, that is currently used for determining the current gateway for an interface.
These files are regenerated when an interface is brought up using the `ifup` command, for interfaces configured using DHCP by Kura, this happens only once on boot. If a DHCP interface is reconfigured at runtime, the information contained in the `coninfo` files can become outdated.

This PR attempts to remove the internal usage of the `coninfo` files to reduce the risk of inconsistencies. The `coninfo` file generation and related APIs are kept.


pierantoniomerlino (approved):

pierantoniomerlino:
Does make sense to update the if-up/down scripts not to create the conninfo files? And deprecate the ConnInfo APIs?

nicolatimeus:
Yes that would make sense.. Probably it should be done with a separate PR to adhere to conventional commits

nicolatimeus:
Probably we cannot remove the scripts until we completely remove the APIs

pierantoniomerlino:
@nicolatimeus Yes, I agree.

PR #4199, opened 2022-10-13
fix(core): Resolved database memory leak

salvatore-coppola:
Specified payload length during the data saving in order to fix lob memory leak.


nicolatimeus:
Please update copyright headers

nicolatimeus (approved):

PR #4178, opened 2022-09-22
refactor: Removed container orchestration bundles from Docker images.

salvatore-coppola:
> **Note**: We are using the Conventional Commits convention for our pull request titles. Please take a look at the [PR title format document](https://github.com/eclipse/kura/blob/develop/CONTRIBUTING.md#submitting-the-changes) for the supported [types](https://github.com/eclipse/kura/blob/develop/CONTRIBUTING.md#type) and [scopes](https://github.com/eclipse/kura/blob/develop/CONTRIBUTING.md#scope).

Removed container orchestration bundles from Docker images and created a new docker specific profile.

PR #3778, opened 2022-02-07
Docker images clean up

salvatore-coppola:
Docker images have bean fixed. 
Base Alpine images has been updated.
ClokService now is disabled by default.
A new profile. ubuntu-20-nn has been added and it is the base package for docker images.

PR #4208, opened 2022-10-18
fix: failing Ubi8-based Docker image build

mattdibi:
Brief description of the PR: This PR fixes the Ubi8-based Docker build of Kura

**Description of the solution adopted:** The package `xmlstarlet` was breaking the build because it was missing from the remote repo. To solve the issue it was removed from the required packages since it is not necessary anymore.

The `xmlstarlet` dependency was first introduced in https://github.com/eclipse/kura/pull/3778 with the sole purpose of disabling the Clock service in the starting snapshot (details [here](https://github.com/eclipse/kura/commit/5c2091d74f7277f7eeafda277709745d2f6deef0#diff-0012b46cf5e2c35e2387bee8539dc0df52f7291cdbca3596e2a6c1205755cb4fR99)).

With https://github.com/eclipse/kura/pull/4178 the need of `xmlstarlet` was removed because we're relying on the `docker_x86_64` installer which has its own starting snapshot for which the ClockService is already disabled (details [here](https://github.com/eclipse/kura/commit/ef505876f35efe7003c573cc5d923fdcf0ee6bbd#diff-8954ca8007020771ed14b11dcdf54a8f226eb28b8c85883670245dcb3032d983R30)).

Therefore the `xmlstarlet` dependency can be safely removed.

### DIFF
--- a/kura/container/kura_alpine/Dockerfile
+++ b/kura/container/kura_alpine/Dockerfile
@@ -24,7 +24,7 @@ RUN echo "$GIT_REPO / $GIT_BRANCH / $KURA_COMMIT" && \
     chmod a+x -R /usr/local/bin && \
     apk update && \
     apk --no-cache add git openjdk8 maven && \
-    apk --no-cache add bash zip curl xmlstarlet gcompat && \
+    apk --no-cache add bash zip curl gcompat && \
     if [ -d /context/.git ]; then \
       mv /context /kura; \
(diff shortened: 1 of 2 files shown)


mattdibi:
Tested locally and I can correctly login to the Kura UI using the Ubi8-based image.

image

salvatore-coppola (approved):

PR #4209, opened 2022-10-19
fix: failing Ubi8-based Docker image build on maintenance branch

mattdibi:
Brief description of the PR. This PR fixes the Ubi8-based Docker build of Kura on the 5.2 maintenance branch

**Related Issue:** https://github.com/eclipse/kura/pull/4208

**Description of the solution adopted:** The package `xmlstarlet` was breaking the build because it was missing from the upstream repo. To solve the issue I refactored the Dockerfile so that the `xmlstarlet` is installed on the `builder` stage where we have more wiggle room on installation options.

**Additional benefits**: The runtime image is now slightly smaller due to the fact that we're not installing `xmlstarlet` there.

### DIFF
--- a/kura/container/kura_ubi8/Dockerfile
+++ b/kura/container/kura_ubi8/Dockerfile
@@ -37,7 +37,7 @@ RUN if [ -n "$KURA_COMMIT" ]; then git checkout "$KURA_COMMIT"; fi && git log -1
 # Replace broken 'nn' script
 RUN cp kura/distrib/src/main/sh/extract.sh kura/distrib/src/main/sh/extract_nn.sh
 
-# Customize stating snapshot and properties
+# Customize starting snapshot and properties
 RUN sed -i 's/kura.primary.network.interface/#kura.primary.network.interface/g' kura/distrib/src/main/resources/intel-up2-ubuntu-20-nn/kura.properties && \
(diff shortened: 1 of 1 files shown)


mattdibi:
Tested locally and I can correctly login to the Kura UI using the Ubi8-based image.

image

```bash
mattia$ docker exec -it 994edb862a27 /bin/sh
sh-4.4# cat /opt/eclipse/kura/framework/kura.properties
...
kura.bios.version=N/A
kura.firmware.version=N/A
#kura.primary.network.interface=enp2s0
# kura.mac.address= Fetch from Java
kura.home=/opt/eclipse/kura
kura.framework.config=/opt/eclipse/kura/framework
kura.user.config=/opt/eclipse/kura
...

salvatore-coppola (approved):
The change is ok but i was wandering: couldn't we use the profile `docker-x86_64-nn` in place of `intel-up2-ubuntu-20-nn`. It has several benefits, the clock service is disabled by default (no need to use the script lines in `Customize starting snapshot and properties` section) and indipendence from intel-up2 profile.

mattdibi:
@salvatore-coppola I thought about it but preferred to use this solution because it minimises code changes. If you prefer to use the docker profile we need to port your PR and mine on here (2k+ LOC changes), here we just have 5 new lines.

salvatore-coppola:
> @salvatore-coppola I thought about it but preferred to use this solution because it minimises code changes. If you prefer to use the docker profile we need to port your PR and mine on here (2k+ LOC changes), here we just have 5 new lines.

It makes sense.

PR #4210, opened 2022-10-19
build: realigned raspberry pi package names to raspberry-pi-armhf and  raspberry-pi-arm64

GregoryIvo:
Changes package naming schemes
'raspberry-pi' -> 'raspberry-pi-armhf'
'raspberry-pi-ubuntu-20' -> 'raspberry-pi-arm64'

------------------------------------------------------
The goal of the PR is to unify package names.

from Preliminary testing. the new raspberry-pi-arm64 package appears to work on both distros raspberry-pi-os and Ubuntu-20

Note, when installing on raspberry-pi-OS 64bit, you must manually add the dependency for jdk8.
this is done with the following.

```
sudo apt install -y wget apt-transport-https gnupg
sudo wget -O - https://packages.adoptium.net/artifactory/api/gpg/key/public | sudo apt-key add -
sudo echo "deb https://packages.adoptium.net/artifactory/deb $(awk -F= '/^VERSION_CODENAME/{print$2}' /etc/os-release) main" | sudo tee /etc/apt/sources.list.d/adoptium.list
sudo apt update
```

Networking appears to work including wifi and DHCP server.


GregoryIvo:
@mattdibi I think you were the last one to touch the profiles. What do you think about this suggested change?

GregoryIvo:
renamed raspberry-pi to raspberry-pi-armhf

GregoryIvo:
armhf builds works on my end.  Hoping this build passes so I can officially begin QA

mattdibi:
So yeah... I like the idea of not having an additional profile to maintain and the renaming of the RPi profile to something more meaningful. Given the QA passes it's a LGTM for me...

@MMaiero what do you think about the naming scheme here?

MMaiero:
@pierantoniomerlino do you remember if we have specific networking features tied to Ubuntu and would affect the raspberry profile?

pierantoniomerlino:
@MMaiero I don't remember any network specific configuration, but I strongly suggest to run a set of QA tests on both devices...

MMaiero:
Agreed. I believe we will need this QA effort to understand what works and what not.
@GregoryIvo @mstankovic can we have an assessment? Can it be done with automation?

Thanks.

GregoryIvo:
It appears everything is working with the logrotate config on both platforms. (raspberry pi OS 64bit & ubuntu 20 64bit)

MMaiero (approved):

PR #4213, opened 2022-11-02
feat: Added WireAsset option for emitting values only on change

nicolatimeus:

Adds two new properties to WireAsset global configuration:

* `emit.on.change`: If set to true, the WireAsset will emit a value on Kura Wires only if it is different than the one obtained from the previous read operation or event for the corresponding channel. Channel errors will always be emitted if emit.errors is set to true.
* `emit.empty.envelopes`: If set to false, the WireAsset will not emit empty envelopes.

### DIFF
--- a/kura/org.eclipse.kura.wire.component.provider/src/main/java/org/eclipse/kura/internal/wire/asset/WireAssetOCD.java
+++ b/kura/org.eclipse.kura.wire.component.provider/src/main/java/org/eclipse/kura/internal/wire/asset/WireAssetOCD.java
@@ -41,7 +41,8 @@ public class WireAssetOCD extends BaseAssetOCD {
     private static final String EMIT_ERRORS_DESCRIPTION = "Specifies wheter errors should be included or not "
             + "in the emitted envelope";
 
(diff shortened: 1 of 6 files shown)


kura/org.eclipse.kura.wire.component.provider/src/main/java/org/eclipse/kura/internal/wire/asset/WireAssetOCD.java
```diff
@@ -39,6 +41,13 @@ public class WireAssetOCD extends BaseAssetOCD {
     private static final String EMIT_ERRORS_DESCRIPTION = "Specifies wheter errors should be included or not "
             + "in the emitted envelope";
 
+    private static final String EMIT_ON_CHANGE_DESCRIPTION = "If set to true, this component will emit a value on Kura Wires"
+            + " only if it is different than the one obtained from the previous read operation or event for the corresponding channel."
```
MMaiero: Is it better like this? "If set to true, this component will include a channel value in the output emitted in Kura Wires only if it value is different than the one from the previous read operation or event."

MMaiero (approved):

PR #4218, opened 2022-11-07
fix: Prevent discarding of NetworkConfigurationService properties

nicolatimeus:

The current implementation of NetworkConfigurationService will discard properties of a configured network interface if it not connected to the system when its activate() or update() methods are called. If this happens, the network interface will be reported as "Disabled" when it is reconnected. This can cause problems expecially with USB modems.

This PR introduces the following changes:
* `NetworkConfigurationService` will no longer drop received configuration properties.
* `ModemMonitorService` will ask the `NetworkConfigurationService` to recompute the network configuration if an USB modem is connected.

### DIFF
--- a/kura/org.eclipse.kura.net.admin/src/main/java/org/eclipse/kura/net/admin/NetworkConfigurationServiceImpl.java
+++ b/kura/org.eclipse.kura.net.admin/src/main/java/org/eclipse/kura/net/admin/NetworkConfigurationServiceImpl.java
@@ -23,7 +23,6 @@ import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Optional;
 import java.util.Set;
-import java.util.StringTokenizer;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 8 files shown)


kura/org.eclipse.kura.net.admin/src/main/java/org/eclipse/kura/net/admin/NetworkConfigurationServiceImpl.java
```diff
@@ -336,16 +349,18 @@ private void decryptPasswordProperties(Map<String, Object> modifiedProps) throws
...
+        // This method returns the network configuration properties without the current
+        // values.
+        // i.e. the ip address that should be applied to the system, but not the actual
+        // one.
         Optional<NetworkConfiguration> networkConfiguration = getNetworkConfiguration(false);
         if (!networkConfiguration.isPresent()) {
             throw new KuraRuntimeException(KuraErrorCode.CONFIGURATION_ERROR,
                     "The network component configuration cannot be retrieved");
-
         }
```
pierantoniomerlino: Since the component configuration is computed from the properties and not from the networkconfiguration object, does it make sense to check if the networkconfiguration is present?

pierantoniomerlino (approved):

nicolatimeus:
The tests should be ok now, the failures in last build are not related to these changes.

PR #4016, opened 2022-06-10
fix(core.deployment): prevent unnecessary modifications inside the packages folder.

kandrej:
On every restart Kura updates the dpa.properties file even if there are no changes in installed packages.

Prevent changes to the dpa.properties file if properties do not change.
Save .dp files with consistent naming "NAME_VERSION.dp".
Fixed a bug in which installed .dp files would be copied to a new dp file named "NAME_VERSION.dp" on startup if they are not already named this way.

Signed-off-by: Andrej Krota <andrej.krota@gmail.com>

PR #4222, opened 2022-11-09
refactor(core.deployment): add writeConfigurationFile method

mattdibi:
**Brief description of the PR**: This PR refactors the recent changes in the `deployment.agent` package to adhere to Sonar code quality suggestions.

**Related Issue:** This PR fixes/closes https://github.com/eclipse/kura/pull/4016

**Description of the solution adopted:** Extracted duplicated code in the newly added `writeDPAPropertiesFile` static method.

### DIFF
--- a/kura/org.eclipse.kura.deployment.agent/src/main/java/org/eclipse/kura/deployment/agent/impl/DeploymentAgent.java
+++ b/kura/org.eclipse.kura.deployment.agent/src/main/java/org/eclipse/kura/deployment/agent/impl/DeploymentAgent.java
@@ -442,21 +442,8 @@ public class DeploymentAgent implements DeploymentAgentService, ConfigurableComp
         oldDeployedPackages.putAll(deployedPackages);
         deployedPackages.setProperty(packageName, packageUrl);
 
-        if (this.dpaConfPath == null) {
-            logger.warn("Configuration file not specified");
-            return;
-        }
-
-        if (oldDeployedPackages.equals(deployedPackages)) {
-            return;
-        }
-
-        try (FileOutputStream fos = new FileOutputStream(this.dpaConfPath)) {
-            deployedPackages.store(fos, null);
-            fos.flush();
-            fos.getFD().sync();
-        } catch (IOException e) {
-            logger.error("Error writing package configuration file", e);
+        if (!oldDeployedPackages.equals(deployedPackages)) {
+            writeConfigurationFile(this.dpaConfPath, deployedPackages);
         }
     }
 
@@ -466,16 +453,18 @@ public class DeploymentAgent implements DeploymentAgentService, ConfigurableComp
         oldDeployedPackages.putAll(deployedPackages);
         deployedPackages.remove(packageName);
 
-        if (this.dpaConfPath == null) {
+        if (!oldDeployedPackages.equals(deployedPackages)) {
+            writeConfigurationFile(this.dpaConfPath, deployedPackages);
+        }
+    }
+
+    private static void writeConfigurationFile(String dpaConfPath, Properties deployedPackages) {
+        if (dpaConfPath == null) {
             logger.warn("Configuration file not specified");
             return;
         }
 
-        if (oldDeployedPackages.equals(deployedPackages)) {
-            return;
-        }
-
-        try (FileOutputStream fos = new FileOutputStream(this.dpaConfPath)) {
+        try (FileOutputStream fos = new FileOutputStream(dpaConfPath)) {
             deployedPackages.store(fos, null);
             fos.flush();
             fos.getFD().sync();
--- a/kura/org.eclipse.kura.deployment.agent/src/main/java/org/eclipse/kura/deployment/agent/impl/DeploymentAgent.java
+++ b/kura/org.eclipse.kura.deployment.agent/src/main/java/org/eclipse/kura/deployment/agent/impl/DeploymentAgent.java
@@ -443,7 +443,7 @@ public class DeploymentAgent implements DeploymentAgentService, ConfigurableComp
         deployedPackages.setProperty(packageName, packageUrl);
 
         if (!oldDeployedPackages.equals(deployedPackages)) {
-            writeConfigurationFile(this.dpaConfPath, deployedPackages);
+            writeDPAPropertiesFile(this.dpaConfPath, deployedPackages);
         }
     }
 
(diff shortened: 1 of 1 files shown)


kura/org.eclipse.kura.deployment.agent/src/main/java/org/eclipse/kura/deployment/agent/impl/DeploymentAgent.java
```diff
@@ -466,16 +453,18 @@ private void removePackageFromConfFile(String packageName) {
...
-            logger.warn("Configuration file not specified");
-            return;
+        if (!oldDeployedPackages.equals(deployedPackages)) {
+            writeDPAPropertiesFile(this.dpaConfPath, deployedPackages);
         }
+    }
 
-        if (oldDeployedPackages.equals(deployedPackages)) {
+    private static void writeDPAPropertiesFile(String dpaConfPath, Properties deployedPackages) {
+        if (dpaConfPath == null) {
```
marcellorinaldo: Can `this.dpaConfPath` be null?

kura/org.eclipse.kura.deployment.agent/src/main/java/org/eclipse/kura/deployment/agent/impl/DeploymentAgent.java
```diff
@@ -466,16 +453,18 @@ private void removePackageFromConfFile(String packageName) {
...
-        if (this.dpaConfPath == null) {
-            logger.warn("Configuration file not specified");
-            return;
+        if (!oldDeployedPackages.equals(deployedPackages)) {
+            writeDPAPropertiesFile(this.dpaConfPath, deployedPackages);
         }
+    }
 
-        if (oldDeployedPackages.equals(deployedPackages)) {
+    private static void writeDPAPropertiesFile(String dpaConfPath, Properties deployedPackages) {
```
marcellorinaldo: Why must the method be static?

mattdibi: It doesn't. I decided to set it static since it doesn't depend on any field of the `DeploymentAgent` class. Now that you questioned it I realised that it doesn't make much sense since it is a private method... fixing.

kura/org.eclipse.kura.deployment.agent/src/main/java/org/eclipse/kura/deployment/agent/impl/DeploymentAgent.java
mattdibi: Don't know. I just refactored the existing code extracting the duplicated lines. I can dig into it but I prefer leaving it as-is...

kura/org.eclipse.kura.deployment.agent/src/main/java/org/eclipse/kura/deployment/agent/impl/DeploymentAgent.java
marcellorinaldo: Wait, I wanted to stimulate a discussion on this refactoring. As it is it makes sense to be static:

- is not dependent on instance creation
- is an "utility" method
- no operation involves instance variables

I wanted to discuss the last point: why doesen't it use the instance variable? Is `this.dpaConfPath` variable even needed in the rest of the class? Does it even make sense to have it as an instance variable?

kura/org.eclipse.kura.deployment.agent/src/main/java/org/eclipse/kura/deployment/agent/impl/DeploymentAgent.java
marcellorinaldo: I would dig deeper, we unlikely will have the chance to refactor this again...

kura/org.eclipse.kura.deployment.agent/src/main/java/org/eclipse/kura/deployment/agent/impl/DeploymentAgent.java
mattdibi: > why doesen't it use the instance variable? Is this.dpaConfPath variable even needed in the rest of the class? Does it make sense to have it as an instance variable?

The variable is needed as it represent the path of the `dpa.properties` file which is used extensively by the `DeploymentAgent`.

As for the `writeDPAPropertiesFile` method: I decided to pass it as a parameter because I wanted to be explicit about what the method depends on. There's no technical reason apart from (subjective) readability.

marcellorinaldo: It is used extensively but unecessarly, in my opinion. The only place besides the methods touched so far by this PR is in `readDeployedPackages`. The read method is unecessarly invoked many times, can't we read the packages at activation and maintain the current properties as an instance variable and update just that? It would be much more efficient and readable.

kura/org.eclipse.kura.deployment.agent/src/main/java/org/eclipse/kura/deployment/agent/impl/DeploymentAgent.java
mattdibi: Soooooo... a few lines above we have [this interesting check](https://github.com/mattdibi/kura/blob/f5165dfd6f0a8884706879f5a939ec60067db1e0/kura/org.eclipse.kura.deployment.agent/src/main/java/org/eclipse/kura/deployment/agent/impl/DeploymentAgent.java#L124-L127)...

I might need a little help here but I'll try to reconstruct what can happen: when the `activate` method is run and there's no dpa properties file path specified an exception is thrown. This means that the class will not publish its services (as it failed to initialise) and thus the code for the `writeDPAPropertiesFile` will not be ran with a null `dpaConfPath`... right?

marcellorinaldo: Correct, also at this point the check becomes useless: https://github.com/eclipse/kura/blob/0c712b5746302df50bfe9d29a250d90877525653/kura/org.eclipse.kura.deployment.agent/src/main/java/org/eclipse/kura/deployment/agent/impl/DeploymentAgent.java#L351

kura/org.eclipse.kura.deployment.agent/src/main/java/org/eclipse/kura/deployment/agent/impl/DeploymentAgent.java
mattdibi: > The read method is unecessarly invoked many times, can't we read the packages at activation and maintain the current properties as an instance variable and update just that? It would be much more efficient and readable.

Arguably the efficiency here is not that important: how many times in the lifetime of a Kura device we see a deployment package installation? I'll take the read from disk over increasing memory usage for storing and maintaining the package list.

I'll leave it as-is.

marcellorinaldo: Agree with that. At this point I would remove the parameter from your write method and use the instance variable.

marcellorinaldo (approved):

Issue #4234, opened 2022-11-11
H2 Database size continues to grow with high data entry rate


salvatore-coppola:
**Describe the bug**
With high data insertion rate (about 20 ms) the size of h2 database in file mode continues to grow and the compact/defrag task seems to do nothing.

**To Reproduce**
1. Create a wire graph with an high data entry rate.
2. Disconnect the cloud publisher.
3. Observe the file size growing.
4. Reconnect the cloud publisher and disconnect the timer.
5. Observe that all data are cleaned from the database.
6. Observe the file size doesn't decrease.

**Expected behavior**
The size remains consistent with the real data.

PR #4236, opened 2022-11-11
fix(core): Fixed abnormal size of file-based database.

salvatore-coppola:

**Description:** Now SHUTDOWN is used to defragment the database instead of SHUTDOWN DEFRAG, since setting the value of MAX_COMPACT_TIME allows the system to perform compaction for a maximum time specified by the value. 
The previous solution did not work because of a bug in H2.

**Related Issue:** This PR fixes/closes {#4234}


MMaiero:
@salvatore-coppola Please provide a PR with updated documentation

MMaiero (approved):

PR #4242, opened 2022-11-15
fix: Filter out modified.interface.names property

nicolatimeus:

* Prevents the `modified.interface.names` property from being stored to / loaded from snapshot

### DIFF
--- a/kura/org.eclipse.kura.net.admin/src/main/java/org/eclipse/kura/net/admin/NetworkConfigurationServiceImpl.java
+++ b/kura/org.eclipse.kura.net.admin/src/main/java/org/eclipse/kura/net/admin/NetworkConfigurationServiceImpl.java
@@ -92,6 +92,7 @@ public class NetworkConfigurationServiceImpl implements NetworkConfigurationServ
     private static final String CONFIG_AUTOCONNECT = ".config.autoconnect";
     private static final String CONFIG_MTU = ".config.mtu";
     private static final String NET_INTERFACES = "net.interfaces";
+    private static final String MODIFIED_INTERFACE_NAMES = "modified.interface.names";
     private static final String MODEM_PORT_REGEX = "^\\d+-\\d+";
     private static final Pattern PPP_INTERFACE = Pattern.compile("ppp[0-9]+");
     private static final Pattern COMMA = Pattern.compile(",");
@@ -196,7 +197,7 @@ public class NetworkConfigurationServiceImpl implements NetworkConfigurationServ
             logger.debug("Got null properties...");
         } else {
             logger.debug("Props...{}", properties);
-            this.properties = new HashMap<>(properties);
+            this.properties = discardModifiedNetworkInterfaces(new HashMap<>(properties));
             updated(this.properties);
         }
     }
@@ -279,6 +280,8 @@ public class NetworkConfigurationServiceImpl implements NetworkConfigurationServ
             updateCurrentNetworkConfiguration();
 
             this.eventAdmin.postEvent(new NetworkConfigurationChangeEvent(modifiedProps));
+
+            this.properties = discardModifiedNetworkInterfaces(this.properties);
 
             if (changed) {
                 this.configurationService.snapshot();
@@ -747,6 +750,16 @@ public class NetworkConfigurationServiceImpl implements NetworkConfigurationServ
             getWifiInfraDefinition(objectFactory, tocd, ifaceName);
             getWifiMasterDefinition(objectFactory, tocd, ifaceName);
         }
+    }
+
+    private static Map<String, Object> discardModifiedNetworkInterfaces(final Map<String, Object> properties) {
+        if (!properties.containsKey(MODIFIED_INTERFACE_NAMES)) {
+            return properties;
+        }
+
+        final Map<String, Object> result = new HashMap<>(properties);
+        result.remove(MODIFIED_INTERFACE_NAMES);
+        return result;
     }
 
     private void getWifiMasterDefinition(ObjectFactory objectFactory, Tocd tocd, String ifaceName) {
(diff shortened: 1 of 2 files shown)


MMaiero:
@mattdibi Can you please review?

kura/org.eclipse.kura.net.admin/src/main/java/org/eclipse/kura/net/admin/NetworkConfigurationServiceImpl.java
```diff
@@ -749,6 +752,16 @@ private void getWifiDefinition(NetInterfaceType type, ObjectFactory objectFactor
...
     }
 
+    private static Map<String, Object> discardModifiedNetworkInterfaces(final Map<String, Object> properties) {
+        if (!properties.containsKey(MODIFIED_INTERFACE_NAMES)) {
+            return properties;
+        }
+
+        final Map<String, Object> result = new HashMap<>(properties);
+        result.remove(MODIFIED_INTERFACE_NAMES);
+        return result;
```
mattdibi, line 762: I think you should be able to just use the `remove` without checking if the Map contains the key since, [from the documentation](https://docs.oracle.com/javase/8/docs/api/java/util/HashMap.html#remove-java.lang.Object-):

> Removes the mapping for the specified key from this map **if present**.

So it should simplify into:

```suggestion
        final Map<String, Object> result = new HashMap<>(properties);
        result.remove(MODIFIED_INTERFACE_NAMES);
        return result;
```

or whatever you like...

It can be done directly on line 200 (see other comment)

kura/org.eclipse.kura.net.admin/src/main/java/org/eclipse/kura/net/admin/NetworkConfigurationServiceImpl.java
```diff
@@ -196,7 +197,7 @@ public void activate(ComponentContext componentContext, Map<String, Object> prop
             logger.debug("Got null properties...");
         } else {
             logger.debug("Props...{}", properties);
-            this.properties = new HashMap<>(properties);
+            this.properties = discardModifiedNetworkInterfaces(new HashMap<>(properties));
```
mattdibi, line 200: Following the other comment:

```suggestion
            this.properties.remove(MODIFIED_INTERFACE_NAMES);
```

kura/org.eclipse.kura.net.admin/src/main/java/org/eclipse/kura/net/admin/NetworkConfigurationServiceImpl.java
```diff
@@ -92,6 +92,7 @@ public class NetworkConfigurationServiceImpl implements NetworkConfigurationServ
     private static final String CONFIG_AUTOCONNECT = ".config.autoconnect";
     private static final String CONFIG_MTU = ".config.mtu";
     private static final String NET_INTERFACES = "net.interfaces";
+    private static final String MODIFIED_INTERFACE_NAMES = "modified.interface.names";
```
mattdibi, line 95: You should be able to easily fix the [issue Sonar reported](https://sonarcloud.io/project/issues?resolved=false&types=CODE_SMELL&branch=PR-4242&id=org.eclipse.kura%3Akura&open=AYR8XSFmk5A3ZVwhEoCN) by using this constant.

mattdibi (changes requested):
Tested on Jetson Nano: the network configuration file `/etc/network/interfaces` is correctly updated even though the `modified.interface.names` property is set on `snapshot_0.xml`. So it is working as expected 👍 Functionality-wise LGTM

I added a few comments for the code, only some minor details.

kura/org.eclipse.kura.net.admin/src/main/java/org/eclipse/kura/net/admin/NetworkConfigurationServiceImpl.java
nicolatimeus: I created the method in this way so that it will be able to handle immutable maps as well, not all Java Map implementation supports the remove operation: https://docs.oracle.com/javase/8/docs/api/java/util/Map.html#remove-java.lang.Object. If they don't, calling remove will throw an [UnsupportedOperationException](https://docs.oracle.com/javase/8/docs/api/java/lang/UnsupportedOperationException.html).

I'm pretty sure that the map that we receive on the activate() and update() methods is immutable.

Probably in the code as it is now it will be always be applied to mutable maps, but I've kept it in this way just to stay on the safe side in case of a refactor.. moreover in the current implementation it is unlikely that the properties will contain the modified interface names property, so we will avoid the map cloning in most cases.

In any case if you want I can double check that it is always applied to mutable map and remove the map cloning

mattdibi: Oh... I didn't know that. TIL

I think it's ok as-is then 👍

kura/org.eclipse.kura.net.admin/src/main/java/org/eclipse/kura/net/admin/NetworkConfigurationServiceImpl.java
mattdibi: As explained [here](https://github.com/eclipse/kura/pull/4242#discussion_r1023658145) this is not a good idea. Consider this as "Resolved".

MMaiero:
Do we need to keep the logging message also reported (for other reasons) by sonar? https://sonarcloud.io/project/issues?resolved=false&severities=CRITICAL&sinceLeakPeriod=true&branch=PR-4242&id=org.eclipse.kura%3Akura&open=AYR8XSFmk5A3ZVwhEoCN

mattdibi (approved):

MMaiero:
Forcing merge to speedup the service release process. The last change has no effect on the build but only on sonar.

PR #4247, opened 2022-11-16
fix(core.data.store): allow storing null payloads

### LINKED ISSUE #4246: DataService not accepting null payload
**Describe the bug**
This is a regression introduced in Kura 5.2.0.
If a message is pushed to the data service with a null payload, this causes the database backend to fail on the following line: [kura/DbDataStore.java at adcf5b813a035c3f7a66c99abda9416d9136d7d7 · eclipse/kura](https://github.com/eclipse/kura/blob/adcf5b813a035c3f7a66c99abda9416d9136d7d7/kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/data/store/DbDataStore.java#L345)
The Eclipse Kura DataService APIs do not report anything in terms of the format of the payload. The only request is to have a byte[] and null is compatible with that.

**To Reproduce**
1. Push a message in the data service with a null payload

**Expected behavior**
The message should be properly processed and stored

**Screenshots**
N/A

 - Any board and system with Kura 5.2.0

**Additional context**
N/A

marcellorinaldo:

This PR allows the `DbDataStore` to store null payloads.

**Related Issue:** This PR fixes/closes https://github.com/eclipse/kura/issues/4246.

**Screenshots:** N/A.

### DIFF
--- a/kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/data/store/DbDataStore.java
+++ b/kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/data/store/DbDataStore.java
@@ -342,12 +342,16 @@ public class DbDataStore implements DataStore {
                 pstmt.setInt(6, -1);                                                // publishedMessageId
                 pstmt.setTimestamp(7, null);                                        // confirmedOn
 
-                if (payload.length < PAYLOAD_BYTE_SIZE_THRESHOLD) {
-                    pstmt.setBytes(8, payload);                                     // smallPayload
+                // smallPayload (=8) vs. largePayload (=9)
+                if (payload == null) {
+                    pstmt.setNull(8, Types.VARBINARY);
+                    pstmt.setNull(9, Types.BLOB);
+                } else if (payload.length < PAYLOAD_BYTE_SIZE_THRESHOLD) {
+                    pstmt.setBytes(8, payload);
                     pstmt.setNull(9, Types.BLOB);
                 } else {
                     pstmt.setNull(8, Types.VARBINARY);
-                    pstmt.setBinaryStream(9, new ByteArrayInputStream(payload), payload.length);    // largePayload
+                    pstmt.setBinaryStream(9, new ByteArrayInputStream(payload), payload.length);
                 }
 
                 pstmt.setInt(10, priority);                                         // priority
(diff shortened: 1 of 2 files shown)


kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/data/store/DbDataStore.java
```diff
@@ -342,12 +342,16 @@ private synchronized DataMessage storeInternal(String topic, byte[] payload, int
                 pstmt.setInt(6, -1);                                                // publishedMessageId
                 pstmt.setTimestamp(7, null);                                        // confirmedOn
 
-                if (payload.length < PAYLOAD_BYTE_SIZE_THRESHOLD) {
-                    pstmt.setBytes(8, payload);                                     // smallPayload
+                // smallPayload (=8) vs. largePayload (=9)
+                if (payload == null) {
```
MMaiero: Can't we have for the null payload the cose as it was in ESF 7? https://github.com/eclipse/kura/blob/d78e70cff2404e292cb9876344d171273966a66c/kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/data/store/DbDataStore.java#L320

MMaiero: so the is would be 
```
if (isNull(payload) || payload.length < PAYLOAD_BYTE_SIZE_THRESHOLD){
```

marcellorinaldo: Like this?

```java
if (payload == null || payload.length < PAYLOAD_BYTE_SIZE_THRESHOLD) {
    pstmt.setBytes(8, payload);
    pstmt.setNull(9, Types.BLOB);
} else ...
```

marcellorinaldo:
Tested as working on Raspberry Pi + example publisher DP.

kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/data/store/DbDataStore.java
MMaiero: Yes. I have put the isNull to have it clean. But works too in the way you have put it

marcellorinaldo: Yes.

nicolatimeus (approved):

PR #4289, opened 2022-12-05
build(deb-installers): added jre 17 as alternative dependencies

marcellorinaldo:

Since Java 17 is now supported, the DEB installers now have Java 17 jre packages as alternative debian dependencies.

**Related Issue:** N/A.

**Screenshots:** N/A.

### DIFF
--- a/kura/distrib/src/main/resources/docker-x86_64-nn/deb/control/control
+++ b/kura/distrib/src/main/resources/docker-x86_64-nn/deb/control/control
@@ -5,7 +5,7 @@ Priority: low
 Depends: dos2unix, unzip, 
  telnet, bluez-hcidump, 
  bluez, chrony, ntpdate, 
- openjdk-8-jre-headless, cron
+ openjdk-8-jre-headless | temurin-8-jdk | openjdk-17-jre-headless | temurin-17-jdk, cron
 Architecture: all
 Maintainer: support@eurotech.com
 Description: Kura is an inclusive software framework that puts a layer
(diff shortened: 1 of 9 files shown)


kura/distrib/src/main/resources/docker-x86_64-nn/deb/control/control
```diff
@@ -5,7 +5,8 @@ Priority: low
 Depends: dos2unix, unzip, 
  telnet, bluez-hcidump, 
  bluez, chrony, ntpdate, 
- openjdk-8-jre-headless, cron
+ openjdk-8-jre-headless | temurin-8-jdk | openjdk-17-jre-headless | temurin-17-jdk,
```
mattdibi, line 8: Could these alternative be too restrictive? I've seen other packages use the following:

```
 openjdk-8-jdk | openjdk-8-jdk-headless | openjdk-8-jre | openjdk-8-jre-headless 
```

instead of only

```
openjdk-8-jre-headless
```

to cover all the bases.

Ditto for `temurin-8` and `openjdk-17`

marcellorinaldo: Done some tests and `openjdk-8-jre-headles` is a subset of `openjdk-8-jre`; installing `openjdk-8-jre` the headless dependency is satisfied.

mattdibi: Can we confidently say that the same goes for `openjdk-8-jdk`? As in: if we specify in our `Depends` only `openjdk-8-jre-headless`, `openjdk-8-jdk` satisfies that dependency?

marcellorinaldo: Yes

mattdibi: After some testing we determined that this is true.

PR #4295, opened 2022-12-07
deprecate: org.eclipse.kura.wire.script.filter.provider

marcellorinaldo:
This PR deprecated the component `org.eclipse.kura.wire.script.filter.provider` because it is not working on Java 17. The component can be replaced with `org.eclipse.kura.wire.script.tools.filter.component.FilterComponent`.

**Related Issue:** N/A.

**Description of the solution adopted:** Besides adding the annotation to the classes of the package, the **metatype** has been modified to inform the user that the component is deprecated.

**Screenshots:** N/A.

### DIFF
--- a/kura/org.eclipse.kura.wire.script.filter.provider/src/main/java/org/eclipse/kura/wire/script/filter/provider/OutputWireRecordListWrapper.java
+++ b/kura/org.eclipse.kura.wire.script.filter.provider/src/main/java/org/eclipse/kura/wire/script/filter/provider/OutputWireRecordListWrapper.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2017, 2020 Eurotech and/or its affiliates and others
+ * Copyright (c) 2017, 2022 Eurotech and/or its affiliates and others
  * 
  * This program and the accompanying materials are made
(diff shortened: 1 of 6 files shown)


mattdibi (approved):

kura/org.eclipse.kura.wire.script.filter.provider/src/main/java/org/eclipse/kura/wire/script/filter/provider/OutputWireRecordListWrapper.java
```diff
@@ -19,6 +19,11 @@
 
 import org.eclipse.kura.wire.WireRecord;
 
+/**
+ * 
+ * @deprecated as of Kura 5.3.0
+ */
+@Deprecated
```
mattdibi: There's an extra `*` for each of these type of comments. Feel free to ignore this if you don't want to fix it.

```suggestion
/**
 * @deprecated as of Kura 5.3.0
 */
@Deprecated
```

marcellorinaldo: Fixed.

mattdibi (approved):

PR #4296, opened 2022-12-07
deprecate: org.eclipse.kura.wire.component.conditional.provider

marcellorinaldo:
This PR deprecated the component `org.eclipse.kura.wire.component.conditional.provider` because it is not working on Java 17. The component can be replaced with `org.eclipse.kura.wire.script.tools.conditional.component.ConditionalComponent`.

**Related Issue:** N/A.

**Description of the solution adopted:** Besides adding the annotation to the classes of the package, the **metatype** has been modified to inform the user that the component is deprecated.

**Screenshots:** N/A.

### DIFF
--- a/kura/org.eclipse.kura.wire.component.conditional.provider/src/main/java/org/eclipse/kura/internal/wire/conditional/ConditionalOptions.java
+++ b/kura/org.eclipse.kura.wire.component.conditional.provider/src/main/java/org/eclipse/kura/internal/wire/conditional/ConditionalOptions.java
@@ -19,7 +19,6 @@ import java.util.Map;
 import javax.script.ScriptException;
 
 /**
- * 
  * @deprecated as of Kura 5.3.0
  */
 @Deprecated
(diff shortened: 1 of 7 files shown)


kura/org.eclipse.kura.wire.component.conditional.provider/src/main/java/org/eclipse/kura/internal/wire/conditional/ConditionalOptions.java
```diff
@@ -18,6 +18,11 @@
 
 import javax.script.ScriptException;
 
+/**
+ * 
+ * @deprecated as of Kura 5.3.0
+ */
+@Deprecated
```
mattdibi: Uuuuh actually, there's an extra `*` for each of these type of comments. Feel free to ignore this if you don't want to fix it.

```suggestion
/**
 * @deprecated as of Kura 5.3.0
 */
@Deprecated
```

mattdibi (approved):

kura/org.eclipse.kura.wire.component.conditional.provider/src/main/java/org/eclipse/kura/internal/wire/conditional/ConditionalOptions.java
marcellorinaldo: Fixed.

mattdibi (approved):

PR #4297, opened 2022-12-07
ci: update meaning of `deprecate` commit type

mattdibi:
**Brief description of the PR.**: Update the meaning of the `deprecate` commit type for a more general use.

**Description of the solution adopted:** Up until now the `deprecate` commit type was underused due to the specificity we decided to give it. 

With this PR we want to generalise the use of the `deprecate` commit type such that it won't be used only for APIs deprecations.

The motivation is that by using the `refactor` commit type for deprecating components we're not giving them the required importance (in the release notes these commits will be hidden in the changelog section). This PR updates the Release notes template and the documentation for supporting the new meaning of the `deprecate` commit type.

### DIFF
--- a/.github/release_notes_template/template.hbs
+++ b/.github/release_notes_template/template.hbs
@@ -41,7 +41,7 @@ Bug Fixes:
 {{! Deprecated APIs/components section }}
 {{#ifContainsType commits type='deprecate'}}
 
-Deprecated:
+Deprecated APIs/components:
   {{#commits}}
     {{#ifCommitType . type='deprecate'}}
   * {{firstLetters hash number='10'}} - {{{commitDescription .}}} ({{authorName}})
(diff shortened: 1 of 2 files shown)


.github/release_notes_template/template.hbs
```diff
@@ -38,10 +38,10 @@ Bug Fixes:
     {{/ifCommitType}}
   {{/commits}}
 {{/ifContainsType}}
-{{! Deprecated APIs section }}
```
MMaiero, line 41: and setting as "Deprecated APIs/components section"?

mattdibi: Fixed. Please note that this is just a comment though. Do you think it would be better to have it as the title of the section? Currently it's simply "Depreated"...

MMaiero: Yes, I was thinking on both

MMaiero (approved):

PR #4300, opened 2022-12-13
feat: sketch implementation of JS Engine Service Loading through Apache SPI Fly

marcellorinaldo:

On Java 17 there is no implementation for `javax.script.ScriptEngineFactory` and `javax.script.ScriptEngine`. The `javax.script.ScriptEngineManager` is the component responsible for retrieving the available engine and factory [1] using the Java Service Provider APIs [2].

Current implementations might rely on such API and won’t work on Java 17. Java Service Provider class loading mechanisms are not supported out-of-the-box by OSGi.

**Related Issue:** https://github.com/eclipse/kura/issues/4301.

**Description of the solution adopted:** The `org.eclipse.kura.wire.script.tool` DP embeds the GraalVM JS Engine, which is compliant with `javax.script.ScriptEngine` [3]. It would be nice to have the bundle make such engine available through those javax APIs.

Service Loader Java APIs are not supported out-of-the box by OSGi, hence if you specify an exported service in a bundle it will not be visible by the others. In this case, the `ScriptEngineManager` will not be able to retrieve a `ScriptEngineFactory` that is not available in the system bundle (which is on the classpath of every bundle).

OSGi defines a specification for using the Service Loader APIs [4]. However, the specification is not implemented in Equinox. Apache provides an OSGi Service Loader called SPI Fly [5] that we can leverage. It will require to specify Service Consumers and Providers modifiying the MANIFEST of the bundles that use/provide the `javax.script.ScriptEngineFactory` components.

**Screenshots:** N/A.

**References**

[1] [ScriptEngineManager Java API](https://docs.oracle.com/javase/7/docs/api/javax/script/ScriptEngineManager.html)
[2] [Java Service Provider](https://docs.oracle.com/javase/7/docs/technotes/guides/jar/jar.html#Service%20Provider)
[3] [GraalVM JS ScriptEngine Implementation](https://www.graalvm.org/22.1/reference-manual/js/ScriptEngine/)
[4] [OSGi Service Loader Mediator Specification](https://docs.osgi.org/specification/osgi.cmpn/7.0.0/service.loader.html)
[5] [Apache SPI Fly](https://aries.apache.org/documentation/modules/spi-fly.html)

### DIFF
--- a/kura/distrib/src/main/ant/build_equinox_distrib.xml
+++ b/kura/distrib/src/main/ant/build_equinox_distrib.xml
@@ -992,6 +992,8 @@ fi]]>
                 value=", reference:file:${kura.install.dir}/${kura.symlink}/${plugins.folder}/com.zaxxer.HikariCP_${hikaricp.version}.jar@3" />
             <entry key="osgi.bundles" operation="+"
                 value=", reference:file:${kura.install.dir}/${kura.symlink}/${plugins.folder}/org.quartz-scheduler.quartz_${quartz.version}.jar@3" />
+            <entry key="osgi.bundles" operation="+"
+                value=", reference:file:${kura.install.dir}/${kura.symlink}/${plugins.folder}/org.apache.aries.spifly_${org.apache.aries.spifly.version}.jar@1:start" />
         </propertyfile>
     </target>
 
--- a/kura/org.eclipse.kura.camel/META-INF/MANIFEST.MF
+++ b/kura/org.eclipse.kura.camel/META-INF/MANIFEST.MF
@@ -4,7 +4,6 @@ Bundle-Name: org.eclipse.kura.camel
 Bundle-SymbolicName: org.eclipse.kura.camel;singleton:=true
 Bundle-Version: 1.6.0.qualifier
 Bundle-Vendor: Eclipse Kura
-Require-Capability: osgi.ee;filter:="(&(osgi.ee=JavaSE)(version=1.8))"
 Bundle-ActivationPolicy: lazy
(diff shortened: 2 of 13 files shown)


marcellorinaldo:
Please do not merge, it is not working.

heyoulin:
Just register ScriptEngineFactory manully in ScriptRunner class

```
private static ScriptEngineManager createManager(final ClassLoader classLoader) throws ScriptException {
        return runWithClassLoader(classLoader, new Callable<ScriptEngineManager>() {

            @Override
            public ScriptEngineManager call() throws Exception {
                // passing null here since this will trigger
                // the script engine lookup to use java basic
                // support for JavaScript
                ScriptEngineManager manager = new ScriptEngineManager();
                List<ScriptEngineFactory> factories = manager.getEngineFactories();
                ServiceLoader<ScriptEngineFactory> svcLoader = ServiceLoader.load(ScriptEngineFactory.class);
                Iterator<ScriptEngineFactory> it = svcLoader.iterator();
                while (it.hasNext()) {
                    ScriptEngineFactory factory = it.next();
                    if (!factories.stream().anyMatch(f -> (f.getClass() == factory.getClass()))) {
                        factory.getNames().forEach(name -> manager.registerEngineName(name, factory));
                    }
                }
                return manager;
            }
        });

    }
```

marcellorinaldo:
> 

Many thanks for the hint.

Also following that approach the bundle is not able to find any ScriptEngines. Maybe I am doing something wrong. I'm commiting your suggestions anyways.

PR #4206, opened 2022-10-17
feat: add Java 17 support

marcellorinaldo:
**Maven plugins updates:**

- `bnd-baseline-maven-plugin`: 3.4.0 -> 6.3.1
- `maven-bundle-plugin`: 3.3.0 -> 5.1.8
- `maven-compiler-plugin`: 3.3 -> 3.10.1
- `target-platform-configuration`: 1.6.0 -> 3.0.0
- `tycho-compiler-plugin`: 1.6.0 -> 3.0.0
- `tycho-maven-plugin`: 1.6.0 -> 3.0.0
- `tycho-p2-extras-plugin`: 0.26.0 -> 3.0.0
- `tycho-source-plugin`: 1.6.0 -> 3.0.0
- `tycho-sourceref-jgit`: 1.6.0 -> 3.0.0

**Dependencies updates:**

Tests
- `junit`: 4.12 -> 4.13.2
- `mockito-core`: 1.10.19 -> 4.8.1
- `objenesis`: 2.4 -> 3.3

**New dependencies:**

- `byte-buddy`: 1.12.18
- `byte-buddy-agent`: 1.12.18

**Summary of changes**

- Target platform: renamed `org.apache.log4j2-api-config` to `log4j2-api-config` since `.` and `-` are not allowed simultaneously in jar names
- Target platform: changed some steps in the definitions of `maven-bundle-plugin` configs
- Target platform & Kura: refactor of poms to have all the versions specified as properties
- Kura: updated manifests for correct log4j versioning
- Kura: updated manifests to remove deprecated required exec header and replaced with capability osgi.ee filter to allow both Java 1.8 and Java 17 execution environments
- Test: migration to mockito 4
- Build: added profile for separating Java 8 build steps from Java 17 ones based on the running jvm
- Build: added `buildAll` property in place of `activeBydefault` for the distrib since the latter is activated only when no other profiles are selected. That is no more the case since a profile between `java8` and `java17` is always active

PR #4306, opened 2022-12-16
fix(web2): non-clickable dialogs in login page after GWT upgrade

mattdibi:
**Brief description of the PR.**: After updating GWT the dialogs in the login page (i.e. the Access Banner and the Alert Dialog) were covered by the backdrop and could not be dismissed. This PR fixes the issue.

**Description of the solution adopted:** After updating GWT in the Java 17 https://github.com/eclipse/kura/pull/4206 PR the dialogs in the login page were covered by the backdrop and rendered unclickable.

kura_530

The issue and its solution are described in [this Stack Overflow post](https://stackoverflow.com/questions/20983110/bootstrap-modal-sitting-behind-backdrop).

Following the suggestion in the first answer I wrapped all the login UI elements inside a `<g:HTMLPanel>` (since the `UiBinder` expect to have only *one* child element) and moved the `Modal` and `AlertDialog` outside the main `Container` element. This seems to fix the issue.

image

> **Note**: Since I was already there I fixed the indentation of the `xml` file.

## Testing

The fix was tested on the Kura user workspace on my workstation.

- **Test 1**: Go to Security -> Web Console -> Enable banner. At logout the banner should be clickabe and not covered by the backdrop.
- **Test 2**: At login use wrong password. The alert dialog should be clickable and not covered by the backdrop.

## Additional notes

A similar issue was reported for the "Identities" UI but I wasn't able to reproduce. I'm waiting for more details before setting this PR as ready for review.

### DIFF
--- a/kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/client/ui/login/LoginUi.ui.xml
+++ b/kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/client/ui/login/LoginUi.ui.xml
@@ -1,17 +1,17 @@
 <!DOCTYPE ui:UiBinder SYSTEM "http://dl.google.com/gwt/DTD/xhtml.ent">
 
-<!-- 
-    
+<!--
+
     Copyright (c) 2019, 2022 Eurotech and/or its affiliates and others
-  
+
     This program and the accompanying materials are made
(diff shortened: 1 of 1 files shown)


MMaiero:
@mstankovic can you please review with the fixture and provide feedback? Thanks

MMaiero (approved):

PR #4309, opened 2022-12-16
refactor(core.cloud): log error if Request ID is null

gnunzi:
* refactor: exceptions of Callable are thrown away, so the error should be logged in Kura


gnunzi (commented):
Two conditions in one row

mattdibi:
Hello there @gnunzi 👋 ,

first and foremost: thank you for your contribution.

Before moving on, could you please spend a few words on:
- what you're trying to achieve with this PR? What is the final goal?
- what's the motivation behind these code changes?

Feel free to refer to our [pull request template](https://github.com/eclipse/kura/blob/develop/PULL_REQUEST_TEMPLATE.md) if you want to.

gnunzi:
If you build a malformed KuraPayload without RequestId or RequesterClientId, currently you don't get any error either in the logs or in the UI.
The code does throw an exception BUT the exception does not show up anywhere. The Exception is basically "lost" in the Java Executor which is executing this code.
Just try to send a mqtt request without RequestId: the request is silently discarded without any error log.

This PR is just to have at least a log error if that error condition (RequestId or RequesterClientId is nul) occurs. The Kura user needs to be notified of this error.

gnunzi:
The error in the workflow above looks like due to a "Unknown host kura-repo.s3.us-west-2.amazonaws.com: Temporary failure in name resolution". It might work in a re-run.

mattdibi:
Hi there @gnunzi,

I think that logging is indeed the correct way to proceed but we should provide more context for this to be useful during debugging. Therefore we should try and log the exception at a higher level.

In this case being `MessageHandlerCallable` a `Callable` task the exception thrown inside should be found inside the `Future` object that is returned by the `submit` call (i.e. around [here](https://github.com/gnunzi/kura-fork/blob/52bd846ee079856544f1a04cec32fc254c734c8f/kura/org.eclipse.kura.cloudconnection.eclipseiot.mqtt.provider/src/main/java/org/eclipse/kura/internal/cloudconnection/eclipseiot/mqtt/cloud/CloudConnectionManagerImpl.java#L475)). Could logging the exception there be more useful? What do you think?

mattdibi:
Thinking about it, using the `Future` to retrieve the exception would block execution to wait for the `Callable` to be ran and that defeat the purpose of that part of the code.

I think your approach is the correct one at this point but I'll propose a few small changes.

kura/org.eclipse.kura.cloudconnection.eclipseiot.mqtt.provider/src/main/java/org/eclipse/kura/internal/cloudconnection/eclipseiot/mqtt/cloud/MessageHandlerCallable.java
```diff
@@ -94,6 +94,8 @@ public Void call() throws Exception {
 
         String requestId = (String) this.kuraMessage.getMetric(METRIC_REQUEST_ID);
         if (requestId == null) {
+            logger.error("Request Id is null");            
+            // The exception below will not show up in Kura's logs
```
mattdibi: Remove unnecessary comment

```suggestion
```

kura/org.eclipse.kura.core.cloud/src/main/java/org/eclipse/kura/core/cloud/MessageHandlerCallable.java
```diff
@@ -99,7 +99,9 @@ public Void call() throws Exception {
 
         String requestId = (String) this.kuraMessage.getMetric(METRIC_REQUEST_ID);
         String requesterClientId = (String) this.kuraMessage.getMetric(REQUESTER_CLIENT_ID);
-        if (requestId == null || requesterClientId == null) {
+        if ( requestId == null || requesterClientId == null ) {
+            logger.error("Request Id or Requester Cliend Id is null" );            
+            // The exception below will not show up in Kura's logs
```
mattdibi: Remove unnecessary comment

```suggestion
```

kura/org.eclipse.kura.cloudconnection.eclipseiot.mqtt.provider/src/main/java/org/eclipse/kura/internal/cloudconnection/eclipseiot/mqtt/cloud/MessageHandlerCallable.java
```diff
@@ -94,6 +94,8 @@ public Void call() throws Exception {
 
         String requestId = (String) this.kuraMessage.getMetric(METRIC_REQUEST_ID);
         if (requestId == null) {
+            logger.error("Request Id is null");
```
mattdibi: Use a lower logging level since it's not critical for execution. I think `debug` would be more apt.

For performance reasons I would check for the log level and create the string only if we're at the correct one. See [here](https://logging.apache.org/log4j/log4j-2.3.2/performance.html) for reference.

```suggestion
        if(logger.isDebugEnabled()) {
            logger.debug("Request Id is null"); 
        }
```

kura/org.eclipse.kura.core.cloud/src/main/java/org/eclipse/kura/core/cloud/MessageHandlerCallable.java
```diff
@@ -99,7 +99,9 @@ public Void call() throws Exception {
 
         String requestId = (String) this.kuraMessage.getMetric(METRIC_REQUEST_ID);
         String requesterClientId = (String) this.kuraMessage.getMetric(REQUESTER_CLIENT_ID);
-        if (requestId == null || requesterClientId == null) {
+        if ( requestId == null || requesterClientId == null ) {
+            logger.error("Request Id or Requester Cliend Id is null" );
```
mattdibi: Same thing here:

```suggestion
        if(logger.isDebugEnabled()) {
            logger.debug("Request Id or Requester Cliend Id is null"); 
        }
```

mattdibi (changes requested):
See comments

mattdibi (approved):

mattdibi:
The Eclipse CI is currently experiencing a major outage and thus the build failed. I was able to build the code locally and it is running fine. Merging.

Issue #4305, opened 2022-12-15
Unable to delete Wire Graph when component no more available


marcellorinaldo:
**Describe the bug**
Unable to delete the wire graph when Deployment Package providing the component is uninstalled.

**To Reproduce**
1. Install a DP that contains a wire component (like `wire.script.filter` or `wire.script.tools`).
2. Go into the Wire Graph and connect a new installed component to a Timer. Apply.
3. Uninstall the previously installed package.
4. Try to delete the Wire Graph. Not possible.

**Expected behavior**
The wire graph should be deletable.

**Screenshots**
N/A.

**Target Environment (please complete the following information):**
 - Raspberry Pi4
 - OS version: Linux raspberry 5.15.61-v7l+ 1579 SMP Fri Aug 26 11:13:03 BST 2022 armv7l GNU/Linux
 - Additional info: N/A.

**Additional context**
N/A.

mattdibi:
Closed by #4310

PR #4310, opened 2022-12-16
fix: web UI bug where invalid component could not be deleted.

GregoryIvo:
**Related Issue:** https://github.com/eclipse/kura/issues/4305

**Description of the solution adopted:**  Web UI throws an error when you try to apply a wire graph with an invalid component. the issue is this error is present when trying to delete this wire graph as well.
check if the invalid component is going to be deleted (not active),  before adding It to the list of invalid components so the error is not thrown. 

**Screenshots:** n/a

**Any side note on the changes made:** n/a


GregoryIvo:
What it looks like now with the fix

https://user-images.githubusercontent.com/29900100/208203068-7ca90390-5808-443b-8d04-87442c5ccbe0.mov

mattdibi (approved):
Tested on my workstation using the development environment and removing the wirescript filter after setting the component in the wiregraph.

**Before this PR**

https://user-images.githubusercontent.com/22748355/208377784-9a5970e6-fba5-462c-9694-f6abb554b55a.mov

**After this PR**

https://user-images.githubusercontent.com/22748355/208377826-bb412993-a3f3-4460-933d-71066e611b2a.mov

LGTM

PR #4318, opened 2022-12-23
fix: missing reported bundle signature status in cloud requests

MMaiero:


MMaiero:
I'll open an item to refactor the old tests (the non Gherkin style ones) as I believe they are not exercising correctly the code. In my opinion we need to convert them to integration tests. Or add them on top of the existing in order to make sure that the actual output of the method is properly verified.

pierantoniomerlino (approved):

Issue #3071, opened 2020-09-14
Stale threads related to Paho


MMaiero:
**Describe the bug**
Eclipse Kura is affected by the following issue in Paho: https://github.com/eclipse/paho.mqtt.java/issues/654
When the broker is not able to manage messages but is still able to accept connections, it can happen that the connection times out, the framework requests disconnection and close of the MQTT client but the related threads are actually still there. After some time the framework will have thousands of threads blocked that will consume CPU and memory causing unpredictable effects.

**To Reproduce**
The broker should be able to receive connections but not able to process messages.
A simple use case can be recreated with `socat TCP-LISTEN:8080,fork,reuseaddr -`

**Expected behavior**
The threads spawn in old failed connections should be closed by pho/framework to prevent strange behaviours of the system.

**Screenshots**
![Screenshot 2020-09-11 at 17 24 58](https://user-images.githubusercontent.com/10448126/93073209-d781a280-f682-11ea-91cd-9999d3a73b65.png)

**Target Environment (please complete the following information):**
 - Board [e.g. Raspberry Pi 3] all
 - OS version: [e.g. provide the result of `uname -a`] all
 - Additional info: 

**Additional context**
N/A

PR #4324, opened 2023-01-11
fix(core): Fix ssl still active after change keystore in SSLServiceManager

salvatore-coppola:
Now in DataServiceImpl shutdownAutoConnectStrategy is invoked org.eclipse.kura.data.transport.listener.DataTransportListener.onConfigurationUpdated(boolean).

Lowered to 4 (lesser than the minimum RECONNECTION_MIN_DELAY) the sum of timeout and quiesce time of org.eclipse.paho.client.mqttv3.MqttAsyncClient.disconnectForcibly.

### DIFF
--- a/kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/data/DataServiceImpl.java
+++ b/kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/data/DataServiceImpl.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2011, 2022 Eurotech and/or its affiliates and others
+ * Copyright (c) 2011, 2023 Eurotech and/or its affiliates and others
  *
  * This program and the accompanying materials are made
  * available under the terms of the Eclipse Public License 2.0
--- a/kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/data/DataServiceImpl.java
+++ b/kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/data/DataServiceImpl.java
@@ -67,7 +67,7 @@ import org.slf4j.LoggerFactory;
 public class DataServiceImpl implements DataService, DataTransportListener, ConfigurableComponent,
         CloudConnectionStatusComponent, CriticalComponent, AutoConnectStrategy.ConnectionManager {
 
-    private static final int RECONNECTION_MIN_DELAY = 5;
+    private static final int RECONNECTION_MIN_DELAY = 1;
 
     private static final Logger logger = LoggerFactory.getLogger(DataServiceImpl.class);
 
--- a/kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/data/DataServiceImpl.java
+++ b/kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/data/DataServiceImpl.java
@@ -1,12 +1,12 @@
 /*******************************************************************************
  * Copyright (c) 2011, 2022 Eurotech and/or its affiliates and others
- * 
+ *
  * This program and the accompanying materials are made
  * available under the terms of the Eclipse Public License 2.0
  * which is available at https://www.eclipse.org/legal/epl-2.0/
- * 
+ *
  * SPDX-License-Identifier: EPL-2.0
- * 
+ *
  * Contributors:
  *  Eurotech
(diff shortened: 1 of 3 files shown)


kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/data/DataServiceImpl.java
```diff
@@ -67,6 +67,8 @@
 public class DataServiceImpl implements DataService, DataTransportListener, ConfigurableComponent,
         CloudConnectionStatusComponent, CriticalComponent, AutoConnectStrategy.ConnectionManager {
 
+    private static final int RECONNECTION_MIN_DELAY = 5;
```
nicolatimeus: Does the workaround still work if we set this to 1 second? In this way we can be still consistent with the description of the connect.retry-interval configuration parameter, which states that the minimum value is 1 second.

kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/data/DataServiceImpl.java
```diff
@@ -1,12 +1,12 @@
 /*******************************************************************************
  * Copyright (c) 2011, 2022 Eurotech and/or its affiliates and others
```
nicolatimeus: Please update end year to 2023

kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/data/transport/mqtt/MqttDataTransport.java
```diff
@@ -1,12 +1,12 @@
 /*******************************************************************************
- * Copyright (c) 2011, 2021 Eurotech and/or its affiliates and others
- * 
+ * Copyright (c) 2011, 2022 Eurotech and/or its affiliates and others
```
nicolatimeus: Please update end year to 2023

kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/data/transport/mqtt/MqttDataTransport.java
```diff
@@ -99,6 +99,9 @@ public class MqttDataTransport implements DataTransportService, MqttCallback, Co
     private static final String TOPIC_ACCOUNT_NAME_CTX_NAME = "account-name";
     private static final String TOPIC_DEVICE_ID_CTX_NAME = "client-id";
 
+    private static final long MQTT_QUIESCE_TIMEOUT = 2;
+    private static final long MQTT_DISCONNECT_TIMEOUT = 2;
```
nicolatimeus: Can you please try to verify if issue https://github.com/eclipse/kura/issues/3071 occurs with the new settings?

kura/test/org.eclipse.kura.core.test/src/test/java/org/eclipse/kura/core/data/DataServiceImplTest.java
```diff
@@ -387,7 +387,7 @@ public void testActivateAndFailConnecting()
         }
 
         // initial checkin + 3 * authentication + (other mqtt + 3)
-        assertEquals(8, count.intValue());
+        assertEquals(4, count.intValue());
```
nicolatimeus: Please update copyright headers

kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/data/transport/mqtt/MqttDataTransport.java
salvatore-coppola: Checked

nicolatimeus (approved):

nicolatimeus:
@MMaiero this PR decreases the timeout of disconnectForcibly that we introduced for fixing #3071. It speeds up the disconnection process and it seems that the thread leak cannot be reproduced. Is it ok for you to merge this? Were there specific reasons why we did not specify a lower timeout when we originally made the fix?

MMaiero:
I have to admit that I don't remember the rationale for not setting a timeout.
Let's merge it and continue to verify during QA that everything works.

nicolatimeus:
There seems to be failures in a DataService test

MMaiero (approved):

Issue #4292, opened 2022-12-07
Possible data duplication on CloudPublisher


LucaDazi:
**Describe the bug**
The CloudPublisher can be configured to put into the KuraMessage body one of the properties received.
When this happens, the original property is left in the properties, which all end up in the list of metrics in the KuraMessage.

This leads to possible duplication of data. If a user wants to publish something in the body of the message, he probably also wants to remove that property from the metrics.

**To Reproduce**
1. Create a WireGraph with some string metric (Timer -> JavascriptFilter -> Publisher)
2. Configure the JavascriptFilter to create a "payload" Stirng metric in output
3. Configure the CloudPublisher to put the "payload" metric into the MQTT message body
4. Verify how the Publisher will publish both a "payload" metric in the metrics, and the content of "payload" in the body

**Expected behavior**
If the CloudPublisher is configured with a non-empty value in the "set.body.from.property" property, the property should not be included in the metrics of the resulting KuraMessage

**Target Environment (please complete the following information):**
 - Board: ANY
 - OS version: ANY

pierantoniomerlino:
Closed by #4328

PR #4328, opened 2023-01-16
feat(wire.CloudPublisher): Set as body a property removing it from metrics

marcomatteo:
Added "remove.body.from.metrics" boolean property in the component metatype.
If set to True and the "set a property as body" is not empty, remove the property from the KuraPayload metrics and KuraMessage properties.
The default value is false and preserve the current behaviour.

Added the test cases and refactor the CloudPublisherTest class with scenarios.

**Related Issue:** This PR fixes the issue #4292

### DIFF
--- a/kura/org.eclipse.kura.wire.component.provider/OSGI-INF/metatype/org.eclipse.kura.wire.CloudPublisher.xml
+++ b/kura/org.eclipse.kura.wire.component.provider/OSGI-INF/metatype/org.eclipse.kura.wire.CloudPublisher.xml
@@ -54,7 +54,7 @@
             cardinality="0"
             required="true"
             default="false"
-            description="If Set body from envelope property is not an empty value, the publisher will remove from the message metrics the property added in the message body.">
+            description="Set true to remove the metric that is sent as body of the message.">
         </AD>
     </OCD>
     
--- a/kura/org.eclipse.kura.wire.component.provider/OSGI-INF/metatype/org.eclipse.kura.wire.CloudPublisher.xml
+++ b/kura/org.eclipse.kura.wire.component.provider/OSGI-INF/metatype/org.eclipse.kura.wire.CloudPublisher.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!--
     
-    Copyright (c) 2016, 2020 Eurotech and/or its affiliates and others
+    Copyright (c) 2016, 2023 Eurotech and/or its affiliates and others
   
     This program and the accompanying materials are made
     available under the terms of the Eclipse Public License 2.0
@@ -46,6 +46,15 @@
             required="false"
             default=""
             description="If set to a non empty value, the publisher will set the message body to the value of the provided STRING or BYTE_ARRAY metric.">
+        </AD>
+        
+        <AD id="remove.body.from.metrics"
+            name="Remove body envelope property from metrics"
+            type="Boolean"
+            cardinality="0"
+            required="true"
+            default="false"
+            description="If Set body from envelope property is not an empty value, the publisher will remove from the message metrics the property added in the message body.">
         </AD>
     </OCD>
     
--- a/kura/org.eclipse.kura.wire.component.provider/src/main/java/org/eclipse/kura/internal/wire/publisher/CloudPublisher.java
+++ b/kura/org.eclipse.kura.wire.component.provider/src/main/java/org/eclipse/kura/internal/wire/publisher/CloudPublisher.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2016, 2020 Eurotech and/or its affiliates and others
+ * Copyright (c) 2016, 2023 Eurotech and/or its affiliates and others
  * 
  * This program and the accompanying materials are made
  * available under the terms of the Eclipse Public License 2.0
@@ -21,10 +21,12 @@ import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.Date;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Optional;
+import java.util.Set;
 
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
@@ -60,8 +62,6 @@ public final class CloudPublisher implements WireReceiver, ConfigurableComponent
 
     private static final Logger logger = LogManager.getLogger(CloudPublisher.class);
 
-    private static final String ASSET_NAME_PROPERTY_KEY = "assetName";
-
     private CloudPublisherOptions cloudPublisherOptions;
 
     private volatile WireHelperService wireHelperService;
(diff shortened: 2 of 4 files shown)


kura/org.eclipse.kura.wire.component.provider/OSGI-INF/metatype/org.eclipse.kura.wire.CloudPublisher.xml
```diff
@@ -47,6 +47,15 @@
...
             description="If set to a non empty value, the publisher will set the message body to the value of the provided STRING or BYTE_ARRAY metric.">
         </AD>
+        
+        <AD id="remove.body.from.metrics"
+            name="Remove body envelope property from metrics"
+            type="Boolean"
+            cardinality="0"
+            required="true"
+            default="false"
+            description="If Set body from envelope property is not an empty value, the publisher will remove from the message metrics the property added in the message body.">
```
marcellorinaldo: Maybe rephrase that, the description it is not so clear to me.

Something like: Set true to remove the metric from the message that is set in the message body.

kura/org.eclipse.kura.wire.component.provider/src/main/java/org/eclipse/kura/internal/wire/publisher/CloudPublisher.java
```diff
@@ -81,7 +81,7 @@ public final class CloudPublisher implements WireReceiver, ConfigurableComponent
      * Binds the Wire Helper Service.
      *
      * @param wireHelperService
-     *            the new Wire Helper Service
+     *                              the new Wire Helper Service
```
marcellorinaldo: What is happening here? Use the kura formatter and cleanup profiles if using Eclipse (https://github.com/eclipse/kura/tree/develop/kura/setups/formatting). Replace tabs with spaces.

kura/org.eclipse.kura.wire.component.provider/src/main/java/org/eclipse/kura/internal/wire/publisher/CloudPublisher.java
```diff
@@ -187,14 +187,14 @@ public void updated(final Wire wire, final Object value) {
...
+     *                       the {@link WireRecord}
      * @return the Kura payload
      * @throws NullPointerException
-     *             if the {@link WireRecord} provided is null
+     *                                  if the {@link WireRecord} provided is null
      */
     private KuraPayload buildKuraPayload(final WireRecord wireRecord) {
         requireNonNull(wireRecord, "Wire Record cannot be null");
-        final KuraPayload kuraPayload = new KuraPayload();
+        KuraPayload kuraPayload = new KuraPayload();
```
marcellorinaldo: Why did you remove the `final` keyword?

kura/org.eclipse.kura.wire.component.provider/src/main/java/org/eclipse/kura/internal/wire/publisher/CloudPublisherOptions.java
```diff
@@ -82,4 +83,9 @@ Optional<String> getBodyProperty() {
 
         return Optional.of(property);
     }
+
+    Boolean getRemoveBodyPropertyFromMetrics() {
```
marcellorinaldo: Use the primitive type here, it eases also the usage later on.

kura/org.eclipse.kura.wire.component.provider/src/main/java/org/eclipse/kura/internal/wire/publisher/CloudPublisherOptions.java
```diff
@@ -82,4 +83,9 @@ Optional<String> getBodyProperty() {
...
+        final Object propertyRaw = this.properties.get(CONF_REMOVE_BODY_PROPERTY);
```
marcellorinaldo: To be on the safe side maybe it is better to do something like:
```java
return (boolean) this.properties.getOrDefault(CONF_REMOVE_BODY_PROPERTY, false);
```

kura/org.eclipse.kura.wire.component.provider/src/main/java/org/eclipse/kura/internal/wire/publisher/CloudPublisher.java
```diff
@@ -280,9 +285,17 @@ private void publish(final List<WireRecord> wireRecords) {
 
     private Map<String, Object> buildKuraMessageProperties(final WireRecord wireRecord) {
         Map<String, TypedValue<?>> wireRecordProps = wireRecord.getProperties();
+        Set<String> metricProperties = new HashSet<>(wireRecordProps.keySet());
```
marcellorinaldo: Can you refactor this method? It is quite cryptic. Maybe add more meaningful names to variables.

marcellorinaldo (changes requested):

marcellorinaldo:
Thanks for the PR @marcomatteo. When you have the time, could you please address the review?

marcomatteo:
Thank you @marcellorinaldo for the review. I'll answer and fix soon.

kura/org.eclipse.kura.wire.component.provider/src/main/java/org/eclipse/kura/internal/wire/publisher/CloudPublisher.java
marcomatteo: You're right, I followed this [documentation page](https://wiki.eclipse.org/Kura/Getting_Started) and was the cleanup/formatter v2 and not the updated v3.

kura/org.eclipse.kura.wire.component.provider/OSGI-INF/metatype/org.eclipse.kura.wire.CloudPublisher.xml
marcomatteo: Changed in something more simple: "Set true to remove the metric that is sent as body of the message."

kura/org.eclipse.kura.wire.component.provider/src/main/java/org/eclipse/kura/internal/wire/publisher/CloudPublisher.java
marcomatteo: I want to know it too.

kura/org.eclipse.kura.wire.component.provider/src/main/java/org/eclipse/kura/internal/wire/publisher/CloudPublisher.java
marcomatteo: Updated with the same logic of metrics: first adding all the metrics to the to the Map, then remove only if the flag is true.

MMaiero:
Hi @marcomatteo,
are the changes to this PR done?
Can we ask @marcellorinaldo to perform another pass for the review?

Thanks

marcomatteo:
Hi @MMaiero,
yes, I'm waiting for the review.
Thank you in advance.

marcellorinaldo:
Also here, the Eclipse CI cannot retrieve Sonar status for Kura project.

image

marcellorinaldo (approved):

PR #4329, opened 2023-01-17
fix: Improved Modem PDP Context parsing

MMaiero:


kura/org.eclipse.kura.net.admin/src/main/java/org/eclipse/kura/net/admin/util/PppPdpUtil.java
```diff
@@ -0,0 +1,32 @@
+/*******************************************************************************
+ * Copyright (c) 2022 Eurotech and/or its affiliates and others
+ *
```
pierantoniomerlino, line 3: Please update the year.

pierantoniomerlino (approved):

PR #4332, opened 2023-01-18

mattdibi:
This PR adds the low level code for interfacing Kura with NetworkManager through Dbus. We're using `dbus-java` 3.3.2 for the Dbus connection.

Changelog:
- Added `NMDbusConnector` (name still WIP) where the business logic resides

Features:
- [X] Support for Wifi AP connections configuration
- [X] Support for Wifi Client connections configuration

Not implemented yet:
- [ ] IPv6 support
- [ ] Support for loopback interface configuration

Known Issues:
- ⚠️ Dbus methods returning `Tuple`-derived objects cannot be de-serialized and throw an exception at runtime. This seems to be a build-related issue since it works completely fine on my testing application.
- Missing unit tests


kura/org.eclipse.kura.net2.admin/src/main/java/org/eclipse/kura/net2/admin/NMDbusConnector.java
```diff
@@ -0,0 +1,223 @@
...
+
+    public static NMDbusConnector getInstance() {
+        if (instance == null) {
+            throw new IllegalStateException(
+                    "Instance not created yet. Please use NMDbusConnector.createInstance() first");
+        }
+
+        return instance;
+    }
+
```
pierantoniomerlino, line 83: I think that it is not so clear how this class should be instantiated/created.

kura/org.eclipse.kura.net2.admin/src/main/java/org/eclipse/kura/net2/admin/NMDbusConnector.java
```diff
@@ -0,0 +1,223 @@
...
+        }
+    }
+
+    public DBusConnection getDbusConnection() {
+        return this.dbusConnection;
+    }
+
+    public void apply(Map<String, Object> networkConfiguration) throws DBusException {
+        logger.info("Applying configuration using NetworkManager Dbus connector");
+
```
pierantoniomerlino, line 114: We'd put this in debug when the code is ready.

kura/org.eclipse.kura.net2.admin/src/main/java/org/eclipse/kura/net2/admin/NMDbusConnector.java
```diff
@@ -0,0 +1,223 @@
...
+            String dhcpClient4GatewayProperty = String.format("net.interface.%s.config.ip4.gateway", iface);
+            if (networkConfiguration.containsKey(dhcpClient4GatewayProperty)) {
+                String dhcpClient4Gateway = (String) networkConfiguration.get(dhcpClient4GatewayProperty);
+                ipv4Map.put("gateway", new Variant<>(dhcpClient4Gateway));
+            }
+
+        } else {
+            ipv4Map.put("method", new Variant<>("auto"));
+        }
```
pierantoniomerlino, line 291: We'd manage the situation when the interface is set as dhcp, but the user has set some custom dns servers.

mattdibi: Yeah... this can be done with a combination of `method` set to `auto` and `ignore-auto-dns` set to false. I was planning to fix it later but I might do this now since I'm already at it xD

kura/org.eclipse.kura.net2.admin/src/main/java/org/eclipse/kura/net2/admin/NMDbusConnector.java
mattdibi: So yeah... I'm not entirely sold on the current implementation.

The [application](https://github.com/mattdibi/NMTest/tree/feat/version_update) I'm using for testing currently hangs on exit if I don't close the dbus connection. Less than ideal...

I don't think this will remains as-is...

mattdibi: Thanks for the early feedback though, really appreciated :D

mattdibi:
Supersided by https://github.com/eclipse/kura/pull/4350

PR #4347, opened 2023-01-27
fix(core): disabling connect auto on startup in data service terminates active cloud connection

GregoryIvo:
Fixes a bug where auto-connect is enabled when there is an active connection to a cloud service and then updated to 'disabled', and the cloud connection disconnects.

**Related Issue:** n/a

**Description of the solution adopted:**  removed the disconnect method call when trying to disable the auto-connect strategy.

**Screenshots:** n/a

**Any side note on the changes made:** n/a

### DIFF
--- a/kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/data/DataServiceImpl.java
+++ b/kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/data/DataServiceImpl.java
@@ -989,7 +989,6 @@ public class DataServiceImpl implements DataService, DataTransportListener, Conf
 
     @Override
     public void stopConnectionTask() {
-        disconnect();
         stopConnectionMonitorTask();
     }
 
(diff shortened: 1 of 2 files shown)


kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/data/DataServiceImpl.java
```diff
@@ -989,7 +989,6 @@ public void startConnectionTask() {
 
     @Override
     public void stopConnectionTask() {
-        disconnect();
```
salvatore-coppola, line 992: @nicolatimeus you added this line in #4023 . Can we remove it safetely ?

nicolatimeus: I don't remember the reason for this. We can try removing it and see if it causes any regression.

GregoryIvo: It seems to be okay with it removed. It appears everything gets shutdown as expected. Will ask QA for some support.

salvatore-coppola (approved):

PR #4348, opened 2023-01-30
feat: Implemented new Db APIs for SQLite

nicolatimeus:

* Implemented new Db APIs for SQLite
* Fixed possible issue with H2 message store identity generator reset
* H2 Wire Record store implementation now always drops and recreates columns if inserted property type is not compatible


kura/org.eclipse.kura.db.sqlite.provider/src/main/java/org/eclipse/kura/internal/db/sqlite/provider/SqliteMessageStoreImpl.java
```diff
@@ -0,0 +1,452 @@
...
+
+        if (sessionId != null) {
+            builder = builder
+                    .withDataTransportToken(new DataTransportToken(rs.getInt("publishedMessageId"), sessionId));
+        }
+
+        return builder.build();
+    }
+
+    private <T> T withConnection(final ConnectionCallable<T> callable, final String exceptionMessage)
```
marcellorinaldo: Does it make sense to have a "`withPreparedStatement`" instead of `withConnection` so to factor in some more common code?

marcellorinaldo: Something like:
```java
private interface PreparedStatementCallable<T> {
    public T call(Connection conn, PreparedStatement preparedStatement) throws SQLException;
}

private <T> T withPreparedStatement(String sql, String exceptionMessage,
            final PreparedStatementCallable<T> callable) throws KuraStoreException {
    try (final Connection connection = this.dbService.getConnection()) {
        try (PreparedStatement preparedStatement = connection.prepareStatement(sql)) {
            return callable.call(connection, preparedStatement);
        }
    } catch (final Exception e) {
        throw new KuraStoreException(e, exceptionMessage);
    }
}
```

kura/org.eclipse.kura.db.sqlite.provider/src/main/java/org/eclipse/kura/internal/db/sqlite/provider/SqliteMessageStoreImpl.java
```diff
@@ -0,0 +1,452 @@
...
+    @Override
+    public synchronized int getMessageCount() throws KuraStoreException {
+
+        return withConnection(c -> {
+            try (final PreparedStatement stmt = c.prepareStatement(this.sqlMessageCount);
+                    final ResultSet rs = stmt.executeQuery()) {
+                if (rs.next()) {
+                    return rs.getInt(1);
+                } else {
+                    return -1;
```
marcellorinaldo: Should we return 0 here?

nicolatimeus: I agree that -1 is not ok here, I have moved the code to a utility class and have it throw an exception instead since it seems that H2 and SQLite always return a result set for this query.

kura/org.eclipse.kura.db.sqlite.provider/src/main/java/org/eclipse/kura/internal/db/sqlite/provider/SqliteMessageStoreImpl.java
nicolatimeus: I have added this to a utility class

kura/org.eclipse.kura.util/src/main/java/org/eclipse/kura/util/message/store/SqlMessageStoreHelper.java
```diff
@@ -0,0 +1,320 @@
...
+                    new String[] { "id" })) {
+                pstmt.setString(1, topic); // topic
+                pstmt.setInt(2, qos); // qos
+                pstmt.setBoolean(3, retain); // retain
+                pstmt.setTimestamp(4, now, this.utcCalendar); // createdOn
+                pstmt.setTimestamp(5, null); // publishedOn
+                pstmt.setInt(6, -1); // publishedMessageId
+                pstmt.setTimestamp(7, null); // confirmedOn
+                if (payload != null) {
+                    pstmt.setBytes(8, payload);
```
marcellorinaldo: I think we need to set the value to null if the condition is false.

kura/org.eclipse.kura.util/src/main/java/org/eclipse/kura/util/message/store/SqlMessageStoreHelper.java
```diff
@@ -0,0 +1,320 @@
...
+
+    public long store(String topic, byte[] payload, int qos, boolean retain, int priority)
+            throws KuraStoreException {
+        validate(topic);
+
+        final Timestamp now = new Timestamp(new Date().getTime());
+
+        return this.connectionProvider.withConnection(c -> {
+
+            long result = -1;
```
marcellorinaldo: As for in one of the previous comments, should we throw an exception instead of returning -1?

kura/org.eclipse.kura.db.sqlite.provider/src/main/java/org/eclipse/kura/internal/db/sqlite/provider/SqliteWireRecordStoreImpl.java
```diff
@@ -0,0 +1,107 @@
...
+
+    private final SqlWireRecordStoreHelper helper;
+
+    public SqliteWireRecordStoreImpl(final ConnectionProvider provider, final String tableName)
+            throws KuraStoreException {
+        final String sanitizedTableName = sanitizeSqlTableAndColumnName(tableName);
+
+        final SqlWireRecordStoreQueries queries = SqlWireRecordStoreQueries.builder()
+                .withSqlAddColumn("ALTER TABLE " + sanitizedTableName + " ADD COLUMN {0} {1};")
+                .withSqlCreateTable("CREATE TABLE IF NOT EXISTS " + sanitizedTableName
```
marcellorinaldo, line 49: Is there a particular reason for using `BIGINT` instead of `DATETIME` as the timestamp type?

nicolatimeus: I have kept it in this way for consistency with the old DbStore and H2DbWireRecordStore implementations. I don't remember the reason for having the timestamp as an integer.

marcellorinaldo (dismissed):

kura/org.eclipse.kura.db.sqlite.provider/src/main/java/org/eclipse/kura/internal/db/sqlite/provider/SqliteMessageStoreImpl.java
```diff
@@ -0,0 +1,198 @@
...
+                .withSqlDeleteDroppedMessages(
+                        DELETE_FROM + this.sanitizedTableName + " WHERE droppedOn <= ? AND droppedOn IS NOT NULL;")
+                .withSqlDeleteConfirmedMessages(
+                        DELETE_FROM + this.sanitizedTableName + " WHERE confirmedOn <= ? AND confirmedOn IS NOT NULL;")
+                .withSqlDeletePublishedMessages(DELETE_FROM + this.sanitizedTableName
+                        + " WHERE qos = 0 AND publishedOn <= ? AND publishedOn IS NOT NULL;")
+                .build();
+
+        this.helper = new SqlMessageStoreHelper(provider, tableName, queries, this::sanitizeSql);
+        this.helper.createTableAndIndexes();
```
marcellorinaldo: A `createTable` without index creation could be helpful in the helper

kura/org.eclipse.kura.util/src/main/java/org/eclipse/kura/util/message/store/SqlMessageStoreQueries.java
```diff
@@ -0,0 +1,276 @@
...
+    private final String sqlSetConfirmed;
+    private final String sqlAllUnpublishedMessages;
+    private final String sqlAllInFlightMessages;
+    private final String sqlAllDroppedInFlightMessages;
+    private final String sqlUnpublishAllInFlightMessages;
+    private final String sqlDropAllInFlightMessages;
+    private final String sqlDeleteDroppedMessages;
+    private final String sqlDeleteConfirmedMessages;
+    private final String sqlDeletePublishedMessages;
+    private final String sqlCreateNextMessageIndex;
```
marcellorinaldo, line 35: These required "index" fields need to be set also if someone don't wants to invoke the "createIndexes" method. Is that on purpose?

nicolatimeus: Currently all fields are checked to ensure that the user explicitly reviews them at least once and provide a value for each of them. If a field is not used, any value can be set to it.

kura/org.eclipse.kura.util/src/main/java/org/eclipse/kura/util/message/store/SqlMessageStoreHelper.java
```diff
@@ -0,0 +1,330 @@
...
+    public void unpublishAllInFlighMessages() throws KuraStoreException {
+        execute(this.queries.getSqlUnpublishAllInFlightMessages());
+    }
+
+    public void dropAllInFlightMessages() throws KuraStoreException {
+        updateTimestamp(this.queries.getSqlDropAllInFlightMessages());
+    }
+
+    public void deleteStaleMessages(int purgeAge) throws KuraStoreException {
+        final long timestamp = System.currentTimeMillis() - TimeUnit.SECONDS.toMillis(purgeAge);
```
marcellorinaldo: A subtle thing can happen here: the long will be set as long in the underlying `setObject()` of the execute method. Some databases might not automatically reparse this as a `Timestamp`. So I would replace this line with:

```java
final long difference = System.currentTimeMillis() - TimeUnit.SECONDS.toMillis(purgeAge);
Timestamp timestamp = new Timestamp(difference);
```

MMaiero:
@marcellorinaldo @salvatore-coppola Please provide your review on this PR.
Once we have review approvals we will merge the code.
The issue reported by Sonar is on code duplication that unfortunately cannot be solved right now due to the fact that the code for SQLite is indeed similar to H2. But H2 is going to be removed in future, so the duplication is acceptable for now.

marcellorinaldo:
> @marcellorinaldo @salvatore-coppola Please provide your review on this PR.
> Once we have review approvals we will merge the code.
> The issue reported by Sonar is on code duplication that unfortunately cannot be solved right now due to the fact that the code for SQLite is indeed similar to H2. But H2 is going to be removed in future, so the duplication is acceptable for now.

Finishing to review the wire part. It will be due soon.

marcellorinaldo (approved):

PR #4350, opened 2023-01-31

mattdibi:
This PR adds the low level code for interfacing Kura with NetworkManager through Dbus. We're using `dbus-java` 3.3.2 for the Dbus connection. The NM auto-generated code was generated by calling the DBus introspection mechanism on NetworkManager 1.30.6.

Changelog:
- Added `dbus-java` 3.3.2 and its dependencies under `/lib`
- **Common**
  - Added `NMDbusConnector` (name still WIP) where the business logic resides. It is a singleton responsible for calling the  property converters and the NM methods for applying the network **configuration** and retrieving the network **status**.
  - Added `NMDeviceAddedHandler` which is responsible for handling the "[DeviceAdded](https://networkmanager.dev/docs/api/latest/gdbus-org.freedesktop.NetworkManager.html#gdbus-signal-org-freedesktop-NetworkManager.DeviceAdded)" signal. In our case it will trigger a network configuration re-apply using the cached configuration.
- **Configuration**
  - Modified `NMConfigurationServiceImpl` to reference `NMDbusConnector` for applying the configuration settings.
- **Status**
  - Added `NMStatusServiceImpl` responsible for implementing the `NetworkStatusService` and requesting information about the network status to the `NMDbusConnector`
  - Added `NMStatusConverter` for converting the network status information provided by NetworkManager into something Kura can understand.
- Utilities
  - Added `NMDeviceState` for representing device status (duh!). See [reference](https://networkmanager.dev/docs/api/latest/nm-dbus-types.html#NMDeviceState).
  - Added `NMDeviceEnable` for representing the configured device status inferred by the `KuraInterfaceStatus` for the ip4/ip6 configuration.

Features:
- [ ] Device type **configuration** support
  - [X] Support for Wifi AP connections configuration (mostly done)
  - [X] Support for Wifi Client connections configuration (mostly done)
  - [X] Support for loopback interface configuration (unmanaged by NM)
- [ ] Device type **status** support
  - [x] Support for Ethernet connections status retrieval
  - [ ] Support for Wifi AP connections status retrieval
  - [ ] Support for Wifi Client connections status retrieval
  - [x] Support for loopback interface status retrieval
  - [ ] Support for PPP connections status retrieval
- [ ] `L2Only` mode
- [ ] "_Renew DHCP lease_" button
- [ ] Event handling
  - [X] **New device connected handling**: when a new network device is connected the "[DeviceAdded](https://networkmanager.dev/docs/api/latest/gdbus-org.freedesktop.NetworkManager.html#gdbus-signal-org-freedesktop-NetworkManager.DeviceAdded)" signal triggers a cached network configuration apply. If the new network device have an entry in the snapshot, the snapshot configuration is applied, it is disabled otherwise. This behaviour relies on two assumptions:
    - If a device is configured in the snapshot, but it's not currently present on the system, its configuration won't be removed from the snapshot.
    - If a device is not configured but it's currently present on the system. It will be disabled and its configuration written on the snapshot only when configured by the user. By default not configured interfaces are **disabled**.
  - [ ] **Kura network configuration enforcement**: we want to re-apply the current Kura configuration in case a user makes changes to the network configuration bypassing Kura. This means that we will capture network configuration change-related signals on DBus and trigger a cached configuration rollback.

Known Issues:
- Missing unit tests
- A better target for generating the NM introspection classes would be version 1.22.16
- `closeConnection()` method **should not be exposed**
- **Ipv6 support**: Ipv6 is currently always set to `netIPv6StatusDisabled/Unmanaged` and the snapshot property ignored while we wait for the support to be added. No support for Ip6 status retrival is in place ATM.
- **Loopback interface support**:
  - loopback interface support for NetworkManager was introduced only from version 1.41.6. Currently we explicitly set the loopback interface as not supported and thus, any configuration applied to it, will be ignored.
  - Loopback interface currently recognized as `NM_DEVICE_TYPE_GENERIC` instead of treating it as a proper `LOOPBACK` type
  - Due to these issues, on startup, the configuration and the status of the loopback interface will be incoherent (i.e. the WebUI will report the Loopback as configured for LAN while the status will report that is unmanaged)
- **Configuration partial write**: Currently, upon receiving the configuration from the ConfigurationService, the code will apply the configuration to all interfaces, even if there's no change in the configuration. We rely on the fact that NM is smart enough to not disrupt the connection, but it would be better to have a partial write.


mattdibi:
Supersided by #4363

PR #4363, opened 2023-02-07

mattdibi:
This PR adds the low level code for interfacing Kura with NetworkManager through Dbus. We're using `dbus-java` 3.3.2 for the Dbus connection. The NM auto-generated code was generated by calling the DBus introspection mechanism on NetworkManager 1.30.6.

Changelog:
- Added `dbus-java` 3.3.2 and its dependencies under `/lib`
- **Common**
  - Added `NMDbusConnector` (name still WIP) where the business logic resides. It is a singleton responsible for calling the  property converters and the NM methods for applying the network **configuration** and retrieving the network **status**.
  - Added `NMDeviceAddedHandler` which is responsible for handling the "[DeviceAdded](https://networkmanager.dev/docs/api/latest/gdbus-org.freedesktop.NetworkManager.html#gdbus-signal-org-freedesktop-NetworkManager.DeviceAdded)" signal. In our case it will trigger a network configuration re-apply using the cached configuration.
- **Configuration**
  - Modified `NMConfigurationServiceImpl` to reference `NMDbusConnector` for applying the configuration settings.
- **Status**
  - Added `NMStatusServiceImpl` responsible for implementing the `NetworkStatusService` and requesting information about the network status to the `NMDbusConnector`
  - Added `NMStatusConverter` for converting the network status information provided by NetworkManager into something Kura can understand.
- Utilities
  - Added `NMDeviceState` for representing device status (duh!). See [reference](https://networkmanager.dev/docs/api/latest/nm-dbus-types.html#NMDeviceState).
  - Added `NMDeviceEnable` for representing the configured device status inferred by the `KuraInterfaceStatus` for the ip4/ip6 configuration.

Features:
- [ ] Device type **configuration** support
  - [X] Support for Wifi AP connections configuration
  - [X] Support for Wifi Client connections configuration (missing support for `pingAccessPoint`)
  - [X] Support for loopback interface configuration (unmanaged by NM)
- [ ] Device type **status** support
  - [x] Support for Ethernet connections status retrieval
  - [ ] Support for Wifi AP connections status retrieval
  - [ ] Support for Wifi Client connections status retrieval
  - [x] Support for loopback interface status retrieval
  - [ ] Support for PPP connections status retrieval
- [ ] `L2Only` mode
- [ ] "_Renew DHCP lease_" button
- [ ] Event handling
  - [X] **New device connected handling**: when a new network device is connected the "[DeviceAdded](https://networkmanager.dev/docs/api/latest/gdbus-org.freedesktop.NetworkManager.html#gdbus-signal-org-freedesktop-NetworkManager.DeviceAdded)" signal triggers a cached network configuration apply. If the new network device have an entry in the snapshot, the snapshot configuration is applied, it is disabled otherwise. This behaviour relies on two assumptions:
    - If a device is configured in the snapshot, but it's not currently present on the system, its configuration won't be removed from the snapshot.
    - If a device is not configured but it's currently present on the system. It will be disabled and its configuration written on the snapshot only when configured by the user. By default not configured interfaces are **disabled**.
  - [ ] **Kura network configuration enforcement**: we want to re-apply the current Kura configuration in case a user makes changes to the network configuration bypassing Kura. This means that we will capture network configuration change-related signals on DBus and trigger a cached configuration rollback.

Known Issues:
- Missing unit tests
- A better target for generating the NM introspection classes would be version 1.22.16
- `closeConnection()` method **should not be exposed**
- **Ipv6 support**: Ipv6 is currently always set to `netIPv6StatusDisabled/Unmanaged` and the snapshot property ignored while we wait for the support to be added. No support for Ip6 status retrival is in place ATM.
- **Loopback interface support**:
  - loopback interface support for NetworkManager was introduced only from version 1.41.6. Currently we explicitly set the loopback interface as not supported and thus, any configuration applied to it, will be ignored.
  - Loopback interface currently recognized as `NM_DEVICE_TYPE_GENERIC` instead of treating it as a proper `LOOPBACK` type
  - Due to these issues, on startup, the configuration and the status of the loopback interface will be incoherent (i.e. the WebUI will report the Loopback as configured for LAN while the status will report that is unmanaged)
- **Configuration partial write**: Currently, upon receiving the configuration from the ConfigurationService, the code will apply the configuration to all interfaces, even if there's no change in the configuration. We rely on the fact that NM is smart enough to not disrupt the connection, but it would be better to have a partial write.

---

### Build

We need to enable the use of the new Kura networking through the `kura.network.provider` property.

Build the target platform:

```bash
mvn -f target-platform/pom.xml clean install
```

build the core components:

```bash
mvn clean install -f kura/pom.xml -Dmaven.test.skip=true -Dkura.network.provider=network-manager
```

build the target profile (in the example below we'll build for the Raspberry Pi 64 bit)

```bash
mvn -f kura/distrib/pom.xml clean install -Praspberry-pi-arm64 -Dmaven.test.skip=true -Dkura.network.provider=network-manager
```

### Running

On a freshly installed Raspberry Pi

#### Enable NetworkManager

Launch the Raspberry Pi Configuration tool

```bash
sudo raspi-config
```

Go to the “5 Advanced Options” option. Navigate to the “A4 Network Config” option and select it. Select “NetworkManager“, then press <kbd>Enter</kbd>.

Reboot when prompted to do so.

On reboot edit `/etc/NetworkManager/NetworkManager.conf`:

```
[main]
plugins=ifupdown,keyfile
dhcp=internal

[ifupdown]
managed=true
```

Then restart NetworkManager to apply the changes:

```bash
sudo systemctl restart NetworkManager
```

#### Install Kura

Install the required JVM before running the `deb` installer. This branch currently only support JVM 1.8

```bash
sudo apt install -y wget apt-transport-https gnupg
```

```bash
sudo wget -O - https://packages.adoptium.net/artifactory/api/gpg/key/public | sudo apt-key add -
```

```bash
sudo echo "deb https://packages.adoptium.net/artifactory/deb $(awk -F= '/^VERSION_CODENAME/{print$2}' /etc/os-release) main" | sudo tee /etc/apt/sources.list.d/adoptium.list
```

```bash
sudo apt update && sudo apt install temurin-8-jdk
```

Then you can install Kura with:

```bash
sudo apt install ./kura_5.3.0-SNAPSHOT_raspberry-pi-arm64_installer.deb
```

After installation **do not restart the device**.

#### Re-enable NetworkManager after Kura installation

Currently the Kura installer disables NetworkManager. We need to re-enable it after Kura installation using:

```bash
sudo systemctl enable NetworkManager
```
```bash
sudo systemctl start NetworkManager
```

#### Start Kura

Finally, you can run Kura and test the new Kura networking

```bash
sudo systemctl start kura
```

#### "DBusExecutionException: failed to update connection"

If, after rebooting the device, you get a `org.freedesktop.dbus.exceptions.DBusExecutionException: failed to update connection: settings plugin does not support modifying connections` in the `kura-console.log` and, upon logging in in the WebUI, a 500 error, you need to delete the `/etc/network/interface` file created by the Kura installer.

This is due to the fact that NetworkManager cannot modify connection settings that are from `/etc/network/interfaces`.


kura/org.eclipse.kura.nm/OSGI-INF/networkStatusService.xml
```diff
@@ -0,0 +1,23 @@
...
+   This program and the accompanying materials are made
+   available under the terms of the Eclipse Public License 2.0
+   which is available at https://www.eclipse.org/legal/epl-2.0/
+ 
+	SPDX-License-Identifier: EPL-2.0
+	
+	Contributors:
+	 Eurotech
+
+-->
```
pierantoniomerlino, line 15: Since all these file are new, please set only 2023 as copyright year. The same applied to all the other files.

kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/configuration/NMConfigurationServiceImpl.java
```diff
@@ -136,6 +147,7 @@ public void activate(ComponentContext componentContext, Map<String, Object> prop
 
     public void deactivate(ComponentContext componentContext) {
         logger.debug("Deactivate NetworkConfigurationService...");
+        this.nmDbusConnector.closeConnection();
```
pierantoniomerlino: Can the nmDbusConnector be null?

kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/configuration/NMConfigurationServiceImpl.java
```diff
@@ -173,7 +185,11 @@ public synchronized void update(Map<String, Object> receivedProperties) {
             decryptAndConvertPasswordProperties(modifiedProps);
             this.networkProperties = new NetworkProperties(discardModifiedNetworkInterfaces(modifiedProps));
 
-            // networkManager.applyConfiguration
+            try {
+                this.nmDbusConnector.apply(modifiedProps);
+            } catch (DBusException e) {
+                logger.error("Couldn't apply network configuration settings due to: ", e);
+            }
```
pierantoniomerlino: What happens if the nmDbusConnector creation has thrown an exception in the activate?

kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/configuration/NMDeviceEnable.java
```diff
@@ -0,0 +1,51 @@
...
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *  Eurotech
+ *******************************************************************************/
+package org.eclipse.kura.nm.configuration;
+
+public enum NMDeviceEnable {
+
```
pierantoniomerlino: Why we need this enum?

kura/org.eclipse.kura.nm/src/main/resources/org/freedesktop/AddAndActivateConnection2Tuple.java
```diff
@@ -0,0 +1,52 @@
+package org.freedesktop;
+
```
pierantoniomerlino: Not sure if the src/main/resources is the best place to put this code.

kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/configuration/KuraInterfaceStatus.java
```diff
@@ -0,0 +1,58 @@
...
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *  Eurotech
+ *******************************************************************************/
+package org.eclipse.kura.nm.configuration;
```
pierantoniomerlino: Is it possible to put together this class and the NMDeviceEnable? Maybe a better name is required.

kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/configuration/NMDbusConnector.java
```diff
@@ -0,0 +1,359 @@
...
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *  Eurotech
+ *******************************************************************************/
+package org.eclipse.kura.nm.configuration;
```
pierantoniomerlino: Please put this class in a different package (i.e. org.eclipse.kura.nm) since this is common for status and configuration.

kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/configuration/NetworkProperties.java
```diff
@@ -21,7 +21,7 @@
 
 public class NetworkProperties {
 
-    private Map<String, Object> properties;
+    private final Map<String, Object> properties;
```
pierantoniomerlino, line 26: Please put this class in a different package (i.e. org.eclipse.kura.nm) since this is common for status and configuration.

pierantoniomerlino:
I think that we should avoid the upload of the jar files in the lib folder. They can be downloaded during build and put in the correct folder. Please take a look [here](https://github.com/eclipse/kura/blob/9e37484295cd6ef541a0bf1130ff8e713802ec71/kura/org.eclipse.kura.ai.triton.server/pom.xml#L40): we used the `maven-dependency-plugin` to retrieve the dependencies and copy them in the lib directory.

kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/configuration/NMDbusConnector.java
mattdibi: I cannot create a package with the same name of another (`org.eclipse.kura.nm`). I will move the common code inside `org.eclipse.kura.nm.common`

kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/configuration/KuraInterfaceStatus.java
mattdibi: I cannot separate the two concepts since one (`NMDeviceEnable`) models whether the device should be disabled/unmanaged/activated, the other (`KuraInterfaceStatus`) models whether the IPv[X] configuration should be set as LAN/WAN/deactivated.

They act at two different abstraction levels:
- `NMDeviceEnable` at `device` a.k.a. `interface` level
- `KuraInterfaceStatus` at IP settings level

and, unfortunately, one(`NMDeviceEnable`) is inferred from the other(`KuraInterfaceStatus`).

I do agree about using a better naming scheme.

Here's my proposal:
- `NMDeviceEnable` -> `KuraDeviceStatus`
- `KuraInterfaceStatus` -> `KuraIpStatus`

pierantoniomerlino: For me it's ok. Probably I prefer:

- `NMDeviceEnable` -> `KuraInterfaceStatus`
- `KuraInterfaceStatus` -> `KuraIpStatus`

`device` looks like the whole system...

mattdibi: I used `device` because NM calls them that way, but, since we're modeling Kura concepts, it makes sense to use the Kura naming convention (i.e. `interface`).

kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/configuration/NMDbusConnector.java
mattdibi: Actually Eclipse is a dum-dum. I updated with the correct package.

kura/org.eclipse.kura.nm/src/main/resources/org/freedesktop/AddAndActivateConnection2Tuple.java
mattdibi: I moved everything in `src/main/java/org/freedesktop`

mattdibi:
> I think that we should avoid the upload of the jar files in the lib folder. They can be downloaded during build and put in the correct folder. Please take a look [here](https://github.com/eclipse/kura/blob/9e37484295cd6ef541a0bf1130ff8e713802ec71/kura/org.eclipse.kura.ai.triton.server/pom.xml#L40): we used the `maven-dependency-plugin` to retrieve the dependencies and copy them in the lib directory.

I was able to do it for all the `jar` with the exception of `jffi-1.3.9-native.jar`. I also updated the `.gitignore` so that after a build the tree won't become dirty from the downloaded artifacts.

**Note**: this is a temporary solution. We'll wrap these dependencies in a bundle and put it in the target-platform in the future.

mattdibi:
@pierantoniomerlino I set the PR ready for review but there's still a pending issue: `NMConfigurationServiceImpl` tests are failing because they're trying to instantiate a `NMDbusConnector` with a real DBus connection (which is missing on the CI and wrong). We should refactor the class to leverage dependency injection and provide a Mock DBus connection for testing.

I will take care of this next week.

kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/NMDbusConnector.java
```diff
@@ -0,0 +1,360 @@
...
+
+        return instance;
+    }
+
+    public DBusConnection getDbusConnection() {
+        return this.dbusConnection;
+    }
+
+    public void closeConnection() {
+        this.dbusConnection.disconnect();
```
mattdibi, line 101:
```suggestion
        if(Objects.nonNull(this.dbusConnection)) {
            this.dbusConnection.disconnect();
            this.dbusConnection = null;
        }
```

mattdibi:
> @pierantoniomerlino I set the PR ready for review but there's still a pending issue: `NMConfigurationServiceImpl` tests are failing because they're trying to instantiate a `NMDbusConnector` with a real DBus connection (which is missing on the CI and wrong). We should refactor the class to leverage dependency injection and provide a Mock DBus connection for testing.
> 
> I will take care of this next week.

Fixed in a554be91736e7e0f42cfd9006adb26dd2c4f6c35 and 6ed8e4f5e2fd15d33d8a621daf5a5d52954ba7b3

kura/org.eclipse.kura.nm/pom.xml
```diff
@@ -31,4 +31,98 @@
...
+
+    <build>
+		<plugins>
+			<plugin>
+				<groupId>org.apache.maven.plugins</groupId>
+				<artifactId>maven-dependency-plugin</artifactId>
+				<version>3.0.0</version>
+				<executions>
+					<execution>
+						<phase>generate-sources</phase>
```
pierantoniomerlino: It seems that the indentation is odd...

kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/KuraIpStatus.java
```diff
@@ -0,0 +1,58 @@
...
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *  Eurotech
+ *******************************************************************************/
+package org.eclipse.kura.nm;
```
pierantoniomerlino, line 13: When we'll support ipv6, this class and the KuraInterfaceStatus will be simplified, isn't it? I mean, probably the Kura status will be sufficient to configure NM?

mattdibi:
@pierantoniomerlino with f902a905e3279e6bedf579e5d158356ccccaf94b and 26f2946669409570c3838efc0fc945df30324a72 I refactored the `NMSettingsConverter` to throw on an unexpected parameter value. This should make it easier to debug a wrong conversion.

With the previous implementation (`Map` based), the code would throw a `cannot convert null to Variant` error which is less useful.

kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/KuraIpStatus.java
mattdibi: Unfortunately no.

The issue here is that we need to model the *interface* (a.k.a. *device* in the NM world) state from both the *ip4* and *ip6* status.

If the *ip4* and *ip6* are disabled then the *device* is disabled. But if any one of them is enabled the *device* is enabled.

This is why I need two conversion stages from the `net.interface.%s.config.ip<ver>.status` parameters to understand if the device is enabled or not.

pierantoniomerlino (approved):

mattdibi:
**Note**: most of the issue reported by Sonar cannot be addressed as they're related to how NetworkManager and DBus are designed.

pierantoniomerlino:
The PR is approved. The Sonar check fails since the code coverage is less than 50%, but we've planned to develop more tests in a separate PR.

MMaiero:
There is what I suppose generated code that is triggering sonar scan.
Can we check if we are able to modify the jenkinsfile to exclude it?

mattdibi:
> There is what I suppose generated code that is triggering sonar scan. Can we check if we are able to modify the jenkinsfile to exclude it?

@MMaiero I should already have done that and it's correctly ignored

MMaiero:
https://sonarcloud.io/project/issues?resolved=false&severities=CRITICAL&sinceLeakPeriod=true&branch=PR-4363&id=org.eclipse.kura%3Akura
Screenshot 2023-02-14 at 18 15 05

mattdibi:
> https://sonarcloud.io/project/issues?resolved=false&severities=CRITICAL&sinceLeakPeriod=true&branch=PR-4363&id=org.eclipse.kura%3Akura Screenshot 2023-02-14 at 18 15 05

@MMaiero 1Yeah but as I said [here](https://github.com/eclipse/kura/pull/4363#issuecomment-1429304546) these are inside my code and due to how the library I'm using are designed

MMaiero:
Can we add a nosonar annotation to the specific issues (or something similar) maybe associated with a note so that we keep an history of that?
I really hate those things but in this case probably it's the only way to go?

pierantoniomerlino:
@mattdibi There are some Sonar issues that can be easily fixed. In particular 2 can be fixed defining a constant in `NMSettingsConverter.java` and a couple of conditions that return always true in `KuraInterfaceStatus.java`

mattdibi:
> Can we add a nosonar annotation to the specific issues (or something similar) maybe associated with a note so that we keep an history of that? I really hate those things but in this case probably it's the only way to go?

I tried adding the "Won't fix" on Sonar but I don't have the option available. Maybe I'm missing a step? I would rather use the Sonar interface than adding the `NOSONAR` tags in the code.

> @mattdibi There are some Sonar issues that can be easily fixed. In particular 2 can be fixed defining a constant in `NMSettingsConverter.java` and a couple of conditions that return always true in `KuraInterfaceStatus.java`

Yeah, I didn't really like what Sonar suggests there as (IMO) they hinder readability in those particular cases (especially in the conditions). Given this and the fact that I already received the approval I left those there.

I can fix both though, as you point out they're really easy to fix. Maybe I'll add a few comments here and there to explain what's going on while making Sonar happy.

mattdibi:
I fixed the remaining critical issues (I really don't like bddb99d958e58df37519344a7228fdbd1db8d379 but whatever).

Regarding the library related ones. On Sonar I have:
image

But I should find:

image

This is probably due to limited permissions.

I would really _really_ like to avoid littering the code with `NOSONAR` comments. Do we have a way around this?

mattdibi:
> The PR is approved. The Sonar check fails since the code coverage is less than 50%, but we've planned to develop more tests in a separate PR.

Regarding coverage there's already a WIP PR addressing the issue here: https://github.com/eclipse/kura/pull/4380

PR #4354, opened 2023-02-01
refactor: FirewallConfigurationService in external bundle

pierantoniomerlino:
> **Note**: We are using the Conventional Commits convention for our pull request titles. Please take a look at the [PR title format document](https://github.com/eclipse/kura/blob/develop/CONTRIBUTING.md#submitting-the-changes) for the supported [types](https://github.com/eclipse/kura/blob/develop/CONTRIBUTING.md#type) and [scopes](https://github.com/eclipse/kura/blob/develop/CONTRIBUTING.md#scope).

This PR copies the `FirewallConfigurationService` in the `org.eclipse.kura.net.admin.firewall` bundle. This will be added only in the profiles with the new Networking (see #4326) based on NetworkManager.
Currently the default configuration is to keep the old Kura Networking. To enable the new one, build the distrib project wih the `-Dkura.network.provider=network-manager` option.

Signed-off-by: pierantoniomerlino <pierantonio.merlino@eurotech.com>

PR #4367, opened 2023-02-09
fix(web2): Set net.admin optional dependency in web2

pierantoniomerlino:
**Note**: We are using the Conventional Commits convention for our pull request titles. Please take a look at the [PR title format document](https://github.com/eclipse/kura/blob/develop/CONTRIBUTING.md#submitting-the-changes) for the supported [types](https://github.com/eclipse/kura/blob/develop/CONTRIBUTING.md#type) and [scopes](https://github.com/eclipse/kura/blob/develop/CONTRIBUTING.md#scope).

This PR fixes the NN profiles where the webUI doesn't start for a missing dependency.

**Description of the solution adopted:** After the merge of #4354, the `FirewallConfigurationService` is extracted from the `net.admin` bundle. The web2 bundle now depends on it for setting the firewall rules (before the PR, it depended from the `NetworkAdminService` that is API...). So, in NN profiles the web2 doesn't start since it needs the `net.admin` package. This PR sets the dependency optional to preserve the bundle resolution.


nicolatimeus (approved):

heyoulin:
I think should move to api package
image

pierantoniomerlino:
@heyoulin Definitely yours is the correct solution. However, we planned to review the FirewallConfigurationService, when the activity on the networking is done. So, this is only a temporary solution to make the NN profiles work.

Thanks!

PR #4385, opened 2023-02-17
test: initial test coverage for NMDbusConnectorTest

GregoryIvo:


GregoryIvo:
@mattdibi please take a look at how my PR is progressing

kura/test/org.eclipse.kura.nm.test/src/test/java/org/eclipse/kura/nm/NMDbusConnectorTest.java
```diff
@@ -0,0 +1,222 @@
...
+	DBusConnection dbusConnectionInternal;
+	
+	Boolean hasDBusExceptionBeenThrown = false;
+	
+	List<String> internalStringList;
+	
+	@BeforeClass
+	public static void setUpPersistantMocks() throws DBusException {
+		givenBasicMockedDbusConnector();
+	}
```
mattdibi: Can you explain why are you doing this?

Setting this fields `static` make them **retain their values between tests**. I understand that you create a new instance with the `@BeforeClass` method but **that's what JUnit is supposed to do out-of-the-box**.

You're opening yourself to side-effects that would make your tests to be a pain to debug and I don't understand why.

Can't you have:

```suggestion
	DBusConnection dbusConnection;
	NetworkManager mockedNetworkManager;
	NMDbusConnector instanceNMDbusConnector;
	DBusConnection dbusConnectionInternal;
	
	Boolean hasDBusExceptionBeenThrown = false;
	
	List<String> internalStringList;
```

and create the mock for each tests?

I know that `NMDbusConnector` is a singleton, but its instance should get destroyed for each test case.

mattdibi: Mmmmmh after thinking about it my assumption might be wrong.

Anyway... i might have an idea on how to fix this.

GregoryIvo:
@mattdibi 
HI! I have three questions for you

1) what do you think about how I implemented the static workaround?

2) What do you think about how I accomplished the mocking any suggestions?

3) any suggestions on what to validate for the apply? 

Thank you!

mattdibi:
Hello there @GregoryIvo,

> 1. what do you think about how I implemented the static workaround?

It doesn't compile. We can't override static method. My bad: I forgot about this little detail when we first talked about it.

It's _reflection_ time then.

We can avoid any change in the `NMDbusConnector` production code and just add a `tearDown` method in the tests.

```java
    @After
    public void tearDown() {
        resetSingleton(NMDbusConnector.class, "instance");
    }

    public static <T> void resetSingleton(Class<T> clazz, String fieldName) {
        Field instance;
        try {
            instance = clazz.getDeclaredField(fieldName);
            instance.setAccessible(true);
            instance.set(null, null);
            instance.setAccessible(false);
        } catch (Exception e) {
            throw new RuntimeException();
        }
    }
```

Again: not super-liking it but it's a common workaround for unit testing singletons.

> 2. What do you think about how I accomplished the mocking any suggestions?

All the basic building block seems to be there. I think they're not that flexible.
What if we had something like:

```java
givenSystemWithMockedDevice(String interfaceName, UInt32 deviceType, UInt32 deviceState);
```

```java
givenMapConfiguringInterface(String interfaceName, boolean dhcpEnabled);
```

So that we can have the following tests:

```java
public void applyShouldWorkWithOneNotConfiguredDevice() {
   givenBasicMockedDbusConnector();
   
   givenSystemWithMockedDevice("wlan0", 2 /* NMDeviceType.NM_DEVICE_TYPE_WIFI */, 100 /*NMDeviceState.NM_DEVICE_STATE_ACTIVATED*/);
   givenSystemWithMockedDevice("eth0", 1 /* NMDeviceType.NM_DEVICE_TYPE_ETHERNET */, 30 /*NMDeviceState.NM_DEVICE_STATE_DISCONNECTED*/);
   
   givenMapWith("net.interfaces", "eth0");
   givenMapWith("net.interface.eth0.config.ip4.status", "netIPv4StatusEnabledWAN");
   givenMapConfiguringInterface("eth0", true); // Here we populate the map with enough settings that the NMSettingConverter doesn't throw

   whenApplyIsCalledWith(this.properties);

   thenNoExceptionOccurrs();
   thenConnectionUpdateIsCalledFor("eth0");
   thenDisconnectIsCalledFor("wlan0");
}
```

This will allow for a more flexible and readable approach. All remaining tests will be a variation of the above one.

For improved readability I would write a converter for UInt32 and NMDeviceType/State so that we don't need comments to read the tests.

```java
public void applyShouldWorkWithOneNotConfiguredDevice() {
   givenBasicMockedDbusConnector();
   
   givenSystemWithMockedDevice("wlan0", NMDeviceType.NM_DEVICE_TYPE_WIFI, NMDeviceState.NM_DEVICE_STATE_ACTIVATED);
   givenSystemWithMockedDevice("eth0", NMDeviceType.NM_DEVICE_TYPE_ETHERNET, NMDeviceState.NM_DEVICE_STATE_DISCONNECTED);
   
   givenMapWith("net.interfaces", "eth0");
   givenMapWith("net.interface.eth0.config.ip4.status", "netIPv4StatusEnabledWAN");
   givenMapConfiguringInterface("eth0", true); // Here we populate the map with enough settings that the NMSettingConverter doesn't throw

   whenApplyIsCalledWith(this.properties);

   thenNoExceptionOccurrs();
   thenConnectionUpdateIsCalledFor("eth0");
   thenDisconnectIsCalledFor("wlan0");
}
```


> 3. any suggestions on what to validate for the apply?

Given that we already completely tested the `NMSettingConverter` we just need to ensure that the `NMDbusConnector`:
- creates and applies the connection when needed
- updates the connection when needed if there's already a connection in place
- should not update anything when there's an unsupported/unmanaged device
- disconnect the device when needed

At this level of abstraction we have:

**Inputs** (a.k.a. `given`s):
- **Devices** available on the system, their **state** and their **type**.
- **Devices** configured (i.e. inside the network properties map) and their desired **state** (enabled for LAN, enabled for WAN, disabled, unmanaged)

**Outputs** (a.k.a. expectations):
- Whether the device is updated (`connection.Update` and `NetworkManager.ActivateConnection` methods are called)
- Whether the device is disabled (`device.Disconnect` and `connection.Delete` methods are called)
- Whether the device is unsupported or unmanaged (none of the above methods is called).

We'll talk about this in further detail later today.

mattdibi:
@GregoryIvo Some additional notes:

- This branch got messed up. Instead of rebasing you merged the latest changes. We'll create a new one and port the commits there.
- Stylecheck failures: your IDE is not applying the Kura formatter and thus the build is failing due to some random tabs. Again we'll talk about it later today.

GregoryIvo:
- Stylecheck failures: your IDE is not applying the Kura formatter and thus the build is failing due to some random tabs. Again we'll talk about it later today. ✅ 
- Reflections ✅

mattdibi:
Supersided by #4408

PR #4386, opened 2023-02-17
fix(web2): construct the net properties from scratch instead of using old data

marcellorinaldo:
This PR fixes a regression that was introduced in some previous PR.

**Related Issue:** N/A.

**Description of the solution adopted:** The properties for the network configuration service are created from zero, instead of starting from the last cached properties object. In this way we prevent the UI from sending unconsistent values. There is still a problem on the Network Configuration Service nm implementation: the Configuration Service merges the properties found on the snapshot prior sending them to the component. Some further checks need to be performed there.

**Screenshots:** N/A.


marcellorinaldo:
Screenshot 2023-02-17 at 11 25 52

PR fails due to CI not being able to access Kura project on Sonar

mattdibi:
LGTM

Tested on RPi with:
image

The log confirms that the config sent is what we expect:
image


As already mentioned the Configuration Service merges the properties found on the snapshot prior sending them to the component and thus, in the attached screenshots, you'll see the `gateway` property updated even though it was not sent from the GUI.

PR #4387, opened 2023-02-17
fix: correct behaviour for `netIPvXStatusUnmanaged`

mattdibi:
Currently, the new Kura networking, when an interface is set to UNMANAGED, will set the NetworkManager `isManaged` flag to false.

We decided that the Kura device unmanaged and the NetworkManaged device unmanaged should be two distinguished concepts.

**Scenario**: A user wants to configure an interface in a way that Kura doesn’t support. He/she can set the network interface as Kura unmanaged and still use NetworkManager (through `nmcli` or `nmtui`) to configure it.

This is also the behaviour of the old Kura networking.

This PR fixes the behaviour on the new Kura networking


mattdibi:
Again: the build is running fine, but Jenkins cannot retrieve status informations from Sonar

image

marcellorinaldo (approved):

PR #4389, opened 2023-02-20
build: added generic-arm32 profile

marcellorinaldo:
This PR adds a generic profile compatible with `arm32hf` architectures. This profile provides a suitable installation for integrating Kura with **NetworkManager**.

Some references:

- [NetworkManager for administrators](https://networkmanager.dev/docs/admins/)

**Related Issue:** N/A.

**Description of the solution adopted:** The generic installer will perform the following actions:

1. Install the "network-manager" bundles instead of the usual "kura.net" ones.
2. Install the `99kura-nm.conf` that contains NM configuration. The file is installed through dpkg, hence it is automatically removed at uninstallation.
3. If python is installed, `snapshot_0.xml`, `iptables.init`, and `kura.properties` are edited by replacing `eth0` and `wlan0` with the actual interface names. If python is not installed or no ethernet interface exists, this susbstitution will not happen. More susbstitutions will be made if the edited files contain also `eth1`, `wlan1`, etc. and there are more than 1 interface for each type (ethernet and wireless).
4. Custom `jdk.dio.properties` file is installed based on the dected board (in this case it is just raspberry)
5. NetworkManager cannot modify connection settings that are from `/etc/network/interfaces`, so this file needs to be removed. The installer moves it to `/etc/network/interfaces.old` and restores it at uninstallation.

**Screenshots:** N/A.


marcellorinaldo:
Line 64 of `kura_install.sh` is to be verified.

marcellorinaldo:
> Line 64 of `kura_install.sh` is to be verified.

Managed by NetworkManager

marcellorinaldo:
Question for reviewer: I have added [Recommends dependencies](https://www.debian.org/doc/debian-policy/ch-relationships.html) for the Raspberry Pi, but raspbian already has those packages installed. Shall we remove such dependencies? So that on boards != raspberry those do not show up?

marcellorinaldo:
Question for reviewer: sometimes it is needed to remove the `/etc/network/interfaces` if present from a previous kura installation (see [troubleshooting section of this PR](https://github.com/eclipse/kura/pull/4363)). Shall we remove the file at installation to be on the safe side? On a fresh image, no problems occur

MMaiero:
@pierantoniomerlino any thoughts on @marcellorinaldo question?

pierantoniomerlino:
> Question for reviewer: sometimes it is needed to remove the `/etc/network/interfaces` if present from a previous kura installation (see [troubleshooting section of this PR](https://github.com/eclipse/kura/pull/4363)). Shall we remove the file at installation to be on the safe side? On a fresh image, no problems occur

It seems that NM cannot manage interfaces that are listed in the `/etc/network/interfaces` file. There is an option (`managed=true`) that can be set in the NM configuration file to avoid it. However, this seems not to work as expected. Moreover, there are situation where we need to keep some info in that file (i.e. can interfaces that cannot be managed by NM), but I don't think it is the case. 

IMHO, we can backup the file in the system, optionally install a new one (probably without any information?) and revert it at the uninstallation.

marcellorinaldo:
> > Question for reviewer: sometimes it is needed to remove the `/etc/network/interfaces` if present from a previous kura installation (see [troubleshooting section of this PR](https://github.com/eclipse/kura/pull/4363)). Shall we remove the file at installation to be on the safe side? On a fresh image, no problems occur
> 
> It seems that NM cannot manage interfaces that are listed in the `/etc/network/interfaces` file. There is an option (`managed=true`) that can be set in the NM configuration file to avoid it. However, this seems not to work as expected. Moreover, there are situation where we need to keep some info in that file (i.e. can interfaces that cannot be managed by NM), but I don't think it is the case.
> 
> IMHO, we can backup the file in the system, optionally install a new one (probably without any information?) and revert it at the uninstallation.

As now, I have tried to set `/etc/network/interfaces` as a dpkg configuration file. This means that, if the file already exists at installation, apt will prompt the user if he wants to replace it, like in the picture below:

Screenshot 2023-02-27 at 11 35 22

If the user inputs `Y`, then a backup of the old file is created by apt. When removing kura with `apt remove kura`, the installed file and the backup are not removed. If, instead, we uninstall using `dpkg --purge kura`, then installed and backup are removed as well.

Is this behavior OK?

@pierantoniomerlino @MMaiero

pierantoniomerlino:
@marcellorinaldo I don't think so. Unfortunately, if Kura is installed in an automatic way, no human intervention is needed. Moreover, since the `interfaces` file is not owned by the Kura package, I don't know if it is safe to to this operation because it can couse some conflict. Is it better to replace it in a post-install script?

marcellorinaldo:
> @marcellorinaldo I don't think so. Unfortunately, if Kura is installed in an automatic way, no human intervention is needed.

There is a flag to specify the behavior (https://manpages.ubuntu.com/manpages/trusty/man1/dpkg.1.html), like `--force-confnew`. But it will introduce more complexity at installation.

MMaiero:
Does it simplify our lives if we consider that this is a new profile that will be installed in systems where Kura has not been installed previously?
Potentially we can also mention the need for a manual cleanup to make things easier

marcellorinaldo:
> Does it simplify our lives if we consider that this is a new profile that will be installed in systems where Kura has not been installed previously?
> Potentially we can also mention the need for a manual cleanup to make things easier

On a freshly installed raspbian, the interfaces file is empty. I think that the code changes to backup and restore the interfaces file are simple, so I would keep them to have the installer more robust.

kura/distrib/src/main/resources/generic-arm32/deb/control/postinst
```diff
@@ -0,0 +1,89 @@
+#!/bin/bash
+
```
MMaiero, line 14: copyright header needed?

kura/distrib/src/main/resources/generic-arm32/deb/control/preinst
```diff
@@ -0,0 +1,94 @@
+#!/bin/bash
+
```
MMaiero, line 14: copyright header needed?

kura/distrib/src/main/resources/generic-arm32/deb/control/prerm
```diff
@@ -0,0 +1,124 @@
+#!/bin/bash
+
```
MMaiero, line 14: copyright header needed?

kura/distrib/src/main/resources/generic-arm32/firewall.init
```diff
@@ -0,0 +1,111 @@
+#!/bin/bash
+# iptables-restore script
```
MMaiero, line 16: copyright header needed?

kura/distrib/src/main/resources/generic-arm32/deb/control/postinst
marcellorinaldo: Don't know if we need to add it, other profiles do not have it....

MMaiero: I think we are required to. Here is an excerpt from the Eclipse Handbook:

> Where possible, all source code, property files, and metadata files (including application, test, and generated source code as well as other types of files such as XML, HTML, etc.) should contain appropriate copyright and license notices as well as information on each contribution.

So maybe we were wrong with the other old files...

marcellorinaldo: Ok, will add then

MMaiero (approved):

PR #4390, opened 2023-02-21
refactor(WireHelperService): Catch case that service pid is an array.

gnunzi:
I want to catch an unhandled error. This is what I saw in my environment:

```
!MESSAGE bundle com.***.service:1.0.0 (185)[com.***Component] : The activate method has thrown an exception
java.lang.ClassCastException: java.util.ArrayList cannot be cast to java.lang.String
       at org.eclipse.kura.internal.wire.helper.WireHelperServiceImpl.newWireSupport(WireHelperServiceImpl.java:147)
       at com.beegy.shaun.event.service.publisher.EventPublisherComponent.activate(EventPublisherComponent.java:68)
       at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
```

Basically, I used mistakenly the service pid twice in my meta-data, Osgi apparently creates an ArrayList for the service pid, so the exception is raised.

The code makes clear that service pid is expected to be a string, not an array. Maybe the developer used the service pid twice and that's an error case that can be catched and reported.

This code can help whoever will face the same configuration mistake in the future.

### DIFF
--- a/kura/org.eclipse.kura.wire.helper.provider/src/main/java/org/eclipse/kura/internal/wire/helper/WireHelperServiceImpl.java
+++ b/kura/org.eclipse.kura.wire.helper.provider/src/main/java/org/eclipse/kura/internal/wire/helper/WireHelperServiceImpl.java
@@ -145,9 +145,9 @@ public final class WireHelperServiceImpl implements WireHelperService {
         }
 
         if (wireComponentRef.getProperty(SERVICE_PID) instanceof java.util.ArrayList) {
-            throw new RuntimeException(
-                    "The service PID of {} is an array. It should be a string. (Maybe you used the same PID twice in OSGi metadata?)",
-                    wireComponentRef.getClass().getCanonicalName());
+            throw new RuntimeException(String.format(
+                    "The service PID of %s is an array. It should be a string. (Maybe you used the same PID twice in OSGi metadata?)",
+                    wireComponentRef.getClass().getCanonicalName()));
         }
         final String servicePid = (String) wireComponentRef.getProperty(SERVICE_PID);
         final String kuraServicePid = (String) wireComponentRef.getProperty(KURA_SERVICE_PID);
--- a/kura/org.eclipse.kura.wire.helper.provider/src/main/java/org/eclipse/kura/internal/wire/helper/WireHelperServiceImpl.java
+++ b/kura/org.eclipse.kura.wire.helper.provider/src/main/java/org/eclipse/kura/internal/wire/helper/WireHelperServiceImpl.java
@@ -144,6 +144,9 @@ public final class WireHelperServiceImpl implements WireHelperService {
             return null;
         }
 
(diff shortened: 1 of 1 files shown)


nicolatimeus:
Hi @gnunzi , thank you for your PR, it seems that the build is failing due to checkstyle, please make sure that you are using the formatter and cleanup profile available at [1].

[1] https://github.com/eclipse/kura/tree/develop/kura/setups/formatting

kura/org.eclipse.kura.wire.helper.provider/src/main/java/org/eclipse/kura/internal/wire/helper/WireHelperServiceImpl.java
```diff
@@ -144,6 +144,11 @@ public WireSupport newWireSupport(final WireComponent wireComponent,
             return null;
         }
 
+        if (wireComponentRef.getProperty(SERVICE_PID) instanceof java.util.ArrayList) {
```
nicolatimeus, line 147: Maybe we can change this to

` (!(wireComponentRef.getProperty(SERVICE_PID) instanceof String))`

since the code expects the service pid to be a string, in this way we can probably catch more error cases.

kura/org.eclipse.kura.wire.helper.provider/src/main/java/org/eclipse/kura/internal/wire/helper/WireHelperServiceImpl.java
```diff
@@ -144,6 +144,11 @@ public WireSupport newWireSupport(final WireComponent wireComponent,
...
+            throw new RuntimeException(String.format(
```
nicolatimeus: Sonar is complaining about this issue https://sonarcloud.io/project/issues?resolved=false&sinceLeakPeriod=true&types=CODE_SMELL&branch=PR-4390&id=org.eclipse.kura%3Akura&open=AYaSLhG3gxuJpOzIn5La. changing the exception type to `IllegalArgumentException` should probably fix.

kura/org.eclipse.kura.wire.helper.provider/src/main/java/org/eclipse/kura/internal/wire/helper/WireHelperServiceImpl.java
gnunzi: I had a look at it, thanks.
Imo if we catch the specific case ArrayList, we can give a strong hint message to the developer ("Maybe you used the same PID twice in OSGi metadata?")

The pid is always string, because it is parsed from the META-INF file. It just happens to be an ArrayList, if it is defined twice by error. You kind of loose this know-how here if you go generic.

Should I make it generic or remain specific? :)

MMaiero:
Now sonar passes. Unfortunately a misconfiguration in the Eclipse CI makes the build fail anyways.
@nicolatimeus If the PR is ok for you, I think we can merge the PR even with that error, not related to this contribution.

nicolatimeus (approved):

PR #4394, opened 2023-02-22
fix: OS indipendent paths in suppression file

mattdibi:
The suppression file was not working for non-Unix OS. This PR should fix the issue.


salvatore-coppola (approved):
tested on win11

PR #4397, opened 2023-02-22
fix(nm): Fix gateway setting in dhcp mode

pierantoniomerlino:

It seems that when in DHCP client mode, the static gateway is applied in any case. The value comes from a previous configuration. This PR fixes this issue, moving the instructions for setting the gateway only in the static configuration (and if the interface is in WAN mode).

### DIFF
--- a/kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/configuration/NMSettingsConverter.java
+++ b/kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/configuration/NMSettingsConverter.java
@@ -85,6 +85,11 @@ public class NMSettingsConverter {
             addressEntry.put("address", new Variant<>(address));
             addressEntry.put("prefix", new Variant<>(new UInt32(prefix)));
 
+            if (ip4Status.equals(KuraIpStatus.ENABLEDWAN)) {
+                Optional<String> gateway = props.getOpt(String.class, "net.interface.%s.config.ip4.gateway", iface);
+                gateway.ifPresent(gatewayAddress -> settings.put("gateway", new Variant<>(gatewayAddress)));
+            }
+
             List<Map<String, Variant<?>>> addressData = Arrays.asList(addressEntry);
             settings.put("address-data", new Variant<>(addressData, "aa{sv}"));
         } else {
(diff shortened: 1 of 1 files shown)


kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/configuration/NMSettingsConverter.java
```diff
@@ -81,6 +82,11 @@ public static Map<String, Variant<?>> buildIpv4Settings(NetworkProperties props,
             String address = props.get(String.class, "net.interface.%s.config.ip4.address", iface);
             Short prefix = props.get(Short.class, "net.interface.%s.config.ip4.prefix", iface);
 
+            if (ip4Status.equals(KuraIpStatus.ENABLEDWAN) && Objects.nonNull(address) && !address.isEmpty()) {
```
mattdibi: No need to check for non null or empty `address`. It's all taken care by the `NetworkProperties.get` method. See: https://github.com/eclipse/kura/blob/db579173be99c70f75354bb2aedf46c90670c7f1/kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/NetworkProperties.java#L43-L53

```suggestion
            if (ip4Status.equals(KuraIpStatus.ENABLEDWAN)) {
```

kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/configuration/NMSettingsConverter.java
```diff
@@ -240,70 +241,71 @@ private static UInt32 convertIp4(String ipAddrString) throws UnknownHostExceptio
...
-        case "MASTER":
-            return "ap";
-        default:
-            throw new IllegalArgumentException(String.format("Unsupported WiFi mode \"%s\"", kuraMode));
+            case "INFRA":
+                return "infrastructure";
+            case "MASTER":
+                return "ap";
+            default:
+                throw new IllegalArgumentException(String.format("Unsupported WiFi mode \"%s\"", kuraMode));
```
mattdibi: I'm going to be a PITA here: I don't like mixing a `fix` commit with a `style` one. Additionally there's the increased risk of merge conflicts.

Furthermore the Eclipse Clean Up profile doesn't like it...

mattdibi (changes requested):
See comments

kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/configuration/NMSettingsConverter.java
mattdibi: This goes for all the refactored `switch` statements

pierantoniomerlino: I've not touch this part. Probably the IDE applied a formatter when I saved the file.

pierantoniomerlino:
@mattdibi Done!

mattdibi (approved):
👍

PR #4419, opened 2023-02-28
fix(nm): handle empty "Gateway" property correctly

mattdibi:
Fix for when the "Gateway" property returns an empty string in the `NMStatusConverter` code.

Due to [how `InetAddress.getByName` works](https://docs.oracle.com/javase/7/docs/api/java/net/InetAddress.html#getByName(java.lang.String)):

> If the host is null then an InetAddress representing an address of the loopback interface is returned. See [RFC 3330](http://www.ietf.org/rfc/rfc3330.txt) section 2 and [RFC 2373](http://www.ietf.org/rfc/rfc2373.txt) section 2.5.3.

the empty string gets converted to "127.0.0.1", but this is **wrong** in our case: the empty string means that the "Gateway" is not set!

### DIFF
--- a/kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/status/NMStatusConverter.java
+++ b/kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/status/NMStatusConverter.java
@@ -18,6 +18,7 @@ import java.util.ArrayList;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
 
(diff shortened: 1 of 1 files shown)


pierantoniomerlino:
Does it make sense to set the gateway to an empty values when in LAN mode? I mean, is there a meaningful value, like 0.0.0.0, as presented [here](https://www.baeldung.com/linux/difference-ip-address#2-routing-and-gateway)?

mattdibi:
I just tested this on my RPi

**`eth0` configured in WAN mode**

```bash
busctl get-property org.freedesktop.NetworkManager /org/freedesktop/NetworkManager/IP4Config/6 org.freedesktop.NetworkManager.IP4Config Gateway
s "192.168.1.1"
```

```bash
route -n
Kernel IP routing table
Destination     Gateway         Genmask         Flags Metric Ref    Use Iface
0.0.0.0         192.168.1.1     0.0.0.0         UG    100    0        0 eth0
172.16.1.0      0.0.0.0         255.255.255.0   U     600    0        0 wlan0
192.168.1.0     0.0.0.0         255.255.255.0   U     100    0        0 eth0
```

**`eth0` configured in LAN mode**

```bash
busctl get-property org.freedesktop.NetworkManager /org/freedesktop/NetworkManager/IP4Config/8 org.freedesktop.NetworkManager.IP4Config Gateway
s ""
```

```bash
route -n
Kernel IP routing table
Destination     Gateway         Genmask         Flags Metric Ref    Use Iface
172.16.1.0      0.0.0.0         255.255.255.0   U     600    0        0 wlan0
192.168.1.0     0.0.0.0         255.255.255.0   U     100    0        0 eth0
```

---

Therefore: if the interface is set as **LAN** the `getGateway` will return `Optional.empty()`. I think this makes sense, what do you think?

pierantoniomerlino:
Yeah, it makes sense. 👍

pierantoniomerlino (approved):

marcellorinaldo (approved):

mattdibi:
As usual the CI reports error because it can't find the project. Merging anyway.

image

PR #4420, opened 2023-03-01
fix(nm): return correct ActiveAccessPoint regardless of mode

mattdibi:
Correctly return [the `ActiveAccessPoint`](https://networkmanager.dev/docs/api/latest/gdbus-org.freedesktop.NetworkManager.Device.Wireless.html#gdbus-property-org-freedesktop-NetworkManager-Device-Wireless.ActiveAccessPoint).

If the interface is set as `MASTER` it will return itself.
If the interface is set as `INFRA` it will return the AP it connects to.


marcellorinaldo (approved):

pierantoniomerlino (approved):

mattdibi:
As usual the CI is not reporting the status correctly. Merging anyway.

image

PR #4417, opened 2023-02-28
feat(web2): added WiFi channels and Country Code support

marcellorinaldo:
This PR:
- adds support for WiFi channels and country code
- fixes also some configuration update bugs

**Related Issue:** N/A.

**Description of the solution adopted:** N/A.

**Screenshots:** N/A.

**Any side note on the changes made:** N/A.

PR #4412, opened 2023-02-27
feat(nm): Added Network Configuration enforcement

salvatore-coppola:
Added NMConnectionChangedHandler managing Network Configuration rollback.

Tests will be added in the pr #4408 once the pr will be merged.

PR #4415, opened 2023-02-27
test: added test coverage for NMDbusConnector

GregoryIvo:
This PR added high test coverage for NMDbusConnector.  This PR will fail to build until #4414 is merged.

> **Note**: We are using the Conventional Commits convention for our pull request titles. Please take a look at the [PR title format document](https://github.com/eclipse/kura/blob/develop/CONTRIBUTING.md#submitting-the-changes) for the supported [types](https://github.com/eclipse/kura/blob/develop/CONTRIBUTING.md#type) and [scopes](https://github.com/eclipse/kura/blob/develop/CONTRIBUTING.md#scope).

Brief description of the PR. [e.g. Added `null` check on `object` to avoid `NullPointerException`]

**Related Issue:** This PR fixes/closes {issue number}

**Description of the solution adopted:** A more detailed description of the changes made to solve/close one or more issues. If the PR is simple and easy to understand this section can be skipped.

**Screenshots:** If applicable, add screenshots to help explain your solution

**Any side note on the changes made:** Description of any other change that has been made, which is not directly linked to the issue resolution [e.g. Code clean up/Sonar issue resolution]

PR #4422, opened 2023-03-01
fix(nm): data race on status retrieval after configuration apply

mattdibi:
All clients of the `NMDBusConnector` expect that, after the `apply` method returns, the network configuration was applied **and the system is in a stable condition**.

Due to the asynchronous nature of the DBus protocol this is not completely true, even though the `dbus-java` library implements some facilities to this end.

We discovered the unexpected behaviour in:
- #4412 : When the signal listener are re-activated after the configuration `apply` they immediately trigger in an infinite loop of `apply` calls. This is because, when the `apply` method returns, the "configuration roll-out" is still in progress on the NetworkManager's end.
- #4417 : When the GUI sends the configuration to the `ConfigurationService` it immediately asks the `NetworkStatusService` for the status. This meant that some DBus object references were getting invalidated due to the "configuration roll-out". Resulting in a lot of `Caused by: org.freedesktop.dbus.errors.UnknownMethod: No such interface “org.freedesktop.DBus.Properties” on object at path /org/freedesktop/NetworkManager/IP4Config/230`

---

### Test performed

System in stable condition (i.e. no pending "configuration roll-out"). Through `busctl` I checked the DBus object for the `wlan0` interface.

```bash
busctl get-property org.freedesktop.NetworkManager /org/freedesktop/NetworkManager/Devices/3 org.freedesktop.NetworkManager.Device Ip4Config
o "/org/freedesktop/NetworkManager/IP4Config/230"
```

When pressing the "Apply" button on the Network configuration UI I find an exception in the log:

```bash
Caused by: org.freedesktop.dbus.errors.UnknownMethod: No such interface “org.freedesktop.DBus.Properties” on object at path /org/freedesktop/NetworkManager/IP4Config/230
```

Re-checking the DBus object for the `wlan0` interface:

```bash
busctl get-property org.freedesktop.NetworkManager /org/freedesktop/NetworkManager/Devices/3 org.freedesktop.NetworkManager.Device Ip4Config
o "/org/freedesktop/NetworkManager/IP4Config/232"
```

This is because, after the apply took effect, the `Ip4Config` object associated to `wlan0` changed **while the `NetworkStatusService` was serving a request and handling a reference to the old object**.

---

This issue should be split in two different problems affecting the two already mentioned PRs

“Vanishing DBus objects”: which affects the GUI when trying to access the network status after applying the new network settings

“Spurious configuration roll-back”: which affects the Configuration enforcement behaviour based on DBus signal captures. When the network configuration is updated we generate a lot of DBus signals. We should ignore them when detecting for external network configuration changes.

---

### Solution

To avoid the issue we need to ensure that:
- Be resilient to the DBus object path getting invalidated.
- When the `apply` method returns, Kura doesn't trigger the Configuration Enforcement mechanism by itself.

The solution is split in 3 different changes:
- "_Object vanishing issue_": In `NMStatusServiceImpl` we catch `UnknownMethod` exceptions and return an `Optional.empty()` if they happen. This is similar to what the old networking was doing when it couldn't retrieve status informations.
- "_Spurious configuration roll-back_": In the `apply` methods we wait for the `NM_DEVICE_STATE_CONFIG` to trigger on the device we're currently configuring before returning.
- "_Configuration enforcement_": We trigger the configuration roll-back only on when `NM_DEVICE_STATE_CONFIG` signal appears on device that Kura manages.

The last two items combined should avoid having the configuration infinite loop issue found in #4412

### DIFF
--- a/kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/NMDbusConnector.java
+++ b/kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/NMDbusConnector.java
@@ -19,7 +19,6 @@ import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Objects;
 import java.util.Optional;
-import java.util.concurrent.CountDownLatch;
 
 import org.eclipse.kura.KuraException;
 import org.eclipse.kura.executor.CommandExecutorService;
@@ -256,7 +255,13 @@ public class NMDbusConnector {
         logger.info("Settings iface \"{}\":{}", iface, deviceType);
 
         if (interfaceStatus == KuraInterfaceStatus.DISABLED) {
-            disable(device);
+            NMDeviceState currentState = getDeviceState(device);
+            if (currentState != NMDeviceState.NM_DEVICE_STATE_DISCONNECTED) {
+                DeviceStateLock dsLock = new DeviceStateLock(this.dbusConnection, device.getObjectPath(),
+                        NMDeviceState.NM_DEVICE_STATE_DISCONNECTED);
+                disable(device);
+                dsLock.waitForSignal();
+            }
         } else if (interfaceStatus == KuraInterfaceStatus.UNMANAGED) {
             logger.info("Iface \"{}\" set as Kura UNMANAGED. Skipping configuration.", iface);
         } else { // NMDeviceEnable.ENABLED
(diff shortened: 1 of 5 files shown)


mattdibi:
Coverage issues will be addressed in #4415

kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/NMDbusConnector.java
```diff
@@ -208,15 +219,31 @@ public synchronized void apply() throws DBusException {
...
+            List<Device> availableInterfaces = getAllDevices();
+            manageNonConfiguredInterfaces(configuredInterfaces, availableInterfaces);
+
+            // TEMPORARY FIX!
+            // All NMDbusConnector client expect that the apply method return only when
+            // the configuration was applied. Unfortunately, due to the asynch nature of DBus,
+            // once the apply method returns, some status/config information can still change.
+            // The correct way to fix this would be by handling a signal coming from the DBus
+            // but this still work 99% of the time.
+            Thread.sleep(1500);  // NOSONAR
```
pierantoniomerlino: I don't think that this is a good solution, even if it is temporary...

pierantoniomerlino (approved):

pierantoniomerlino:
My only concern is whether 5s are sufficient in every situation. Let's try on slow devices.

marcellorinaldo (approved):

mattdibi:
> My only concern is whether 5s are sufficient in every situation. Let's try on slow devices.

It's an easy fix in case we need to change it. Keep in mind that this wait is not for the complete activation (which can take up to 5/10 seconds in my experience, even more if there's a problem) of the interface but just for the event that signals the configuration.

I agree that further testing is needed though.

PR #4429, opened 2023-03-03
fix: Extract sqlite native lib in bundle storage area

nicolatimeus:

Currently sqlite-jdbc extracts the native libraries in the `/tmp` folder, this may cause problems for example if `/tmp` is cleaned up by `systemd-tmfiles`. This PR sets the `org.sqlite.tmpdir` system property to the bundle storage area (which is also in tmpfs in Kura by default), so that native libraries are extracted there. It should still be possible to override the `org.sqlite.tmpdir` property if needed.


salvatore-coppola (approved):
LGTM

MMaiero:
The Jenkins code and test build as long as Sonar pass.
The Jenkins build reports a communication error in getting the results back from Sonar.
Eclipse IT is informed of this issue and working on solving this. Merging on our end as it is not an issue of the contributed code.

PR #4440, opened 2023-03-10
feat(web2): Radio Mode is replaced by Band when using NetworkManager

marcellorinaldo:
This PR replaces the *Radio Mode* parameter with a more user-friendly *Band* when using a device with `networkmanager`. The reason of this change is that it is not really possible to configure the band in `networkmanager`, (some fine-grained radio modes are collapsed to a generic one) so there might be a discrepancy on what you set on the UI and what is effectively applied. Moreover, a simple *Band* is more user friendly.

**Related Issue:** N/A.

**Description of the solution adopted:** Under the hoods, we continue using the radio modes to not break retrocompatibility. The UI is changed just to display the frequency band that corresponds to the configured radio mode.

**Screenshots:** Screenshot EN:
en

Screenshot JP:
jp


marcellorinaldo:
Translation was reviewed by our japanese collegues.

mattdibi:
LGTM but we need to discuss how to map "2.4GHz/5GHz" to what NetworkManager exposes.

Here you can see what I do in the code:

https://github.com/eclipse/kura/blob/40f9fef6f6118361a2aeed7599b800c491dc64bf/kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/configuration/NMSettingsConverter.java#L252-L267

Because NetworkManager only exposes "a" and "bg" for the band options:

image

See: https://networkmanager.dev/docs/api/latest/nm-settings-dbus.html

marcellorinaldo:
Merging the PR: build, tests, and Sonar are passing. Errors seems to be related to Jenkins.

PR #4446, opened 2023-03-14
fix(nm): trigger configuration rollback only for managed devices

mattdibi:
**Brief description of the PR**: This PR changes the Configuration Enforcement mechanism so that it is more selective on when to trigger. It will now trigger only if it detect changes on devices that are currently managed by Kura. 

I refactored `NMDbusConnector` such that:
- `getDeviceIpInterface` is now exposed as `public`
- added `getManagedDevices` which return the list of the currently managed devices. It corresponds to the list of interfaces `net.interfaces` contained in the snapshot.

With the two methods above the `NMConfigurationEnforcementHandler` is able to check whether a device is currently managed by Kura.

This renders the Configuration Enforcement mechanism a little more robust and less affected by configuration roll-back loops.


mattdibi:
Writing the description I just realised that this PR is conceptually **WRONG**:

If a new device gets plugged in and activated, Kura will ignore it. But we don't want that: we want Kura to disconnect devices that are not managed by it!

salvatore-coppola (approved):

PR #4448, opened 2023-03-14
feat(web2): added automatic channel selection support for WiFi on net2

marcellorinaldo:
This PR allows to select `Automatic` in the channel field; which corresponds to channel 0 with frequency 0. This makes it compatible with [NM](https://networkmanager.dev/docs/api/latest/nm-settings-dbus.html), which allows ACS by setting channel to 0.

**Related Issue:** N/A.

**Screenshots:** ACS:
Screenshot 2023-03-14 at 11 52 24

### DIFF
--- a/kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/client/ui/network/TabWirelessUi.java
+++ b/kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/client/ui/network/TabWirelessUi.java
@@ -107,12 +107,13 @@ public class TabWirelessUi extends Composite implements NetworkTab {
     private static final String WIFI_CIPHERS_CCMP_TKIP_MESSAGE = MessageUtils
             .get(GwtWifiCiphers.netWifiCiphers_CCMP_TKIP.name());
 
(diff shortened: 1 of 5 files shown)


mattdibi (changes requested):
Multiple bugs found:
- UI always report "automatic" channel
- After a few "Apply" the wifi interface gets disabled on its own

Some further debugging is required

marcellorinaldo:
@mattdibi Made some fixes and now it should be working.

mattdibi:
Tested on RPi 4 with NM 1.30.6. Everything working as expected.

AP mode:
- Setting the band and the channel, the AP is correctly set
- Setting the band and the automatic channel, the AP is set to the first available channel for that band
- Setting the channel and the band as 2.4/5 result in the channel dictating the band as expected.
- Setting the band as 2.4/5 and the channel as automatic result in the AP to be set in the first channel of the 2.4band and advertising as 802.11n

Station mode:
- Setting the band as 2.4/5 and the channel as automatic, if the remaining parameters are correct the device connects to the AP as expected.

mattdibi (approved):

mattdibi (changes requested):
Nevermind

mattdibi (approved):

PR #4449, opened 2023-03-14
feat(api): Added new KeyValueDbService API.

salvatore-coppola:
Added new KeyValueDbService API.


kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/connection/listener/ConnectionListener.java
```diff
@@ -11,13 +11,12 @@
...
 
-package org.eclipse.kura.store.listener;
+package org.eclipse.kura.connection.listener;
 
 import org.osgi.annotation.versioning.ConsumerType;
 
 /**
- * Listener interface to be implemented by applications that needs to be notified about connection events in the
- * {@link MessageStoreProvider} and {@link WireRecordStoreProvider}.
+ * Listener interface to be implemented by applications that needs to be notified about connection.
```
nicolatimeus:
```suggestion
 * Listener interface to be implemented by applications that needs to be notified about connection events.
```

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/db/keyvalue/KeyValueDbService.java
```diff
@@ -0,0 +1,112 @@
...
+ * Contributors:
+ *  Eurotech
+ *******************************************************************************/
+
+import org.eclipse.kura.KuraException;
+import org.eclipse.kura.connection.listener.ConnectionListener;
+import org.osgi.annotation.versioning.ProviderType;
+
+/**
+ * {@link KeyValueDbService} provides APIs to use the functionalities a of a KeyValue database.
```
nicolatimeus: Maybe we can explain better what we mean with KeyValue database.. something like this?
```suggestion
 * {@link KeyValueDbService} provides APIs to use the functionalities a of a database capable of storing data in form of key-value pairs.
```

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/db/keyvalue/KeyValueDbService.java
```diff
@@ -0,0 +1,112 @@
...
+@ProviderType
+public interface KeyValueDbService {
+
+    /**
+     * Adds a {@link ConnectionListener}.
+     *
+     * @param listener
+     *            to add
+     *
+     * @since 2.5.0
```
nicolatimeus: There should be no need for this, all methods will be added in 2.5.0, and we already mentioned that with the class-level annotation

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/db/keyvalue/KeyValueDbService.java
```diff
@@ -0,0 +1,112 @@
...
+     */
+    public void addListener(ConnectionListener listener);
+
+    /**
+     * Removes a {@link ConnectionListener}
+     *
+     * @param listener
+     *            to remove
+     *
+     * @since 2.5.0
```
nicolatimeus: same as above

nicolatimeus (approved):

PR #4450, opened 2023-03-14
fix: Manage Ubuntu cloud-init

MMaiero:
The PR introduces a couple of checks to identify and disable cloud-init (installed by default in Ubuntu server) and to give permissions to Kura to manage the networking interfaces via network-manager


mattdibi:
@MMaiero we'll probably need to document the fact the we need to restart NetworkManager after installation otherwise it will not pick up the changes.

mattdibi (approved):

MMaiero:
I think after installation, it clearly states to reboot the device
Also because, for what I have seen it will not able to apply the firewall rules if not after a reboot

marcellorinaldo:
Waiting for merging of https://github.com/eclipse/kura/pull/4451 so we can make the changes also for x86_64.

MMaiero:
@marcellorinaldo Updated

PR #4457, opened 2023-03-15
feat(nm): NetworkStatusService implementation for modems

pierantoniomerlino:

This PR adds the implementation of the `NetworkStatusService` for modems


kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/net/status/modem/Bearer.java
```diff
@@ -26,17 +27,17 @@ public class Bearer {
     private final String name;
     private final boolean connected;
     private final String apn;
-    private final BearerIpType ipType;
+    private final Set<BearerIpType> ipTypes;
```
mattdibi, line 30: This code was autogenerated on my RPi with version 1.14.12. Why did you need to change it?

There are two possible causes for this and one scares me:
1. The autogen script fumbled and the code was not correct
2. You used another MM version for which the signature of these methods is different <- This might be an issue

kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/MMModem3gppRegistrationState.java
```diff
@@ -0,0 +1,108 @@
...
+    MM_MODEM_3GPP_REGISTRATION_STATE_SEARCHING(2),
+    MM_MODEM_3GPP_REGISTRATION_STATE_DENIED(3),
+    MM_MODEM_3GPP_REGISTRATION_STATE_UNKNOWN(4),
+    MM_MODEM_3GPP_REGISTRATION_STATE_ROAMING(5),
+    MM_MODEM_3GPP_REGISTRATION_STATE_HOME_SMS_ONLY(6),
+    MM_MODEM_3GPP_REGISTRATION_STATE_ROAMING_SMS_ONLY(7),
+    MM_MODEM_3GPP_REGISTRATION_STATE_EMERGENCY_ONLY(8),
+    MM_MODEM_3GPP_REGISTRATION_STATE_HOME_CSFB_NOT_PREFERRED(9),
+    MM_MODEM_3GPP_REGISTRATION_STATE_ROAMING_CSFB_NOT_PREFERRED(10),
+    MM_MODEM_3GPP_REGISTRATION_STATE_ATTACHED_RLOS(11);
```
mattdibi, line 31: Why did you add a value for all enum if you don't use it? Is it a good idea to have values for the enums?

kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/MMModemBand.java
```diff
@@ -0,0 +1,731 @@
...
+ *  Eurotech
+ *******************************************************************************/
+package org.eclipse.kura.nm;
+
+import org.eclipse.kura.net.status.modem.ModemBand;
+import org.freedesktop.dbus.types.UInt32;
+
+public enum MMModemBand {
+
+    MM_MODEM_BAND_UNKNOWN(0),
```
mattdibi, line 20: Also here there's no need for the value

kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/NMDbusConnector.java
```diff
@@ -146,71 +150,114 @@ public synchronized List<String> getInterfaces() throws DBusException {
...
-
-            return NMStatusConverter.buildLoopbackStatus(interfaceName, loopbackPropertiesWrapper, ip4configProperties);
-        } else if (deviceType == NMDeviceType.NM_DEVICE_TYPE_WIFI) {
-            Wireless wirelessDevice = this.dbusConnection.getRemoteObject(NM_BUS_NAME, device.getObjectPath(),
-                    Wireless.class);
-            Properties wirelessDeviceProperties = this.dbusConnection.getRemoteObject(NM_BUS_NAME,
-                    wirelessDevice.getObjectPath(), Properties.class);
+        NetworkInterfaceStatus networkInterfaceStatus = null;
+        Optional<Device> device = getDeviceByInterfaceName(interfaceName);
+        if (device.isPresent()) {
```
mattdibi, line 157: This is just personal preference but you can add an early return here:

```suggestion
        if (!device.isPresent()) {
            return null;
        }
```

Personally I think it's more readable, up to you really.

kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/NMDbusConnector.java
```diff
@@ -146,71 +150,114 @@ public synchronized List<String> getInterfaces() throws DBusException {
...
+                    networkInterfaceStatus = NMStatusConverter.buildEthernetStatus(interfaceName,
+                            ethernetPropertiesWrapper, ip4configProperties);
+                    break;
+                case NM_DEVICE_TYPE_LOOPBACK:
+                    DevicePropertiesWrapper loopbackPropertiesWrapper = new DevicePropertiesWrapper(deviceProperties,
+                            Optional.empty(), NMDeviceType.NM_DEVICE_TYPE_LOOPBACK);
+
+                    networkInterfaceStatus = NMStatusConverter.buildLoopbackStatus(interfaceName,
+                            loopbackPropertiesWrapper, ip4configProperties);
+                    break;
```
mattdibi, line 195: My OCD requires me to comment on the fact that, for consistency, I would add the respective `create` methods here:

```suggestion
                    networkInterfaceStatus = createWiredStatus(...);
                    break;
                case NM_DEVICE_TYPE_LOOPBACK:
                    networkInterfaceStatus = createLoopbackStatus(...);
                    break;
```

kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/NMDbusConnector.java
```diff
@@ -259,56 +306,66 @@ private synchronized void manageConfiguredInterfaces(List<String> configuredInte
...
-
-        if (!CONFIGURATION_SUPPORTED_DEVICE_TYPES.contains(deviceType)
-                || !CONFIGURATION_SUPPORTED_STATUSES.contains(ip4Status)
-                || !CONFIGURATION_SUPPORTED_STATUSES.contains(ip6Status)) {
-            logger.warn("Device \"{}\" of type \"{}\" with status \"{}\"/\"{}\" currently not supported", iface,
-                    deviceType, ip4Status, ip6Status);
-            return;
-        }
+        Optional<Device> device = getDeviceByInterfaceName(iface);
+        if (device.isPresent()) {
```
mattdibi, line 310: Again, style preference for early return so that you can also add a log entry

```suggestion
        if (!device.isPresent()) {
            log.info("Device \"{}\" is missing. Skipping configuration.", iface);
            return;
        } 
```

kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/NMDbusConnector.java
```diff
@@ -438,9 +495,21 @@ private String getDeviceIpInterface(Device device) throws DBusException {
         return deviceProperties.Get(NM_DEVICE_BUS_NAME, NM_DEVICE_PROPERTY_INTERFACE);
     }
 
-    private Device getDeviceByIpIface(String iface) throws DBusException {
-        DBusPath ifaceDevicePath = this.nm.GetDeviceByIpIface(iface);
-        return this.dbusConnection.getRemoteObject(NM_BUS_NAME, ifaceDevicePath.getPath(), Device.class);
+    private Optional<Device> getDeviceByInterfaceName(String interfaceName) throws DBusException {
+        Optional<Device> device = Optional.empty();
+        if (!interfaceName.startsWith("ppp")) {
```
mattdibi: I don't like this.

We already have the `CONFIGURATION_SUPPORTED_DEVICE_TYPES` AND `STATUS_SUPPORTED_DEVICE_TYPES`. Can't we rely on those?

We should remove the `PPP` device type from those.

See:

https://github.com/eclipse/kura/blob/04d3f581e9073e646324208abc30692fec5420ba/kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/NMDbusConnector.java#L73-L74

and:

https://github.com/eclipse/kura/blob/04d3f581e9073e646324208abc30692fec5420ba/kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/NMDbusConnector.java#L78-L80

kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/NMDbusConnector.java
```diff
@@ -474,4 +543,103 @@ private void configurationEnforcementDisable() throws DBusException {
...
+
+    private String getDeviceIdFromNM(String devicePath) throws DBusException {
+        Properties nmModemProperties = this.dbusConnection.getRemoteObject(NM_BUS_NAME, devicePath,
+                Properties.class);
+        String deviceId = (String) nmModemProperties.Get(NM_DEVICE_BUS_NAME + ".Modem", "DeviceId");
+        logger.debug("Found DeviceId {} for device {}", deviceId, devicePath);
+        return deviceId;
+    }
+
+    private Map<DBusPath, Map<String, Map<String, Variant<?>>>> getManagedObjects() throws DBusException {
```
mattdibi: I would add a `MM` for making it obvious that this refers to ModemManager. 

```suggestion
    private Map<DBusPath, Map<String, Map<String, Variant<?>>>> getMMManagedObjects() throws DBusException {
```

or

```suggestion
    private Map<DBusPath, Map<String, Map<String, Variant<?>>>> getManagedObjectsFromMM() throws DBusException {
```

I think that shortly I'll need to perform the same operation for wpa_supplicant and a name clash is likely

kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/NMDbusConnector.java
```diff
@@ -474,4 +543,103 @@ private void configurationEnforcementDisable() throws DBusException {
...
+                            .getValue();
+                    return modemDeviceId.equals(deviceId);
+                }).findFirst();
+        if (modemEntry.isPresent()) {
+            modemPath = Optional.of(modemEntry.get().getKey().getPath());
+        }
+        return modemPath;
+    }
+
+    private Optional<String> getModemPath(String devicePath) throws DBusException {
```
mattdibi: Should we add some form of convention to distinguish methods that are called for MM or NM?

NM has a modem type and a modem path too... From the method name alone I don't know what I'm calling.

kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/status/NMStatusConverter.java
```diff
@@ -57,36 +86,36 @@ public class NMStatusConverter {
...
+    private static final String MM_MODEM_BUS_NAME = "org.freedesktop.ModemManager1.Modem";
+    private static final String MM_SIM_BUS_NAME = "org.freedesktop.ModemManager1.Sim";
+    private static final String MM_BEARER_BUS_NAME = "org.freedesktop.ModemManager1.Bearer";
+    private static final String MM_MODEM_LOCATION_BUS_NAME = "org.freedesktop.ModemManager1.Modem.Location";
+    private static final String MM_MODEM_3GPP_BUS_NAME = "org.freedesktop.ModemManager1.Modem.Modem3gpp";
     private static final String EMPTY_MAC_ADDRESS = "00:00:00:00:00:00";
-
-    private static final String NM_DEVICE_PROPERTY_STATE = "State";
+    private static final byte[] EMPTY_MAC_ADDRESS_BYTES = { 0x00, 0x00, 0x00, 0x00, 0x00, 0x00 };
+    private static final String STATE = "State";
```
mattdibi, line 102: Use the already established convention

```suggestion
    private static final String NM_DEVICE_PROPERTY_STATE = "State";
```

kura/test/org.eclipse.kura.nm.test/src/test/java/org/eclipse/kura/nm/MMBearerIpFamilyTest.java
```diff
@@ -0,0 +1,194 @@
...
+        public MMBearerIpFamilyToMMBearerIpFamilyTest(UInt32 intValue, MMBearerIpFamily ipFamily) {
+            this.inputIntValue = intValue;
+            this.expectedIpFamily = ipFamily;
+        }
+
+        @Test
+        public void shouldReturnCorrectIpFamily() {
+            whenCalculateMMBearerIpFamily();
+            thenCalculatedMMBearerIpFamilyIsCorrect();
+        }
```
mattdibi, line 63: Personally I prefer to see the input and output of the unit tests as parameters of the `given`/`when`/`then` methods to emphasise what we're testing:

```suggestion
        public void shouldReturnCorrectIpFamily() {
            whenConverterIsCalledWith(this.inputIntValue);
            thenCalculatedMMBearerIpFamilyEquals(this.expectedIpFamily);
        }
```

This is less relevant here since the test is parametrised, but it might be easier for debugging since it's perfectly clear where the input and expected values are stored.

kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/NMDbusConnector.java
pierantoniomerlino: IMHO, this class needs a refactor since it has too many responibilities. This suggestion can be done in that activity.

kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/NMDbusConnector.java
pierantoniomerlino: I totally agree on this. I don't like this implementation.

kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/NMDbusConnector.java
pierantoniomerlino: That's true. However, this method involves both NM and MM. I can call it `getMMModemPath`, since there is a `getModemPathFromMM`...

kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/status/NMStatusConverter.java
pierantoniomerlino: I call it simply `STATE` (replacing the old one) since this constant is used both for NM and MM. So, what could be the correct name? `DEVICE_PROPERTY_STATE`? Define a `NM_DEVICE_PROPERTY_STATE` and a `MM_DEVICE_PROPERTY_STATE` with the same value :-( ?

mattdibi (approved):
LGTM

PR #4458, opened 2023-03-15
build: Updated Apache Fileupload to v1.5.0

MMaiero:
Updated dependency and added setFileCountMax(2L)


kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/server/servlet/FileServlet.java
```diff
@@ -858,6 +858,7 @@ class UploadRequest extends ServletFileUpload {
     public UploadRequest(DiskFileItemFactory diskFileItemFactory) {
         super(diskFileItemFactory);
         setSizeMax(FileServlet.getFileUploadSizeMax());
+        setFileCountMax(2L);
```
marcellorinaldo, line 861: Why is this needed? And why are max. 2 files allowed per request?

marcellorinaldo: The first question is answered here: https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2023-24998

But why 2 files max.? Isn't 1 enough?

marcellorinaldo:
Tested as working on Raspberry Pi 4.

kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/server/servlet/FileServlet.java
MMaiero: I have tried with 1 and it does not allow to upload. I have not checked in the code but maybe the limit is exclusive

marcellorinaldo (approved):

PR #4459, opened 2023-03-16
fix: Fixed possible deadlock with SQLite

nicolatimeus:

The existing synchronization logic in ConnectionPoolManager might cause a deadlock interacting with the synchronization logic provided by the Hikari pool. This PR removes the synchronization logic from ConnectionPoolManager to rely only on Hikari pool synchronization.


salvatore-coppola (approved):
LGTM

PR #4461, opened 2023-03-17
feat: First implementation of NetworkStatusService REST APIs

nicolatimeus:

Adds NetworkStatusService REST APIs and cloud request handler

* Introduces a new `kura.permission.rest.network.status` permission for accessing the  REST APIs
* The current methods are provisional and need to be changed when NetworkStatusService APIs will be finalised
* Includes a minor refactor to make immutable some of the network status API classes


kura/org.eclipse.kura.rest.network.status.provider/src/main/java/org/eclipse/kura/network/status/provider/api/InterfaceStatusListDTO.java
```diff
@@ -0,0 +1,32 @@
...
+
+    private final List<NetworkInterfaceStatusDTO> interfaces;
+    private final List<FailureDTO> failures;
+
+    public InterfaceStatusListDTO(final List<NetworkInterfaceStatus> interfaces, final List<FailureDTO> failures) {
+        this.interfaces = interfaces.stream().map(NetworkInterfaceStatusDTO::fromNetworkInterfaceStatus)
+                .collect(Collectors.toList());
+        this.failures = (failures == null || failures.isEmpty()) ? null : failures;
+    }
+
```
pierantoniomerlino, line 31: Is it better to return a null or an empty list?

kura/org.eclipse.kura.rest.network.status.provider/src/main/java/org/eclipse/kura/internal/network/status/provider/NetworkStatusRestServiceImpl.java
```diff
@@ -0,0 +1,135 @@
...
+                interfaces.add(status);
+            } catch (final Exception e) {
+                failures.add(new FailureDTO(interfaceId, e));
+            }
+        }
+
+        return new InterfaceStatusListDTO(interfaces, failures);
+    }
+
+}
```
pierantoniomerlino, line 135: Please run the cleanup.

kura/org.eclipse.kura.rest.network.status.provider/src/main/java/org/eclipse/kura/internal/network/status/provider/NetworkStatusRestServiceImpl.java
```diff
@@ -0,0 +1,135 @@
...
+    }
+
+    private InterfaceStatusListDTO getInterfaceStatusInternal(final List<String> interfaceIds) {
+        final List<NetworkInterfaceStatus> interfaces = new ArrayList<>();
+        final List<FailureDTO> failures = new ArrayList<>();
+
+        for (final String interfaceId : interfaceIds) {
+            try {
+                final NetworkInterfaceStatus status = networkStatusService.getNetworkStatus(interfaceId)
+                        .orElseThrow(() -> new KuraException(KuraErrorCode.NOT_FOUND, "Interface not found"));
```
pierantoniomerlino, line 124: Is the `orElseThrow` compatible with Java8? Or, must we be compatible with Java8?

nicolatimeus: It should be available in Java8: https://docs.oracle.com/javase/8/docs/api/java/util/Optional.html#orElseThrow-java.util.function.Supplier-. We are already using this in Kura codebase.

kura/org.eclipse.kura.rest.network.status.provider/src/main/java/org/eclipse/kura/network/status/provider/api/InterfaceStatusListDTO.java
nicolatimeus: Setting the filed as null will cause GSON to not serialize it, so the failures property will not be added at all if there are no failures, it will not appear as `"failures": null`. But I don't have a strong preference, if you want we can also always serialize the field as an empty JSON array.

pierantoniomerlino: I've neither a strong preference. I only prefer to implement the same behavior of other Kura RestAPis. What is the default behavior with the others?

nicolatimeus: We don't have a consistent behavior. Since all other arrays in this resta apis are always serialized we could always serialize also this one even if there are no failures.

pierantoniomerlino (approved):

pierantoniomerlino (approved):

pierantoniomerlino:
It looks ok for me. Tested with a RPi.

PR #4462, opened 2023-03-17
refactor(api): updated NetworkStatusService and implementation to use the newly introduced device ID

pierantoniomerlino:

In order to correctly support modem devices, this PR updates the `NetworkStatusService` methods to use an `id` instead of an `interfaceName` to identify a network device. The `id` coincides with the `interfaceName` for ethernet and wifi interfaces, while for modems is the usb/pci path. The `interfaceName` for modems is the main port used for communication (i.e. ttyACM0, ttyUSB2, ...) and is used for identifying them by NetworkManager at configuration time.

### DIFF
--- a/kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/NMDbusConnector.java
+++ b/kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/NMDbusConnector.java
@@ -228,7 +228,7 @@ public class NMDbusConnector {
             modemDeviceProperties = getModemProperties(modemPath.get());
             if (modemDeviceProperties.isPresent()) {
                 simProperties = getModemSimProperties(modemDeviceProperties.get());
-//                bearerProperties = getModemBearersProperties(modemPath.get(), modemDeviceProperties.get());
+                bearerProperties = getModemBearersProperties(modemPath.get(), modemDeviceProperties.get());
             }
         }
         DevicePropertiesWrapper modemPropertiesWrapper = new DevicePropertiesWrapper(deviceProperties,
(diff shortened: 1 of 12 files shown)


kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/NMDbusConnector.java
```diff
@@ -140,20 +140,29 @@ public synchronized List<String> getInterfaces() throws DBusException {
...
+                    if (modemPath.isPresent()) {
+                        Optional<Properties> modemDeviceProperties = getModemProperties(modemPath.get());
+                        modemDeviceProperties.ifPresent(properties -> supportedDeviceNames
+                                .add(NMStatusConverter.getModemDeviceHwPath(properties)));
+                    }
+                } else {
+                    Properties deviceProperties = this.dbusConnection.getRemoteObject(NM_BUS_NAME,
+                            device.getObjectPath(), Properties.class);
+                    supportedDeviceNames.add(deviceProperties.Get(NM_DEVICE_BUS_NAME, NM_DEVICE_PROPERTY_INTERFACE));
+                }
```
mattdibi: I think you're duplicating this in the `getDeviceByInterfaceId` method. I would extract this code into a method `getIdFrom(Device device)` so that we have a single source of truth for the ID.

I know that, by the looks of it, it gets inefficient (we call a lot of the same properties on the DBus), but I think it's more important to have a single method for getting the ID given a device

mattdibi: Furthermore, this is more of a `getDeviceIds` than a `getInterfaces` method.

mattdibi:
Note: the PR is currently in a bad state. It is working but test are failing because they were not updated. Will fix on monday!

kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/NMDbusConnector.java
```diff
@@ -292,7 +290,7 @@ private synchronized void doApply(Map<String, Object> networkConfiguration) thro
 
     private synchronized void manageConfiguredInterfaces(List<String> configuredInterfaces,
             NetworkProperties properties) throws DBusException {
-        List<String> availableInterfaces = getAllDeviceInterfaceNames();
+        List<String> availableInterfaces = getInterfaces();
```
pierantoniomerlino: Does it make sense to call it `getDeviceIds`? Indeed, under the hood it calls `getDeviceId` for all the available and supported devices.

mattdibi (approved):

PR #4466, opened 2023-03-20
fix: NM logging verbosity

MMaiero:
Applied formatting to all NM classes


mattdibi (changes requested):
Split this PR in two:
- One for fixing the logging verbosity (commit type: `fix`)
- One for fixing the formatting (commit type: `style`)

mattdibi (approved):

PR #4468, opened 2023-03-20
refactor(web2): hide verify wifi credentials button if net provider is network-manager

marcellorinaldo:
If the network provider is NetworkManager, we currently do not have any clean way to implement the "Verify WiFi credentials" feature of the UI. Hence, the button that exposes that feature is hidden.

**Related Issue:** N/A.

**Screenshots:** N/A.

### DIFF
--- a/kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/client/ui/network/TabWirelessUi.java
+++ b/kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/client/ui/network/TabWirelessUi.java
@@ -512,9 +512,9 @@ public class TabWirelessUi extends Composite implements NetworkTab {
             }
 
             @Override
-            public void onSuccess(Boolean result) {
-                TabWirelessUi.this.isNet2 = result;
-                changeRadioModeToBand(result);
+            public void onSuccess(Boolean isNet2) {
+                TabWirelessUi.this.isNet2 = isNet2;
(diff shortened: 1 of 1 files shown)


marcellorinaldo:
Tests made:

- Old networking: the button is shown and it is working.
- NM: the button is correctly hidden.

Raspberry Pi 4, Raspbian 11 Bullseye.

kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/client/ui/network/TabWirelessUi.java
```diff
@@ -512,9 +512,9 @@ public void onFailure(Throwable caught) {
             }
 
             @Override
-            public void onSuccess(Boolean result) {
-                TabWirelessUi.this.isNet2 = result;
-                changeRadioModeToBand(result);
+            public void onSuccess(Boolean isNet2) {
```
GregoryIvo, line 515: Can we change the boolean name to something more descriptive? isNetworkManagerAvailable?

marcellorinaldo: Maybe in another PR. It is a differentiator for the package name in `kura.web`, we should rename also them in case.

GregoryIvo (approved):

PR #4479, opened 2023-03-22
fix(nm): apply configuration to all device even if one fails

salvatore-coppola:
Added exception capturing in [NMDbusConnector.java](https://github.com/eclipse/kura/compare/develop...salvatore-coppola:fix-fail-config-apply-iface-down?expand=1#diff-044ee4653a90a5b82546d6756de84757b52913a1630267773f0afa53bb66a38b)

### DIFF
--- a/kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/NMDbusConnector.java
+++ b/kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/NMDbusConnector.java
@@ -19,6 +19,7 @@ import java.util.EnumSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
+import java.util.NoSuchElementException;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
@@ -303,7 +304,7 @@ public class NMDbusConnector {
 
             try {
                 manageConfiguredInterface(iface, properties);
-            } catch (Exception e) {
+            } catch (DBusException | DBusExecutionException | IllegalArgumentException | NoSuchElementException e) {
                 logger.error("Unable to configure iface {}, skipping", iface, e);
             }
         }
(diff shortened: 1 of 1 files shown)


salvatore-coppola:
@marcellorinaldo need a little test on some devices.

salvatore-coppola:
DO NOT MERGE before NMDbusConnector refactor (probably will be overcomed

kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/NMDbusConnector.java
```diff
@@ -300,7 +297,11 @@ private synchronized void manageConfiguredInterfaces(List<String> configuredInte
                 continue;
             }
 
-            manageConfiguredInterface(iface, properties);
+            try {
+                manageConfiguredInterface(iface, properties);
+            } catch (Exception e) {
+                logger.error("Unable to configure iface {}, skipping", iface, e);
+            }
```
mattdibi: This is correct but we need to do the same for `manageNonConfiguredInterfaces`. I think you can just mirror the logic you have here there.

mattdibi (changes requested):
In addition to the above comment please rebase your branch so that you have the formatting fixes

kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/NMDbusConnector.java
salvatore-coppola: I've added the same logic in manageNonConfiguredInterfaces

mattdibi (approved):

kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/NMDbusConnector.java
```diff
@@ -301,7 +301,11 @@ private synchronized void manageConfiguredInterfaces(List<String> configuredInte
                 continue;
             }
 
-            manageConfiguredInterface(iface, properties);
+            try {
+                manageConfiguredInterface(iface, properties);
+            } catch (Exception e) {
```
mattdibi: Can we have a more specific catch? `DBusException` and `DBusRuntimeException`.

kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/NMDbusConnector.java
```diff
@@ -381,26 +385,32 @@ private void enableInterface(String deviceId, NetworkProperties properties, Devi
     private synchronized void manageNonConfiguredInterfaces(List<String> configuredInterfaces,
             List<Device> availableInterfaces) throws DBusException {
         for (Device device : availableInterfaces) {
-            NMDeviceType deviceType = getDeviceType(device);
-            String ipInterface = getDeviceId(device);
+            try {
+                NMDeviceType deviceType = getDeviceType(device);
+                String ipInterface = getDeviceId(device);
```
mattdibi: Would it be a good idea to extract this into its own method so that we have symmetry with what the `manageConfiguredInterface` method does?

```suggestion
            try {
                manageNonConfiguredInterface(device);
```

MMaiero:
@mattdibi Are additional changes needed or te PR can be merged?

mattdibi:
@MMaiero requested changes still pending

MMaiero:
@salvatore-coppola Can we sort them out, so that we can proceed with the merging process? Thanks

mattdibi (approved):

PR #4482, opened 2023-03-22
fix(core.crypto): UTF characters wrongly encoded/decoded with AES

salvatore-coppola:
Different fix for the problem in #4481 but in this case we create encoders and decoders capable of raise exceptions to detect invalid UTF-8 characters and try to fallback to the default charset.

### DIFF
--- a/kura/org.eclipse.kura.core.crypto/src/main/java/org/eclipse/kura/core/crypto/CryptoServiceImpl.java
+++ b/kura/org.eclipse.kura.core.crypto/src/main/java/org/eclipse/kura/core/crypto/CryptoServiceImpl.java
@@ -22,7 +22,6 @@ import java.io.UnsupportedEncodingException;
 import java.nio.ByteBuffer;
 import java.nio.CharBuffer;
 import java.nio.charset.CharacterCodingException;
-import java.nio.charset.Charset;
 import java.nio.charset.CharsetDecoder;
 import java.nio.charset.CharsetEncoder;
 import java.nio.charset.CodingErrorAction;
@@ -74,9 +73,6 @@ public class CryptoServiceImpl implements CryptoService {
     private CharsetEncoder utf8Encoder;
     private CharsetDecoder utf8Decoder;
 
-    private CharsetEncoder platformEncoder;
-    private CharsetDecoder platformDecoder;
-
     public CryptoServiceImpl() {
         this.utf8Encoder = StandardCharsets.UTF_8.newEncoder();
         this.utf8Encoder.onMalformedInput(CodingErrorAction.REPORT);
@@ -85,14 +81,6 @@ public class CryptoServiceImpl implements CryptoService {
         this.utf8Decoder = StandardCharsets.UTF_8.newDecoder();
         this.utf8Decoder.onMalformedInput(CodingErrorAction.REPORT);
         this.utf8Decoder.onUnmappableCharacter(CodingErrorAction.REPORT);
-
-        this.platformEncoder = Charset.defaultCharset().newEncoder();
-        this.platformEncoder.onMalformedInput(CodingErrorAction.REPLACE);
-        this.platformEncoder.onUnmappableCharacter(CodingErrorAction.REPLACE);
-
-        this.platformDecoder = Charset.defaultCharset().newDecoder();
-        this.platformDecoder.onMalformedInput(CodingErrorAction.REPLACE);
-        this.platformDecoder.onUnmappableCharacter(CodingErrorAction.REPLACE);
     }
 
     public void setSystemService(SystemService systemService) {
@@ -142,7 +130,7 @@ public class CryptoServiceImpl implements CryptoService {
             byteBuffer = this.utf8Encoder.encode(CharBuffer.wrap(value));
         } catch (CharacterCodingException e) {
             // fallback for backward compatibility
-            byteBuffer = this.platformEncoder.encode(CharBuffer.wrap(value));
+            byteBuffer = ByteBuffer.wrap(new String(value).getBytes());
         }
         byte[] encodedBytes = new byte[byteBuffer.limit()];
         byteBuffer.get(encodedBytes);
@@ -156,7 +144,7 @@ public class CryptoServiceImpl implements CryptoService {
             charBuffer = this.utf8Decoder.decode(ByteBuffer.wrap(value));
         } catch (CharacterCodingException e) {
             // fallback for backward compatibility
-            charBuffer = this.platformDecoder.decode(ByteBuffer.wrap(value));
+            charBuffer = CharBuffer.wrap(new String(value).toCharArray());
         }
         char[] decodedChar = new char[charBuffer.limit()];
         charBuffer.get(decodedChar);
--- a/kura/org.eclipse.kura.core.crypto/src/main/java/org/eclipse/kura/core/crypto/CryptoServiceImpl.java
+++ b/kura/org.eclipse.kura.core.crypto/src/main/java/org/eclipse/kura/core/crypto/CryptoServiceImpl.java
@@ -70,19 +70,6 @@ public class CryptoServiceImpl implements CryptoService {
     private final SecureRandom random = new SecureRandom();
     private SystemService systemService;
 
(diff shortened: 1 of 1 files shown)


kura/org.eclipse.kura.core.crypto/src/main/java/org/eclipse/kura/core/crypto/CryptoServiceImpl.java
```diff
@@ -89,21 +120,50 @@ public char[] encryptAes(char[] value) throws KuraException {
...
-        } catch (InvalidKeyException | IllegalBlockSizeException | BadPaddingException e) {
+        } catch (InvalidKeyException | IllegalBlockSizeException | BadPaddingException | CharacterCodingException e) {
             throw new KuraException(KuraErrorCode.ENCODE_ERROR, VALUE_EXCEPTION_CAUSE);
         } catch (InvalidAlgorithmParameterException e) {
             throw new KuraException(KuraErrorCode.ENCODE_ERROR, PARAMETER_EXCEPTION_CAUSE);
         }
 
     }
 
+    private byte[] charArrayToByteArray(char[] value) throws CharacterCodingException {
```
MMaiero, line 113: Is this needed?
Can the encoding to utf8 fail due to CharacterCodingException?

kura/org.eclipse.kura.core.crypto/src/main/java/org/eclipse/kura/core/crypto/CryptoServiceImpl.java
```diff
@@ -64,6 +71,30 @@ public class CryptoServiceImpl implements CryptoService {
     private final SecureRandom random = new SecureRandom();
     private SystemService systemService;
 
+    private CharsetEncoder utf8Encoder;
```
MMaiero: Please see that CharsetEncoder is not thread-safe: https://docs.oracle.com/javase/7/docs/api/java/nio/charset/CharsetEncoder.html

kura/org.eclipse.kura.core.crypto/src/main/java/org/eclipse/kura/core/crypto/CryptoServiceImpl.java
```diff
@@ -64,6 +71,30 @@ public class CryptoServiceImpl implements CryptoService {
...
+    private CharsetDecoder utf8Decoder;
```
MMaiero: Same for decoder

kura/org.eclipse.kura.core.crypto/src/main/java/org/eclipse/kura/core/crypto/CryptoServiceImpl.java
```diff
@@ -89,21 +120,50 @@ public char[] encryptAes(char[] value) throws KuraException {
...
     }
 
+    private byte[] charArrayToByteArray(char[] value) throws CharacterCodingException {
+
+        ByteBuffer byteBuffer;
+        try {
+            byteBuffer = this.utf8Encoder.encode(CharBuffer.wrap(value));
+        } catch (CharacterCodingException e) {
+            // fallback for backward compatibility
+            byteBuffer = this.platformEncoder.encode(CharBuffer.wrap(value));
```
MMaiero: Does the behaviour here change compared to the past?
The encoder is set to CodingErrorAction.REPORT so it fails on issues on encoding.
But in the current code it acts silently replacing characters. So, to keep backward compatibility, we should use the old code

kura/org.eclipse.kura.core.crypto/src/main/java/org/eclipse/kura/core/crypto/CryptoServiceImpl.java
```diff
@@ -89,21 +120,50 @@ public char[] encryptAes(char[] value) throws KuraException {
...
+        return encodedBytes;
+    }
+
+    private char[] byteArrayToCharArray(byte[] value) throws CharacterCodingException {
+        CharBuffer charBuffer;
+        try {
+            charBuffer = this.utf8Decoder.decode(ByteBuffer.wrap(value));
+        } catch (CharacterCodingException e) {
+            // fallback for backward compatibility
+            charBuffer = this.platformDecoder.decode(ByteBuffer.wrap(value));
```
MMaiero: Same as for the previous comment:

> Does the behaviour here change compared to the past?
The encoder is set to CodingErrorAction.REPORT so it fails on issues on encoding.
But in the current code it acts silently replacing characters. So, to keep backward compatibility, we should use the old code

We should probably use the old code to keep the current behaviour

kura/org.eclipse.kura.core.crypto/src/main/java/org/eclipse/kura/core/crypto/CryptoServiceImpl.java
salvatore-coppola: the encoder has been instruced to thrown the exception in the CryptoServiceImpl contructor.

kura/org.eclipse.kura.core.crypto/src/main/java/org/eclipse/kura/core/crypto/CryptoServiceImpl.java
salvatore-coppola: i'll synchronize the methods using them.

kura/org.eclipse.kura.core.crypto/src/main/java/org/eclipse/kura/core/crypto/CryptoServiceImpl.java
salvatore-coppola: I'll change the CodingErrorAction with replace for the platform encoder/decoder

kura/org.eclipse.kura.core.crypto/src/main/java/org/eclipse/kura/core/crypto/CryptoServiceImpl.java
nicolatimeus: Is it better to make the methods synchronized or using a dedicated encoder/decoder for each encrypt/decrypt operation?

kura/org.eclipse.kura.core.crypto/src/main/java/org/eclipse/kura/core/crypto/CryptoServiceImpl.java
nicolatimeus: The Javadoc of [getBytes()](https://docs.oracle.com/javase/7/docs/api/java/lang/String.html#getBytes()) says that "The behavior of this method when this string cannot be encoded in the default charset is unspecified.", so it is not given that switching to REPLACE will bring back the old behavior.. probably we should switch back to `getBytes()` here.

kura/org.eclipse.kura.core.crypto/src/main/java/org/eclipse/kura/core/crypto/CryptoServiceImpl.java
nicolatimeus: We should probably go back to `new String(value)` also here

kura/org.eclipse.kura.core.crypto/src/main/java/org/eclipse/kura/core/crypto/CryptoServiceImpl.java
salvatore-coppola: > Is it better to make the methods synchronized or using a dedicated encoder/decoder for each encrypt/decrypt operation?

I don't know, it seems too expensive to me to create an encoder/decoder every time.

nicolatimeus: It doesn't seem that expensive looking at openjdk impl https://github.com/openjdk/jdk/blob/jdk8-b120/jdk/src/share/classes/sun/nio/cs/UTF_8.java#L57 https://github.com/openjdk/jdk/blob/jdk8-b120/jdk/src/share/classes/java/nio/charset/Charset-X-Coder.java.template

nicolatimeus (approved):

PR #4486, opened 2023-03-24
feat(dhcp-server): added dnsmasq support

marcellorinaldo:
This PR adds support for _dnsmasq_, for devices that do not have dhcpd (installed by _isc-dhcp-server_ or _dhcpserver_).

**Related Issue:** N/A.

**Screenshots:** N/A.

**Any side note on the changes made:**:

- Refactor of `DhcpServerManager` class to support tools that use systemd.
- UI fix on the router mode and NAT selection.

### DIFF
--- a/kura/distrib/src/main/resources/generic-aarch64/deb/control/control
+++ b/kura/distrib/src/main/resources/generic-aarch64/deb/control/control
@@ -19,7 +19,7 @@ Depends: openjdk-8-jre-headless | temurin-8-jdk | openjdk-17-jre-headless | temu
  polkit | policykit-1, ssh | openssh, openssl, busybox, openvpn,
  bluez | bluez5, bluez-hcidump | bluez5-noinst-tools,
  ntpdate, chrony, chronyc | chrony, cron | cronie,
- network-manager | networkmanager, bind9 | bind, isc-dhcp-server | dhcp-server, isc-dhcp-server | dhcp-client, iw, iptables, modemmanager,
+ network-manager | networkmanager, bind9 | bind, dnsmasq | isc-dhcp-server | dhcp-server, dnsmasq | isc-dhcp-server | dhcp-client, iw, iptables, modemmanager,
  logrotate,
  gpsd
 Recommends: python3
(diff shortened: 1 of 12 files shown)


marcellorinaldo:
Currently testing:

- device with _dnsmasq_, generic profile, with network manager
- device with kura networking, testing for regressions

If all manual tests pass, the review will be put as ready for review.

kura/distrib/src/main/resources/generic-aarch64/deb/control/control
```diff
@@ -19,7 +19,7 @@ Depends: openjdk-8-jre-headless | temurin-8-jdk | openjdk-17-jre-headless | temu
  polkit | policykit-1, ssh | openssh, openssl, busybox, openvpn,
  bluez | bluez5, bluez-hcidump | bluez5-noinst-tools,
  ntpdate, chrony, chronyc | chrony, cron | cronie,
- network-manager | networkmanager, bind9 | bind, isc-dhcp-server | dhcp-server, isc-dhcp-server | dhcp-client, iw, iptables, modemmanager,
+ network-manager | networkmanager, bind9 | bind, dnsmasq | isc-dhcp-server | dhcp-server, dnsmasq | isc-dhcp-server | dhcp-client, iw, iptables, modemmanager,
```
MMaiero, line 22: if we add the dnsmasq to the generic profiles, does it make sense to keep the old isc-dhcp-server and dhcp-server code?

marcellorinaldo: Not really, but that way we are excluding devices that have isc-dhcp-server but not dnsmasq. I don't know if any of the supported devices fall in such case.

MMaiero:
https://sonarcloud.io/project/issues?resolved=false&severities=BLOCKER%2CCRITICAL%2CMAJOR%2CMINOR&sinceLeakPeriod=true&types=BUG&branch=PR-4486&id=org.eclipse.kura%3Akura&open=AYcUZ1UCebN_proxRZ7U

https://sonarcloud.io/project/issues?resolved=false&types=CODE_SMELL&sinceLeakPeriod=true&branch=PR-4486&id=org.eclipse.kura%3Akura

marcellorinaldo:
Current problems:

- DNS is passed is passed correctly to the connecting clients, but it works only when it comes from `eth0`;
- NAT and forwarding rules are not added by `NetworkManager`.

Those issues do not affect only `dnsmasq`, but are more general. Adding the missing firewall rules does not solve the issue 1. There might be some other stuff going on in `NetworkManager` under the hoods.

MMaiero:
Forcing merge to be feature complete for the next steps.
@marcellorinaldo will create a new PR with the additional tests and further fixes as long as the bugs are found during the QA process that we are about to start

PR #4487, opened 2023-03-24
fix: Fixed polkit rules

salvatore-coppola:
- Fixed ModemManager action name.
- Added kura-nm.rules for network profiles for the new polkit versions.
- Formatted code replacing tab with spaces


MMaiero:
@salvatore-coppola I have made a change to apply the kura-nm file in polkit only if NN == No. Please have a look if that make sense also to you.

pierantoniomerlino (approved):

PR #4490, opened 2023-03-27
fix(distrib): manage netplan configuration on installation

mattdibi:
Upon installing Kura on Ubuntu Server 20.04 devices, the following exception was thrown:

```
Connection 'kura-enp3s0-connection' is not available on device enp3s0 because device is strictly unmanaged
```

This was due to the fact that for OSes using Netplan, the default network renderer was set to `networkd`. This PR changes the Netplan configuration so that NetworkManager is set as the default network renderer, allowing it to manage all the network interfaces.

~@MMaiero @marcellorinaldo This PR has only one issue: after configuring Netplan we perform a `systemctl reset NetworkManager`. This triggers the renewal of the DHCP lease and the IP changes. This means that the installation process looks like got stuck but in reality it **succeeds.**~ Updated: no longer an issue.

⚠️ **Note**: we rely on the user performing the reboot for all changes to be applied.

### DIFF
--- a/kura/distrib/src/main/resources/generic-aarch64/kura_install.sh
+++ b/kura/distrib/src/main/resources/generic-aarch64/kura_install.sh
@@ -85,13 +85,26 @@ sed -i "s|/bin/sh KURA_DIR|/bin/bash ${INSTALL_DIR}/kura|" /lib/systemd/system/f
 systemctl daemon-reload
 systemctl enable firewall
 
-# disables cloud-init if exists and allows interface management to network-manager
+# disables cloud-init network management if exists, sets netplan network renderer to NetworkManager allowing interface management to NetworkManager
 if [ -d /etc/cloud/cloud.cfg.d ]; then
     echo "network: {config: disabled}" | sudo tee -a /etc/cloud/cloud.cfg.d/99-disable-network-config.cfg > /dev/null
 fi
+if [ -d /etc/netplan/ ]; then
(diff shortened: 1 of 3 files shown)


mattdibi:
Tested on RPi 4 running Ubuntu Server 20.04.5 LTS

MMaiero:
I think that in case of successful Kura installation, the installer already asks for a reboot.
If the user experience is better if we don't reset network manager, I would go with that.

kura/distrib/src/main/resources/generic-aarch64/kura_install.sh
```diff
@@ -85,13 +85,26 @@ sed -i "s|/bin/sh KURA_DIR|/bin/bash ${INSTALL_DIR}/kura|" /lib/systemd/system/f
...
+    cat << EOF >> /etc/netplan/00-installer-config.yaml
+# This file describes the network interfaces available on your system
+# For more information, see netplan(5).
+network:
+  version: 2
+  renderer: NetworkManager
+EOF
+    netplan generate
+    netplan apply
+    systemctl restart NetworkManager
```
marcellorinaldo: I would remove this line, since the user is prompt to restart the gateways anyways. Also the firewall requires a restart, network-manager is not the only one requiring a restart.

marcellorinaldo (changes requested):

mattdibi:
On Ubuntu 22.04 the Netplan configuration file changed name but it works anyway.

MMaiero (approved):

PR #4492, opened 2023-03-27
fix: Improved local image existence test with exact match

MMaiero:


MMaiero:
@GregoryIvo Please have a look at this PR as well. I think we had a bad check for the image existence.

kura/org.eclipse.kura.container.orchestration.provider/src/main/java/org/eclipse/kura/container/orchestration/provider/impl/ContainerOrchestrationServiceImpl.java
```diff
@@ -736,14 +736,16 @@ private boolean doesImageExist(String imageName, String imageTag) {
...
-            if (image.getRepoTags() != null) {
-                for (String tags : image.getRepoTags()) {
-                    if (tags.contains(imageName + ":" + imageTag)) {
-                        return true;
-                    }
+            if (isNull(image.getRepoTags())) {
+                continue;
+            }
+            for (String tag : image.getRepoTags()) {
+                if (tag.equalsIgnoreCase(requiredImage)) {
```
GregoryIvo: looks good, my only concern is it appears that container tags can be case-sensitive.

- https://www.nocentino.com/posts/2019-09-03-docker-pull-is-case-sensitive/
- https://docs.docker.com/engine/reference/commandline/tag/

image

MMaiero: ok, I'll change the equalsignorecase to a plain equals

GregoryIvo (approved):
With these new changes, LGTM!

PR #4503, opened 2023-03-28
chore(protocol.can): Added libsocket-can to classpath

salvatore-coppola:
Just to remove the annoying error messages in Eclipse IDE but maybe there's a reason why it was not added..

### DIFF
--- a/kura/org.eclipse.kura.protocol.can/META-INF/MANIFEST.MF
+++ b/kura/org.eclipse.kura.protocol.can/META-INF/MANIFEST.MF
@@ -16,4 +16,5 @@ Import-Package: de.entropia.can;version="1.0",
 Export-Package: org.eclipse.kura.protocol.can;version="2.0.0"
 Service-Component: OSGI-INF/*.xml
 Bundle-ActivationPolicy: lazy
-Bundle-ClassPath: .
+Bundle-ClassPath: .,
+ lib/libsocket-can-java-1.0.0.jar


nicolatimeus:
This is probably not necessary, the bundle imports the `de.entropia.can` package, which is exported by `libsocket-can-osgi` [1]. Moreover, the `lib/libsocket-can-java-1.0.0.jar` file does not exist in the repository.
`libsocket-can-osgi` is not built by default, Kura must be built with the `can` profile for having it in the target platform.

[1] https://github.com/eclipse/kura/blob/d411154f3dd08f1b5d0e435650d06e8b06566b00/target-platform/org.eclipse.kura.libsocket-can-osgi/pom.xml#L109

salvatore-coppola:
> This is probably not necessary, the bundle imports the `de.entropia.can` package, which is exported by `libsocket-can-osgi` [1]. Moreover, the `lib/libsocket-can-java-1.0.0.jar` file does not exist in the repository. `libsocket-can-osgi` is not built by default, Kura must be built with the `can` profile for having it in the target platform.
> 
> [1]
> 
> https://github.com/eclipse/kura/blob/d411154f3dd08f1b5d0e435650d06e8b06566b00/target-platform/org.eclipse.kura.libsocket-can-osgi/pom.xml#L109

Yeah that lib is a leftover on my filesystem, btw i don't get why eclipse don't recognize the exported package by the library present in the target-platform.

PR #4504, opened 2023-03-28
fix(nm): added conversion from signal quality to signal strength

pierantoniomerlino:

This PR adds a conversion method for converting the wifi/modem signal quality (in percentage) in signal strength (in dBm).

### DIFF
--- a/kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/net/status/wifi/WifiAccessPoint.java
+++ b/kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/net/status/wifi/WifiAccessPoint.java
@@ -34,6 +34,7 @@ public class WifiAccessPoint {
     private final WifiMode mode;
     private final long maxBitrate;
     private final int signalQuality;
+    private final int signalStrength;
     private final Set<WifiSecurity> wpaSecurity;
     private final Set<WifiSecurity> rsnSecurity;
 
@@ -44,6 +45,7 @@ public class WifiAccessPoint {
         this.mode = builder.mode;
         this.maxBitrate = builder.maxBitrate;
         this.signalQuality = builder.signalQuality;
+        this.signalStrength = builder.signalStrength;
         this.wpaSecurity = builder.wpaSecurity;
         this.rsnSecurity = builder.rsnSecurity;
     }
@@ -72,6 +74,10 @@ public class WifiAccessPoint {
         return this.signalQuality;
     }
 
+    public int getSignalStrength() {
+        return this.signalStrength;
+    }
+
     public Set<WifiSecurity> getWpaSecurity() {
         return this.wpaSecurity;
     }
@@ -92,6 +98,7 @@ public class WifiAccessPoint {
         private WifiMode mode;
         private long maxBitrate;
         private int signalQuality;
+        private int signalStrength;
         private Set<WifiSecurity> wpaSecurity = Collections.emptySet();
         private Set<WifiSecurity> rsnSecurity = Collections.emptySet();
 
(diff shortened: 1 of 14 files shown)


MMaiero:
https://sonarcloud.io/project/issues?resolved=false&types=BUG&sinceLeakPeriod=true&branch=PR-4504&id=org.eclipse.kura%3Akura

mattdibi (approved):

kura/test/org.eclipse.kura.nm.test/src/test/java/org/eclipse/kura/nm/status/NMDbusConnectorSignalConverterTest.java
```diff
@@ -0,0 +1,297 @@
...
+import java.util.List;
+
+import org.junit.Test;
+import org.junit.experimental.runners.Enclosed;
+import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
+import org.junit.runners.Parameterized.Parameters;
+
+@RunWith(Enclosed.class)
+public class NMDbusConnectorSignalConverterTest {
```
mattdibi: The name of the class is misleading IMO. You're testing `NMStatusConverter` methods here.

I would call this class `NMStatusConverterTest` and it will later be used to store all the tests for the `NMStatusConverter` class.

If you want to keep these tests separate from the other I would still rename this class `NMStatusConverterSignalConverterTest`

kura/test/org.eclipse.kura.nm.test/src/test/java/org/eclipse/kura/nm/status/NMDbusConnectorSignalConverterTest.java
```diff
@@ -0,0 +1,297 @@
...
+
+    @RunWith(Parameterized.class)
+    public static class NMDbusConnectorModemSignalStrengthConverterTest {
```
mattdibi: Do we really need all these informations in this method name? Wouldn't `ModemSignalStrengthConverterTest` suffice since its parent class name is always reported?

kura/test/org.eclipse.kura.nm.test/src/test/java/org/eclipse/kura/nm/status/NMDbusConnectorSignalConverterTest.java
```diff
@@ -0,0 +1,297 @@
...
+            this.calculatedSignalStrength = NMStatusConverter.convertToModemSignalStrength(inputSignalQuality);
+        }
+
+        private void thenCalculatedModemSignalStrengthIs(int expectedSignalStrength) {
+            assertEquals(expectedSignalStrength, this.calculatedSignalStrength);
+        }
+    }
+
+    @RunWith(Parameterized.class)
+    public static class NMDbusConnectorWifiSignalStrengthConverterTest {
```
mattdibi: Same here.

Do we really need all these informations in this method name? Wouldn't `WifiSignalStrengthConverterTest` suffice since its parent class name is always reported?

kura/test/org.eclipse.kura.nm.test/src/test/java/org/eclipse/kura/nm/status/NMDbusConnectorSignalConverterTest.java
pierantoniomerlino: Yeah... I'd keep these tests separated because of the parametric testing.

mattdibi (approved):

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/net/status/wifi/WifiAccessPoint.java
```diff
@@ -28,13 +28,23 @@
 @ProviderType
 public class WifiAccessPoint {
 
+    /** The Service Set IDentifier of the WiFi network */
```
nicolatimeus: Javadocs on private fields will not be visible, please move them to the getter methods instead.

marcellorinaldo (approved):

nicolatimeus (approved):

PR #4516, opened 2023-03-31
fix(nm): avoid error when `Modem.Location` interface is missing

mattdibi:
As per [ModemManager DBus API documentation](https://www.freedesktop.org/software/ModemManager/doc/latest/ModemManager/gdbus-org.freedesktop.ModemManager1.Modem.Location.html), the `Modem.Location` interface:

>  ...allows devices to provide location information to client applications. **Not all devices can provide this information, or even if they do, they may not be able to provide it while a data session is active.**
>
> This interface will only be available once the modem is ready to be registered in the cellular network...

This meant that, it was possible to find this error during Modem configuration:

```
org.freedesktop.dbus.errors.UnknownMethod: No such interface “org.freedesktop.ModemManager1.Modem.Location” on object at path /org/freedesktop/ModemManager1/Modem/8
```

This PR add the required checks to avoid throwing errors due to the lack of the `Modem.Location` interface during Modem configuration.

### DIFF
--- a/kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/NMDbusConnector.java
+++ b/kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/NMDbusConnector.java
@@ -460,7 +460,7 @@ public class NMDbusConnector {
             currentLocationSources = MMModemLocationSource
                     .toMMModemLocationSourceFromBitMask(modemLocationProperties.Get(MM_LOCATION_BUS_NAME, "Enabled"));
         } catch (DBusExecutionException e) {
-            logger.debug("Cannot retrive Modem.Location capabilities for {}. Caused by: ",
+            logger.warn("Cannot retrive Modem.Location capabilities for {}. Caused by: ",
                     modemLocationProperties.getObjectPath(), e);
             return;
         }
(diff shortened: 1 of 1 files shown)


kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/NMDbusConnector.java
```diff
@@ -449,10 +449,22 @@ private void handleModemManagerGPSSetup(Device device, Optional<Boolean> enableG
...
+                .of(MMModemLocationSource.MM_MODEM_LOCATION_SOURCE_NONE);
+
+        try {
+            availableLocationSources = MMModemLocationSource.toMMModemLocationSourceFromBitMask(
+                    modemLocationProperties.Get(MM_LOCATION_BUS_NAME, "Capabilities"));
+            currentLocationSources = MMModemLocationSource
+                    .toMMModemLocationSourceFromBitMask(modemLocationProperties.Get(MM_LOCATION_BUS_NAME, "Enabled"));
+        } catch (DBusExecutionException e) {
+            logger.debug("Cannot retrive Modem.Location capabilities for {}. Caused by: ",
+                    modemLocationProperties.getObjectPath(), e);
```
pierantoniomerlino, line 464: Is it better to use `error` or `warning` instead of `debug` here?

mattdibi: I was in doubt about this. It can be missing because the interface is not yet ready, or because the Modem doesn't support the Location interface... it is not necessarily an error.

I can change it to `warning` if you want. I looks like it would be more consistent with the other similar log entries in the same file.

mattdibi: Fixed in: 186be0ffbe630c90deee6487d25d136fc5ff4a96

pierantoniomerlino: I supposed that this could be not an error, but I agree that a warning is more appropriate.

pierantoniomerlino (approved):

PR #4517, opened 2023-03-31
refactor(nm): NMDBusConnector class refactor

pierantoniomerlino:

This PR perform a refactor of the `NMDbusConnector` class. In particular, it adds the `apply(String deviceId)` method that is used by the `NMConfigurationEnforcementHandler` to apply the configuration to a single device/interface and it rewrites the `doApply` method in a simpler way.

### DIFF
--- a/kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/NMConfigurationEnforcementHandler.java
+++ b/kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/NMConfigurationEnforcementHandler.java
@@ -47,7 +47,7 @@ public class NMConfigurationEnforcementHandler implements DBusSigHandler<Device.
         if (deviceIsConnectingToANewNetwork || deviceDisconnectedBecauseOfConfigurationEvent) {
             try {
                 logger.info("Network change detected on interface {}. Roll-back to cached configuration", s.getPath());
-                String deviceId = this.nm.getDeviceId(s.getPath());
+                String deviceId = this.nm.getDeviceIdByDBusPath(s.getPath());
                 this.nm.apply(deviceId);
             } catch (DBusException e) {
                 logger.error("Failed to handle network configuration change event for device: {}. Caused by:",
--- a/kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/NMConfigurationEnforcementHandler.java
+++ b/kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/NMConfigurationEnforcementHandler.java
@@ -47,7 +47,8 @@ public class NMConfigurationEnforcementHandler implements DBusSigHandler<Device.
         if (deviceIsConnectingToANewNetwork || deviceDisconnectedBecauseOfConfigurationEvent) {
             try {
                 logger.info("Network change detected on interface {}. Roll-back to cached configuration", s.getPath());
-                this.nm.apply();
+                String deviceId = this.nm.getDeviceId(s.getPath());
+                this.nm.apply(deviceId);
             } catch (DBusException e) {
                 logger.error("Failed to handle network configuration change event for device: {}. Caused by:",
                         s.getPath(), e);
--- a/kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/NMDbusConnector.java
+++ b/kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/NMDbusConnector.java
@@ -184,35 +184,35 @@ public class NMDbusConnector {
             }
 
             switch (deviceType) {
-                case NM_DEVICE_TYPE_ETHERNET:
-                    Wired wiredDevice = this.dbusConnection.getRemoteObject(NM_BUS_NAME, device.get().getObjectPath(),
-                            Wired.class);
-                    Properties wiredDeviceProperties = this.dbusConnection.getRemoteObject(NM_BUS_NAME,
-                            wiredDevice.getObjectPath(), Properties.class);
+            case NM_DEVICE_TYPE_ETHERNET:
+                Wired wiredDevice = this.dbusConnection.getRemoteObject(NM_BUS_NAME, device.get().getObjectPath(),
+                        Wired.class);
+                Properties wiredDeviceProperties = this.dbusConnection.getRemoteObject(NM_BUS_NAME,
+                        wiredDevice.getObjectPath(), Properties.class);
 
-                    DevicePropertiesWrapper ethernetPropertiesWrapper = new DevicePropertiesWrapper(deviceProperties,
-                            Optional.of(wiredDeviceProperties), NMDeviceType.NM_DEVICE_TYPE_ETHERNET);
+                DevicePropertiesWrapper ethernetPropertiesWrapper = new DevicePropertiesWrapper(deviceProperties,
+                        Optional.of(wiredDeviceProperties), NMDeviceType.NM_DEVICE_TYPE_ETHERNET);
 
-                    networkInterfaceStatus = NMStatusConverter.buildEthernetStatus(interfaceId,
-                            ethernetPropertiesWrapper, ip4configProperties);
-                    break;
-                case NM_DEVICE_TYPE_LOOPBACK:
-                    DevicePropertiesWrapper loopbackPropertiesWrapper = new DevicePropertiesWrapper(deviceProperties,
-                            Optional.empty(), NMDeviceType.NM_DEVICE_TYPE_LOOPBACK);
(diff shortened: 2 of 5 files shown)


kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/NMDbusConnector.java
```diff
@@ -274,53 +274,64 @@ public synchronized void apply() throws DBusException {
...
+                    } catch (DBusException | DBusExecutionException | IllegalArgumentException
+                            | NoSuchElementException e) {
+                        logger.error("Unable to configure device {}, skipping", deviceId, e);
+                    }
+                }
+            });
         } finally {
             configurationEnforcementEnable();
         }
     }
```
mattdibi, line 325: This is way too unnecessarily complicated. I propose the following:

```suggestion
            List<String> configuredInterfaceIds = properties.getStringList("net.interfaces");

            List<Device> availableDevices = getAllDevices();
            availableDevices.forEach(device -> {
                String deviceId = getDeviceId(device.getObjectPath());
                if (shouldBeConfigured(deviceId, deviceIdToBeConfigured)) {
                    if (configuredInterfaceIds.contains(deviceId)) {
                        manageConfiguredInterface(device, deviceId, properties);
                    } else {
                        manageNonConfiguredInterface(device, deviceId);
                    }
                }
            });
        } catch (DBusException | DBusExecutionException | IllegalArgumentException
                | NoSuchElementException e) {
            logger.error("Unable to configure device {}, skipping", deviceId, e); // If you don't have the deviceId you can use the DBus path here
        } finally {
            configurationEnforcementEnable();
        }
    }
```

kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/NMDbusConnector.java
```diff
@@ -470,23 +465,23 @@ private void handleModemManagerGPSSetup(Device device, Optional<Boolean> enableG
         modemLocation.Setup(MMModemLocationSource.toBitMaskFromMMModemLocationSource(currentLocationSources), false);
     }
 
-    private String getDeviceId(Device device) throws DBusException {
-        NMDeviceType deviceType = getDeviceType(device);
+    protected String getDeviceId(String devicePath) throws DBusException {
```
mattdibi: Which device path? OS path? DBus path? If it's dbus path is it for the NM Device path or for the MM Modem path?

```suggestion
    protected String getDeviceId(String nmDeviceDbusPath) throws DBusException {
```

kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/NMDbusConnector.java
```diff
@@ -555,16 +550,16 @@ private Boolean isDeviceManaged(Device device) throws DBusException {
         return deviceProperties.Get(NM_DEVICE_BUS_NAME, NM_DEVICE_PROPERTY_MANAGED);
     }
 
-    private NMDeviceType getDeviceType(Device device) throws DBusException {
-        Properties deviceProperties = this.dbusConnection.getRemoteObject(NM_BUS_NAME, device.getObjectPath(),
+    private NMDeviceType getDeviceType(String devicePath) throws DBusException {
```
mattdibi: Same thing here:

```suggestion
    private NMDeviceType getDeviceType(String nmDeviceDbusPath) throws DBusException {
```

kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/NMConfigurationEnforcementHandler.java
```diff
@@ -47,7 +47,8 @@ public void handle(Device.StateChanged s) {
         if (deviceIsConnectingToANewNetwork || deviceDisconnectedBecauseOfConfigurationEvent) {
             try {
                 logger.info("Network change detected on interface {}. Roll-back to cached configuration", s.getPath());
-                this.nm.apply();
+                String deviceId = this.nm.getDeviceIdByDBusPath(s.getPath());
```
mattdibi, line 50:
```suggestion
                String deviceId = this.nm.getDeviceIdByDBusPath(s.getDevicePath());
```

kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/NMDeviceAddedHandler.java
```diff
@@ -33,7 +33,8 @@ public NMDeviceAddedHandler(NMDbusConnector connector) {
     public void handle(NetworkManager.DeviceAdded s) {
         try {
             logger.info("New network device connected at {}", s.getDevicePath());
-            this.nm.apply();
+            String deviceId = this.nm.getDeviceIdByDBusPath(s.getPath());
```
mattdibi:
```suggestion
            String deviceId = this.nm.getDeviceIdByDBusPath(s.getDevicePath().toString());
```

kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/NMDbusConnector.java
```diff
@@ -269,63 +269,76 @@ ip4configProperties, new AccessPointsProperties(activeAccessPoint, accessPoints)
     }
 
     public synchronized void apply(Map<String, Object> networkConfiguration) throws DBusException {
-        doApply(networkConfiguration);
-        this.cachedConfiguration = networkConfiguration;
+        try {
+            configurationEnforcementDisable();
+            doApply(networkConfiguration);
```
mattdibi: You need to clear the modem reset timers here.

```suggestion
            configurationEnforcementDisable();
            modemResetHandlersDisable();
            doApply(networkConfiguration);
```

kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/NMDbusConnector.java
```diff
@@ -269,63 +269,76 @@ ip4configProperties, new AccessPointsProperties(activeAccessPoint, accessPoints)
...
+        if (Objects.isNull(deviceId) || deviceId.isEmpty()) {
+            throw new IllegalArgumentException("DeviceId cannot be null or empty.");
+        }
+        if (Objects.isNull(this.cachedConfiguration)) {
+            logger.warn("No cached network configuration found.");
+            return;
+        }
+        try {
+            configurationEnforcementDisable();
+            doApply(deviceId, this.cachedConfiguration);
```
mattdibi: You need to clear possible pending reset timers here:

```suggestion
            configurationEnforcementDisable();
            modemResetHandlersDisable(deviceId);
            doApply(deviceId, this.cachedConfiguration);
```

Currently pending timers are tracked in a `List<NMModemResetHandler>`. I think you just need to move to a `Map<String, NMModemResetHandler>` so that you can keep track of the available reset handlers and remove only the ones you need

kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/NMDbusConnector.java
pierantoniomerlino: Ahhh... maybe I miss it during the rebase.

kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/NMDeviceAddedHandler.java
pierantoniomerlino: Is`s.getDevicePath().getPath()` better than the `toString()` method?

mattdibi:
For documentation purpose I will report here my findings on a new behaviour with this PR.

I performed a couple of tests using a Telit HE910 modem. When connected this device creates an additional NetworkManager device of type PPP. 

image

image


When a new configuration is issued (for example disconnecting the modem) the following stacktrace appears in the log:

MicrosoftTeams-image (9)

This is due to the fact that the PPP device (which is created upon the modem connecting), disappears and its dbus path invalidated. Since we're iterating over devices here:

https://github.com/eclipse/kura/blob/2cdcc7677f408857165ff164a60c784805639d91/kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/NMDbusConnector.java#L316-L324

when we get to the PPP device, it doesn't exist anymore and the DBus path is invalidated. This is all correctly handled by the code (the exception is handled and all remaining operation work as expected). No code chages are required here IMO.

mattdibi (approved):
LGTM

PR #4518, opened 2023-03-31
fix: Allowed loading certificate chain in private key entry

nicolatimeus:

### DIFF
--- a/kura/org.eclipse.kura.core.keystore/src/main/java/org/eclipse/kura/core/keystore/util/KeystoreRemoteService.java
+++ b/kura/org.eclipse.kura.core.keystore/src/main/java/org/eclipse/kura/core/keystore/util/KeystoreRemoteService.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2021 Eurotech and/or its affiliates and others
+ * Copyright (c) 2021, 2023 Eurotech and/or its affiliates and others
  *
  * This program and the accompanying materials are made
  * available under the terms of the Eclipse Public License 2.0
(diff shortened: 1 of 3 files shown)


MMaiero:
@nicolatimeus Are there changes needed in the REST APIs?

nicolatimeus:
As far as I could see, the current rest APIs and request hander do not allow to modify a `PrivateKeyEntry` in any way by design. They only allow to create a PrivateKeyEntry by generating its private key and certificate on the device basing on some parameters provided by the user [1] [2] [3]. The generated `PrivateKeyEntry` will always have a single certificate in its chain. In fact it is not possible to create a `PrivateKeyEntry` with a certificate chain with length > 1 in any way. `TrustedCertificateEntries` can have only a single certificate associated with it.

In my opinion there are no bugfix level changes needed to them. We may consider if we want to change their behavior to allow modifying the certificate chain associated with a `PrivateKeyEntry`. Without this functionality, the workflow of creating a key pair on the device, get the CSR, sign it with an external CA and copy back the signed certificate chain to the device may not be possible using only the REST APIs/request handler, since there is no way to associate the signed chain with the private key on the device.

[1] https://github.com/eclipse/kura/blob/b759503eee453eea9774658aa7bcde24c4b3e6cb/kura/org.eclipse.kura.core.keystore/src/main/java/org/eclipse/kura/core/keystore/rest/provider/KeystoreRestService.java#L108
[2] https://github.com/eclipse/kura/blob/b759503eee453eea9774658aa7bcde24c4b3e6cb/kura/org.eclipse.kura.core.keystore/src/main/java/org/eclipse/kura/core/keystore/request/handler/KeystoreServiceRequestHandlerV1.java#L151
[3] https://github.com/eclipse/kura/blob/b759503eee453eea9774658aa7bcde24c4b3e6cb/kura/org.eclipse.kura.core.keystore/src/main/java/org/eclipse/kura/core/keystore/util/KeyPairInfo.java

MMaiero (approved):

PR #4520, opened 2023-04-03
test(nm.configuration.writer): refactor and added tests dnsmasq cases for DhcpServerConfigWriter

marcellorinaldo:
This PR adds tests cases for testing `DhcpServerConfigWriter` in case `dnsmasq` is used. Moreover, a refactor of the testing class is made to allow more parametrization of test cases and usage of Junit's temporary files.

**Related Issue:** N/A.

**Screenshots:** N/A.


kura/test/org.eclipse.kura.nm.test/src/test/java/org/eclipse/kura/nm/configuration/writer/DhcpServerConfigWriterTest.java
```diff
@@ -347,14 +424,32 @@ private void thenKuraExceptionIsCaught() {
...
     @After
-    public void clean() {
-        File dhcpConfigFile = new File("/tmp/dhcpd-eth0.conf");
-        if (dhcpConfigFile.exists()) {
-            dhcpConfigFile.delete();
+    public void cleanUp() throws IOException {
+        dhcpServerMock.close();
+        this.mockFiles.delete();
+        this.mockFiles.create();
+    }
```
mattdibi: Given how JUnit works I don't think you need these....

JUnit should clean up all the `mockFiles` on its own since it's a TemporaryFolder. I'm not so sure about the `dhcpServerMock` since it's `static` though

kura/test/org.eclipse.kura.nm.test/src/test/java/org/eclipse/kura/nm/configuration/writer/DhcpServerConfigWriterTest.java
```diff
@@ -347,14 +424,32 @@ private void thenKuraExceptionIsCaught() {
...
-        if (dhcpServerMock != null) {
-            dhcpServerMock.close();
+
+        public PropertiesBuilder with(String name, Object value) {
+            this.properties.put(name, value);
+            return this;
+        }
+
+        public void build() {
+            DhcpServerConfigWriterTest.this.networkProperties = this.properties;
```
mattdibi: Even though I really like the use of the builder pattern here, I don't like this. 

I find somewhat surprising that a `build()` method can have side effects on its wrapping class. Is there a better way?

kura/test/org.eclipse.kura.nm.test/src/test/java/org/eclipse/kura/nm/configuration/writer/DhcpServerConfigWriterTest.java
```diff
@@ -347,14 +424,32 @@ private void thenKuraExceptionIsCaught() {
...
-        File dhcpConfigFile = new File("/tmp/dhcpd-eth0.conf");
-        if (dhcpConfigFile.exists()) {
-            dhcpConfigFile.delete();
+    public void cleanUp() throws IOException {
+        dhcpServerMock.close();
+        this.mockFiles.delete();
+        this.mockFiles.create();
+    }
+
+    private class PropertiesBuilder {
```
mattdibi: This next comment ties together with https://github.com/eclipse/kura/pull/4520/files#r1156204051.

Again: I really like the builder pattern but... is it really necessary here? Couldn't a

```java
private void givenNetworkPropertiesWith(String key, Object value) {
   this.map.put(key, value);
}
```

method suffice and obtain the overall same results without the need of an additional class?

Keep in mind that this is how it was already implemented in all other tests using `NetworkProperties`. See:

https://github.com/eclipse/kura/blob/b01a8133d4d34ee2533b7efa0bd82825bfa2a89e/kura/test/org.eclipse.kura.nm.test/src/test/java/org/eclipse/kura/nm/configuration/NMSettingsConverterTest.java#L96-103

https://github.com/eclipse/kura/blob/b01a8133d4d34ee2533b7efa0bd82825bfa2a89e/kura/test/org.eclipse.kura.nm.test/src/test/java/org/eclipse/kura/nm/NetworkPropertiesTest.java#L72-74

mattdibi (approved):
LGTM.

Take a look at the comments I added and feel free to ignore them if you don't have the time or the will (or both!) for implementing them.

kura/test/org.eclipse.kura.nm.test/src/test/java/org/eclipse/kura/nm/configuration/writer/DhcpServerConfigWriterTest.java
marcellorinaldo: The static mock needs the close here. For the tmp files I will remove those lines.

kura/test/org.eclipse.kura.nm.test/src/test/java/org/eclipse/kura/nm/configuration/writer/DhcpServerConfigWriterTest.java
marcellorinaldo: Yes it might be more intuitive this way. I wanted to try something new but it seems to be not the right choice.

marcellorinaldo:
> LGTM.
> 
> Take a look at the comments I added and feel free to ignore them if you don't have the time or the will (or both!) for implementing them.

No time but will for sure!

mattdibi (approved):

PR #4529, opened 2023-04-05
feat(dnsmasq): support for multiple DHCP server interfaces

marcellorinaldo:
This PR writes the `dnsmasq` configuration so that each property has associated a connection label, allowing for multiple interfaces to be set as DHCP server at the same time.

**Related Issue:** N/A.

**Description of the solution adopted:** The `dhcp-range` property creates a label for each activated interface. Each DHCP interface configuration is put on a file that follows the pattern:

```
/etc/dnsmasq.d/dnsmasq-<ifname>.conf
```

It is not allowed in dnsmasq to have global settings specified more than once across all files. Hence, a global configuration is written and maintained by the DHCP server manager. The created global configuration file is:

```
/etc/dnsmasq.d/dnsmasq-globals.conf
```

**Screenshots:** N/A.

**Any side note on the changes made:** Created more robust way to detect change in configuration in the method `isRunning`. It wouldn't trigger on NM configuration update.


kura/org.eclipse.kura.linux.net/src/main/java/org/eclipse/kura/linux/net/dhcp/server/DnsmasqTool.java
```diff
@@ -80,8 +92,31 @@ public boolean disableInterface(String interfaceName) throws KuraProcessExecutio
         }
```
pierantoniomerlino, line 91: Do you prefer to use the `KuraProcessExecutionErrorException(Throwable cause, Object argument)` signature instead?

kura/org.eclipse.kura.linux.net/src/main/java/org/eclipse/kura/linux/net/dhcp/server/DnsmasqTool.java
```diff
@@ -42,17 +49,22 @@ public DnsmasqTool(CommandExecutorService service) {
     public boolean isRunning(String interfaceName) {
         CommandStatus status = this.executorService.execute(
                 new Command(new String[] { "systemctl", "is-active", "--quiet", DhcpServerTool.DNSMASQ.getValue() }));
-        boolean isRunning = status.getExitStatus().isSuccessful() && this.activeInterfaces.contains(interfaceName);
 
-        logger.debug("DNSMASQ - Is dnsmasq running for interface {}? {}", interfaceName, isRunning);
+        boolean isRunning = status.getExitStatus().isSuccessful() && !isConfigFileAlteredOrNonExistent(interfaceName);
+
```
pierantoniomerlino, line 53: Does it make sense to return false even if the dnsmasq is running, but the configuration file is altered? It depends, of course, on how this method is used...

kura/test/org.eclipse.kura.linux.net.test/META-INF/MANIFEST.MF
```diff
@@ -19,9 +19,10 @@ Import-Package: org.eclipse.kura;version="[1.0,2.0)",
  org.eclipse.kura.net.wifi;version="[2.0,3.0)",
  org.eclipse.kura.system;version="[1.0,2.0)",
  org.eclipse.kura.usb;version="[1.0,2.0)",
+ org.junit;version="[4.12,5.0)",
  org.mockito;version="[4.0.0,5.0.0)",
  org.mockito.invocation;version="[4.0.0,5.0.0)",
- org.mockito.mock;version="1.10.19",
+ org.mockito.mock;version="[4.0.0,5.0.0)",
  org.mockito.stubbing;version="[4.0.0,5.0.0)",
  org.osgi.framework;version="1.4.0",
```
pierantoniomerlino, line 27: Please, remove the micro from the version: [4.0,5.0) is correct here.

kura/org.eclipse.kura.linux.net/src/main/java/org/eclipse/kura/linux/net/dhcp/server/DnsmasqTool.java
marcellorinaldo: The `isRunning` method is used by the monitor to check whether we have running the updated configuration. Maybe the name `isUpdatedAndRunning` is more appropriate but I wouldn't change it in this PR.

kura/org.eclipse.kura.linux.net/src/main/java/org/eclipse/kura/linux/net/dhcp/server/DnsmasqTool.java
marcellorinaldo: Yes, maybe that is more informative.

marcellorinaldo:
Tests missing, will add in different PR

pierantoniomerlino (approved):

MMaiero:
@marcellorinaldo The tests are failing. Can you please check?

marcellorinaldo:
> @marcellorinaldo The tests are failing. Can you please check?

Forgot to update couple of test cases.

marcellorinaldo:

MMaiero:
Merging due to the need to build RC2. @marcellorinaldo please add coverage as required https://sonarcloud.io/component_measures?id=org.eclipse.kura%3Akura&branch=PR-4529&metric=new_coverage&view=list

PR #4532, opened 2023-04-05
fix: Fixed NetworkConfigurationService returning passwords as strings

nicolatimeus:

The NetworkConfigurationService is currently returning password fields as strings instead of Password objects for the configuration of the network interfaces that are present in the snapshot and not currently attached to the system.

This adds a check before its configuration is returned to wrap any string password property into a Password object.


marcellorinaldo (approved):

MMaiero:
Sonar reports issues, can you please check? Thanks

PR #4535, opened 2023-04-05
refactor(nm): `org.eclipse.kura.nm` log cleanup/update

mattdibi:
This PR:

- removes two logs in the `enableInterface` method that were not carrying much useful informations and cluttering the log
- lower the log entry for the `getAssociatedConnections` method from `info` to `debug`
- adds a log entry for when the scheduled timer is cancelled due to a `clearTimer` method call. This allows for a better understanding of the overall timer status on the system and is consistent with the rest of the code.

image

This is the resulting configuration log:

image


pierantoniomerlino (approved):

pierantoniomerlino:
Do you prefer the info or debug level here?

mattdibi:
@pierantoniomerlino `info` since this should be at the same level as the timer scheduling and the other log entry where the timer is cancelled.  I think all of these are pretty important and are worth enough to be on the `info` level.

pierantoniomerlino (approved):

MMaiero:
Tests seem to fail. Expected?

mattdibi:
> Tests seem to fail. Expected?

Uh... quite the opposite actually. Will look into it.

mattdibi:

PR #4544, opened 2023-04-07
fix(nm.test): Fixed nm tests.

salvatore-coppola:
- Added CommandExecutorServiceMock
- Added mockito-inline dependency.


marcellorinaldo (commented):
Does it make sense to have the dependency available to all tests?

Moreover, if adding such dependency is the file in `test/resources/mockito-extensions` still needed?

mattdibi (approved):

mattdibi:
@marcellorinaldo due to current timing constraints, I would look into your suggestions on another time and merge this as-is (there are a lot of PRs that depends on this one ATM). What do you think?

PR #4545, opened 2023-04-07
fix(web2.server.net2): wrong enum default leaves wireless interface as station

marcellorinaldo:
This PR fixes the default case of an enum that sets the wireless interface as station while it should be disabled. This causes the client part of the UI to set a wrong validation state to the password property when the interface is disabled making it impossible to switch to AP mode.

**Related Issue:** N/A.

**Screenshots:** N/A.


marcellorinaldo:
@MMaiero can you test on a device with two wireless interfaces?

MMaiero:
Yes, let me finish with the test for the dyna serial and I'll jump on this

MMaiero:
Seems to be working for me

MMaiero (approved):

PR #4548, opened 2023-04-07
fix(installer): Added ModemManager enable command in install script

pierantoniomerlino:

This PR adds the enable and start command of the ModemManager service in the install scripts for the generic profiles.

### DIFF
--- a/kura/distrib/src/main/resources/generic-aarch64/kura_install.sh
+++ b/kura/distrib/src/main/resources/generic-aarch64/kura_install.sh
@@ -61,7 +61,6 @@ systemctl disable chrony
 systemctl enable NetworkManager
 systemctl start NetworkManager
 systemctl enable ModemManager
-systemctl start ModemManager
 
 # set up users and grant permissions
 cp ${INSTALL_DIR}/kura/install/manage_kura_users.sh ${INSTALL_DIR}/kura/.data/manage_kura_users.sh
--- a/kura/distrib/src/main/resources/generic-aarch64/kura_install.sh
+++ b/kura/distrib/src/main/resources/generic-aarch64/kura_install.sh
@@ -60,6 +60,8 @@ systemctl stop chrony
 systemctl disable chrony
 systemctl enable NetworkManager
 systemctl start NetworkManager
+systemctl enable ModemManager
+systemctl start ModemManager
 
 # set up users and grant permissions
 cp ${INSTALL_DIR}/kura/install/manage_kura_users.sh ${INSTALL_DIR}/kura/.data/manage_kura_users.sh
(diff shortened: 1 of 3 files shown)


marcellorinaldo (commented):
Isn't it sufficient to just enable it? The device needs a reboot anyways

pierantoniomerlino:
I followed the same convention used for NetworkManager: enable and start. But, yes, I agree that it is sufficient to enable the services, since a reboot is required (but not forced). However, this should be done for all the services, not only MM.

MMaiero:
@pierantoniomerlino I think you have placed the wrong back port label. I have changed it since it is a code contribution and not a docs contribution

marcellorinaldo:
> I followed the same convention used for NetworkManager: enable and start. But, yes, I agree that it is sufficient to enable the services, since a reboot is required (but not forced). However, this should be done for all the services, not only MM.

NM is started just because the installer needs it for determining the main net interface. We should keep the services as disabled when possible since it is not given that prior kura start they have the correct configuration.

marcellorinaldo:
Moreover, like `wpa_supplicant`, `ModemManager` seems to be started as `NetworkManager` starts. Therefore, this PR might not be needed.

pierantoniomerlino:
@marcellorinaldo which device you used for the tests? I checked on a RPi3 and NM doesn't start MM: the modem is visible on the system as a USB device, but Kura cannot see it. After the installation and reboot, MM is not started.

BTW, to perform the test I disabled MM before installing Kura.

marcellorinaldo:
> @marcellorinaldo which device you used for the tests? I checked on a RPi3 and NM doesn't start MM: the modem is visible on the system as a USB device, but Kura cannot see it. After the installation and reboot, MM is not started.
> 
> BTW, to perform the test I disabled MM before installing Kura.

Is it the same also when disabling wpa_supplicant on a fresh image? In that case, shall we take into account all the cases in which a user disables some services before installing kura?

pierantoniomerlino:
IMHO, Kura installer should enable all the services that are needed for its correct operation. I don't know if it is the same for `wpa_supplicant`. Is it a service or is it started by NM when needed?

marcellorinaldo:
> IMHO, Kura installer should enable all the services that are needed for its correct operation. I don't know if it is the same for `wpa_supplicant`. Is it a service or is it started by NM when needed?

I think it is used by network manager like modem manager. To be verified.

pierantoniomerlino:
It seems that `wpa_supplicant` and `dnsmasq` are started by NM (but the service is not enabled). MM and `systemd-resolved` don't.

marcellorinaldo:
> It seems that `wpa_supplicant` and `dnsmasq` are started by NM (but the service is not enabled). MM and `systemd-resolved` don't.

So I guess it is better that we enable it.

marcellorinaldo (approved):

PR #4568, opened 2023-04-12
fix(installer): Added dnsmasq default configuration file

pierantoniomerlino:

This PR adds the `dnsmasq` default configuration file. The Kura installer will copy it in the `/etc/default/` folder.

### DIFF
--- a/kura/distrib/src/main/resources/common/dnsmasq
+++ b/kura/distrib/src/main/resources/common/dnsmasq
@@ -1,3 +1,3 @@
 ENABLED=1
-CONFIG_DIR=/etc/dnsmasq.d,.dpkg-dist,.dpkg-old,.dpkg-new
+CONFIG_DIR=/etc/dnsmasq.d,*.conf
 DNSMASQ_EXCEPT="lo"
--- a/kura/distrib/src/main/resources/common/dnsmasq
+++ b/kura/distrib/src/main/resources/common/dnsmasq
@@ -1,0 +1,3 @@
+ENABLED=1
+CONFIG_DIR=/etc/dnsmasq.d,.dpkg-dist,.dpkg-old,.dpkg-new
+DNSMASQ_EXCEPT="lo"
--- a/kura/distrib/src/main/resources/generic-aarch64/kura_install.sh
+++ b/kura/distrib/src/main/resources/generic-aarch64/kura_install.sh
@@ -103,6 +103,12 @@ if [ -d /usr/lib/NetworkManager/conf.d/ ]; then
     fi
 fi
 
+# install dnsmasq default configuration
+if [ -f /etc/default/dnsmasq ]; then
+    mv /etc/default/dnsmasq /etc/default/dnsmasq.old
+fi
+cp ${INSTALL_DIR}/kura/install/dnsmasq /etc/default/dnsmasq
+
 # disable NTP service
 if command -v timedatectl > /dev/null ;
   then
--- a/kura/distrib/src/main/ant/build_equinox_distrib.xml
+++ b/kura/distrib/src/main/ant/build_equinox_distrib.xml
@@ -612,6 +612,9 @@ fi]]>
             <zipfileset file="src/main/resources/common/ifup-local.raspbian"
                 prefix="${build.output.name}/${install.folder}" />
 
+            <zipfileset file="src/main/resources/common/dnsmasq"
+                prefix="${build.output.name}/${install.folder}" />
+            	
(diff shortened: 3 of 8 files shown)


kura/distrib/src/main/resources/generic-aarch64/kura_install.sh
```diff
@@ -103,6 +103,12 @@ if [ -d /usr/lib/NetworkManager/conf.d/ ]; then
     fi
 fi
 
+# install dnsmasq default configuration
+if [ -f /etc/default/dnsmasq ]; then
```
marcellorinaldo, line 107: We should also check that we currently have `dnsmasq` installed and perform this copy only if needed (we might have installed the alternative `isc-dhcp-server`).

kura/distrib/src/main/resources/common/dnsmasq
```diff
@@ -0,0 +1,3 @@
+ENABLED=1
+CONFIG_DIR=/etc/dnsmasq.d,.dpkg-dist,.dpkg-old,.dpkg-new
```
marcellorinaldo: Shall we include `.old` exclusion here? We might risk to have the `/etc/default/dnsmasq.old` overriding ours.

marcellorinaldo (changes requested):

kura/distrib/src/main/resources/generic-aarch64/kura_install.sh
pierantoniomerlino: We can decide how to proceed here. This is a lazy approach: copy the file even if the package is not installed on the system. In that case the presence of the config file is harmless. Otherwise, we have to check if dnsmasq is installed (is the package name different according to the OSs?), and copy the file only if needed.
What do you prefer?

kura/distrib/src/main/resources/common/dnsmasq
pierantoniomerlino: Do you mean `dnsmasq.old` in the same folder of `dnsmasq`? I checked on my RPi and it seems to work. We can copy the backup file, but where?

kura/distrib/src/main/resources/generic-aarch64/kura_install.sh
marcellorinaldo: Maybe the lazy approach is the best

kura/distrib/src/main/resources/common/dnsmasq
marcellorinaldo: Wait I made some confusion here. This directive tells `dnsmasq` to accept as config files in `/etc/dnsmasq.d` that do not end with `.dpkg-dist`, `.dpkg-old`, or `.dpkg-new`? Does it make sense to change it to accept only `.conf`? I think we can do it with `CONFIG_DIR=/etc/dnsmasq.d,*.conf`

marcellorinaldo (changes requested):

kura/distrib/src/main/resources/common/dnsmasq
pierantoniomerlino: The documentation is:

> By default search this drop directory for configuration options.
> Libvirt leaves a file here to make the system dnsmasq play nice.
> Comment out this line if you don't want this. The dpkg-* are file
> endings which cause dnsmasq to skip that file. This avoids pulling
> in backups made by dpkg.
> CONFIG_DIR=/etc/dnsmasq.d,.dpkg-dist,.dpkg-old,.dpkg-new

This is the default configuration in my RPi.

marcellorinaldo: > -7, --conf-dir=<directory>[,<file-extension>......],
Read all the files in the given directory as configuration files. If extension(s) are given, any files which end in those extensions are skipped. Any files whose names end in ~ or start with . or start and end with # are always skipped. If the extension starts with * then only files which have that extension are loaded. So --conf-dir=/path/to/dir,*.conf loads all files with the suffix .conf in /path/to/dir. This flag may be given on the command line or in a configuration file. If giving it on the command line, be sure to escape * characters. Files are loaded in alphabetical order of filename.

marcellorinaldo: - `/dir/path,*.conf` matches all and only `.conf` files.
- `/dir/path,.conf` matches all except `.conf` files.

marcellorinaldo (approved):
Seems to be working on Raspberry Pi 4, Raspbian 11.

PR #4584, opened 2023-04-20
fix: check if schedule is enabled before parsing

GregoryIvo:
a minor change suggested by @marcellorinaldo  that prevents the parsing of the connection schedule cron string unless the feature is enabled. With this change, if you have a bad string, the error will not show when you are not using the connection schedule.

### DIFF
--- a/kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/data/DataServiceImpl.java
+++ b/kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/data/DataServiceImpl.java
@@ -750,7 +750,11 @@ public class DataServiceImpl implements DataService, DataTransportListener, Conf
             return;
         }
 
-        final Optional<CronExpression> schedule = this.dataServiceOptions.getConnectionScheduleExpression();
+        Optional<CronExpression> schedule = Optional.empty();
+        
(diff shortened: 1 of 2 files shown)


GregoryIvo:
@marcellorinaldo Kindly take a look at this when you get a chance!

kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/data/DataServiceImpl.java
```diff
@@ -750,7 +750,11 @@ private void createAutoConnectStrategy() {
             return;
         }
 
-        final Optional<CronExpression> schedule = this.dataServiceOptions.getConnectionScheduleExpression();
+        Optional<CronExpression> schedule = Optional.empty();
+        
+        if (this.dataServiceOptions.isConnectionScheduleEnabled()) {
```
marcellorinaldo: I would not do the check here but rather in the options class since it is already returning an Optional: https://github.com/eclipse/kura/blob/31ef269ef081bedd72772b8508dcc98cc960ba37/kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/data/DataServiceOptions.java#L183

marcellorinaldo (changes requested):

marcellorinaldo (approved):

PR #4585, opened 2023-04-20
fix(installer): comment configured interfaces in interfaces file

pierantoniomerlino:

This PR adds a (python) script for commenting the configuration of some network interfaces in the `/etc/network/interfaces/` file. This is done only for generic profiles where the NM can conflict with the `ifupdown` scritps.

### DIFF
--- a/kura/distrib/src/main/resources/common/comment-interfaces-file.py
+++ b/kura/distrib/src/main/resources/common/comment-interfaces-file.py
@@ -14,6 +14,7 @@
 import subprocess
 import sys
 import shutil
+import logging
 from os.path import exists
 
 ETHERNET_TYPES = ['ethernet', 'eth', 'wired']
@@ -21,15 +22,6 @@ WIRELESS_TYPES = ['wifi', 'wireless']
 INTERFACES_PATH = "/etc/network/interfaces"
 INTERFACES_OLD_PATH = INTERFACES_PATH + ".old"
 INTERFACES_TMP_PATH = INTERFACES_PATH + ".tmp"
-
-logging.basicConfig(
-    format='[comment_interfaces_file.py] %(asctime)s %(levelname)s %(message)s',
-    level=logging.INFO,
-    datefmt='%Y-%m-%d %H:%M:%S',
-    handlers=[
-        logging.StreamHandler()
-    ]
-)
 
 def get_eth_wlan_interfaces_names():
     """Reads the network interface names using 'nmcli dev' command.
@@ -81,6 +73,15 @@ def comment_paragraph(paragraph):
     return (commented)
 
 def main():
+    logging.basicConfig(
+        format='[comment_interfaces_file.py] %(asctime)s %(levelname)s %(message)s',
+        level=logging.INFO,
+        datefmt='%Y-%m-%d %H:%M:%S',
+        handlers=[
+            logging.StreamHandler()
+        ]
+    )
+
     if (not exists(INTERFACES_PATH)):
         logging.info("File %s does not exist.", INTERFACES_PATH)
         exit(0)
--- a/kura/distrib/src/main/resources/common/comment-interfaces-file.py
+++ b/kura/distrib/src/main/resources/common/comment-interfaces-file.py
@@ -1,0 +1,101 @@
+#!/usr/bin/env python3
+#
+# Copyright (c) 2023 Eurotech and/or its affiliates and others
+#
+# This program and the accompanying materials are made
+# available under the terms of the Eclipse Public License 2.0
+# which is available at https://www.eclipse.org/legal/epl-2.0/
+#
+# SPDX-License-Identifier: EPL-2.0
+#
+# Contributors:
+#  Eurotech
+#
+import subprocess
+import sys
+import shutil
+from os.path import exists
+
+LOG_MSG_PREFIX = "[comment_interfaces_file.py] "
+ETHERNET_TYPES = ['ethernet', 'eth', 'wired']
+WIRELESS_TYPES = ['wifi', 'wireless']
+INTERFACES_PATH = "/etc/network/interfaces"
+INTERFACES_OLD_PATH =  INTERFACES_PATH + ".old"
+INTERFACES_TMP_PATH = INTERFACES_PATH + ".tmp"
+
+def get_eth_wlan_interfaces_names():
+    """Reads the network interface names using 'nmcli dev' command.
+
+    It is assumed that at least one ethernet interface is present.
+
+    Returns:
+        tuple of lists (eth_names, wlan_names) where:
+            'eth_names' are the found ethernet interface names ordered by name;
+            'wlan_names' are the found wireless interface names ordered by name, might be an empty list.
+    """
+    cmd_output = subprocess.check_output(['nmcli', 'dev']).decode(sys.stdout.encoding).strip()
+    # list comprehension to remove empty items
+    lines = [x.strip() for x in cmd_output.split('\n') if len(x.strip()) > 0]
+
+    # removing header
+    del lines[0]
+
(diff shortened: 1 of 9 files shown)


kura/distrib/src/main/resources/common/comment-interfaces-file.py
```diff
@@ -0,0 +1,101 @@
...
+            print(LOG_MSG_PREFIX + "Comment " + name + " configuration.");
+            output += commentParagraph(paragraph)
+        else:
+            output += paragraph + "\n\n"
+    file = open(INTERFACES_TMP_PATH, "w")
+    file.write(output)
+    file.close()
+
+print(LOG_MSG_PREFIX + "Replace " + INTERFACES_PATH + " file.");
+shutil.move(INTERFACES_TMP_PATH, INTERFACES_PATH)
```
mattdibi: For redability, it can be useful to define a `main` function:

```suggestion
def main():
  if (not exists(INTERFACES_PATH)):
      print(LOG_MSG_PREFIX + "File " + INTERFACES_PATH + " does not exist.")
      exit(1)
  
  print(LOG_MSG_PREFIX + "Backup " + INTERFACES_PATH + " file.")
  shutil.copyfile(INTERFACES_PATH, INTERFACES_OLD_PATH)
  shutil.copyfile(INTERFACES_PATH, INTERFACES_TMP_PATH)
  
  print(LOG_MSG_PREFIX + "Search for interfaces...");
  interface_names = get_eth_wlan_interfaces_names()
  
  for name in interface_names:
      with open(INTERFACES_TMP_PATH, 'r+', encoding='utf-8') as file_to_edit:
          contentParagraph = file_to_edit.read().split("\n\n")
      output = ""
      for paragraph in contentParagraph:
          if name in paragraph:                
              print(LOG_MSG_PREFIX + "Comment " + name + " configuration.");
              output += commentParagraph(paragraph)
          else:
              output += paragraph + "\n\n"
      file = open(INTERFACES_TMP_PATH, "w")
      file.write(output)
      file.close()
  
  print(LOG_MSG_PREFIX + "Replace " + INTERFACES_PATH + " file.");
  shutil.move(INTERFACES_TMP_PATH, INTERFACES_PATH)

if __name__ == "__main__":
    main()
```

kura/distrib/src/main/resources/common/comment-interfaces-file.py
```diff
@@ -0,0 +1,105 @@
...
+import subprocess
+import sys
+import shutil
+from os.path import exists
+
+LOG_MSG_PREFIX = "[comment_interfaces_file.py] "
+ETHERNET_TYPES = ['ethernet', 'eth', 'wired']
+WIRELESS_TYPES = ['wifi', 'wireless']
+INTERFACES_PATH = "/etc/network/interfaces"
+INTERFACES_OLD_PATH =  INTERFACES_PATH + ".old"
```
marcellorinaldo: 2 blank spaces here

kura/distrib/src/main/resources/common/comment-interfaces-file.py
```diff
@@ -0,0 +1,105 @@
...
+            commented += "#" + line + "\n" 
+        else:
+            commented += line + "\n" 
+    commented += "\n"
+    return (commented)
+
+def main():
+    if (not exists(INTERFACES_PATH)):
+        print(LOG_MSG_PREFIX + "File " + INTERFACES_PATH + " does not exist.")
+        exit(1)
```
marcellorinaldo: I would exit normally here, the absence of the file is not an error.

kura/distrib/src/main/resources/common/comment-interfaces-file.py
```diff
@@ -0,0 +1,105 @@
...
+        if (not line.startswith("#")):
+            commented += "#" + line + "\n" 
+        else:
+            commented += line + "\n" 
+    commented += "\n"
+    return (commented)
+
+def main():
+    if (not exists(INTERFACES_PATH)):
+        print(LOG_MSG_PREFIX + "File " + INTERFACES_PATH + " does not exist.")
```
mattdibi: A much less important note here (feel free to ignore this comment): you might want to have %-formatting here instead of relying on string concatenation. String concatenation might result in unexpected errors during type casting.

Example:

```suggestion
        print("%s File %s does not exist." % (LOG_MSG_PREFIX, INTERFACES_PATH))
```

Other ways to achieve the same thing:
- f-strings (introduced in Python 3.6)
- `str.format` function

Furthermore, instead of simply `print`-ing you can rely on the [built-in logger](https://docs.python.org/3/library/logging.html). It requires some setup but it is a built-in solution for this exact case.

```python
logging.basicConfig(
    format='[comment_interfaces_file.py] %(asctime)s %(levelname)s %(message)s',
    level=logging.INFO,
    datefmt='%Y-%m-%d %H:%M:%S',
    handlers=[
        logging.StreamHandler()
    ]
)

...
logging.info("File %s does not exist.", INTERFACES_PATH)
```

kura/distrib/src/main/resources/common/comment-interfaces-file.py
```diff
@@ -0,0 +1,105 @@
...
+    """Reads the network interface names using 'nmcli dev' command.
+
+    It is assumed that at least one ethernet interface is present.
+
+    Returns:
+        tuple of lists (eth_names, wlan_names) where:
+            'eth_names' are the found ethernet interface names ordered by name;
+            'wlan_names' are the found wireless interface names ordered by name, might be an empty list.
+    """
+    cmd_output = subprocess.check_output(['nmcli', 'dev']).decode(sys.stdout.encoding).strip()
```
mattdibi: This might be a silly comment but: what if, instead of relying on the `nmcli` command, you go ask to NM directly through dbus? `dbusctl` allows for JSON output so that you can parse it directly using the built-in json library and avoid a lot of headaches in parsing the output while being sure to have a reliable API under the hood.

This assumes `dbusctl` is available on the target machine ofc

Again, feel free to ignore this.

mattdibi: Actually... thinking about it, it would require multiple calls to the DBus interface. Quite more convoluted that what we have here. Ignore my comment.

kura/distrib/src/main/resources/common/comment-interfaces-file.py
pierantoniomerlino: So... I'm a Python noob!

kura/distrib/src/main/resources/common/comment-interfaces-file.py
pierantoniomerlino: Yeah... I simply adapted the function in `find-net-interfaces.py`...

marcellorinaldo (approved):

kura/distrib/src/main/resources/common/comment-interfaces-file.py
```diff
@@ -0,0 +1,114 @@
...
+import logging
+from os.path import exists
+
+ETHERNET_TYPES = ['ethernet', 'eth', 'wired']
+WIRELESS_TYPES = ['wifi', 'wireless']
+INTERFACES_PATH = "/etc/network/interfaces"
+INTERFACES_OLD_PATH = INTERFACES_PATH + ".old"
+INTERFACES_TMP_PATH = INTERFACES_PATH + ".tmp"
+
+def get_eth_wlan_interfaces_names():
```
mattdibi: Instead of duplicating the code here, you can import the function from `find-net-interfaces` and proceed from there.

mattdibi (changes requested):
See comment

mattdibi (approved):

marcellorinaldo (approved):

PR #4588, opened 2023-04-21
fix(linux.net): Replaced timestamp check for dnsmasq configuration files with a hash check.

salvatore-coppola:
Replaced timestamp check for dnsmasq configuration files with a hash check.
This change will make the check clock indipendent.

### DIFF
--- a/kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/configuration/monitor/DhcpServerMonitor.java
+++ b/kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/configuration/monitor/DhcpServerMonitor.java
@@ -69,10 +69,14 @@ public class DhcpServerMonitor {
         this.dhcpServerInterfaceConfiguration.entrySet().forEach(entry -> {
             String interfaceName = entry.getKey();
             boolean enable = entry.getValue();
-            if (enable && !this.dhcpServerManager.isRunning(interfaceName)) {
-                startDhcpServer(interfaceName);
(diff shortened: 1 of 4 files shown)


kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/configuration/monitor/DhcpServerMonitor.java
```diff
@@ -69,10 +69,14 @@ private void monitor() {
...
-            } else if (!enable && this.dhcpServerManager.isRunning(interfaceName)) {
-                stopDhcpServer(interfaceName);
+            try {
+                if (enable && !this.dhcpServerManager.isRunning(interfaceName)) {
+                    startDhcpServer(interfaceName);
+                } else if (!enable && this.dhcpServerManager.isRunning(interfaceName)) {
+                    stopDhcpServer(interfaceName);
+                }
+            } catch (KuraException e) {
+                logger.warn("Failed to chech DHCP server status for the interface " + interfaceName, e);
```
marcellorinaldo, line 79: We should log an error here since the monitor is not working.

salvatore-coppola: in the other similar condition we use a warning, i replied the behavior.

marcellorinaldo:
Made some quick tests and it seems to be working: every DHCP server-related property update triggers correctly a service reload.

marcellorinaldo (approved):
Tests planned in future PR.

PR #4589, opened 2023-04-21
fix(setups): Fix oomph setup

salvatore-coppola:
Fix the oomph setup replacing m2e-tycho and eclemma repositories.

This eliminates the need to use older versions of Eclipse.

If merged the main README need an update

### DIFF
--- a/README.md
+++ b/README.md
@@ -165,7 +165,7 @@ mvn -f kura/distrib/pom.xml clean install -DbuildAllContainers
 After this command runs, images can be found in your preferred container engine image list.
 
 ### Eclipse IDE
-The simplest way to start developing on Eclipse Kura is to use an [Eclipse Installer](https://www.eclipse.org/downloads/) based setup. A detailed installation and setup guide is available on the [official documentation](http://eclipse.github.io/kura/dev/kura-setup.html). Here you'll find a brief explaination of the required steps.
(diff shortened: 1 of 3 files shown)


MMaiero:
Did you find anything useful in terms of this issue? https://github.com/eclipse-tycho/tycho/issues/146
So that we are able to use the upstream repo instead of a generic fork?

MMaiero:
Testing with the latest Eclipse it seems that we still have problems. Maybe I've missed something, but even after setting the target platform everything is marked as error:

image

salvatore-coppola:
> Testing with the latest Eclipse it seems that we still have problems. Maybe I've missed something, but even after setting the target platform everything is marked as error:
> 
> image

I'll check.

salvatore-coppola:
@MMaiero now everything should be fine

GregoryIvo:
I just gave this a try on my system and I have encountered this:
image
and after setting the target platform:
image

salvatore-coppola:
> I just gave this a try on my system and I have encountered this: image and after setting the target platform: image

very strange, on my system works without problem, let's check together later

marcellorinaldo:
Works on my PC.

marcellorinaldo (approved):
Works on my setup. MacOS 13.4 x86, Java 1.8.

PR #4600, opened 2023-04-26
fix(core): Watchdog is not refreshed when connection to DB missing

GregoryIvo:
This PR fixes the following issue:

when the DB connection is misconfigured, and then the configuration is changed to be valid it is possible that the watchdog will still reboot the device. This PR added appropriate watchdog check-ins to the data service so that it will not erroneously restart the device. 

### DIFF
--- a/kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/data/DataServiceImpl.java
+++ b/kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/data/DataServiceImpl.java
@@ -253,7 +253,6 @@ public class DataServiceImpl implements DataService, DataTransportListener, Conf
         } catch (KuraStoreException e) {
             logger.error("Failed to start store", e);
             DataServiceImpl.this.disconnectDataTransportAndLog(e);
-            DataServiceImpl.this.watchdogService.checkin(DataServiceImpl.this);
         }
     }
 
@@ -459,7 +458,6 @@ public class DataServiceImpl implements DataService, DataTransportListener, Conf
             } catch (KuraStoreException e) {
                 logger.error("Failed to unpublish in-flight messages", e);
                 DataServiceImpl.this.disconnectDataTransportAndLog(e);
-                DataServiceImpl.this.watchdogService.checkin(DataServiceImpl.this);
             }
         } else if (this.storeState.isPresent()) {
             logger.info("New session established. Dropping all in-flight messages.");
@@ -556,7 +554,6 @@ public class DataServiceImpl implements DataService, DataTransportListener, Conf
             } catch (KuraStoreException e) {
                 logger.error("Cannot confirm message to store", e);
                 disconnectDataTransportAndLog(e);
-                DataServiceImpl.this.watchdogService.checkin(DataServiceImpl.this);
             }
 
             // Notify the listeners
@@ -600,7 +597,6 @@ public class DataServiceImpl implements DataService, DataTransportListener, Conf
         try {
             this.storeState.get().getOrOpenMessageStore();
         } catch (KuraStoreException e) {
-            DataServiceImpl.this.watchdogService.checkin(DataServiceImpl.this);
             throw new KuraConnectException(e, MESSAGE_STORE_NOT_CONNECTED_MESSAGE);
         }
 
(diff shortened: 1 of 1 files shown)


kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/data/DataServiceImpl.java
```diff
@@ -467,6 +467,7 @@ private void unpublishOrDropInFlightMessages(boolean publishInFlightMessages) {
             } catch (KuraStoreException e) {
                 logger.error("Failed to drop in-flight messages", e);
                 DataServiceImpl.this.disconnectDataTransportAndLog(e);
+                DataServiceImpl.this.watchdogService.checkin(DataServiceImpl.this);
```
nicolatimeus: There should be no need for this

kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/data/DataServiceImpl.java
```diff
@@ -913,7 +914,11 @@ public void run() {
                 }
                 connected = true;
             } catch (KuraConnectException | KuraStoreException e) {
-                logger.warn("Connect failed", e);
+                logger.warn("DataService failure occured: ", e);
+                
+                if (e instanceof KuraStoreException) {
+                    DataServiceImpl.this.watchdogService.checkin(DataServiceImpl.this);
```
nicolatimeus: We should do this only if `DataServiceImpl.this.dataServiceOptions.isConnectionRecoveryEnabled()`

nicolatimeus: Is this equivalent to adding an additional `|| e instanceof KuraStoreException` to line 924?

GregoryIvo: Seems to be so, making that change

kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/data/DataServiceImpl.java
```diff
@@ -913,13 +913,13 @@ public void run() {
                 }
                 connected = true;
             } catch (KuraConnectException | KuraStoreException e) {
-                logger.warn("Connect failed", e);
-
+                logger.warn("DataService failure occured: ", e);
```
nicolatimeus: Maybe we can also add the exception class name as part of the message since it is not easily visible if the user looks at systemd journal, something like `logger.warn("Connection attempt failed: {}", e.getClass().getSimpleName(), e);`

kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/data/DataServiceImpl.java
```diff
@@ -913,13 +913,13 @@ public void run() {
...
+                
                 if (DataServiceImpl.this.dataServiceOptions.isConnectionRecoveryEnabled()) {
 
                     if (isAuthenticationException(e) || DataServiceImpl.this.connectionAttempts
                             .getAndIncrement() < DataServiceImpl.this.dataServiceOptions
-                                    .getRecoveryMaximumAllowedFailures()) {
+                                    .getRecoveryMaximumAllowedFailures() || e instanceof KuraStoreException) {
```
nicolatimeus: We should move the new check before of the `this.connectionAttempts.getAndIncrement()` call otherwise the remaining attempt counter will be decreased unnecessarily.

kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/data/DataServiceImpl.java
```diff
@@ -913,13 +913,14 @@ public void run() {
                 }
                 connected = true;
             } catch (KuraConnectException | KuraStoreException e) {
-                logger.warn("Connect failed", e);
+                logger.warn("Connection attempt failed with exception {}: {}", e.getClass().getSimpleName(), e);
```
nicolatimeus: The last `: {}` suffix in the format string is not necessary

nicolatimeus (approved):

PR #4601, opened 2023-04-26
fix(dnsmasq): added protections against wrong configurations

marcellorinaldo:
This PR includes checks in the piece of code that manages `dnsmasq` to prevent wrong configurations from being applied.

**Related Issue:** N/A.

**Description of the solution adopted:** Several changes are introduced in this PR:

- `dnsmasq` is disabled at installation so it won't start at system reboot causing conflicts with other DNS server tools.
- The global configuration is written only once.
- When a wrong configuration is written in `/etc/dnsmasq.d/`, `DnsmasqTool` detects the defect and removes it.

**Screenshots:** N/A.

**Manual Tests**: The following manual tests have been executed on a Raspberry Pi 4 - Raspbian 11 to validate the solution:

1. On fresh installation after Kura starts the `dnsmasq` service is not failing.
2. Applying a wrong snapshot with an unexisting interface configured in DHCP server mode does not keep a wrong configuration, the directory `/etc/dnsmasq.d/` will not contain the configuration for such interface.
3. `dnsmasq` does not fail and is running even if there is no interface configured in DHCP server mode; no specific configuration is present in `/etc/dnsmasq.d/`.
4. Configs creation and service reloading is done consistently with what is set on the UI (tried combinations of DHCP&NAT, only NAT, etc.).
5. `/etc/dnsmasq.d/dnsmasq-globals.conf` is rewritten at kura restart. Modifying it after Kura is started will trigger the monitor to rewrite it.
6. If after first installation there is already a configuration present, like `/etc/dnsmasq.d/dnsmasq-wlan0.conf`, such configuration is rewritten if configured from the UI.

Cases that we do not care about:

1. If an interface is disabled but a user manually puts a valid config in `/etc/dnsmasq.d/`, like `/etc/dnsmasq.d/dnsmasq-wlan0.conf`,  then the file `/etc/dnsmasq.d/dnsmasq-wlan0.conf` is not deteled and `dnsmasq` fails because the interface is not existent. This case is specular to the nr.1 of the previous list.
2. If a user removes a configuration inside `/etc/dnsmasq.d/` except for the global one, the monitor will detect the change but the configuration is not rewritten upon next configuration change. The `dnsmasq` service is not restarted in this case so it will keep running with the last configuration.

### DIFF
--- a/kura/org.eclipse.kura.linux.net/src/main/java/org/eclipse/kura/linux/net/dhcp/server/DnsmasqTool.java
+++ b/kura/org.eclipse.kura.linux.net/src/main/java/org/eclipse/kura/linux/net/dhcp/server/DnsmasqTool.java
@@ -142,7 +142,9 @@ public class DnsmasqTool implements DhcpLinuxTool {
     private void writeGlobalConfig() {
         try {
             Path dnsmasqGlobalsPath = Paths.get(DNSMASQ_GLOBAL_CONFIG_FILE);
-            Files.write(dnsmasqGlobalsPath, GLOBAL_CONFIGURATION.getBytes(StandardCharsets.UTF_8));
+            if (Files.notExists(dnsmasqGlobalsPath)) {
+                Files.write(dnsmasqGlobalsPath, GLOBAL_CONFIGURATION.getBytes(StandardCharsets.UTF_8));
+            }
         } catch (IOException e) {
             logger.warn("DNSMASQ - Failed setting in DHCP-only mode.", e);
         }
--- a/kura/org.eclipse.kura.linux.net/src/main/java/org/eclipse/kura/linux/net/dhcp/server/DnsmasqTool.java
+++ b/kura/org.eclipse.kura.linux.net/src/main/java/org/eclipse/kura/linux/net/dhcp/server/DnsmasqTool.java
@@ -77,17 +77,22 @@ public class DnsmasqTool implements DhcpLinuxTool {
 
         writeGlobalConfig();
 
-        return this.executorService.execute(systemctlRestartCommand());
+        CommandStatus restartStatus = this.executorService.execute(systemctlRestartCommand());
+
+        if (!restartStatus.getExitStatus().isSuccessful()) {
+            removeInterfaceConfig(interfaceName);
+            restartStatus = this.executorService.execute(systemctlRestartCommand());
+        }
+
+        return restartStatus;
     }
 
     @Override
     public boolean disableInterface(String interfaceName) throws KuraProcessExecutionErrorException {
         try {
-            File configFile = new File(DhcpServerManager.getConfigFilename(interfaceName));
-
             boolean isInterfaceDisabled = true;
 
-            if (Files.deleteIfExists(configFile.toPath())) {
+            if (removeInterfaceConfig(interfaceName)) {
                 CommandStatus status = this.executorService.execute(systemctlRestartCommand());
                 isInterfaceDisabled = status.getExitStatus().isSuccessful();
             }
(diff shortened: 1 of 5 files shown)


marcellorinaldo:
Question for reviewer: when no interface is set in DHCP server mode should we disable `dnsmasq` or keep it running with an empty configuration as it is right now?

kura/org.eclipse.kura.linux.net/src/main/java/org/eclipse/kura/linux/net/dhcp/server/DnsmasqTool.java
```diff
@@ -50,8 +55,7 @@ public DnsmasqTool(CommandExecutorService service) {
 
     @Override
     public boolean isRunning(String interfaceName) throws KuraProcessExecutionErrorException {
-        CommandStatus status = this.executorService.execute(
-                new Command(new String[] { "systemctl", "is-active", "--quiet", DhcpServerTool.DNSMASQ.getValue() }));
+        CommandStatus status = this.executorService.execute(IS_ACTIVE_COMMAND);
 
         boolean isRunning;
         try {
```
pierantoniomerlino, line 63: Could you please catch only the relevant exceptions in the catch branch (NoSuchAlgorithmException, IOException)? Moreover, the message in the KuraProcessExecutionErrorException is wrong.

kura/org.eclipse.kura.linux.net/src/main/java/org/eclipse/kura/linux/net/dhcp/server/DnsmasqTool.java
```diff
@@ -77,18 +81,23 @@ public CommandStatus startInterface(String interfaceName) throws KuraProcessExec
```
pierantoniomerlino, line 87: Same here.

kura/org.eclipse.kura.linux.net/src/main/java/org/eclipse/kura/linux/net/dhcp/server/DnsmasqTool.java
```diff
@@ -121,14 +130,21 @@ private boolean isConfigFileAlteredOrNonExistent(String interfaceName)
...
+        } catch (Exception e) {
+            return false;
+        }
     }
 
     private void writeGlobalConfig() {
         try {
             Path dnsmasqGlobalsPath = Paths.get(DNSMASQ_GLOBAL_CONFIG_FILE);
-            Files.write(dnsmasqGlobalsPath, GLOBAL_CONFIGURATION.getBytes(StandardCharsets.UTF_8));
+            if (Files.notExists(dnsmasqGlobalsPath)) {
```
pierantoniomerlino: What happens if someone changes the content of the file outside Kura? Is it possible to check the hash of this file as well?

kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/configuration/NMConfigurationServiceImpl.java
```diff
@@ -389,11 +389,6 @@ private void writeDhcpServerConfiguration(Set<String> interfaceNames) {
         });
     }
 
-    protected DhcpServerConfigWriter buildDhcpServerConfigWriter(final String interfaceName,
```
pierantoniomerlino: I think that this method is used only for testing. Indeed the tests are failing.

MMaiero:
The PR fails due to the quality gate: https://sonarcloud.io/component_measures?id=org.eclipse.kura%3Akura&branch=PR-4601&metric=new_coverage&view=list

marcellorinaldo:
> The PR fails due to the quality gate: https://sonarcloud.io/component_measures?id=org.eclipse.kura%3Akura&branch=PR-4601&metric=new_coverage&view=list

Will not add tests now but in a future PR

MMaiero:
@pierantoniomerlino Can you please check this PR? If ok we are going to merge, @marcellorinaldo will add a following PR to complete the testing of the changes

pierantoniomerlino (approved):

PR #4606, opened 2023-04-27
fix(nm): cannot configure ethernet interfaces if the cable is disconnected

mattdibi:
When creating a new connection for an ethernet device without a connected cable the following exception was thrown:

```
Wed 2023-04-19 13:21:32.979015 UTC [s=c82f5507362a4bcabd3597251333b2c2;i=4e53c;b=3b9b181c4f434fb3a0649bc9d54ba0fd;m=3719928b;t=5f9b04d7a5f72;x=27124d60895e6d09]
    MESSAGE=Unable to apply configuration to the device path /org/freedesktop/NetworkManager/Devices/2
    STACKTRACE=org.freedesktop.dbus.exceptions.DBusExecutionException: Connection 'kura-eno1-connection' is not available on device eno1 because device has no carrier
                at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
                at sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
...
```

which is correct.

Unfortunately, due to the fact that the connection was created using the `AddAndActivate` method, the connection was not created and saved on disk due to the above mentioned error.

To fix the issue I split the operation in two steps:
- I create and save on disk the connection using the `AddConnection` method of the `Settings` interface
- Then activate the connection using the `ActivateConnection` method of the `NetworkManager` interface

**Tests**: tested working on my RPi4 running Raspbian 64bit NM 1.30.6

### DIFF
--- a/kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/NMDbusConnector.java
+++ b/kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/NMDbusConnector.java
@@ -35,7 +35,6 @@ import org.eclipse.kura.nm.status.AccessPointsProperties;
 import org.eclipse.kura.nm.status.DevicePropertiesWrapper;
 import org.eclipse.kura.nm.status.NMStatusConverter;
 import org.eclipse.kura.nm.status.SupportedChannelsProperties;
-import org.freedesktop.AddAndActivateConnectionTuple;
 import org.freedesktop.NetworkManager;
 import org.freedesktop.dbus.DBusPath;
 import org.freedesktop.dbus.connections.impl.DBusConnection;
@@ -395,10 +394,12 @@ public class NMDbusConnector {
             this.nm.ActivateConnection(new DBusPath(connection.get().getObjectPath()),
                     new DBusPath(device.getObjectPath()), new DBusPath("/"));
         } else {
-            AddAndActivateConnectionTuple createdConnectionTuple = this.nm.AddAndActivateConnection(
-                    newConnectionSettings, new DBusPath(device.getObjectPath()), new DBusPath("/"));
+            Settings settings = this.dbusConnection.getRemoteObject(NM_BUS_NAME, NM_SETTINGS_BUS_PATH, Settings.class);
+            DBusPath createdConnectionPath = settings.AddConnection(newConnectionSettings);
+            this.nm.ActivateConnection(createdConnectionPath, new DBusPath(device.getObjectPath()), new DBusPath("/"));
+
             Connection createdConnection = this.dbusConnection.getRemoteObject(NM_BUS_NAME,
-                    createdConnectionTuple.getPath().getPath(), Connection.class);
+                    createdConnectionPath.getPath(), Connection.class);
             connection = Optional.of(createdConnection);
         }
 
--- a/kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/NMDbusConnector.java
+++ b/kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/NMDbusConnector.java
@@ -391,16 +391,20 @@ public class NMDbusConnector {
 
         if (connection.isPresent()) {
             connection.get().Update(newConnectionSettings);
-            this.nm.ActivateConnection(new DBusPath(connection.get().getObjectPath()),
-                    new DBusPath(device.getObjectPath()), new DBusPath("/"));
         } else {
             Settings settings = this.dbusConnection.getRemoteObject(NM_BUS_NAME, NM_SETTINGS_BUS_PATH, Settings.class);
             DBusPath createdConnectionPath = settings.AddConnection(newConnectionSettings);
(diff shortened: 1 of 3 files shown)


kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/NMDbusConnector.java
```diff
@@ -395,10 +394,12 @@ private void enableInterface(String deviceId, NetworkProperties properties, Devi
             this.nm.ActivateConnection(new DBusPath(connection.get().getObjectPath()),
                     new DBusPath(device.getObjectPath()), new DBusPath("/"));
         } else {
-            AddAndActivateConnectionTuple createdConnectionTuple = this.nm.AddAndActivateConnection(
-                    newConnectionSettings, new DBusPath(device.getObjectPath()), new DBusPath("/"));
+            Settings settings = this.dbusConnection.getRemoteObject(NM_BUS_NAME, NM_SETTINGS_BUS_PATH, Settings.class);
+            DBusPath createdConnectionPath = settings.AddConnection(newConnectionSettings);
+            this.nm.ActivateConnection(createdConnectionPath, new DBusPath(device.getObjectPath()), new DBusPath("/"));
```
nicolatimeus: Should we catch the exception thrown here and go on with the remaining parts of the method if the connection is added successfully but cannot be activated for some reason?

mattdibi: I thought about it and decided against it since it should be safe as it is right now.

If an exception is thrown on the connection activation I cannot proceed with the housekeeping and the modem watchdog setup.

What do you think?

nicolatimeus: I'm in doubt.. I wonder if the configuration logic should depend or not on whether the interfaces can be brought up successfully at the instant when the configuration is applied. for example I would enable the modem monitor logic even if cannot connect at the moment.. what happens if the ActivateConnection call is completely removed?

kura/test/org.eclipse.kura.nm.test/src/test/java/org/eclipse/kura/nm/NMDbusConnectorTest.java
```diff
@@ -1337,10 +1328,13 @@ private void whenApplyIsCalledWith(Map<String, Object> networkConfig) {
         try {
             this.instanceNMDbusConnector.apply(networkConfig);
         } catch (DBusException e) {
+            e.printStackTrace();
             this.hasDBusExceptionBeenThrown = true;
         } catch (NoSuchElementException e) {
+            e.printStackTrace();
             this.hasNoSuchElementExceptionThrown = true;
         } catch (NullPointerException e) {
+            e.printStackTrace();
```
mattdibi: Remove leftover stacktraces

GregoryIvo:
Perhaps this is unrelated:

Stop kura service -> disconnect ethernet -> (SSH via Wi-Fi Dies)

when I am testing this, I am noticing that when my ethernet cable is pulled while connected via wifi (ssh via wifi), I lose all network connectivity. 

in my console:
image

GregoryIvo:
Good News: 

After connecting to the device with a display and performing the test as described in the ticket, the issue the PR's solves seems to be resolved.

image
(Eth0 -> shows up)

no active connection
image

---

When eth is then plugged back in:
image
image

image



**TLDR: looks like it is working on my end (RPI 4 AARCH64 Generic)!**

kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/NMDbusConnector.java
mattdibi: Mh... I see what you mean but I have reservations about changing this particular piece of code this far in the release process.

> what happens if the ActivateConnection call is completely removed?

Since we do not set the `connection.autoconnect` parameter it defaults to `true` and thus the connection that are created are brought up automagically by NM, without throwing any error in the case of the no carrier interface.

~The same goes for the connection Update I guess (still testing it).~ Nopes: the connection needs to be activated for the changes to take effect.

This also is what they do in the python examples for NM:
- [Add new connection example](https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/blob/main/examples/python/dbus/add-connection.py)
- [Update connection example](https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/blob/main/examples/python/dbus/update-ip4-method.py)

... they do not perform the call to the `ActivateConnection` method. 

The drawback here is that, by removing the Activate call, we don't log the error which might be very useful for debugging purposes... also the wait for connection completion breaks.

mattdibi: > Should we catch the exception thrown here and go on with the remaining parts of the method if the connection is added successfully but cannot be activated for some reason?

Addressed in 6ad61b89cbda386ed5ca5667a709e8e1cd548b4a

nicolatimeus (approved):

PR #4609, opened 2023-04-27
fix(web2): fixed authType and pdpType enums conversion on net2

marcellorinaldo:
This PR fixes the conversion between the enums held GWT-side `GwtModemAuthType`, `GwtModemPdpType` and the `AuthType`, `PdpType` enums used in the configuration service.

**Related Issue:** N/A.

**Screenshots:** N/A.

**Manual Tests**: Using a generic profile verify with a cellular connection:

1. Configure a modem with a Auth Type parameter value different than None
2. Save the configuration.
3. Refresh the web ui and go back to the network configuration section
4. No exceptions should appear on the logs.

### DIFF
--- a/kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/server/net2/configuration/NetworkConfigurationServicePropertiesBuilder.java
+++ b/kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/server/net2/configuration/NetworkConfigurationServicePropertiesBuilder.java
@@ -259,9 +259,9 @@ public class NetworkConfigurationServicePropertiesBuilder {
             GwtModemInterfaceConfig gwtModemConfig = (GwtModemInterfaceConfig) this.gwtConfig;
 
             this.properties.setModemAuthType(this.ifname,
-                    EnumsParser.getAuthType(Optional.ofNullable(gwtModemConfig.getAuthType().name())));
+                    EnumsParser.getAuthType(Optional.ofNullable(gwtModemConfig.getAuthType())));
             this.properties.setModemPdpType(this.ifname,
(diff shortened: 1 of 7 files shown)


kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/server/net2/configuration/NetworkConfigurationServicePropertiesBuilder.java
```diff
@@ -258,8 +258,10 @@ private void setModemProperties() {
         if (this.gwtConfig instanceof GwtModemInterfaceConfig) {
             GwtModemInterfaceConfig gwtModemConfig = (GwtModemInterfaceConfig) this.gwtConfig;
 
-            this.properties.setModemAuthType(this.ifname, Optional.ofNullable(gwtModemConfig.getAuthType().name()));
-            this.properties.setModemPdpType(this.ifname, Optional.ofNullable(gwtModemConfig.getPdpType().name()));
+            this.properties.setModemAuthType(this.ifname,
+                    EnumsParser.getAuthType(Optional.ofNullable(gwtModemConfig.getAuthType().name())));
```
nicolatimeus: `AuthType.name()` should never return `null`, can `gwtModemConfig.getAuthType()` return `null`? If yes probably we should do `Optional.ofNullable(gwtModemConfig.getAuthType()).map(GwtAuthType::name)`

nicolatimeus:
This one must be updated as well, it is currently using the GWT enum variant names:

https://github.com/eclipse/kura/blob/559b9b9ff1a422fc7153b01bae89667b4bf9bbd7/kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/configuration/NMSettingsConverter.java#L272

nicolatimeus (approved):
If the build passes we can merge

PR #4615, opened 2023-04-28
fix(monitor.dns-server): added check for interface as LAN before starting DNS server

marcellorinaldo:
This PR prevents the DNS server monitors from starting the DNS server when the `pass.dns` property is true but the interface is actually not set as LAN. This issue affected both the Kura networking profiles and the new NetworkManager-based ones.

**Related Issue:** N/A.

**Screenshots:** N/A.

**Manual Tests**: On `generic-arm32` and `raspberry-pi-armhf` it has been verified that:

1. Pass DNS functionality is still working: with the gateway set as AP with `passDNS=true` a laptop is able to obtain an IP address and DNS list. `named` is correctly running on gateway.
2. Setting the wireless interface from AP to DISABLED will shut down the `named` service. Prior this PR this was not the case. Re-enabling the interface as AP with `passDNS=true` will correctly restart `named` service and clients are able to connect.
3. Switching `passDNS` property whenin AP mode is correctly shutting down/starting `named` service.
4. Behaviour with multiple interfaces is consistent: if at least one interface is set in DHCP mode with pass DNS then `named` service keeps running. If both interfaces are disabled, then `named` will be stopped.

### DIFF
--- a/kura/org.eclipse.kura.net.admin/src/main/java/org/eclipse/kura/net/admin/monitor/DnsMonitorServiceImpl.java
+++ b/kura/org.eclipse.kura.net.admin/src/main/java/org/eclipse/kura/net/admin/monitor/DnsMonitorServiceImpl.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2011, 2022 Eurotech and/or its affiliates and others
+ * Copyright (c) 2011, 2023 Eurotech and/or its affiliates and others
  *
  * This program and the accompanying materials are made
  * available under the terms of the Eclipse Public License 2.0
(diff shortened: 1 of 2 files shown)


pierantoniomerlino:
@marcellorinaldo have you tested this PR with multiple interfaces configured as LANs and passDNS? Or combination of these?

marcellorinaldo:
> @marcellorinaldo have you tested this PR with multiple interfaces configured as LANs and passDNS? Or combination of these?

Just tested on a`generic-arm32` profile. The behaviour is consistent with the other tests. Disabling an interface does not stop `named` if there is another interface set with pass DNS. Disabling both will stop the service.

If needed, I will later do the same test also on `raspberry-pi-armhf` profile.

kura/org.eclipse.kura.net.admin/src/main/java/org/eclipse/kura/net/admin/monitor/DnsMonitorServiceImpl.java
```diff
@@ -236,9 +236,10 @@ private void updateDnsProxyConfig() {
             List<NetInterfaceConfig<? extends NetInterfaceAddressConfig>> netInterfaceConfigs = this.networkConfiguration
                     .getNetInterfaceConfigs();
             for (NetInterfaceConfig<? extends NetInterfaceAddressConfig> netInterfaceConfig : netInterfaceConfigs) {
-                if (netInterfaceConfig.getType() == NetInterfaceType.ETHERNET
+                if ((netInterfaceConfig.getType() == NetInterfaceType.ETHERNET
```
pierantoniomerlino, line 239: Please update copyright header

marcellorinaldo:
@pierantoniomerlino I made further testing also on the `raspberry-pi-armhf` profiles and everything seems fine. I have updated the PR description with these new tests.

pierantoniomerlino (approved):

PR #4638, opened 2023-05-11
fix(web2): GPS modem config is done independently of set IP status

marcellorinaldo:
Prior this PR, the configuration of the **GPS enable** property was not performed if the device was set as _Disabled_ or _Unmanaged_. A user might want to enable GPS without wanting the modem as IP interface.

**Related Issue:** N/A.

**Screenshots:** N/A.

**Manual Tests**: N/A.


mattdibi (approved):
Tested with device running latest develop version of Kura and a Quectel modem supporting GPS.

When the enable field is set on the UI the snapshot is correctly updated and the NMDbusConnector receives the expected value.

LGTM

PR #4639, opened 2023-05-11
feat(nm): post modem gps event on configuration

mattdibi:
**Related Issue:** #4633 

To correctly support the `serial` position provider of Kura, it is required to post a `ModemGpsEnabled/DisabledEvent` upon configuring the modem. This PR adds the event posting.

**Note**: This PR depends on #4633.

Pending:
- [x] Merge #4633 
- [ ] Update to avoid conflicts
- [ ] Match supported models with known serial parameters instead of relying on hardcoded ones.

---

### 1. Enable the Position service in the Kura UI

Set the parameters as follows:
- *enabled*: `true`
- *static*: `false`
- *provider*: `serial`

image

### 2. Go in the "Network" tab, select the modem and enable the GPS

image

... and you're done!


mattdibi:
Superseded by https://github.com/eclipse/kura/pull/4644

PR #4645, opened 2023-05-12
fix(net.admin): added dhclient hook files

pierantoniomerlino:

This PR adds a script hook for `dhclient`. If an interface is configured as LAN, the hook prevent `dhclient` to write the DNS address to the `resolv.conf` file.


kura/distrib/src/main/ant/build_equinox_distrib.xml
```diff
@@ -554,6 +554,10 @@ fi]]>
 				prefix="${build.output.name}/${install.folder}" />
 			<zipfileset file="src/main/resources/common/ifup-local.raspbian"
 				prefix="${build.output.name}/${install.folder}" />
+	        <zipfileset file="src/main/resources/common/kura-dhclient-enter-hooks"
```
MMaiero: formatting seems to be broken

nicolatimeus (approved):

PR #4655, opened 2023-05-16
fix(web2): DHCP renew button is now disabled except in WAN/LAN mode using dhcp

salvatore-coppola:
DHCP renew button is now disabled except in WAN/LAN mode using dhcp.


mattdibi (approved):
Testing on RPi Raspberry Pi OS 64bit generic profile.

**Before this PR**:

image

**After this PR**:

image

The button is enabled for WAN/LAN modes.

LGTM

PR #4664, opened 2023-05-18
feat: equinox ECF remote services experiment

marcellorinaldo:
N/A.

**Related Issue:** N/A.

**Screenshots:** N/A.

**Manual Tests**: N/A.

### DIFF
--- a/kura/distrib/src/main/ant/build_equinox_distrib.xml
+++ b/kura/distrib/src/main/ant/build_equinox_distrib.xml
@@ -256,6 +256,9 @@
             property="is.web2.jar.available" />
         <antcall target="web2-config" />
 
+        <!-- Remote services -->
+        <antcall target="ecf-config" />
+
         <available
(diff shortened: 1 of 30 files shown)


marcellorinaldo:
Compiling Kura with:

```
mvn -f target-platform/pom.xml clean install
mvn -f kura/pom.xml clean install -DskipTests
mvn -f kura/distrib/pom.xml -Pgeneric-arm32
```

Should produce a deb that contains all what it's needed to run OSGi remote services.

I wanted to create two demo bundles: one client and one provider as DPs.

To test you should install two Kura instances on two devices connected together. On one you install the client and on the other the provider.

`org.eclipse.kura.remote.service.demo.client` and `org.eclipse.kura.remote.service.demo.provider` are not yet working. They install correctly but I think the OSGI-INF is wrong. They should write on the log when the `update` methods are called.

PR #4665, opened 2023-05-18
fix(web2): save configured wifi channel in GWT interface

marcellorinaldo:
Prior this PR, the selected WiFi Channel was saved in the snapshot but not picked up when populating the channels list.

There is a convention that the element on position 0 of the integer list retuned by `gwtConfig.getChannels()` is considered from the UI the user-selected one.

**Related Issue:** N/A.

**Screenshots:** N/A.

**Manual Tests**: N/A.


salvatore-coppola (approved):
successfully tested on RaspberryPi 4B with generic-arm32

PR #4671, opened 2023-05-19
test(linux.net.dhcp.server): added `DnsmasqToolTest` and `DhcpdToolTest`

marcellorinaldo:
This PR adds some tests for the `linux.net.dhcp.server` package.

**Related Issue:** N/A.

**Screenshots:** N/A.

**Manual Tests**: N/A.

**Any side note on the changes made:** Refactor of `DnsmasqTool` for better testability.


marcellorinaldo:
Static mocking involved, not a big fan of this solution. Maybe we need to refactor the tested classes.

salvatore-coppola (commented):
Instead of `setDnsmasqGlobalConfigFile` could we have 2 constructors ? One with the `globalConfigFilename` as parameter and another one without, using the default config filename (`/etc/dnsmasq.d/dnsmasq-globals.conf`).

So using the class without setting the config file would not produce an error.

MMaiero:
The quality gate remark has been resolved. Merging now. @marcellorinaldo to catch the improvement suggested by @salvatore-coppola please open a new PR.

MMaiero (approved):

PR #4689, opened 2023-05-26
fix(core.deployment): avoid duplicate DPs in packages

marcellorinaldo:
This PR avoid having duplicate DPs inside the `/opt/eclipse/kura/packages` directory with different filenames.

**Related Issue:** N/A.

**Screenshots:** N/A.

**Manual Tests**: The test procedure is as follows:
1. Create a local server reachable from gateway containing a large DP.
2. Connect the device to Kapua.
3. From Kapua, install a package specifying the URL to the server containing the DP.

Tested cases with _DeploymentService_  configured with property **downloads.directory=/tmp**:
- [x] Fresh install, no DP in `packages` and no entry in `dpa.properties`, perform request and restart framework.
- [x] Keep DP in `packages` and entry in `dpa.properties`, resubmit request and restart framework.
- [x] Remove the DP from `packages` but keep entry in `dpa.properties`, resubmit request and restart framework.
- [x] Keep the DP in `packages`, remove entry from `dpa.properties`, resubmit request and restart framework.
- [x] Keep the DP in `packages`, edit `dpa.properties` to have the value of the entry to some random string, resubmit request and restart framework.
- [x] Remove the DP from `packages`, edit `dpa.properties` to have the value of the entry to some random string, resubmit request and restart framework.

Tested cases with _DeploymentService_ configured with property **downloads.directory=/opt/eclipse/kura/packages**:
- [x] Fresh install, no DP in `packages` and no entry in `dpa.properties`.
- [x] Keep DP in `packages` and entry in `dpa.properties`, resubmit request and restart framework.
- [x] Remove the DP from `packages` but keep entry in `dpa.properties`, resubmit request and restart framework.
- [x] Keep the DP in `packages`, remove entry from `dpa.properties`, resubmit request and restart framework.
- [x] Keep the DP in `packages`, edit `dpa.properties` to have the value of the entry to some random string, resubmit request and restart framework.
- [x] Remove the DP from `packages`, edit `dpa.properties` to have the value of the entry to some random string, resubmit request and restart framework.

To perform the manual tests it might be useful to inspect the downloads and `packages` folder with command:

```shell
watch -n .1 ls -al <path>
```

In all the above test the following conditions hold after the request completes:
- Exactly one DP in `/opt/eclipse/kura/packages`.
- `/opt/eclipse/kura/packages/dpa.properties` contains an entry for the installed dp with path to `packages`.

Additional test:
- [x] Create and execute two different requests that install two different DPs but with the same _name_ and _version_ metrics. After installation and after framework restart the DPs are correctly installed in `dpa.properties` and in `packages` with the correct name (not the one specified in request's metrics).


marcellorinaldo:
I'm wondering how these tests passed before. Currently they pass on my Eclipse but fail on my local Maven build. Let's see the CI.

nicolatimeus:
Does this change the behavior of the verifier script part?https://github.com/eclipse/kura/blob/28941cb627f14624e6e57024a4d48bd1ff43d8d4/kura/org.eclipse.kura.core.deployment/src/main/java/org/eclipse/kura/core/deployment/download/impl/DownloadImpl.java#L281

MMaiero:
The sonar cloud errors are related to other parts of the code that have not been changed in this PR. If @nicolatimeus is ok with the changes, we can proceed with merging

MMaiero:
@marcellorinaldo @nicolatimeus Does these changes in the dp name affect upgrades and current deployments?

marcellorinaldo:
> @marcellorinaldo @nicolatimeus Does these changes in the dp name affect upgrades and current deployments?

We need to keep "-" instead of "_" as separator since the verifier scripts rely on that and changing it may affect existing installations. So we decided to change the naming convention for the DPs to use "-" as well. In this way we can keep existing installations working. However, we want to perform some more tests.

nicolatimeus:
Please do not merge this for now, it requires more testing. I will also try to make some modifications to minimize the changes in the name conventions.

nicolatimeus:
I have tested the following:
1. Marcello's tests in previous comment, plus some Kura restarts to verify that the dps are still loaded
2. Deployment job with verifier (both downloaded from the cloud and deployed by the main installation script)
3. Performed an upgrade with dp installed and repeated 1. and 2. with same scripts

nicolatimeus:
With the latest change the following separators in package names are used:
* The DeploymentAgent uses `_` for dps (same as previous version)
* DEPLOY-V2 uses `_` for dps like DeploymentAgent (changed since previous version)
* DEPLOY-V2 uses `-` for shell scripts (same as previous version)

marcellorinaldo:
@nicolatimeus I have redone the tests with the latest changes, everything seems OK.

nicolatimeus (approved):

PR #4720, opened 2023-06-15
fix(network): Fixed issue with concurrent dhcp client and the default route

salvatore-coppola:
This pr fix an issue with concurrent dhcp client and the default route.
Before this pr dhclient made request to the server asking the default routers even if it was in lan mode.
Now an hook script remove the dhclient option from the request.

### DIFF
--- a/kura/distrib/src/main/ant/build_equinox_distrib.xml
+++ b/kura/distrib/src/main/ant/build_equinox_distrib.xml
@@ -614,13 +614,15 @@ fi]]>
                 prefix="${build.output.name}/${install.folder}" />
             <zipfileset file="src/main/resources/common/kura-dhclient-enter-hook"
                 prefix="${build.output.name}/${install.folder}" />
-        	
+            <zipfileset file="src/main/resources/common/kura-dhclient-enter-route-hook"
+                prefix="${build.output.name}/${install.folder}" />
+
             <zipfileset file="src/main/resources/common/dnsmasq"
                 prefix="${build.output.name}/${install.folder}" />
-        	
+
             <zipfileset file="${project.build.directory}/${build.output.name}/selinuxKura.pp"
             	prefix="${build.output.name}/${install.folder}" />
-        	
+
             <zipfileset
                 file="${project.build.directory}/${build.output.name}/start_kura.sh"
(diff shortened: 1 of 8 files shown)


mattdibi:
# Prior to this PR

We noticed that after rebooting a device running an interface set as LAN, the default route was not reinstated.

I tested the following setup on a RPi 64bit running Kura 5.3.0:
- eth0: set as DHCP/LAN
- wlan0: set as Station Mode/DHCP/WAN

image

image

Before rebooting we have:

```bash
route -n
Kernel IP routing table
Destination     Gateway         Genmask         Flags Metric Ref    Use Iface
0.0.0.0         192.168.2.1     0.0.0.0         UG    0      0        0 wlan0
192.168.1.0     0.0.0.0         255.255.255.0   U     0      0        0 eth0
192.168.2.0     0.0.0.0         255.255.255.0   U     0      0        0 wlan0
```

After reboot:

```bash
route -n
Kernel IP routing table
Destination     Gateway         Genmask         Flags Metric Ref    Use Iface
192.168.1.0     0.0.0.0         255.255.255.0   U     0      0        0 eth0
192.168.2.0     0.0.0.0         255.255.255.0   U     0      0        0 wlan0
```

Which meant that the device was not able to ping the internet anymore.

```
pi@raspberrypi:~ $ ping google.com
ping: connect: Cannot assign requested address
pi@raspberrypi:~ $ ping 8.8.8.8
ping: connect: Network is unreachable
```

The problem is related to the concurrence of multiple dhclient.

Routes are always requested to the dhcp server and then removed if the interface is set to LAN mode.

The default route is always requested and when another one is present, with the same metric, the dhclient fails to set the route (in this case the one that is set and later removed by Kura).

# After this PR

After rebooting the route table is correct:

```bash
pi@raspberrypi:~ $ route -n
Kernel IP routing table
Destination     Gateway         Genmask         Flags Metric Ref    Use Iface
0.0.0.0         192.168.2.1     0.0.0.0         UG    0      0        0 wlan0
192.168.1.0     0.0.0.0         255.255.255.0   U     0      0        0 eth0
192.168.2.0     0.0.0.0         255.255.255.0   U     0      0        0 wlan0
```

mattdibi (approved):
LGTM

PR #4733, opened 2023-06-23
refactor(nm): move signal handlers into `org.eclipse.kura.nm.signal.handlers` package

mattdibi:
Similarly to #4730 here we group together the signal handlers in the same package.

I don't really like [this](https://github.com/eclipse/kura/compare/develop...mattdibi:refactor/signal_handlers_package?expand=1#diff-044ee4653a90a5b82546d6756de84757b52913a1630267773f0afa53bb66a38bL549-L552)

@pierantoniomerlino what do you think? Is it worth to change the access modifier just for this refactoring?


pierantoniomerlino:
@mattdibi I agree on your opinion regarding the access modifier. However, I'll keep this modification since we planned to refactor the `NMDbusConnector` too.

pierantoniomerlino (approved):

PR #4737, opened 2023-06-26
fix(deployment.agent ): Added check on URI schema in the agent configuration file.

salvatore-coppola:
Added check on URI schema in the agent configuration file.

Now if the schema is `http` or `https` the package is rejected.

### DIFF
--- a/kura/org.eclipse.kura.deployment.agent/src/main/java/org/eclipse/kura/deployment/agent/impl/DeploymentAgent.java
+++ b/kura/org.eclipse.kura.deployment.agent/src/main/java/org/eclipse/kura/deployment/agent/impl/DeploymentAgent.java
@@ -340,9 +340,8 @@ public class DeploymentAgent implements DeploymentAgentService, ConfigurableComp
             try {
                 String packageUri = (String) deployedPackages.get(packageName);
 
-                if (isHttp(packageUri)) {
-                    throw new KuraRuntimeException(KuraErrorCode.SECURITY_EXCEPTION,
-                            "For security reason is not allowed install package from remote repository in config file.");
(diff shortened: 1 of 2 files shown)


kura/org.eclipse.kura.deployment.agent/src/main/java/org/eclipse/kura/deployment/agent/impl/DeploymentAgent.java
```diff
@@ -335,17 +337,27 @@ private void installPackagesFromConfFile() {
...
+
+                logger.info("Deploying package name {} at URI {}", packageName, packageUri);
+                installDeploymentPackageAsync(packageUri);
             } catch (Exception e) {
                 logger.error("Error installing package {}", packageName, e);
             }
         }
     }
 
+    private boolean isHttp(String packageUri) throws URISyntaxException {
```
marcellorinaldo: I would only allow file URIs, not just exclude http protocols (we leave out ftp and others). Something like:

```java
private boolean isFile(String packageUri) {
    new File(packageUri).isFile();
}
```

salvatore-coppola: @marcellorinaldo don't know, however only file and http are supported by our internal installer:
https://github.com/eclipse/kura/blob/c2800924ab338cf4036087224ad1404366c901f7/kura/org.eclipse.kura.deployment.agent/src/main/java/org/eclipse/kura/deployment/agent/impl/DeploymentAgent.java#L359

marcellorinaldo (approved):

PR #4755, opened 2023-07-10
refactor: use `Device.Udi` to retrieve ModemManager device dbus path

mattdibi:
Currently, in the Kura codebase, to retrieve the ModemManager dbus path for a NetworkManager device we:
- Gather all the objects managed by ModemManager using the `GetManagedObjects` method
- Iterate over the managed objects
- Look for a match between the ModemManager `DeviceIdentifier` and the NetworkManager `DeviceId`

This PR introduces a better way to achieve the same results leveraging the `Device.Udi` property of NetworkManager. Per [NetworkManager documentation](https://networkmanager.dev/docs/api/latest/gdbus-org.freedesktop.NetworkManager.Device.html#gdbus-property-org-freedesktop-NetworkManager-Device.Udi):

> Operating-system specific transient device hardware identifier. This is an opaque string representing the underlying hardware for the device, and shouldn't be used to keep track of individual devices. **For some device types (Bluetooth, Modems) it is an identifier used by the hardware service (ie bluez or ModemManager) to refer to that device, and client programs use it get additional information from those services which NM does not provide.** The Udi is not guaranteed to be consistent across reboots or hotplugs of the hardware. If you're looking for a way to uniquely track each device in your application, use the object path. If you're looking for a way to track a specific piece of hardware across reboot or hotplug, use a MAC address or USB serial number.

In the case of ModemManager it is the dbus path that we need to call the API methods on that object. Note that we don't need it to be consistent between reboots since it's only used to manipulate the object once we have uniquely identified using our internal representation.

---

Tests performed:
- [X] RPi4 w/ NM 1.30.6, MM  1.14.12
- [X] RPi4 w/ NM 1.14.6, MM  1.14.12
- [X] Dy1014 w/ NM 1.30.4, MM 1.16.2

```
2023-07-11T09:33:46,214 [Start Level: Equinox Container: 7b4847fc-4f84-469c-ab74-64bea6cd38cf] WARN  o.e.k.n.NMDbusConnector - Device "1-1.1" of type "NM_DEVICE_TYPE_MODEM" not configured. Disabling...
2023-07-11T09:33:46,260 [Start Level: Equinox Container: 7b4847fc-4f84-469c-ab74-64bea6cd38cf] INFO  o.e.k.n.NMDbusConnector - ModemDbusPath for device /org/freedesktop/NetworkManager/Devices/4 is /org/freedesktop/ModemManager1/Modem/0
```

### DIFF
--- a/kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/NMDbusConnector.java
+++ b/kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/NMDbusConnector.java
@@ -49,7 +49,6 @@ import org.freedesktop.dbus.DBusPath;
 import org.freedesktop.dbus.connections.impl.DBusConnection;
 import org.freedesktop.dbus.exceptions.DBusException;
 import org.freedesktop.dbus.exceptions.DBusExecutionException;
-import org.freedesktop.dbus.interfaces.ObjectManager;
 import org.freedesktop.dbus.interfaces.Properties;
 import org.freedesktop.dbus.types.UInt32;
 import org.freedesktop.dbus.types.Variant;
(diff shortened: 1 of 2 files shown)


mattdibi:
@sfiorani can you take a look at this? Tests on other devices are welcome :)

sfiorani (approved):

PR #4759, opened 2023-07-12
fix: support for no Wireless Security

mattdibi:
During testing we found a couple of bugs affecting the Generic profiles when set for connecting to a Wireless Access Point with no security.

The first bug was affecting the backend, more precisely the NetworkManager-interfacing code. It was found that, when connecting or providing an AP without security, the `802-11-wireless-security` settings **should not be set**. Fixed in 1e31329

The second bug was affecting the UI. When parsing the configuration returned by the backend, the parser would not take into account the `NONE` option. Resulting in a UI reporting "WPA2" as the configured security type but actually using "NONE" under the hood. Fixed in 3445fba

> **Warning**: WEP encryption is currently broken. We're not correctly setting the password when setting up the connection and thus NetworkManager will not create the connection. We'll leave this as a known issue for now.


salvatore-coppola (approved):

marcellorinaldo (approved):

salvatore-coppola:
Tested successfully on Raspberry Pi 32 bit

mattdibi:
Merging as is. The Sonar issue will be addressed in a future PR

PR #4772, opened 2023-07-19
fix(nm): WEP support for Generic Profiles

mattdibi:
As reported in https://github.com/eclipse/kura/pull/4759

> Warning: WEP support is currently broken. We're not correctly setting the password when setting up the connection and thus NetworkManager will not create the connection. We'll leave this as a known issue for now.

This PR addresses the issue by setting the correct parameters for the WEP connection.

Please note that, as we did on the non-generic profiles, we're only using `NM_WEP_KEY_TYPE_KEY`:

> in which case the key is either a 10- or 26-character hexadecimal string, or a 5- or 13-character ASCII password

`NM_WEP_KEY_TYPE_PASSPHRASE` is not currently supported.

Likewise we're only using index 0 WEP key. This is the WEP key used in most networks. See [NetworkManager documentation](https://networkmanager.dev/docs/api/latest/nm-settings-dbus.html) for additional informations.

**Test performed**: Tested working on RPi4:
- [X] DUT connecting to a device running Kura 5.2.3 acting as a WEP Access Point.
- [X] using my workstation to connect to DUT acting as WEP Access Point
- [x] using a device running Kura 5.2.3 to connect to DUT acting as WEP Access Point

### DIFF
--- a/kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/configuration/NMSettingsConverter.java
+++ b/kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/configuration/NMSettingsConverter.java
@@ -48,6 +48,9 @@ public class NMSettingsConverter {
 
     private static final int NM_WEP_KEY_TYPE_KEY = 1;
 
+    private static final String KURA_PROPS_KEY_WIFI_MODE = "net.interface.%s.config.wifi.mode";
+    private static final String KURA_PROPS_KEY_WIFI_SECURITY_TYPE = "net.interface.%s.config.wifi.%s.securityType";
+
     private NMSettingsConverter() {
         throw new IllegalStateException("Utility class");
     }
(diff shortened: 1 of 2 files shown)


MMaiero:
@GregoryIvo Can you please have a look at this?

GregoryIvo:
Tested this PR on raspberry Pi 4 aarch64,  and a WEP AP, seems to be working well!

kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/configuration/NMSettingsConverter.java
```diff
@@ -175,23 +181,50 @@ public static Map<String, Variant<?>> build80211WirelessSettings(NetworkProperti
     }
 
     public static Map<String, Variant<?>> build80211WirelessSecuritySettings(NetworkProperties props, String deviceId) {
+        String propMode = props.get(String.class, KURA_PROPS_KEY_WIFI_MODE, deviceId);
+        String securityType = props.get(String.class, KURA_PROPS_KEY_WIFI_SECURITY_TYPE, deviceId,
+                propMode.toLowerCase());
+
+        if ("SECURITY_WEP".equals(securityType)) {
```
GregoryIvo, line 188: I feel like this if structure might grow out of control in the future.

What if we introduce a WifiSecurityType class, and a new class for each security type to encapsulate all of the security setups?

MMaiero (approved):

kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/configuration/NMSettingsConverter.java
mattdibi: I think you're correct. Adding a small `KuraWifiSecurityType` enum that represents the possible values we can find in the snapshot would be a better option than comparing Strings. This way it would be less error-prone, more type safe and cleaner.

If I have time I'll try to submit a new PR to refactor this.

PR #4785, opened 2023-07-25
fix(nm): "unmanaged", "unknown" and "disabled" interface status handling inside NMSettingsConverter

mattdibi:
> **Warning**
> Wait #4786 before merging to avoid merge conflicts on that PR

This PR fixes two issues related to the `net.interface.%s.config.ip4.status` and `net.interface.%s.config.ip6.status`.

#### `Unmanaged` handling

The first one concerns a wrong behaviour that was encoded in the unit tests: in the `buildSettings` tests, when the interface is set as `Unmanaged` we still check for the settings to be sent to NetworkManager. This is *wrong* at the design level. When the device is set as `Unmanaged` none of the `buildSettings` method should be called because *we shouldn't set anything*. If we do we would overwrite what the user set and thus lose the meaning of the `Unmanaged` property.

To prevent similar issues, in addition to updating the tests,  I added an exception for when the `buildSettings` methods are called with an `Unmanaged` interface.

#### `Disabled` handling

Due to the fact that, until now, `ip6.status` was always set to `Disabled` or `Unmanaged` we did not take into account the case in which one ip version is enabled and one disabled. Therefore we collapsed the concept of **interface disabled** with **ipv4 disabled**. In reality, when only one ip version is disabled, the **interface** is enabled we just need to set the corresponding `ipvX.method` as disabled.

This PR adds a simple check inside the `buildIp4Settings` and `buildIp6Settings` methods so that when the corresponding ip is disabled they will set the correct NetworkManager property.

#### `Unknown` handling

Similarly to what happened to the `Unmanaged` status we're not correctly handling the `Unknown` status in the `buildSettings`. Again, this is due to the fact that `buildSettings` assumes the statuses it sees were already checked by the `NMDbusConnector`. Unfortunately in the tests these assumptions are violeted and there are some curious behaviour.

See: https://github.com/eclipse/kura/blob/1acc7838278e6dbe430bc9d6d70cfb00bcbf2059/kura/test/org.eclipse.kura.nm.test/src/test/java/org/eclipse/kura/nm/configuration/NMSettingsConverterTest.java#L745

`netIPv4StatusManagedLan` is not correct, but the tests works anyway.

To prevent similar issues, in addition to updating the tests,  I added an exception for when the `buildSettings` methods are called with an `Unknown` status.

### DIFF
--- a/kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/configuration/NMSettingsConverter.java
+++ b/kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/configuration/NMSettingsConverter.java
@@ -156,19 +156,10 @@ public class NMSettingsConverter {
     }
 
     public static Map<String, Variant<?>> buildIpv6Settings(NetworkProperties props, String deviceId) {
-        KuraIpStatus ip6Status = KuraIpStatus
-                .fromString(props.get(String.class, "net.interface.%s.config.ip6.status", deviceId));
-
-        if (ip6Status == KuraIpStatus.UNMANAGED || ip6Status == KuraIpStatus.UNKNOWN) {
-            throw new IllegalArgumentException("IPv6 status is not supported: " + ip6Status
-                    + ". Build settings should be called only for managed interfaces.");
-        }
-
         Map<String, Variant<?>> settings = new HashMap<>();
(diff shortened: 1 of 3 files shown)


mattdibi:
@sfiorani please take a look at this

pierantoniomerlino (approved):

PR #4791, opened 2023-08-02
feat: Added trigger scan button

salvatore-coppola:
Added the capability to force a wireless networks refresh with a button in the SSID Modal window.

Furthermore polkit rule and dbuse authorization have been added in order to use the wpasupplicant api on dbus


mattdibi (dismissed):
About the NMDbusConnector changes:

I think the rescan should be ran only if the device is a WiFi interface. As of right now I think you’ll get an exception thrown if the recompute is requested for a different device.

I think you just need to move the wpa supplicant call inside here https://github.com/eclipse/kura/blob/f1859f1c7ea5bf2c6c8581ed020849b00ec4584d/kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/NMDbusConnector.java#L241

Furthermore I wouldn’t instantiate the wpa_supplicant wrapper at every net status request but rather add it as a private member of the NMDbusConnector as we did for all the other wrappers. See:

https://github.com/eclipse/kura/blob/f1859f1c7ea5bf2c6c8581ed020849b00ec4584d/kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/NMDbusConnector.java#L93-L94

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/net/status/NetworkStatusService.java
```diff
@@ -34,14 +34,33 @@ public interface NetworkStatusService {
...
      *             when an error occurs while retrieving the status of the
      *             given interface
      */
-    public Optional<NetworkInterfaceStatus> getNetworkStatus(String interfaceName) throws KuraException;
+    public Optional<NetworkInterfaceStatus> getNetworkStatus(String interfaceId) throws KuraException;
+
+    /**
+     * Return an optional {@link NetworkInterfaceStatus} of the given network
+     * interface selected by its name.
+     *
```
pierantoniomerlino, line 52: I'd would like to use the same javadoc of the `getNetworkStatus(String interfaceId)` method, adding a comment regarding the recompute parameter.

kura/test/org.eclipse.kura.nm.test/src/test/java/org/eclipse/kura/nm/NMDbusConnectorTest.java
```diff
@@ -927,6 +949,16 @@ public void givenBasicMockedDbusConnector() throws DBusException, IOException {
...
+        
+        when(this.dbusConnection.getRemoteObject(eq("fi.w1.wpa_supplicant1"),
+                eq("/fi/w1/wpa_supplicant1"), any())).thenReturn(this.mockedWpaSupplicant);
+        
+        DBusPath fakeDbusInterfacePath = new DBusPath("/fake/dbus/path");
+        
+        when(this.mockedWpaSupplicant.GetInterface(anyString())).thenReturn(fakeDbusInterfacePath);
+        
+        when(this.dbusConnection.getRemoteObject(eq("fi.w1.wpa_supplicant1"), eq(fakeDbusInterfacePath.getPath()), any()))
+        .thenReturn(this.mockedInterface);
```
mattdibi: This is not the correct way to mock the device. 

You're creating a single `wpa_supplicant` interface for the entire test which gets returned for every interfaceId. This is not the correct way to model the test. What if we want to test that, among multiple wifi devices, only the one requested perform a rescan?

You should create and return the dbus path only for Wifi Devices and only for the interfaceId you expect.

```suggestion
        
        when(this.dbusConnection.getRemoteObject(eq("fi.w1.wpa_supplicant1"),
                eq("/fi/w1/wpa_supplicant1"), any())).thenReturn(this.mockedWpaSupplicant);
        
```

You should mock the interface in the [`givenMockedDevice` method](https://github.com/eclipse/kura/blob/177821520ceb573539fd6f0aad461990a98d0ec5/kura/test/org.eclipse.kura.nm.test/src/test/java/org/eclipse/kura/nm/NMDbusConnectorTest.java#L986) where you can mock the interface only for WiFi devices:

https://github.com/eclipse/kura/blob/177821520ceb573539fd6f0aad461990a98d0ec5/kura/test/org.eclipse.kura.nm.test/src/test/java/org/eclipse/kura/nm/NMDbusConnectorTest.java#L1017

There you'll have:

```java
    if (type == NMDeviceType.NM_DEVICE_TYPE_WIFI) {

        DBusPath fakeDbusInterfacePath = new DBusPath("/fake/dbus/path");
        
        when(this.mockedWpaSupplicant.GetInterface(interfaceId)).thenReturn(fakeDbusInterfacePath);

        // create mocked interface here
        
        when(this.dbusConnection.getRemoteObject(eq("fi.w1.wpa_supplicant1"), eq(fakeDbusInterfacePath.getPath()), any()))
        .thenReturn(mockedInterface);

         simulateIwCommandOutputs(interfaceId, mockedProperties1);
}
```

To mock the interface you'll need a `Map<String, Interface>` so that you can keep track of the interfaces you mocked as we did for the mocked devices. Use the dbus path as key to the map

kura/test/org.eclipse.kura.nm.test/src/test/java/org/eclipse/kura/nm/NMDbusConnectorTest.java
```diff
@@ -91,10 +91,15 @@
...
 
+import fi.w1.Wpa_supplicant1;
+import fi.w1.wpa_supplicant1.Interface;
+
 public class NMDbusConnectorTest {
 
     private static final String MM_MODEM_BUS_NAME = "org.freedesktop.ModemManager1.Modem";
     private final DBusConnection dbusConnection = mock(DBusConnection.class, RETURNS_SMART_NULLS);
+    private final Wpa_supplicant1 mockedWpaSupplicant = mock(Wpa_supplicant1.class);
+    private final Interface mockedInterface = mock(Interface.class);
```
mattdibi: As I said in my previous comment this is not correct. There's only one interface for the entire test.

Please, refer to how we handled the Device here:

https://github.com/eclipse/kura/blob/805c45b5e2ee8b258d192bb589762987494fa7a3/kura/test/org.eclipse.kura.nm.test/src/test/java/org/eclipse/kura/nm/NMDbusConnectorTest.java#L121

You should have:
```suggestion
    private final Map<String, Interface> mockedInterfaces = new HashMap<>();
```

and in the `givenMockedDevice` code **create** the mock interface adding it to the above mentioned map for tracking.

Later in the `then` method you'll check the expectation using the "interfaceId" as key to retrieve the mocked object from the map.

kura/test/org.eclipse.kura.nm.test/src/test/java/org/eclipse/kura/nm/NMDbusConnectorTest.java
```diff
@@ -984,6 +1008,12 @@ private void givenMockedDevice(String deviceId, String interfaceId, NMDeviceType
 
         if (type == NMDeviceType.NM_DEVICE_TYPE_WIFI) {
             simulateIwCommandOutputs(interfaceId, mockedProperties1);
+
+            doReturn(mockedInterface).when(this.dbusConnection).getRemoteObject("fi.w1.wpa_supplicant1",
+                    "/mock/device/" + interfaceId, Interface.class);
+
+            when(this.mockedWpaSupplicant.GetInterface(interfaceId))
+                    .thenReturn(new DBusPath("/mock/device/" + interfaceId));
```
mattdibi:
```suggestion
            Interface mockedInterface = mock(Interface.class);
            mockedInterfaces.add(interfaceId, mockedInterface);
            
            DBusPath mockedInterfaceDbusPath = new DBusPath("/mock/device/" + interfaceId);
            
            when(this.mockedWpaSupplicant.GetInterface(interfaceId))
                    .thenReturn(mockedInterfaceDbusPath);

            doReturn(mockedInterface).when(this.dbusConnection).getRemoteObject("fi.w1.wpa_supplicant1",
                    mockedInterfaceDbusPath, Interface.class);

```

kura/test/org.eclipse.kura.nm.test/src/test/java/org/eclipse/kura/nm/NMDbusConnectorTest.java
```diff
@@ -1466,6 +1513,10 @@ private void thenLocationSetupWasCalledWith(EnumSet<MMModemLocationSource> expec
                 .Setup(MMModemLocationSource.toBitMaskFromMMModemLocationSource(expectedLocationSources), expectedFlag);
     }
 
+    private void thenScanIsTriggered() {
+        verify(this.mockedInterface, times(1)).Scan(any());
+    }
```
mattdibi:
```suggestion
    private void thenScanIsTriggered(String interfaceId) {
        verify(this.mockedInterfaces.get(interfaceId), times(1)).Scan(any());
    }
```

mattdibi (approved):
LGTM 👌

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/net/status/NetworkStatusService.java
```diff
@@ -45,7 +45,10 @@ public interface NetworkStatusService {
 
     /**
      * Return an optional {@link NetworkInterfaceStatus} of the given network
-     * interface selected by its name.
+     * interface selected by its name.For Ethernet and WiFi interfaces, the
+     * identifier is typically the interface name. For the modems, instead,
```
pierantoniomerlino, line 49: Use id instead of name.

pierantoniomerlino: In the javadoc:

```
@param recompute
     *            If set to true, the NetworkStatusService will perform some additional **expansive** operations (like AP
     *            scanning) to ensure the status returned by the method is up to date. If set to false, it will return
* the status as returned by the network configuration provider.
```

do you mean expansive or expensive?

pierantoniomerlino:
It seems that in https://github.com/eclipse/kura/blob/0beca995f07c92427b4cc504dfc0bb64c83966d2/kura/distrib/src/main/resources/common/manage_kura_users.sh#L145 a space is missing. This prevents the scripts to run correctly.

pierantoniomerlino (approved):

PR #4793, opened 2023-08-02
feat(api): Firewall IPv6 API

pierantoniomerlino:

This PR adds some new APIs to manage IPv6 address in firewall.

**Description of the solution adopted:** The following classes have been added:

- `FirewallOpenPortConfig6` and `FirewallOpenPortConfigIP6`
- `FirewallPortForwardingConfig6` and `FirewallPortForwardingConfigIP6`

All the Firewall rules classes have been updated to use builders instead of constructors. The implementations are updated accordingly.

Some code duplication should be accepted.


kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/net/IP4Address.java
```diff
@@ -12,6 +12,8 @@
  ******************************************************************************/
 package org.eclipse.kura.net;
 
+import java.net.UnknownHostException;
```
marcellorinaldo, line 15: Please update copyright headers

marcellorinaldo (changes requested):

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/net/IP4Address.java
```diff
@@ -26,6 +28,16 @@ public class IP4Address extends IPAddress {
         super(addr, jnAddress);
     }
 
+    /**
+     * Returns the default IPv4 address (0.0.0.0/0).
+     * 
+     * @return the 0.0.0.0/0 IPv4 address
+     * @throws UnknownHostException
+     */
```
MMaiero, line 37: We are missing a ```@since```

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/net/IP6Address.java
```diff
@@ -26,6 +28,16 @@ public class IP6Address extends IPAddress {
         super(addr, jnAddress);
     }
 
+    /**
+     * Returns the default IPv6 address (::/0).
+     * 
+     * @return the ::/0 IPv6 address
+     * @throws UnknownHostException
+     */
+    public static IP6Address getDefaultAddress() throws UnknownHostException {
```
MMaiero, line 38: Same here, I think

nicolatimeus:
Please check again all new public methods and classes in the API bundle, there are still few of them that are missing since annotations and/or javadocs. For new classes (also nested ones) it is enough to add since annotation at class level.

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/net/firewall/FirewallOpenPortConfigIP6.java
```diff
@@ -0,0 +1,63 @@
...
+ * @since 2.6
+ */
+@ProviderType
+public class FirewallOpenPortConfigIP6 extends FirewallOpenPortConfigIP<IP6Address> implements FirewallOpenPortConfig6 {
+
+    private FirewallOpenPortConfigIP6(FirewallOpenPortConfigIP6Builder builder) {
+        super(builder);
+    }
+
+    public static FirewallOpenPortConfigIP6Builder builder() {
```
MMaiero, line 34: Javadoc needed here?

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/net/firewall/FirewallOpenPortConfigIP6.java
```diff
@@ -0,0 +1,63 @@
...
+        return new FirewallOpenPortConfigIP6Builder();
+    }
+
+    /**
+     * The builder class for the IPv6 firewall open port configuration
+     * 
+     * @since 2.6
```
MMaiero: probably the since is not needed as it is already stated at the beginning of the class

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/net/firewall/FirewallPortForwardConfigIP6.java
```diff
@@ -0,0 +1,64 @@
...
+import java.net.UnknownHostException;
+
+import org.eclipse.kura.net.IP6Address;
+import org.eclipse.kura.net.NetworkPair;
+import org.osgi.annotation.versioning.ProviderType;
+
+/**
+ * The implementation of IPv6 firewall port forward configurations
+ *
+ * @noextend This class is not intended to be subclassed by clients.
```
MMaiero, line 24: We can add the since here

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/net/firewall/FirewallOpenPortConfigIP6.java
pierantoniomerlino: Do we really need a javadoc for this builder?

MMaiero: It's public so it could be used, right?
Otherwise let's not expose it or let's deprecate it if it is the case.

nicolatimeus (approved):

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/net/firewall/FirewallOpenPortConfigIP6.java
pierantoniomerlino: Yes, it is public and can be used. But its meaning is straightforward and the since annotation is at class level. My only concern is to not add useless information.

MMaiero:
The build fails due to code duplication. This is accepted by the team: the IPv4 code has been duplicated and adapted to support IPv6. The duplication is accepted for now as we are planning to rework this in a more organic way in a later release.

PR #4795, opened 2023-08-04
feat: Remove IPv6 protocol disabling from generic installers

sfiorani:
**Brief description of the PR.**

This PR removes the IPv6 protocol disabling steps in the generic profiles installers. This way, the protocol IPv6 is not disabled by default at installation.

**Related Issue:**

**Screenshots:**

**Manual Tests:** Tested on Raspberry Pi - Kura Version: generic-aarch64


marcellorinaldo (changes requested):
I wouldn't remove the `kura/distrib/src/main/resources/common/sysctl.kura.conf` and `kura/distrib/src/main/resources/common/patch_sysctl.sh` files since they are used in the non-generic profiles.

marcellorinaldo (approved):

PR #4796, opened 2023-08-07
feat(net.configuration): Updated metatype with new IPv6 properties

sfiorani:
**Brief description of the PR:**

This PR updates those methods that generate metatype for network interfaces, in order to update also the new IPv6 protocol configuraions.

**Related Issue:**

**Screenshots:**

**Manual Tests**: 


pierantoniomerlino (approved):

kura/org.eclipse.kura.net.configuration/src/main/java/org/eclipse/kura/net/configuration/NetworkConfigurationPropertyNames.java
```diff
@@ -22,8 +22,15 @@ public enum NetworkConfigurationPropertyNames {
     CONFIG_IPV4_ADDRESS,
     CONFIG_IPV4_PREFIX,
     CONFIG_IPV4_GATEWAY,
-    CONFIG_DNS_SERVERS,
```
MMaiero, line 25: Isn't this a breaking change? @pierantoniomerlino

pierantoniomerlino: @MMaiero The package is exported as internal. Is it still a breaking change?

MMaiero: if it has implications to the snapshot, probably it is a problem, right?

sfiorani: @marcellorinaldo had similar doubt while development

pierantoniomerlino: AFAIK, this enum is used only to retrieve the Messages in the `/resources /.../NetworkConfigurationMessagesBundle.properties`, so I think that we can change it. I'd to admit that the name of the enum is a bit misleading.

MMaiero: ok for me if we have the confidence that this is not going to be an issue for users that will use old snapshots in this new version

pierantoniomerlino (approved):

PR #4802, opened 2023-08-10
feat: Firewall ipv6 implementation

pierantoniomerlino:

After the merge of #4793, this PR adds the support for the IPv6 firewall. The implementation was done tring to re-use as much code as possible, creating abstract classes to share common methods. The new feature is added to the `org.eclipse.kura.net.admin.firewall` package used only in generics profiles.
Standard profiles will not support the IPv6 firewall.

Since we decided to keep the old firewall implementation in the `org.eclipse.kura.net.admin` bundle, some degree of code duplication should be tolerated.


pierantoniomerlino:
To test the IPv6 firewall implementation, try to apply a snapshot containing the configuration. The code below is an example (the syntax for the rules are the same than the IPv4 firewall):
 
```xml
  <esf:configuration pid="org.eclipse.kura.net.admin.ipv6.FirewallConfigurationServiceIPv6">
      <esf:properties>
          <esf:property array="false" encrypted="false" name="firewall.ipv6.open.ports" type="String">
              <esf:value>8001,tcp,::0/0,wlan0,,,,#</esf:value>
          </esf:property>
          <esf:property array="false" encrypted="false" name="firewall.ipv6.nat" type="String">
              <esf:value>eth0,wlan0,tcp,::/0,::/0,true,#;</esf:value>
          </esf:property>
          <esf:property array="false" encrypted="false" name="firewall.ipv6.port.forwarding" type="String">
              <esf:value>eth0,wlan0,2a00:1450:400c:c01::71,tcp,2020,1010,true,2001:db8::/64,aa:bb:cc:dd:ee:ff,10100:10200,#</esf:value>
          </esf:property>
          <esf:property array="false" encrypted="false" name="kura.service.pid" type="String">
              <esf:value>org.eclipse.kura.net.admin.ipv6.FirewallConfigurationServiceIPv6</esf:value>
          </esf:property>
          <esf:property array="false" encrypted="false" name="service.pid" type="String">
              <esf:value>org.eclipse.kura.net.admin.ipv6.FirewallConfigurationServiceIPv6</esf:value>
          </esf:property>
      </esf:properties>
  </esf:configuration>
```

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/net/firewall/FirewallOpenPortConfigIP.java
```diff
@@ -172,10 +244,70 @@ public String getSourcePortRange() {
         return this.sourcePortRange;
     }
 
+    /**
+     * @deprecated since 2.6. se the FirewallOpenPortConfigIP builder
```
marcellorinaldo: MIspelling on the comment, *se* -> *see*. Same applies for `setPermittedMac`, `setUnpermittedInterfaceName`, `setUnpermittedInterfaceName`, `setPermittedInterfaceName`, `setPermittedNetwork`, `setProtocol`, and `setPortRange`.

kura/org.eclipse.kura.net.admin.firewall/src/main/java/org/eclipse/kura/net/admin/AbstractFirewallConfigurationServiceImpl.java
```diff
@@ -0,0 +1,360 @@
...
+    }
+
+    protected void deactivate(ComponentContext componentContext) {
+        logger.info("Deactivating FirewallConfigurationService...");
+        logger.info("Deactivating FirewallConfigurationService... Done.");
+    }
+
+    public synchronized void updated(Map<String, Object> properties) {
+        logger.debug("updated()");
+        for (Entry<String, Object> entry : properties.entrySet()) {
```
marcellorinaldo: We should check the log level before entering this for loop. Something like:
```java
if (logger.getLogLevel() == LogLevel.DEBUG) {
    for (Entry ...
}
```

kura/org.eclipse.kura.net.admin.firewall/OSGI-INF/firewallConfigurationServiceIPv6.xml
```diff
@@ -0,0 +1,32 @@
...
+    immediate="true"
+    modified="updated">
+   <implementation class="org.eclipse.kura.net.admin.ipv6.FirewallConfigurationServiceIPv6Impl"/>
+   <property name="service.pid" value="org.eclipse.kura.net.admin.ipv6.FirewallConfigurationServiceIPv6"/>
+   <property name="kura.service.pid" value="org.eclipse.kura.net.admin.ipv6.FirewallConfigurationServiceIPv6"/>
+   <service>
+      <provide interface="org.eclipse.kura.net.admin.ipv6.FirewallConfigurationServiceIPv6"/>
+      <provide interface="org.eclipse.kura.configuration.SelfConfiguringComponent"/>
+   </service>
+   <reference bind="setEventAdmin" cardinality="1..1" interface="org.osgi.service.event.EventAdmin" name="EventAdmin" policy="static" unbind="unsetEventAdmin"/>
```
marcellorinaldo: Unbind methods are not necessary for static service references

kura/org.eclipse.kura.linux.net/src/main/java/org/eclipse/kura/linux/net/iptables/LocalRule.java
```diff
@@ -55,8 +54,9 @@ public class LocalRule {
      * @param permittedMAC
      *            MAC address from which connection is allowed (such as AA:BB:CC:DD:EE:FF)
      */
-    public LocalRule(int port, String protocol, NetworkPair<IP4Address> permittedNetwork, String permittedInterfaceName,
-            String unpermittedInterfaceName, String permittedMAC, String sourcePortRange) {
+    public LocalRule(int port, String protocol, NetworkPair<? extends IPAddress> permittedNetwork,
+            String permittedInterfaceName, String unpermittedInterfaceName, String permittedMAC,
+            String sourcePortRange) {
         this.port = port;
         this.portRange = null;
```
marcellorinaldo: Can we use an optional here? Or does it affect old implementations?

kura/org.eclipse.kura.core.net/src/main/java/org/eclipse/kura/core/net/FirewallConfiguration.java
```diff
@@ -68,8 +68,8 @@ public FirewallConfiguration(Map<String, Object> properties) {
     private void parseNatRules(Map<String, Object> properties) {
         String str;
         String[] astr;
```
marcellorinaldo: I would suggest to use more meaningful names for some variables.

`str` -> `openPorts`
`astr` -> `rules`
`private void parseOpenPortRule(String sop)` -> `private void parseOpenPortRule(String **rule**)`

kura/org.eclipse.kura.core.net/src/main/java/org/eclipse/kura/core/net/FirewallConfiguration.java
```diff
@@ -19,46 +19,46 @@
...
-    private final List<FirewallOpenPortConfigIP<? extends IPAddress>> openPortConfigs;
-    private final List<FirewallPortForwardConfigIP<? extends IPAddress>> portForwardConfigs;
-    private final List<FirewallNatConfig> natConfigs;
-    private final List<FirewallAutoNatConfig> autoNatConfigs;
+    private List<FirewallOpenPortConfigIP<? extends IPAddress>> openPortConfigs;
+    private List<FirewallPortForwardConfigIP<? extends IPAddress>> portForwardConfigs;
+    private List<FirewallNatConfig> natConfigs;
+    private List<FirewallAutoNatConfig> autoNatConfigs;
 
-    public FirewallConfiguration() {
```
nicolatimeus, line 56: This is a breaking change in terms of API

nicolatimeus: Should we rename this to AbstractFirewallConfigration and rename FirewallConfigurationIPv4 to FirewallConfiguration and keep the same public interface of this class? otherwise we have to bump the exported version of the org.eclipse.kura.core.net package to 2.0.0

kura/org.eclipse.kura.linux.net/META-INF/MANIFEST.MF
```diff
@@ -55,7 +55,7 @@ Export-Package: org.eclipse.kura.internal.board;version="1.0.0";x-internal:=true
    org.eclipse.kura.usb",
  org.eclipse.kura.linux.net.dhcp;version="2.1.0";uses:="org.eclipse.kura.net.dhcp",
  org.eclipse.kura.linux.net.dns;version="1.1.0";uses:="org.eclipse.kura.net,org.eclipse.kura.net.dns",
- org.eclipse.kura.linux.net.iptables;version="1.1.0";uses:="org.eclipse.kura.net",
```
nicolatimeus, line 58: This PR adds breaking changes in this package.

nicolatimeus (approved):
Please address the lates issues in the API PR, then we can merge

marcellorinaldo (approved):

pierantoniomerlino:
The build fails due to code duplication. This is accepted by the team: the IPv4 code has been duplicated and adapted to support IPv6. The duplication is accepted for now as we are planning to rework this in a more organic way in a later release.

PR #4800, opened 2023-08-09
feat(web2): add IPv6 configuration tab

marcellorinaldo:
This PR adds a net tab "IPv6" to the networking section that allows to configure IPv6-related settings.

**Related Issue:** N/A.

**Description of the solution adopted:** Summary of changes:
- Added UI tab
- Added validations
- Refactor of the network tabs orcherstrator code
- Backend code not implemented in this PR, will be separate

**Screenshots:**

Screenshot 2023-08-10 at 12 08 29

**Manual Tests**:
- [x] *Address Generation Mode* and *Privacy* are enabled only when *Configure = SLAAC*
- [x] When *Status = Enabled for LAN* and *Configure = Manually* then only *IP address* and *Subnet Mask* are editable.
- [x] When *Status = Enabled for WAN* and *Configure = Manually* then *WAN Priority*, *IP address*, *Subnet Mask*, *Gateway*, and *DNS Servers* are editable.
- [x] When *Status = Enabled for WAN* and *Configure != Manually* the *DNS Servers* is editable
- [x] *WAN Priority* is enabled only when *Status = Enabled for WAN*
- [x] All fields are disabled when *Status = Disabled* or a Loopback interface is selected
- [x] Tabs order (IPv4, IPv6, Wireless, DHCPv4&NAT, Modem, etc.) is the same as prior this PR
- [x] *WAN Priority* accepts only integers > -1
- [x] *Subnet mask* accepts only integers between 1 and 128
- [x] Fields that accept IPv6 addresses have working validation, even with shortened addresses. Some test cases can be found below
- [x] IPv6 tab is shown only on generic profiles
- [x] *Information* hints are correctly shown when hovering over the enabled elements

IPv6 addresses (positive) test cases:
- `2001:0000:3238:DFE1:0063:0000:0000:FEFB`
- `2001:0000:3238:DFE1:63:0000:0000:FEFB`
- `2001:0000:3238:DFE1:63::FEFB`
- `2001:0:3238:DFE1:63::FEFB`
- `::`
- `::1`
- `FE80::`

**Any side note on the changes made:** N/A.

PR #4804, opened 2023-08-11
feat(web.server.net2): add IPv6 configuration properties support

marcellorinaldo:
This PR adds the support for the new IPv6 properties in the UI server part, it is a continuation of this PR: https://github.com/eclipse/kura/pull/4800.

**Related Issue:** N/A.

**Screenshots:** N/A.

**Manual Tests**: N/A.

### DIFF
--- a/kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/server/net2/configuration/GwtNetInterfaceConfigBuilder.java
+++ b/kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/server/net2/configuration/GwtNetInterfaceConfigBuilder.java
@@ -59,6 +59,7 @@ public class GwtNetInterfaceConfigBuilder {
         this.gwtConfig = createGwtNetInterfaceConfigSubtype();
         setCommonProperties();
         setIpv4Properties();
+        setIpv6Properties();
         setIpv4DhcpClientProperties();
         setIpv4DhcpServerProperties();
         setRouterMode();
@@ -116,6 +117,42 @@ public class GwtNetInterfaceConfigBuilder {
         this.gwtConfig.setSubnetMask(this.properties.getIp4Netmask(this.ifName));
         this.gwtConfig.setGateway(this.properties.getIp4Gateway(this.ifName));
         this.gwtConfig.setDnsServers(this.properties.getIp4DnsServers(this.ifName));
+    }
+
+    private void setIpv6Properties() {
+        Optional<String> status = this.properties.getIp6Status(this.ifName);
+        if (status.isPresent()) {
+            this.gwtConfig.setIpv6Status(status.get());
+        }
+
+        Optional<Integer> priority = this.properties.getIp6WanPriority(this.ifName);
+        if (priority.isPresent()) {
+            this.gwtConfig.setIpv6WanPriority(priority.get());
+        }
+
+        Optional<String> configure = this.properties.getIp6AddressMethod(this.ifName);
+        if (configure.isPresent()) {
+            this.gwtConfig.setIpv6ConfigMode(configure.get());
+        }
+
+        Optional<String> autoconfiguration = this.properties.getIp6AddressGenMode(this.ifName);
+        if (autoconfiguration.isPresent()) {
+            this.gwtConfig.setIpv6AutoconfigurationMode(autoconfiguration.get());
+        }
+
+        this.gwtConfig.setIpv6Address(this.properties.getIp6Address(this.ifName));
+
+        Optional<Integer> netmask = this.properties.getIp6Netmask(this.ifName);
+        if (netmask.isPresent()) {
+            this.gwtConfig.setIpv6SubnetMask(netmask.get());
+        }
+
+        this.gwtConfig.setIpv6Gateway(this.properties.getIp6Gateway(this.ifName));
+
+        Optional<String> privacy = this.properties.getIp6Privacy(this.ifName);
+        if (privacy.isPresent()) {
+            this.gwtConfig.setIpv6Privacy(privacy.get());
+        }
     }
 
     private void setIpv4DhcpClientProperties() {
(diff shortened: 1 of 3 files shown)


kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/server/net2/configuration/NetworkConfigurationServicePropertiesBuilder.java
```diff
@@ -104,6 +105,37 @@ private void setIpv4Properties() {
...
+    private void setIpv6Properties() {
+        this.properties.setIp6Status(this.ifname, this.gwtConfig.getIpv6Status());
+        this.properties.setIp6AddressMethod(this.ifname, this.gwtConfig.getIpv6ConfigMode());
+
+        boolean isManual = this.gwtConfig.getIpv6ConfigMode().equals("netIPv6MethodManual");
+        boolean isWan = this.gwtConfig.getIpv6Status().equals("netIPv6StatusEnabledWAN");
+        boolean isAuto = this.gwtConfig.getIpv6ConfigMode().equals("netIPv6MethodAuto");
+
+        if (isWan) {
+            if (!Objects.isNull(this.gwtConfig.getIpv6WanPriority())) {
```
mattdibi: Really nitpicking here: for readability reasons I would avoid negating the call and directly use the `nonNull` method.

Don't feel obliged to change this. I would merge this as-is since it's a really small thing.

```suggestion
            if (Objects.nonNull(this.gwtConfig.getIpv6WanPriority())) {
```

kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/server/net2/configuration/NetworkConfigurationServiceProperties.java
```diff
@@ -145,6 +145,97 @@ public void setIp4DnsServers(String ifname, String dnsServers) {
...
+        this.properties.put(String.format(NET_INTERFACE_CONFIG_IP6_GATEWAY, ifname), gateway);
+    }
+
+    public String getIp6DnsServers(String ifname) {
+        return (String) this.properties.getOrDefault(String.format(NET_INTERFACE_CONFIG_IP6_DNS_SERVERS, ifname), "");
+    }
+
+    public void setIp6DnsServers(String ifname, String dnsServers) {
+        this.properties.put(String.format(NET_INTERFACE_CONFIG_IP6_DNS_SERVERS, ifname), dnsServers);
+    }
```
mattdibi, line 237: Actually, we represent the DNS servers as a list of strings. See

https://github.com/eclipse/kura/blob/8566aa9cd5feaf9165d6e6d2e352a1b0f491f6e0/kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/configuration/NMSettingsConverter.java#L242-L243

Am I missing something?

kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/server/net2/configuration/GwtNetInterfaceConfigBuilder.java
```diff
@@ -118,6 +119,42 @@ private void setIpv4Properties() {
         this.gwtConfig.setDnsServers(this.properties.getIp4DnsServers(this.ifName));
     }
 
+    private void setIpv6Properties() {
+        Optional<String> status = this.properties.getIp6Status(this.ifName);
+        if (status.isPresent()) {
+            this.gwtConfig.setIpv6Status(status.get());
+        }
```
mattdibi, line 126: Keep in mind that if the `ip6.status` is missing we have a special way to handle it. See:

https://github.com/eclipse/kura/blob/8566aa9cd5feaf9165d6e6d2e352a1b0f491f6e0/kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/NMDbusConnector.java#L389-L393

We might want to replicate the logic on the UI side to represent it correctly to the user.

kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/server/net2/configuration/NetworkConfigurationServiceProperties.java
marcellorinaldo: Actually the `props.getOptStringList` does the conversion into `List<String>`, which is especting a comma-separated string.

mattdibi: Yeah I know. I deduce that, in this code, you don't care about what the comma-separated-string is representing. You just want the comma-separated-string. Is this correct? If so, you can ignore my comment.

marcellorinaldo: Correct

kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/server/net2/configuration/GwtNetInterfaceConfigBuilder.java
marcellorinaldo: Maybe it's better to implement those checks on the `NetworkTabsUi` and disable the IPv6 tab in case ipv4 is set unmanaged, so that the adapter code does not need to perform any checks. What do you think?

mattdibi: Yeah. This is just a representation issue therefore should be handled only by the frontend.

mattdibi (approved):

PR #4809, opened 2023-08-17
feat(web2): ipv6 status information display

mattdibi:
This PR adds the required code to display the ipv6 status information. It is a follow-up of #4806

> **Note**
While performing these changes I encountered a bug that let me discover a wrong assumption that was made in #4800. In [TabIp6Ui.getUpdatedNetInterface](https://github.com/marcellorinaldo/kura/blob/1cc4ad77dfa148a93f27952e14450c18bc39930b/kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/client/ui/network/TabIp6Ui.java#L570) we assumed that `updatedNetIf` was freshly created and empty. **This is not correct** as we can see from [`NetworkTabsUi`](https://github.com/marcellorinaldo/kura/blob/1cc4ad77dfa148a93f27952e14450c18bc39930b/kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/client/ui/network/NetworkTabsUi.java#L478). This led to the changes in af8a8a02d03128903f1bf44d7a25b2417e4f5d7b

> **Warning**
The PR in its current form is not 100% functional. While working on this we found that there were a couple of bugs affecting the frontend and UI validation. These will be addressed in a separate PR.

<details>
<summary>

## Tests

</summary>

To verify everything is working correctly I ran this PR code on a RPi 4 w/ Raspberry Pi OS

image

The output of Kura is consistent with what `nmcli` and `ip` provide:

image

image

---

With DHCP server on the LAN advertising:

image

image

</details>

### DIFF
--- a/kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/client/ui/network/TabIp6Ui.java
+++ b/kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/client/ui/network/TabIp6Ui.java
@@ -470,9 +470,6 @@ public class TabIp6Ui extends Composite implements NetworkTab {
             refreshFieldsBasedOnInterface(this.selectedNetIfConfig.get());
             refreshFieldsBasedOnSelectedValues();
         }
-
-        // Show read-only dns field when there are no custom DNS entries
-        this.dnsRead.setVisible(this.dns.getValue() == null || this.dns.getValue().isEmpty());
     }
 
     private void refreshFieldsBasedOnInterface(GwtNetInterfaceConfig config) {
@@ -546,6 +543,9 @@ public class TabIp6Ui extends Composite implements NetworkTab {
             this.autoconfiguration.setEnabled(false);
             this.privacy.setEnabled(false);
         }
+
+        // Show read-only dns field when there are no custom DNS entries
+        this.dnsRead.setVisible(this.dns.getValue() == null || this.dns.getValue().isEmpty());
     }
 
     @Override
--- a/kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/client/ui/network/TabIp6Ui.java
+++ b/kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/client/ui/network/TabIp6Ui.java
@@ -470,6 +470,16 @@ public class TabIp6Ui extends Composite implements NetworkTab {
             refreshFieldsBasedOnInterface(this.selectedNetIfConfig.get());
             refreshFieldsBasedOnSelectedValues();
         }
+
+        // Show read-only dns field when DHCP is selected and there are no
+        // custom DNS entries
+        String configureValue = this.configure.getSelectedItemText();
+        if (configureValue.equals("netIPv6MethodDhcp")
+                && (this.dns.getValue() == null || this.dns.getValue().isEmpty())) {
+            this.dnsRead.setVisible(true);
+        } else {
+            this.dnsRead.setVisible(false);
+        }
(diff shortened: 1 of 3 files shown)


marcellorinaldo (dismissed):

marcellorinaldo (dismissed):

kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/client/ui/network/TabIp6Ui.java
```diff
@@ -470,6 +470,16 @@ private void refreshForm() {
             refreshFieldsBasedOnInterface(this.selectedNetIfConfig.get());
             refreshFieldsBasedOnSelectedValues();
         }
+
```
marcellorinaldo: Doe it make sense to put this into the `refreshFieldsBasedOnSelectedValues` method?

kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/client/ui/network/TabIp6Ui.java
```diff
@@ -470,6 +470,16 @@ private void refreshForm() {
...
+        // Show read-only dns field when DHCP is selected and there are no
+        // custom DNS entries
+        String configureValue = this.configure.getSelectedItemText();
+        if (configureValue.equals("netIPv6MethodDhcp")
+                && (this.dns.getValue() == null || this.dns.getValue().isEmpty())) {
```
marcellorinaldo: `getValue` returns `null` if the field is empty or the parsing fails, so there is no need to call `getValue().isEmpty()`

mattdibi: 🤔 I tested with only checking for `null` and it didn't work correctly. I'll leave it as-is for now.

kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/client/ui/network/TabIp6Ui.java
mattdibi: Addressed in e3010a7ac9a202fc84f7b99fa5bfc199a567da27

kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/client/ui/network/TabIp6Ui.java
```diff
@@ -471,15 +471,8 @@ private void refreshForm() {
...
-        // custom DNS entries
-        String configureValue = this.configure.getSelectedItemText();
-        if (configureValue.equals("netIPv6MethodDhcp")
-                && (this.dns.getValue() == null || this.dns.getValue().isEmpty())) {
-            this.dnsRead.setVisible(true);
-        } else {
-            this.dnsRead.setVisible(false);
-        }
+        // Show read-only dns field when there are no custom DNS entries
+        this.dnsRead.setVisible(this.dns.getValue() == null || this.dns.getValue().isEmpty());
```
marcellorinaldo: ` this.dns.getValue().isEmpty()` is not needed

marcellorinaldo: Nevermind, didn't see the previous reply

marcellorinaldo (approved):
Will solve frontend bugs in a separate PR

PR #4814, opened 2023-08-23
fix(web2): correctly report gateway and WAN priority when not set

mattdibi:
This PR fixes an issue that emerged while testing IpV6 w/ SLAAC and DHCPv6.

## Issue description

Steps to reproduce:
- On an interface set the IpV6 Configuration as “Manually” and fill all the required fields.
- Press “Apply”
- Check that the configuration was correctly applied through `nmcli`
- On that same interface set IpV6 Configuration as “DHCPv6” and press “Apply”. Be sure that, on your LAN, nothing provides the gateway address
- The gateway address shown by the UI **will still be the one provided at the step before (i.e. the manually set one).**
- `nmcli` will report **no gateway set**

<details>
<summary>Screenshots</summary>

`nmcli con show kura-eth0-connection` reports:

image-20230822-072338

Meanwhile the UI reports the following:

image-20230822-072453

</details>

---

Similar behaviour for the "WAN priority" field

https://github.com/eclipse/kura/assets/22748355/4f09b1b3-f5b8-46ff-b5d4-ce0d62a9610a

## Fix description

I updated `NetworkStatusServiceAdapter` so that when the IP address of the gateway is missing from the status **it actually sets it as empty** instead of not setting it at all. This prevents the UI to fill the field with the cached configuration.

This behaviour is similar to what we do with the DNS servers: if the DNS server list is empty **we actually set it to empty** instead of not setting it at all. See:

https://github.com/eclipse/kura/blob/93ad9a86ba00b41b8b86a5ba0a29330141c47e8f/kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/server/net2/status/NetworkStatusServiceAdapter.java#L216

> **Note**
> Notice that the issue affected the IpV4 Gateway too and therefore I updated that accordingly.

---

The fix is similar for the "WAN priority" case.

### DIFF
--- a/kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/server/net2/status/NetworkStatusServiceAdapter.java
+++ b/kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/server/net2/status/NetworkStatusServiceAdapter.java
@@ -59,6 +59,8 @@ import org.slf4j.LoggerFactory;
  */
 public class NetworkStatusServiceAdapter {
 
+    private final static String EMPTY_STRING = "";
+
     private final Logger logger = LoggerFactory.getLogger(NetworkStatusServiceAdapter.class);
     private final NetworkStatusService networkStatusService;
 
@@ -210,9 +212,11 @@ public class NetworkStatusServiceAdapter {
                     gwtConfig.setIpAddress(firstAddress.getAddress().getHostAddress());
                     gwtConfig.setSubnetMask(NetworkUtil.getNetmaskStringForm(firstAddress.getPrefix()));
                 }
-                if (address.getGateway().isPresent()) {
-                    gwtConfig.setGateway(address.getGateway().get().getHostAddress());
-                }
+
+                String gatewayAddress = address.getGateway().isPresent() ? address.getGateway().get().getHostAddress()
+                        : EMPTY_STRING;
+                gwtConfig.setGateway(gatewayAddress);
+
                 gwtConfig.setReadOnlyDnsServers(prettyPrintDnsServers(address.getDnsServerAddresses()));
             });
         }
@@ -236,9 +240,11 @@ public class NetworkStatusServiceAdapter {
                     gwtConfig.setIpv6Address(firstAddress.getAddress().getHostAddress());
                     gwtConfig.setIpv6SubnetMask((int) firstAddress.getPrefix());
                 }
-                if (address.getGateway().isPresent()) {
-                    gwtConfig.setIpv6Gateway(address.getGateway().get().getHostAddress());
-                }
+
+                String gatewayAddress = address.getGateway().isPresent() ? address.getGateway().get().getHostAddress()
+                        : EMPTY_STRING;
+                gwtConfig.setIpv6Gateway(gatewayAddress);
+
                 gwtConfig.setIpv6ReadOnlyDnsServers(prettyPrintDnsServers(address.getDnsServerAddresses()));
             });
         }
--- a/kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/server/net2/configuration/GwtNetInterfaceConfigBuilder.java
+++ b/kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/server/net2/configuration/GwtNetInterfaceConfigBuilder.java
@@ -31,6 +31,7 @@ import org.eclipse.kura.web.shared.model.GwtWifiWirelessMode;
 public class GwtNetInterfaceConfigBuilder {
 
     private static final String NA = "N/A";
+    private static final Integer DEFAULT_WAN_PRIORITY = -1;
 
     private final NetworkConfigurationServiceProperties properties;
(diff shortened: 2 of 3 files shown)


kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/server/net2/status/NetworkStatusServiceAdapter.java
```diff
@@ -210,9 +212,11 @@ private void setIpv4DhcpClientProperties(GwtNetInterfaceConfig gwtConfig,
                     gwtConfig.setIpAddress(firstAddress.getAddress().getHostAddress());
                     gwtConfig.setSubnetMask(NetworkUtil.getNetmaskStringForm(firstAddress.getPrefix()));
                 }
-                if (address.getGateway().isPresent()) {
-                    gwtConfig.setGateway(address.getGateway().get().getHostAddress());
-                }
+
+                String gatewayAddress = address.getGateway().isPresent() ? address.getGateway().get().getHostAddress()
+                        : EMPTY_STRING;
```
marcellorinaldo, line 217: do we really need a constant for the empty string?

mattdibi: Eh... not really. It's just for better readability. It doesn't hurt though...

marcellorinaldo: IMHO it's not better readable, one can interpret is as a constant for 'empty' and it feels like:

```java
private static final int ONE = 1;
```

But if you want, we can keep it.

marcellorinaldo (commented):
Since we're on it in this PR, it might have sense to do the same for the WAN priority: https://github.com/eclipse/kura/blob/93ad9a86ba00b41b8b86a5ba0a29330141c47e8f/kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/server/net2/configuration/GwtNetInterfaceConfigBuilder.java#L114C10-L114C10

mattdibi:
> Since we're on it in this PR, it might have sense to do the same for the WAN priority: https://github.com/eclipse/kura/blob/93ad9a86ba00b41b8b86a5ba0a29330141c47e8f/kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/server/net2/configuration/GwtNetInterfaceConfigBuilder.java#L114C10-L114C10

Mh... I'd like to keep things separate if possible and address this in a separate PR. What do you think?

marcellorinaldo:
> > Since we're on it in this PR, it might have sense to do the same for the WAN priority: https://github.com/eclipse/kura/blob/93ad9a86ba00b41b8b86a5ba0a29330141c47e8f/kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/server/net2/configuration/GwtNetInterfaceConfigBuilder.java#L114C10-L114C10
> 
> Mh... I'd like to keep things separate if possible and address this in a separate PR. What do you think?

It's the same issue, I think we can keep the fixes all here. Maybe we might change the PR title to "fix(web2): correct priority and gateway fields reset"

mattdibi:
> > > Since we're on it in this PR, it might have sense to do the same for the WAN priority: https://github.com/eclipse/kura/blob/93ad9a86ba00b41b8b86a5ba0a29330141c47e8f/kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/server/net2/configuration/GwtNetInterfaceConfigBuilder.java#L114C10-L114C10
> > 
> > 
> > Mh... I'd like to keep things separate if possible and address this in a separate PR. What do you think?
> 
> It's the same issue, I think we can keep the fixes all here. Maybe we might change the PR title to "fix(web2): correct priority and gateway fields reset"

Fixed in db4e9578e4d9af959ca215fb66938e7d9853051d

marcellorinaldo (changes requested):
On a fresh installation I got errors only the first time I access the Network section. Nothing is shown in the logs. Usually this is a NPE on the UI side since the tabs are not loaded.

Screenshot 2023-08-30 at 13 56 14

mattdibi:
> On a fresh installation I got errors only the first time I access the Network section. Nothing is shown in the logs. Usually this is a NPE on the UI side since the tabs are not loaded.
> 
> Screenshot 2023-08-30 at 13 56 14

I... wasn't able to reproduce? ... I just installed a fresh version of my branch and no error was thrown...

Ah... the joys of working on the UI

kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/server/net2/status/NetworkStatusServiceAdapter.java
mattdibi: I'll keep it

marcellorinaldo (approved):
Let's merge it, if it happens again we will sort it out.

PR #4816, opened 2023-08-24
feat: adding support for vlan via networkmanager

fdizazzo:
This enabled management of VLANs via NetworkManager as discussed in #4808 

A VLAN can be created by uploading a snapshot XML file, attached is an example creating a vlan with tag 40 on physical interface eth0.
Created VLAN can then be managed just as Ethernet interfaces.

New configuration properties, mapping fields described [in NetworkManager api documentation](https://networkmanager.dev/docs/api/latest/settings-vlan.html), are:
`net.interface.{vlanInterfaceName}.config.vlan.parent` [required]: Physical interface Vlan is bound to (String)
`net.interface.{vlanInterfaceName}.config.vlan.id` [required]: Vlan tag identifier, between 0 and 4094 (Integer)
`net.interface.{vlanInterfaceName}.config.vlan.ingress` [optional]: Incoming traffic priorities in the format from:to, as comma separated pairs of un unsigned integer (String)
`net.interface.{vlanInterfaceName}.config.vlan.egress` [optional]: Outgoing traffic priorities in the format from:to, as comma separated pairs of un unsigned integer (String)
`net.interface.{vlanInterfaceName}.config.vlan.flags` [optional]: Configuration flags, between 0 and 15, default 1 (Integer)

image

[vlan_creation_example.zip](https://github.com/eclipse/kura/files/12430058/vlan_creation_example.zip)


kura/distrib/src/main/resources/common/99-kura-nm.conf
```diff
@@ -5,12 +5,12 @@ plugins=ifupdown,keyfile
 dhcp=dhclient
 
 [ifupdown]
-managed=true
+managed=false
```
pierantoniomerlino: Why do do change this property? In this way NM cannot see any interface listed in `/etc/network/interfaces'.

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/net/status/vlan/VlanInterfaceStatus.java
```diff
@@ -0,0 +1,106 @@
...
+
+import org.eclipse.kura.net.status.NetworkInterfaceStatus;
+import org.eclipse.kura.net.status.NetworkInterfaceType;
+import org.osgi.annotation.versioning.ProviderType;
+
+/**
+ * Class that contains specific properties to describe the status of a
+ * Vlan.
+ *
+ */
```
pierantoniomerlino, line 26: Add `@since 2.6` annotation please.

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/net/status/vlan/VlanInterfaceStatus.java
```diff
@@ -0,0 +1,106 @@
...
+    private VlanInterfaceStatus(VlanInterfaceStatusBuilder builder) {
+        super(builder);
+        this.vlanId = builder.vlanId;
+        this.parentInterface = builder.parentInterface;
+        this.flags = builder.flags;
+    }
+
+    public int getVlanId() {
+        return vlanId;
+    }
```
pierantoniomerlino, line 41: Please `this.` before the variable name. The same applies below. Did you applied the clean-up?

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/net/status/vlan/VlanInterfaceStatus.java
```diff
@@ -0,0 +1,106 @@
...
+ * Vlan.
+ *
+ */
+@ProviderType
+public class VlanInterfaceStatus extends NetworkInterfaceStatus {
+
+    private final int flags;
+    private final String parentInterface;
+    private final int vlanId;
+
```
pierantoniomerlino, line 32: Are these status properties or configuration? If this is only configuration, probably this class is not needed.

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/net/vlan/VlanInterface.java
```diff
@@ -0,0 +1,48 @@
...
+     */
+    public String getParentInterface();
+    
+    /**
+     * Indicates the configured Vlan tag.
+     *
+     * @return
+     */
+    public int getVlanId();
+
```
pierantoniomerlino, line 65: Are the `ingress` and `egress` properties configurable? Should they added here?

kura/org.eclipse.kura.core.net/src/main/java/org/eclipse/kura/core/net/vlan/VlanInterfaceImpl.java
```diff
@@ -0,0 +1,123 @@
+/*******************************************************************************
+ * Copyright (c) 2011, 2023 Eurotech and/or its affiliates and others
+ *
```
pierantoniomerlino, line 3: Set the copyright year to 2023 only since it is a new file.

pierantoniomerlino:
@fdizazzo Cloud you please provide us a (partial) snapshot with a VLAN configuration for testing?

kura/distrib/src/main/resources/common/99-kura-nm.conf
fdizazzo: This was indeed a mistake in testing, such a change is not needed. Reverting it on next commit.

fdizazzo:
Here's two:
[vlan_creation_example.txt](https://github.com/eclipse/kura/files/12452058/vlan_creation_example.txt) creates a vlan with tagId 40 over interface ens33, using just the required properties `vlan.id` and `vlan.parent` named _ens33.40_

[vlan_creation_complete.txt](https://github.com/eclipse/kura/files/12452069/vlan_creation_complete.txt) creates a vlan with tagId 41 over interface ens33 using all supported properties including `vlan.flags`, `vlan.ingress` and `vlan.egress` named _vlanFull_

For your tests be sure to change the content of net.interfaces property so it contains the names of the interfaces on your test environment, plus the names of vlan you're creating/already created. Likewise be sure that `vlan.parent` property references an existing network interface.

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/net/status/vlan/VlanInterfaceStatus.java
fdizazzo: Won't this be needed to display the vlan status on Kura Status page? vlanId and parentInterface could be useful for quick reference by the user

fdizazzo: Removed `flags` variable, as it's strictly related to configuration

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/net/vlan/VlanInterface.java
fdizazzo: `ingress` and `egress` are configurable. Addredd here and in its implementation class

mattdibi:
Hello there @fdizazzo! Sorry for taking so long...

First and foremost: good job! I tested it and it's working fine on my RPi. There's only one issue I want to talk about (but I think I also have a solution for that).

The issue is, as per our discussion #4808, you're duplicating the entire network configuration flow to handle the VLAN case and thus the VLAN connection gets configured two times:

image

Furthermore this also reflect in the code duplication: `checkVlanConfiguration`, `setupVlan` and `getInterfaceStatusFromProperties` are just duplicating the normal flow of operation.

...  but here comes the neat part...

As already discussed, we only need to handle the special case in which the device to be configured is not present on the system *and* is of VLAN type. Everything else is already done by Kura.

If, in the `doApply` method, we iterate over the superset of the available devices *and* the configured devices we can handle *all* our required cases.

Here's the idea:

```java
private synchronized void doApply(Map<String, Object> networkConfiguration) throws DBusException {
    logger.info("Applying configuration using NetworkManager Dbus connector");
    NetworkProperties properties = new NetworkProperties(networkConfiguration);
    List<String> configuredInterfaceIds = properties.getStringList("net.interfaces");
    List<String> availableDevicesIds = getInterfaceIds();

    List<String> availableDevices = configuredInterfaceIds + availableDevicesIds; // Merge without duplicates

    availableDevices.forEach(deviceId -> {
        try {
            doApply(deviceId, networkConfiguration);
        } catch (DBusException | DBusExecutionException | IllegalArgumentException | NoSuchElementException e) {
            logger.error("Unable to apply configuration to the device {}", deviceId, e);
        }
    });
}

private synchronized void doApply(String deviceIdToBeConfigured, Map<String, Object> networkConfiguration)
        throws DBusException {
    NetworkProperties properties = new NetworkProperties(networkConfiguration);
    List<String> configuredInterfaceIds = properties.getStringList("net.interfaces");

    Optional<Device> device = getNetworkManagerDeviceByInterfaceId(deviceIdToBeConfigured);
    if (device.isPresent()) {
        if (configuredInterfaceIds.contains(deviceIdToBeConfigured)) {
            manageConfiguredInterface(device.get(), deviceIdToBeConfigured, properties);
        } else {
            manageNonConfiguredInterface(device.get(), deviceIdToBeConfigured);
        }
    } else {
        if (isVlanType(deviceIdToBeConfigured)) {
            manageConfiguredInterface(Optional.empty(), deviceIdToBeConfigured, properties);
        } else {
            logger.warn("Device \"{}\" not found. Skipping configuration.", deviceIdToBeConfigured);
        }
    }
}
```

> **Note** 
This code is just for explaination purpose. It doesn't compile.

The amount of changes required should be minimal and we should still be able to handle the VLAN correctly.

1. If the device exists on the system and is configured (i.e. in the snapshot) -> it gets configured
2. If the device exists on the system and is not confgiured -> it gets disabled
3. if the device doesn't exists, but it's not a VLAN -> we do nothing
4. if the device doesn't exists and it's of VLAN type -> it gets configured

What do you think?

fdizazzo:
Doable and done. No difference in behaviour in both manual and existing junit tests, but i had to change several (private) methods' signatures in NMDbusConnector so an additional look at it would be nice :)

kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/NMDbusConnector.java
```diff
@@ -359,13 +382,18 @@ public synchronized void apply(String deviceId) throws DBusException {
...
     private synchronized void doApply(Map<String, Object> networkConfiguration) throws DBusException {
         logger.info("Applying configuration using NetworkManager Dbus connector");
-        List<Device> availableDevices = this.networkManager.getAllDevices();
+        NetworkProperties properties = new NetworkProperties(networkConfiguration);
+        List<String> availableDeviceIds = getInterfaceIds();
+        Set<String> availableDevices = new LinkedHashSet<String>(availableDeviceIds);
+        Optional<List<String>> configuredInterfaceIds = properties.getOptStringList("net.interfaces");
+        if (configuredInterfaceIds.isPresent()) {
+            availableDevices.addAll(configuredInterfaceIds.get());
+        }
```
mattdibi, line 391: I think this can be simplified since `net.interface` property is not optional.

```suggestion
        NetworkProperties properties = new NetworkProperties(networkConfiguration);
        List<String> availableDeviceIds = getInterfaceIds();
        List<String>> configuredInterfaceIds = properties.getStringList("net.interfaces");

        Set<String> availableDevices = new LinkedHashSet<String>(availableDeviceIds);
        availableDevices.addAll(configuredInterfaceIds);
```

fdizazzo, line 393: I agree, but it breaks this test
https://github.com/eclipse/kura/blob/f99573b3fe5efd6ff7ea2411e54982f33d3cce8e/kura/test/org.eclipse.kura.nm.test/src/test/java/org/eclipse/kura/nm/NMDbusConnectorTest.java#L327-L341

Could a better way be something like this?
```
try{
    List<String>> configuredInterfaceIds = properties.getStringList("net.interfaces");
} catch (NoSuchElementException e) {
    //No configured interface, doing nothing
    return;
}
```

mattdibi:
@fdizazzo Yeah sorry, I need to put more thought into this.

In the meantime thanks for the contribution! This is exactly the approach I was thinking. On Monday (hopefully) I'll provide a more in depth review :)

kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/NMDbusConnector.java
```diff
@@ -378,16 +406,35 @@ private synchronized void doApply(String deviceIdToBeConfigured, Map<String, Obj
...
-                manageConfiguredInterface(device.get(), deviceIdToBeConfigured, properties);
+                manageConfiguredInterface(device, deviceIdToBeConfigured, properties, Optional.empty());
             } else {
-                manageNonConfiguredInterface(device.get(), deviceIdToBeConfigured);
+                manageNonConfiguredInterface(device, deviceIdToBeConfigured);
+            }
+        } else {
+            if (isVlanType(deviceIdToBeConfigured, networkConfiguration)) {
+                manageConfiguredInterface(device, deviceIdToBeConfigured, properties, 
+                        Optional.of(NMDeviceType.NM_DEVICE_TYPE_VLAN));
```
mattdibi: There's a couple of issues here:
1. I think we can keep things a little bit more generic. This now works for VLAN device types, but the approach should work (in theory) with every device that can be created by NetworkManager (bridges for example).
2. To retrieve the device type I suggest you to not use the ` NetworkConfigurationServiceCommon.getNetworkTypeFromProperties` method since it belongs to the old networking implementation and there's no actual need for the `NetInterfaceType` here. I would add a new method inside [the `NMDeviceType` enum](https://github.com/eclipse/kura/blob/develop/kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/enums/NMDeviceType.java) so that it can parse the type reported in the snapshot.
3. I would not change the `manageConfiguredInterface` method signature to accept the device type since it's something that belong to the lower abstraction levels (apart from the VLAN).
4. Just for readability, instead of passing the `device` (which is an `Optional.empty()` in this case), I would explicitly pass its value

Putting everything together this is what I propose:

```suggestion
        } else {
             // Handle the case in which the property is not set as you prefer
            NMDeviceType propertyDeviceType = NMDeviceType.fromProperties(properties.getString("net.interface.%s.type", deviceIdToBeConfigured)); 
            if (VIRTUAL_DEVICE_TYPE.contains(propertyDeviceType)) {
                manageConfiguredInterface(Optional.empty(), deviceIdToBeConfigured, properties);
```

Where:

```java
    private static final List<NMDeviceType> VIRTUAL_DEVICE_TYPES = Arrays.asList(NMDeviceType.NM_DEVICE_TYPE_VLAN);
```

kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/NMDbusConnector.java
```diff
@@ -378,16 +406,35 @@ private synchronized void doApply(String deviceIdToBeConfigured, Map<String, Obj
...
+            Optional<NMDeviceType> knownDeviceType) throws DBusException {
+        NMDeviceType deviceType;
+        if (device.isPresent()) {
+            deviceType = this.networkManager.getDeviceType(device.get().getObjectPath());
+        } else if (knownDeviceType.isPresent()) {
+            deviceType = knownDeviceType.get();
+        } else {
+            logger.error("Unable to apply configuration to device {}: no device nor device type provided");
+            return;
+        }
```
mattdibi: Following what was said in the previous comment:

```suggestion
    private synchronized void manageConfiguredInterface(Optional<Device> device, String deviceId, NetworkProperties properties) throws DBusException {
        NMDeviceType deviceType;
        if (device.isPresent()) {
            deviceType = this.networkManager.getDeviceType(device.get().getObjectPath());
        } else {
            deviceType = NMDeviceType.fromProperties(properties.getString("net.interface.%s.type", deviceId)); 
        ...
            // Again, handle the case in which the type is not set as you prefer
            logger.error("Unable to apply configuration to device {}: no device nor device type provided");
            return;
        }
```

kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/NMDbusConnector.java
```diff
@@ -415,36 +462,46 @@ private synchronized void manageConfiguredInterface(Device device, String device
...
                     deviceId);
-            Optional<String> mmDbusPath = this.networkManager.getModemManagerDbusPath(device.getObjectPath());
+            Optional<String> mmDbusPath = this.networkManager.getModemManagerDbusPath(device.get().getObjectPath());
             this.modemManager.setGPS(mmDbusPath, enableGPS);
         }
 
     }
 
-    private void enableInterface(String deviceId, NetworkProperties properties, Device device, NMDeviceType deviceType)
+    private void enableInterface(String deviceId, NetworkProperties properties, Optional<Device> optDevice, NMDeviceType deviceType)
```
mattdibi: I think this is way more complicated than it needs to be (or I might be oversimplifying things :D)

The logic is the following:
- If there's no device we just need to create the connection (since when the connection gets created the virtual device will spawn too)
- If there's no device there cannot be an associated connection for which we need to update the settings
- If there's no device there cannot be any additional associated connections to cleanup in the housekeeping stage

Therefore this should be simplified as follows:

```java
    private void enableInterface(String deviceId, NetworkProperties properties, Optional<Device> device, NMDeviceType deviceType)
            throws DBusException {
        if(device.isPresent()) {
            enableInterface(deviceId, properties, device.get(), deviceType);
        } else {
            enableVirtualInterface(deviceId, properties, deviceType);
        }
    }

    private void enableVirtualInterface(String deviceId, NetworkProperties properties, NMDeviceType deviceType)
            throws DBusException {
        Map<String, Map<String, Variant<?>>> newConnectionSettings = NMSettingsConverter.buildSettings(properties,
                Optional.empty(), deviceId, deviceId, deviceType);
        try {
            this.networkManager.AddAndActivateConnection(
                    newConnectionSettings, new DBusPath("/"), new DBusPath("/"));
        } catch (DBusExecutionException e) {
            logger.warn("Couldn't complete activation of {} interface, caused by:", deviceId, e);
        }
    }

    private void enableInterface(String deviceId, NetworkProperties properties, Device device, NMDeviceType deviceType)
            throws DBusException {
        // this is the old method without any change
```

> **Warning**
Here I'm assuming that, since we're creating and activating a connection on a device that not yet exists, we need to pass to the `AddAndActivateConnection` method a "/" dbus path. I didn't have the time to test this assumption though.

Additionally here there's no signal that we can wait for to understand if the connection was applied (since there's no device yet). A possible solution would be to wait for the `DeviceAdded` signal?

mattdibi (changes requested):

kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/NMDbusConnector.java
mattdibi: Yeah after looking into it my suggestion was not correct. Using `getOptStringList` is the right choice here.

kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/NMDbusConnector.java
mattdibi: Mh... apparently this fails with:

```
2023-09-05T08:41:01,589 [ConfigurationListener Event Queue] WARN  o.e.k.n.NMDbusConnector - Couldn't complete activation of eth0.40 interface, caused by:
org.freedesktop.dbus.exceptions.DBusExecutionException: Failed to find a compatible device for this connection
	at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
	at sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
```

So it needs to be a little more complicated than what I wrote before but I think the principle should still stand...

mattdibi: This seems to be working:

```java
   private void enableInterface(String deviceId, NetworkProperties properties, Optional<Device> device,
            NMDeviceType deviceType) throws DBusException {
        if (device.isPresent()) {
            enableInterface(deviceId, properties, device.get(), deviceType);
        } else {
            enableVirtualInterface(deviceId, properties, deviceType);
        }
    }

    private void enableVirtualInterface(String deviceId, NetworkProperties properties, NMDeviceType deviceType)
            throws DBusException {
        Map<String, Map<String, Variant<?>>> newConnectionSettings = NMSettingsConverter.buildSettings(properties,
                Optional.empty(), deviceId, deviceId, deviceType);

        Settings settings = this.dbusConnection.getRemoteObject(NM_BUS_NAME, NM_SETTINGS_BUS_PATH, Settings.class);
        DBusPath createdConnectionPath = settings.AddConnection(newConnectionSettings); // The virtual device is created HERE
        Connection createdConnection = this.dbusConnection.getRemoteObject(NM_BUS_NAME, createdConnectionPath.getPath(),
                Connection.class);

        try {
            this.networkManager.activateConnection(createdConnection, new DBusPath("/"));
        } catch (DBusExecutionException e) {
            logger.warn("Couldn't complete activation of {} interface, caused by:", deviceId, e);
        }
    }

    private void enableInterface(String deviceId, NetworkProperties properties, Device device, NMDeviceType deviceType)
            throws DBusException {
   // This is the old method without changes
```

Where `NetworkManagerDbusWrapper.java` was added:
```java
public void activateConnection(Connection connection, DBusPath dBusPath) {
        this.networkManager.ActivateConnection(new DBusPath(connection.getObjectPath()), dBusPath, new DBusPath("/"));
    }
```

but I would update the existing `activateConnection` method in the NetworkManager wrapper to just use `DBusPath` instead of `Device` as argument.

<details>

<summary> Note </summary>

For completeness I add here another working version:

```java
    private void enableVirtualInterface(String deviceId, NetworkProperties properties, NMDeviceType deviceType)
            throws DBusException {
        Map<String, Map<String, Variant<?>>> newConnectionSettings = NMSettingsConverter.buildSettings(properties,
                Optional.empty(), deviceId, deviceId, deviceType);

        Settings settings = this.dbusConnection.getRemoteObject(NM_BUS_NAME, NM_SETTINGS_BUS_PATH, Settings.class);
        DBusPath createdConnectionPath = settings.AddConnection(newConnectionSettings); // The virtual device is created HERE
        Connection createdConnection = this.dbusConnection.getRemoteObject(NM_BUS_NAME, createdConnectionPath.getPath(),
                Connection.class);
        Device device = this.networkManager.getDeviceByIpIface(deviceId); // This is just a call to the GetDeviceByIpIface method exposed by NM

        try {
            this.networkManager.activateConnection(createdConnection, device);
        } catch (DBusExecutionException e) {
            logger.warn("Couldn't complete activation of {} interface, caused by:", deviceId, e);
        }
    }
```

Where:

```java
public Device getDeviceByIpIface(String ipIface) throws DBusException {
        DBusPath devicePath = this.networkManager.GetDeviceByIpIface(ipIface);
        return this.dbusConnection.getRemoteObject(NM_BUS_NAME, devicePath.getPath(), Device.class);
    }
```

</details>

kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/NMDbusConnector.java
```diff
@@ -378,16 +406,35 @@ private synchronized void doApply(String deviceIdToBeConfigured, Map<String, Obj
...
+    
+    private synchronized void manageConfiguredInterface(Optional<Device> device, String deviceId, NetworkProperties properties, 
+            Optional<NMDeviceType> knownDeviceType) throws DBusException {
+        NMDeviceType deviceType;
+        if (device.isPresent()) {
+            deviceType = this.networkManager.getDeviceType(device.get().getObjectPath());
+        } else if (knownDeviceType.isPresent()) {
+            deviceType = knownDeviceType.get();
+        } else {
+            logger.error("Unable to apply configuration to device {}: no device nor device type provided");
```
mattdibi: Missing argument

```suggestion
            logger.error("Unable to apply configuration to device {}: no device nor device type provided", deviceId);
```

kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/NMDbusConnector.java
mattdibi: **Note**: This change also addresses some of the critical issues reported [by Sonar](https://github.com/eclipse/kura/pull/4816/checks?check_run_id=16418938420) so we're kinda killing two birds with one stone.

MMaiero:
Hello, there are conflicts that need to be updated as well

kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/NMDbusConnector.java
```diff
@@ -481,8 +534,42 @@ private void enableInterface(String deviceId, NetworkProperties properties, Devi
...
+            }
+            Device createdDevice = returnedDevice.get();
+            if (Boolean.FALSE.equals(this.networkManager.isDeviceManaged(createdDevice))) {
+                this.networkManager.setDeviceManaged(createdDevice, true);
+            }
+            DeviceStateLock dsLock = new DeviceStateLock(this.dbusConnection, createdDevice.getObjectPath(),
+                    NMDeviceState.NM_DEVICE_STATE_ACTIVATED);
+            this.networkManager.activateConnection(createdConnection, createdDevice);
+            dsLock.waitForSignal();
+        } catch (DBusException | TimeoutException e) {
```
mattdibi: For debugging reason I would add also `DBusExecutionException` here.

When something breaks at runtime, the underlying library usually throws a `DBusExecutionException` instead of a `DBusException`.

```suggestion
        } catch (DBusExecutionException | DBusException | TimeoutException e) {
```

kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/signal/handlers/DeviceCreationLock.java
```diff
@@ -0,0 +1,57 @@
...
+        this.deviceQueue = new ArrayBlockingQueue<Device>(1);
+        this.dbusConnection = nm.getDbusConnection();
+        this.handler = new NMDeviceCreationHandler(nm, deviceId, this.deviceQueue);
+
+        this.dbusConnection.addSigHandler(NetworkManager.DeviceAdded.class, this.handler);
+    }
+
+    public Optional<Device> waitForDeviceCreation() throws DBusException, TimeoutException {
+        try {
+            return Optional.ofNullable(deviceQueue.poll(5L, TimeUnit.SECONDS));
```
mattdibi: I would add the timeout as a parameter for testability reasons. Like we did [in the `WPAScanLock`](https://github.com/eclipse/kura/blob/develop/kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/signal/handlers/WPAScanLock.java)

```suggestion
    public Optional<Device> waitForDeviceCreation(long deviceCreationTimeoutSeconds) throws DBusException, TimeoutException {
        try {
            return Optional.ofNullable(deviceQueue.poll(deviceCreationTimeoutSeconds, TimeUnit.SECONDS));
```

Additionally, I think a 1 second timeout might be enough and allow for quicker unit tests but I might be wrong. Feel free to correct me on this.

kura/org.eclipse.kura.net.configuration/src/main/java/org/eclipse/kura/net/configuration/NetworkConfigurationServiceCommon.java
```diff
@@ -108,6 +108,15 @@ public static Tocd getDefinition(Map<String, Object> properties, Optional<List<U
                     getDnsDefinition(tocd, ifaceName);
                     getModemDefinition(tocd, ifaceName);
                     break;
+                case VLAN:
+                    if (usbNetDevices.isPresent()) {
+                        getUsbDeviceDefinition(usbNetDevices.get(), tocd, ifaceName);
+                    }
```
mattdibi: Sonar is whining [about Cognitive Complexity](https://sonarcloud.io/project/issues?resolved=false&severities=CRITICAL&types=CODE_SMELL&branch=PR-4816&id=org.eclipse.kura%3Akura&open=AYooGPqpeo7TZ9t3jvzn) here.

I'll teach you a simple trick to make it shut up: if you substitute each `if` statement with a `ifPresent` call it won't add to the cognitive complexity computation 🤫 😉 

For example:

```suggestion
                    usbNetDevices.ifPresent(usbNetDevice -> getUsbDeviceDefinition(usbNetDevice, tocd, ifaceName));
```

This should take care of this Critical Issue

mattdibi (changes requested):
Good job @fdizazzo! I think we're almost there with the PR, we just need to make Sonar happy a final round of tests and I think we can merge (if @pierantoniomerlino agrees that is).

We have 3 issues reported by Sonar:
1. [Cognitive complexity on the `NetworkConfigurationServiceCommon.java`](https://sonarcloud.io/project/issues?resolved=false&severities=CRITICAL&types=CODE_SMELL&branch=PR-4816&id=org.eclipse.kura%3Akura&open=AYooGPqpeo7TZ9t3jvzn). It should be taken care of by my suggestion.
2. [Usage of wildcard type](https://sonarcloud.io/project/issues?resolved=false&severities=CRITICAL&types=CODE_SMELL&branch=PR-4816&id=org.eclipse.kura%3Akura&open=AYooGPpyeo7TZ9t3jvzm). It is dictated by the underlying library so we'll just set it as "Won't fix" (@MMaiero can you take care of that?)
3. Code coverage.

For 3. we can tackle the low hanging fruits:
1. Add the tests for the `NMSettingsConverter` inside [`NMSettingsConverterTest`](https://github.com/eclipse/kura/blob/develop/kura/test/org.eclipse.kura.nm.test/src/test/java/org/eclipse/kura/nm/configuration/NMSettingsConverterTest.java). The convention used for test names and unit tests should be pretty easy to follow. Feel free to ping us if anything is not clear.
2. Add the tests for the NMDBusConnector. I would say that a single test that verify that:
   - Given a NM instance with a single ethernet interface `eth0`
   - Given a configuration creating a VLAN on `eth0`
   - When `apply` is called
   - Then the `Settings.AddConnection` method is called
   - Then the `NetworkManager.ActivateConnection` method is called

Hopefully this is enough to bring us above threshold

MMaiero:
> 2\. [Usage of wildcard type](https://sonarcloud.io/project/issues?resolved=false&severities=CRITICAL&types=CODE_SMELL&branch=PR-4816&id=org.eclipse.kura%3Akura&open=AYooGPpyeo7TZ9t3jvzm). It is dictated by the underlying library so we'll just set it as "Won't fix" (@MMaiero can you take care of that?)

Done

pierantoniomerlino (approved):

mattdibi:
Hello there @fdizazzo! Looks like we're almost ready to merge 🥳 

The code LGTM but, as per [Eclipse Foundation policy](https://www.eclipse.org/projects/handbook/#ip-project-content), since this PR contains more than 1k LOC we'll need to perform an IP review. You can track the progress of the request [here](https://gitlab.eclipse.org/eclipsefdn/emo-team/iplab/-/issues/10343).

Meanwhile I'll perform some more in depth testing.

mattdibi:
Testing WIP

Static IP works as expected

image

Currently investigating DHCP server...

mattdibi:
Hello there @fdizazzo!

During my testing I discovered the *tiiiiiiiniest* of bugs on the DHCP server.

When configuring a VLAN and enabling the DHCP server, no `dnsmasq` configuration was written on disk. This behaviour is caused by:

https://github.com/eclipse/kura/blob/d2fe2ff2b1e9bbba4eac91c334068091e5cf7e54/kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/configuration/NMConfigurationServiceImpl.java#L432-L434

The bug can be fixed by this simple line change:

```diff
diff --git a/kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/configuration/NMConfigurationServiceImpl.java b/kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/configuration/NMConfigurationServiceImpl.java
index 7ecd623a8..3eda684bc 100644
--- a/kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/configuration/NMConfigurationServiceImpl.java
+++ b/kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/configuration/NMConfigurationServiceImpl.java
@@ -427,7 +427,8 @@ public class NMConfigurationServiceImpl implements SelfConfiguringComponent {
         final NetInterfaceStatus status = getNetInterfaceStatus(interfaceName)
                 .orElse(NetInterfaceStatus.netIPv4StatusUnknown);
 
-        if ((type != NetInterfaceType.ETHERNET && type != NetInterfaceType.WIFI) || !isDhcpServerEnabled) {
+        if ((type != NetInterfaceType.ETHERNET && type != NetInterfaceType.VLAN && type != NetInterfaceType.WIFI)
+                || !isDhcpServerEnabled) {
             return false;
         }
 
```

I already tested it and it's working. 

image

I think that after fixing this and receiving approval by the Eclipse Foundation we should be ready to merge.

fdizazzo:
On it!
I guess it should also be changed here for nat support?
https://github.com/eclipse/kura/blob/d2fe2ff2b1e9bbba4eac91c334068091e5cf7e54/kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/configuration/NMConfigurationServiceImpl.java#L388-L389

mattdibi:
> On it! I guess it should also be changed here for nat support?
> 
> https://github.com/eclipse/kura/blob/d2fe2ff2b1e9bbba4eac91c334068091e5cf7e54/kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/configuration/NMConfigurationServiceImpl.java#L388-L389

I think so yeah. @pierantoniomerlino thoughts?

pierantoniomerlino:
Yeah... I think that a VLAN interface acts as an ethernet one, so the same configuration can be applied.

fdizazzo:
Just pushed this change.
But i've noticed an issue on this:
https://github.com/eclipse/kura/blob/d2fe2ff2b1e9bbba4eac91c334068091e5cf7e54/kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/configuration/NMConfigurationServiceImpl.java#L233-L240
as right now the legacy LinuxNetworkUtil is unable to handle vlan interfaces due to a naming mismatch in the underlying system calls:
```
ip -o link show
5: vlanFull@ens33: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue state UP mode DEFAULT group default qlen 1000\    link/ether 00:0c:29:5c:c6:ff brd ff:ff:ff:ff:ff:ff
6: ens33.40@ens33: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue state UP mode DEFAULT group default qlen 1000\    link/ether 00:0c:29:5c:c6:ff brd ff:ff:ff:ff:ff:ff
ip -o -4 addr show
5: vlanFull    inet 192.168.41.100/24 brd 192.168.41.255 scope global noprefixroute vlanFull\       valid_lft forever preferred_lft forever
6: ens33.40    inet 10.209.55.242/24 brd 10.209.55.255 scope global noprefixroute ens33.40\       valid_lft forever preferred_lft forever
```
While it should be never invoked in NMConfigurationServiceImpl and had no impact within the scope of this PR I'm not sure if it's used elsewhere and needs to be fixed.

mattdibi:
Watch out @fdizazzo, something went wrong with 780fa7de414a6f5989b4cb189b95eb736e746209 and Sonar is reporting a bug: https://sonarcloud.io/project/issues?resolved=false&types=BUG&branch=PR-4816&id=org.eclipse.kura%3Akura&open=AYp0lmiafPqe6Yh5bc9k

I think there's a simple copy paste mishap in that commit.

kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/configuration/NMConfigurationServiceImpl.java
```diff
@@ -383,8 +390,7 @@ private boolean isNatValid(String interfaceName) {
         Optional<NetInterfaceStatus> status = getNetInterfaceStatus(interfaceName);
 
         if (type.isPresent() && isNatEnabled.isPresent() && status.isPresent()) {
-            boolean isSupportedType = type.get() == NetInterfaceType.ETHERNET || type.get() == NetInterfaceType.WIFI
-                    || type.get() == NetInterfaceType.MODEM;
+            boolean isSupportedType = SUPPORTED_NAT_INTERFACE_TYPES.contains(status);
```
mattdibi: Shouldn't this be `type`?

```suggestion
            boolean isSupportedType = SUPPORTED_NAT_INTERFACE_TYPES.contains(type.get());
```

mattdibi:
> Just pushed this change. But i've noticed an issue on this:
> 
> https://github.com/eclipse/kura/blob/d2fe2ff2b1e9bbba4eac91c334068091e5cf7e54/kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/configuration/NMConfigurationServiceImpl.java#L233-L240
> 
> 
> as right now the legacy LinuxNetworkUtil is unable to handle vlan interfaces due to a naming mismatch in the underlying system calls:
> ```
> ip -o link show
> 5: vlanFull@ens33: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue state UP mode DEFAULT group default qlen 1000\    link/ether 00:0c:29:5c:c6:ff brd ff:ff:ff:ff:ff:ff
> 6: ens33.40@ens33: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue state UP mode DEFAULT group default qlen 1000\    link/ether 00:0c:29:5c:c6:ff brd ff:ff:ff:ff:ff:ff
> ip -o -4 addr show
> 5: vlanFull    inet 192.168.41.100/24 brd 192.168.41.255 scope global noprefixroute vlanFull\       valid_lft forever preferred_lft forever
> 6: ens33.40    inet 10.209.55.242/24 brd 10.209.55.255 scope global noprefixroute ens33.40\       valid_lft forever preferred_lft forever
> ```
> 
> While it should be never invoked in NMConfigurationServiceImpl and had no impact within the scope of this PR I'm not sure if it's used elsewhere and needs to be fixed.

Regarding this, it shouldn't be a problem. That is mainly taking care of modems and VLAN will never be a USB device. So... leave it as is and thanks for bringing this up :)

mattdibi (approved):
LGTM.

Waiting https://gitlab.eclipse.org/eclipsefdn/emo-team/iplab/-/issues/10343 for merging

kura/org.eclipse.kura.core.net/META-INF/MANIFEST.MF
```diff
@@ -15,11 +15,14 @@ Import-Package: org.eclipse.kura;version="[1.0,2.0)",
  org.eclipse.kura.net.dhcp;version="[1.0,2.0)",
  org.eclipse.kura.net.firewall;version="[2.0,3.0)",
  org.eclipse.kura.net.modem;version="[2.4,2.5)",
+ org.eclipse.kura.net.status.vlan;version="1.0.0",
```
MMaiero: Range needed

kura/org.eclipse.kura.core.net/META-INF/MANIFEST.MF
```diff
@@ -15,11 +15,14 @@ Import-Package: org.eclipse.kura;version="[1.0,2.0)",
...
+ org.eclipse.kura.net.vlan;version="1.0.0",
```
MMaiero: Range needed

kura/org.eclipse.kura.core.net/src/main/java/org/eclipse/kura/core/net/vlan/VlanInterfaceConfigImpl.java
```diff
@@ -0,0 +1,51 @@
+/*******************************************************************************
+ * Copyright (c) 2011, 2023 Eurotech and/or its affiliates and others
```
MMaiero: 2023 only

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/net/status/vlan/VlanInterfaceStatus.java
```diff
@@ -0,0 +1,97 @@
...
+ *
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *  Areti
+ ******************************************************************************/
```
mattdibi:
```suggestion
/*******************************************************************************
 * Copyright (c) 2023 Areti and others
 *
 * This program and the accompanying materials are made
 * available under the terms of the Eclipse Public License 2.0
 * which is available at https://www.eclipse.org/legal/epl-2.0/
 *
 * SPDX-License-Identifier: EPL-2.0
 *
 * Contributors:
 *  Areti
 *  Eurotech
 ******************************************************************************/
```

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/net/status/vlan/package-info.java
```diff
@@ -0,0 +1,20 @@
...
+ * Contributors:
+ *  Areti
+ ******************************************************************************/
+/**
+ *
+ */
+/**
+ * Provides classes for describing the status of a Vlan.
+ *
+ */
```
mattdibi:
```suggestion
/*******************************************************************************
 * Copyright (c) 2023 Areti and others
 *
 * This program and the accompanying materials are made
 * available under the terms of the Eclipse Public License 2.0
 * which is available at https://www.eclipse.org/legal/epl-2.0/
 *
 * SPDX-License-Identifier: EPL-2.0
 *
 * Contributors:
 *  Areti
 *  Eurotech
 ******************************************************************************/
/**
 * Provides classes for describing the status of a Vlan.
 *
 */
```

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/net/vlan/VlanInterface.java
```diff
@@ -0,0 +1,65 @@
...
+ * 
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ * 
+ * SPDX-License-Identifier: EPL-2.0
+ * 
+ * Contributors:
+ *  Areti
+ ******************************************************************************/
```
mattdibi:
```suggestion
/*******************************************************************************
 * Copyright (c) 2023 Areti and others
 * 
 * This program and the accompanying materials are made
 * available under the terms of the Eclipse Public License 2.0
 * which is available at https://www.eclipse.org/legal/epl-2.0/
 * 
 * SPDX-License-Identifier: EPL-2.0
 * 
 * Contributors:
 *  Areti
 *  Eurotech
 ******************************************************************************/
```

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/net/vlan/package-info.java
```diff
@@ -0,0 +1,20 @@
...
+ * Contributors:
+ *  Areti
+ ******************************************************************************/
+/**
+ *
+ */
+/**
+ * Provides interfaces for instances and configurations of  Vlan.
+ *
+ */
```
mattdibi:
```suggestion
/*******************************************************************************
 * Copyright (c) 2023 Areti and others
 *
 * This program and the accompanying materials are made
 * available under the terms of the Eclipse Public License 2.0
 * which is available at https://www.eclipse.org/legal/epl-2.0/
 *
 * SPDX-License-Identifier: EPL-2.0
 *
 * Contributors:
 *  Areti
 *  Eurotech
 ******************************************************************************/
/**
 * Provides interfaces for instances and configurations of  Vlan.
 *
 */
```

kura/org.eclipse.kura.core.net/src/main/java/org/eclipse/kura/core/net/vlan/VlanInterfaceConfigImpl.java
```diff
@@ -0,0 +1,51 @@
...
+ * 
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ * 
+ * SPDX-License-Identifier: EPL-2.0
+ * 
+ * Contributors:
+ *  Areti
+ *******************************************************************************/
```
mattdibi:
```suggestion
/*******************************************************************************
 * Copyright (c) 2023 Areti and others
 * 
 * This program and the accompanying materials are made
 * available under the terms of the Eclipse Public License 2.0
 * which is available at https://www.eclipse.org/legal/epl-2.0/
 * 
 * SPDX-License-Identifier: EPL-2.0
 * 
 * Contributors:
 *  Areti
 *  Eurotech
 *******************************************************************************/
```

kura/org.eclipse.kura.core.net/src/main/java/org/eclipse/kura/core/net/vlan/VlanInterfaceImpl.java
```diff
@@ -0,0 +1,144 @@
...
+ * 
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ * 
+ * SPDX-License-Identifier: EPL-2.0
+ * 
+ * Contributors:
+ *  Areti
+ *******************************************************************************/
```
mattdibi:
```suggestion
/*******************************************************************************
 * Copyright (c) 2023 Areti and others
 * 
 * This program and the accompanying materials are made
 * available under the terms of the Eclipse Public License 2.0
 * which is available at https://www.eclipse.org/legal/epl-2.0/
 * 
 * SPDX-License-Identifier: EPL-2.0
 * 
 * Contributors:
 *  Areti
 *  Eurotech
 *******************************************************************************/
```

kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/signal/handlers/DeviceCreationLock.java
```diff
@@ -0,0 +1,61 @@
...
+ *
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *  Areti
+ *******************************************************************************/
```
mattdibi:
```suggestion
/*******************************************************************************
 * Copyright (c) 2023 Areti and others
 * 
 * This program and the accompanying materials are made
 * available under the terms of the Eclipse Public License 2.0
 * which is available at https://www.eclipse.org/legal/epl-2.0/
 * 
 * SPDX-License-Identifier: EPL-2.0
 * 
 * Contributors:
 *  Areti
 *  Eurotech
 *******************************************************************************/
```

kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/signal/handlers/NMDeviceCreationHandler.java
```diff
@@ -0,0 +1,55 @@
...
+ *
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *  Areti
+ *******************************************************************************/
```
mattdibi:
```suggestion
/*******************************************************************************
 * Copyright (c) 2023 Areti and others
 * 
 * This program and the accompanying materials are made
 * available under the terms of the Eclipse Public License 2.0
 * which is available at https://www.eclipse.org/legal/epl-2.0/
 * 
 * SPDX-License-Identifier: EPL-2.0
 * 
 * Contributors:
 *  Areti
 *  Eurotech
 *******************************************************************************/
```

mattdibi (changes requested):

kura/org.eclipse.kura.core.net/src/main/java/org/eclipse/kura/core/net/vlan/VlanInterfaceConfigImpl.java
mattdibi: Fixed in 3208904aeee9f9fe120043a16d2bf4053bb5689a

kura/org.eclipse.kura.core.net/META-INF/MANIFEST.MF
mattdibi: Use ranges

```suggestion
 org.eclipse.kura.net.status.vlan;version="[1.0,2.0)",
 org.eclipse.kura.net.vlan;version="[1.0,2.0)",
```

kura/org.eclipse.kura.core.net/META-INF/MANIFEST.MF
mattdibi: Fixed in 37e65d3b731f6be23a857778f060b78e4564ed6e

kura/org.eclipse.kura.core.net/META-INF/MANIFEST.MF
mattdibi: Fixed in 37e65d3b731f6be23a857778f060b78e4564ed6e

mattdibi (approved):
@fdizazzo I took the liberty to update the PR with the feedback from @MMaiero and https://gitlab.eclipse.org/eclipsefdn/emo-team/iplab/-/issues/10343

Take a look and double check I didn't screw up anything 😆 

Commits:
- 3208904aeee9f9fe120043a16d2bf4053bb5689a
- 37e65d3b731f6be23a857778f060b78e4564ed6e

[Here](https://github.com/eclipse/kura/compare/4a027d74a0474b411734c4f814767f9c83544bce..37e65d3b731f6be23a857778f060b78e4564ed6e) you can see my changes.

fdizazzo:
Looks good, thanks for your support :)

mattdibi:
The IP Team approved the PR. Merging...

@fdizazzo 🥳  Congratulations and thank you for the contribution!

PR #4819, opened 2023-08-28
fix(network.threat.manager): Added default flooding protection rules for filter table

pierantoniomerlino:

This PR adds a default set of rules for the firewall filter table if the flooding protection is enabled.

**Description of the solution adopted:** The `network.threat.manager` adds a set of firewall rules for flooding protection. These rules are mainly applied to the mangle table, but there are situations where more rules are needed to the filter tables. The current `FirewalConfigurationServiceImpl.addFloodingProtectionRules` method doesn't allow this, so this PR adds a basic set of default rules to the filter tables when the flooding protection is enabled.

The code is replicated both in the old `org.eclipse.kura.net.admin` and the new `org.eclipse.kura.net.admin.firewall`.


marcellorinaldo (approved):

pierantoniomerlino:
Superseeded by #4825

PR #4833, opened 2023-09-04
feat(api): add java.vm.vendor and jdk.vendor.version to SystemService APIs

marcellorinaldo:
This PR adds the `java.vm.vendor` and `jdk.vendor.version` Java `System` properties to the `SystemService` APIs. Minor version of bundle is increased to 1.7.0 and implementations are updated accordingly.

**Related Issue:** N/A.

**Screenshots:** N/A.

**Manual Tests**: N/A.


kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/system/SystemService.java
```diff
@@ -121,6 +121,12 @@ public interface SystemService {
      */
     public static final String KEY_LEGACY_PPP_LOGGING = "kura.legacy.ppp.logging.enabled";
 
+    /**
+     * @since 2.6
+     */
+    public static final String KEY_JAVA_VM_VENDOR = "java.vm.vendor";
+    public static final String KEY_JDK_VENDOR_VERSION = "jdk.vendor.version";
```
pierantoniomerlino, line 132: I think that we usually add the since annotation for every entry.

PR #4872, opened 2023-09-28
feat(nm): Backend implementation to support EAP-TLS + Minor WebUI fixes

GregoryIvo:
This PR adds some basic support for TLS.

add your certs to the KeystoreManager,

point the keystone, and cert names to the NM service

certs are decoded passed to NM, and a config is established.

Note: there may be a possible race condition with NM and the keystore service, which needs to be investigated.


GregoryIvo:
The snapshot I used to test this PR:

[wifi-ee-snapshot.json.zip](https://github.com/eclipse/kura/files/12754112/wifi-ee-snapshot.json.zip)

kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/configuration/NMSettingsConverter.java
```diff
@@ -171,15 +173,32 @@ private static void create8021xTls(NetworkProperties props, String deviceId, Map
...
+            logger.info(identity + " client cert: " + clientCert.getEncoded());
+        } catch (Exception e) {
+            logger.error("Enable to find or decode Client Certificate", e);
+        }
+
+        try {
+            PrivateKey privateKey = props.get(PrivateKey.class, "net.interface.%s.config.802-1x.private-key-name",
+                    deviceId);
+            settings.put("private-key", new Variant<>(privateKey.getEncoded()));
+            logger.info(identity + " private key: " + privateKey.getEncoded());
```
mattdibi: Remove this for the real production code. Logging plaintext certs is not... ideal

```suggestion
```

kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/configuration/NMSettingsConverter.java
```diff
@@ -171,15 +173,32 @@ private static void create8021xTls(NetworkProperties props, String deviceId, Map
         String identity = props.get(String.class, "net.interface.%s.config.802-1x.identity", deviceId);
         settings.put("identity", new Variant<>(identity));
 
-        String clientCert = props.get(String.class, "net.interface.%s.config.802-1x.client-cert", deviceId);
-        settings.put("client-cert", new Variant<>(clientCert.getBytes(StandardCharsets.UTF_8)));
+        try {
+            Certificate clientCert = props.get(Certificate.class, "net.interface.%s.config.802-1x.client-cert-name",
+                    deviceId);
+            settings.put("client-cert", new Variant<>(clientCert.getEncoded()));
+            logger.info(identity + " client cert: " + clientCert.getEncoded());
```
mattdibi: Remove this for the real production code. Logging plaintext certs is not... ideal

```suggestion
```

kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/configuration/NMSettingsConverter.java
```diff
@@ -171,15 +173,32 @@ private static void create8021xTls(NetworkProperties props, String deviceId, Map
...
-        settings.put("private-key", new Variant<>(privateKey.getBytes(StandardCharsets.UTF_8)));
+        Optional<Password> privateKeyPassword = props.getOpt(Password.class,
+                "net.interface.%s.config.802-1x.private-key-password", deviceId);
 
-        String privateKeyPassword = props
-                .get(Password.class, "net.interface.%s.config.802-1x.private-key-password", deviceId).toString();
-        settings.put("private-key-password", new Variant<>(privateKeyPassword));
+        if (privateKeyPassword.isPresent()) {
+            settings.put("private-key-password", new Variant<>(privateKeyPassword));
+        }
```
mattdibi: Prefer the `ifPresent` syntax

```suggestion
        privateKeyPassword.ifPresent(value -> {
            settings.put("private-key-password", new Variant<>(value));
        });
```

kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/configuration/NMSettingsConverter.java
```diff
@@ -191,9 +210,14 @@ private static void create8021xOptionalCaCertAndAnonIdentity(NetworkProperties p
...
 
-        Optional<String> caCert = props.getOpt(String.class, "net.interface.%s.config.802-1x.ca-cert", deviceId);
-        if (caCert.isPresent()) {
-            settings.put("ca-cert", new Variant<>(caCert.get().getBytes(StandardCharsets.UTF_8)));
+        try {
+            Optional<Certificate> caCert = props.getOpt(Certificate.class,
+                    "net.interface.%s.config.802-1x.ca-cert-name", deviceId);
+            if (caCert.isPresent()) {
+                settings.put("ca-cert", new Variant<>(caCert.get().getEncoded()));
+            }
```
mattdibi: Prefer the `ifPresent` syntax

```suggestion
            caCert.ifPresent(value -> {
                settings.put("ca-cert", new Variant<>(value.getEncoded()));
            });
```

mattdibi (changes requested):
See comments.

Furthermore there are test failures. You'll probably need to update some unit tests to take into account the newly introduced dependency on the KeystoreService and actually test the newly introduced functionality.

kura/org.eclipse.kura.nm/OSGI-INF/networkConfigurationService.xml
```diff
@@ -24,4 +24,5 @@
    <reference bind="setExecutorService" cardinality="1..1" interface="org.eclipse.kura.executor.PrivilegedExecutorService" name="PrivilegedExecutorService" policy="static" unbind="unsetExecutorService"/>
    <reference bind="setDnsServerService" cardinality="1..1" interface="org.eclipse.kura.internal.linux.net.dns.DnsServerService" name="DNSService" policy="static" />
    <reference name="CryptoService" interface="org.eclipse.kura.crypto.CryptoService" bind="setCryptoService" unbind="unsetCryptoService" cardinality="1..1" policy="static"/>
+   <reference name="KeystoreService" interface="org.eclipse.kura.security.keystore.KeystoreService" bind="setKeystoreService" unbind="unsetKeystoreService" cardinality="0..n" policy="dynamic"/>
```
mattdibi, line 27: Are you sure about this cardinality? 🤔

GregoryIvo: Yessir! We want all the Keystore instances, so we can select which one to use. Unfortunately, we can't use a target filter because NM does not use the traditional meta-type UI.

GregoryIvo:
How I set up EAP-TLS for testing:
1) pull ```https://github.com/sxiii/fralp```
2) Generate certs with ```docker run -it -v pki:/easyrsa/pki easyrsa build-client-full toshiba```
3) build the container and mount pki folder as instructed in the git repo
4) set PC IP as radius server on the AP
5) use open SSL to decrypt the Private key you generated before
6) in kura create a new wifiKeystore
7) add certificate and add CA.cert created above
8) add private key (use the cert, and the decrypted key)
9) update the snapshot to include the name of the keystore.pid, ca.cert name, client-cert-name (private key name), and private-key-name (private-key-name)
10) configure wifi settings in the snapshot
11) apply snapshot

GregoryIvo:
what a working config looks like:
```
nmcli> print
===============================================================================
              Connection profile details (kura-wlan0-connection)
===============================================================================
connection.id:                          kura-wlan0-connection
connection.uuid:                        1d60629e-c907-40e0-8e7e-1bdf01db27a9
connection.stable-id:                   --
connection.type:                        802-11-wireless
connection.interface-name:              wlan0
connection.autoconnect:                 yes
connection.autoconnect-priority:        0
connection.autoconnect-retries:         1
connection.multi-connect:               0 (default)
connection.auth-retries:                -1
connection.timestamp:                   1696362353
connection.read-only:                   no
connection.permissions:                 --
connection.zone:                        --
connection.master:                      --
connection.slave-type:                  --
connection.autoconnect-slaves:          -1 (default)
connection.secondaries:                 --
connection.gateway-ping-timeout:        0
connection.metered:                     unknown
connection.lldp:                        default
connection.mdns:                        -1 (default)
connection.llmnr:                       -1 (default)
connection.wait-device-timeout:         -1
-------------------------------------------------------------------------------
802-1x.optional:                        no
802-1x.eap:                             tls
802-1x.identity:                        pc
802-1x.anonymous-identity:              --
802-1x.pac-file:                        --
802-1x.ca-cert:                         --
802-1x.ca-cert-password:                <hidden>
802-1x.ca-cert-password-flags:          0 (none)
802-1x.ca-path:                         --
802-1x.subject-match:                   --
802-1x.altsubject-matches:              --
802-1x.domain-suffix-match:             --
802-1x.domain-match:                    --
802-1x.client-cert:                     /etc/NetworkManager/system-connections/1d60629e-c907-40e0-8e7e-1bdf01db27a9-client-cert.der
802-1x.client-cert-password:            <hidden>
802-1x.client-cert-password-flags:      0 (none)
802-1x.phase1-peapver:                  --
802-1x.phase1-peaplabel:                --
802-1x.phase1-fast-provisioning:        --
802-1x.phase1-auth-flags:               0x0 (none)
802-1x.phase2-auth:                     --
802-1x.phase2-autheap:                  --
802-1x.phase2-ca-cert:                  --
802-1x.phase2-ca-cert-password:         <hidden>
802-1x.phase2-ca-cert-password-flags:   0 (none)
802-1x.phase2-ca-path:                  --
802-1x.phase2-subject-match:            --
802-1x.phase2-altsubject-matches:       --
802-1x.phase2-domain-suffix-match:      --
802-1x.phase2-domain-match:             --
802-1x.phase2-client-cert:              --
802-1x.phase2-client-cert-password:     <hidden>
802-1x.phase2-client-cert-password-flags:0 (none)
802-1x.password:                        <hidden>
802-1x.password-flags:                  0 (none)
802-1x.password-raw:                    <hidden>
802-1x.password-raw-flags:              0 (none)
802-1x.private-key:                     /etc/NetworkManager/system-connections/1d60629e-c907-40e0-8e7e-1bdf01db27a9-private-key.pem
802-1x.private-key-password:            <hidden>
802-1x.private-key-password-flags:      4 (not required)
802-1x.phase2-private-key:              --
802-1x.phase2-private-key-password:     <hidden>
802-1x.phase2-private-key-password-flags:0 (none)
802-1x.pin:                             <hidden>
802-1x.pin-flags:                       0 (none)
802-1x.system-ca-certs:                 no
802-1x.auth-timeout:                    0
-------------------------------------------------------------------------------
802-11-wireless.ssid:                   greg-enterprise-external
802-11-wireless.mode:                   infrastructure
802-11-wireless.band:                   --
802-11-wireless.channel:                0
802-11-wireless.bssid:                  --
802-11-wireless.rate:                   0
802-11-wireless.tx-power:               0
802-11-wireless.mac-address:            --
802-11-wireless.cloned-mac-address:     --
802-11-wireless.generate-mac-address-mask:--
802-11-wireless.mac-address-blacklist:  --
802-11-wireless.mac-address-randomization:default
802-11-wireless.mtu:                    auto
802-11-wireless.seen-bssids:            62:22:32:98:12:54,62:22:32:A8:12:54,DC:A6:32:A6:E0:C2
802-11-wireless.hidden:                 no
802-11-wireless.powersave:              0 (default)
802-11-wireless.wake-on-wlan:           0x1 (default)
802-11-wireless.ap-isolation:           -1 (default)
-------------------------------------------------------------------------------
802-11-wireless-security.key-mgmt:      wpa-eap
802-11-wireless-security.wep-tx-keyidx: 0
802-11-wireless-security.auth-alg:      --
802-11-wireless-security.proto:         --
802-11-wireless-security.pairwise:      --
802-11-wireless-security.group:         --
802-11-wireless-security.pmf:           0 (default)
802-11-wireless-security.leap-username: --
802-11-wireless-security.wep-key0:      <hidden>
802-11-wireless-security.wep-key1:      <hidden>
802-11-wireless-security.wep-key2:      <hidden>
802-11-wireless-security.wep-key3:      <hidden>
802-11-wireless-security.wep-key-flags: 0 (none)
802-11-wireless-security.wep-key-type:  unknown
802-11-wireless-security.psk:           <hidden>
802-11-wireless-security.psk-flags:     0 (none)
802-11-wireless-security.leap-password: <hidden>
802-11-wireless-security.leap-password-flags:0 (none)
802-11-wireless-security.wps-method:    0x0 (default)
802-11-wireless-security.fils:          0 (default)
-------------------------------------------------------------------------------
ipv4.method:                            auto
ipv4.dns:                               --
ipv4.dns-search:                        --
ipv4.dns-options:                       --
ipv4.dns-priority:                      0
ipv4.addresses:                         --
ipv4.gateway:                           --
ipv4.routes:                            --
ipv4.route-metric:                      -1
ipv4.route-table:                       0 (unspec)
ipv4.routing-rules:                     --
ipv4.ignore-auto-routes:                no
ipv4.ignore-auto-dns:                   no
ipv4.dhcp-client-id:                    --
ipv4.dhcp-iaid:                         --
ipv4.dhcp-timeout:                      0 (default)
ipv4.dhcp-send-hostname:                yes
ipv4.dhcp-hostname:                     --
ipv4.dhcp-fqdn:                         --
ipv4.dhcp-hostname-flags:               0x0 (none)
ipv4.never-default:                     no
ipv4.may-fail:                          yes
ipv4.dad-timeout:                       -1 (default)
ipv4.dhcp-vendor-class-identifier:      --
ipv4.dhcp-reject-servers:               --
-------------------------------------------------------------------------------
ipv6.method:                            disabled
ipv6.dns:                               --
ipv6.dns-search:                        --
ipv6.dns-options:                       --
ipv6.dns-priority:                      0
ipv6.addresses:                         --
ipv6.gateway:                           --
ipv6.routes:                            --
ipv6.route-metric:                      -1
ipv6.route-table:                       0 (unspec)
ipv6.routing-rules:                     --
ipv6.ignore-auto-routes:                no
ipv6.ignore-auto-dns:                   no
ipv6.never-default:                     no
ipv6.may-fail:                          yes
ipv6.ip6-privacy:                       -1 (unknown)
ipv6.addr-gen-mode:                     stable-privacy
ipv6.ra-timeout:                        0 (default)
ipv6.dhcp-duid:                         --
ipv6.dhcp-iaid:                         --
ipv6.dhcp-timeout:                      0 (default)
ipv6.dhcp-send-hostname:                yes
ipv6.dhcp-hostname:                     --
ipv6.dhcp-hostname-flags:               0x0 (none)
ipv6.token:                             --
-------------------------------------------------------------------------------
proxy.method:                           none
proxy.browser-only:                     no
proxy.pac-url:                          --
proxy.pac-script:                       --
-------------------------------------------------------------------------------
```

kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/configuration/NMConfigurationServiceImpl.java
```diff
@@ -295,6 +313,77 @@ private void decryptAndConvertPasswordProperties(Map<String, Object> modifiedPro
...
+
+        if (keystoreService == null) {
+            logger.error("Cannot find keystore service for interface {}", interfaceName);
+            return;
+        }
+
+        for (Entry<String, Object> prop : modifiedProps.entrySet()) {
+            String clientCertString = String.format("%s.config.802-1x.client-cert-name", interfaceName);
+            String caCertString = String.format("%s.config.802-1x.802-1x.ca-cert-name", interfaceName);
+            String privateKeyString = String.format("%s.config.802-1x.private-key-name", interfaceName);
```
mattdibi: Why instantiate these strings here if they only depend on the `interfaceName`? They get instantiated for every property... this is a waste of clock cycles.

```suggestion
        final String clientCertString = String.format("%s.config.802-1x.client-cert-name", interfaceName);
        final String caCertString = String.format("%s.config.802-1x.802-1x.ca-cert-name", interfaceName);
        final String privateKeyString = String.format("%s.config.802-1x.private-key-name", interfaceName);

        for (Entry<String, Object> prop : modifiedProps.entrySet()) {
```

kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/configuration/NMConfigurationServiceImpl.java
```diff
@@ -295,6 +313,77 @@ private void decryptAndConvertPasswordProperties(Map<String, Object> modifiedPro
...
+                    logger.error("Enable to decode certificate {} from keystore.", value.toString(), e);
+                    modifiedProps.put(prop.getKey(), value);
+                }
+            } else if (prop.getKey().contains(privateKeyString)) {
+                Object value = prop.getValue();
+                try {
+                    String valueString = (String) prop.getValue();
+                    modifiedProps.put(prop.getKey(), decryptPrivateKey(valueString, keystoreService));
+                } catch (Exception e) {
+                    logger.error("Enable to decode private key {} from keystore.", value.toString(), e);
```
mattdibi: Probably a typo

```suggestion
                    logger.error("Unable to decode private key {} from keystore.", value.toString(), e);
```

kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/configuration/NMConfigurationServiceImpl.java
```diff
@@ -295,6 +313,77 @@ private void decryptAndConvertPasswordProperties(Map<String, Object> modifiedPro
...
+            String clientCertString = String.format("%s.config.802-1x.client-cert-name", interfaceName);
+            String caCertString = String.format("%s.config.802-1x.802-1x.ca-cert-name", interfaceName);
+            String privateKeyString = String.format("%s.config.802-1x.private-key-name", interfaceName);
+            if (prop.getKey().contains(clientCertString) || prop.getKey().contains(caCertString)) {
+                Object value = prop.getValue();
+                try {
+                    String valueString = (String) prop.getValue();
+                    modifiedProps.put(prop.getKey(), decryptCertificate(valueString, keystoreService));
+                } catch (Exception e) {
+                    logger.error("Enable to decode certificate {} from keystore.", value.toString(), e);
```
mattdibi: No need to call "toString()" method as formatting and string conversion is done by the Formatter

```suggestion
                    logger.error("Enable to decode certificate {} from keystore.", value, e);
```

kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/configuration/NMConfigurationServiceImpl.java
mattdibi: No need to call "toString()" method as formatting and string conversion is done by the Formatter

```suggestion
                    logger.error("Enable to decode private key {} from keystore.", value, e);
```

kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/configuration/NMConfigurationServiceImpl.java
```diff
@@ -295,6 +313,77 @@ private void decryptAndConvertPasswordProperties(Map<String, Object> modifiedPro
...
+
+    private void findAndDecodeCertificatesForInterface(String interfaceName, Map<String, Object> modifiedProps,
+            KeystoreService keystoreService) {
+
+        if (keystoreService == null) {
+            logger.error("Cannot find keystore service for interface {}", interfaceName);
+            return;
+        }
+
+        for (Entry<String, Object> prop : modifiedProps.entrySet()) {
```
mattdibi: Also... why are you iterating over a `Map`... couldn't you just call `containsKey` and act accordingly?

kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/configuration/NMConfigurationServiceImpl.java
```diff
@@ -295,6 +313,77 @@ private void decryptAndConvertPasswordProperties(Map<String, Object> modifiedPro
...
+            if (prop.getKey().contains("802-1x.keystore.pid")) {
+                String keystorePid = (String) prop.getValue();
+
+                String interfaceName = prop.getKey().split("\\.")[2];
+                findAndDecodeCertificatesForInterface(interfaceName, modifiedProps, getKeystore(keystorePid));
+            }
+        }
+    }
+
+    private void findAndDecodeCertificatesForInterface(String interfaceName, Map<String, Object> modifiedProps,
```
mattdibi, line 336: Finally... I think you kinda inverted the logic of this method.

You want, for the interface name passed as argument, decrypt every instance of certificate you find in the properties. I don't think iterating over the properties is the right way. You already know the properties you need to decrypt, you only need to check if they exists and then decrypt them.

Therefore:

```java
private void findAndDecodeCertificatesForInterface(String interfaceName, Map<String, Object> modifiedProps,
        KeystoreService keystoreService) {

    if (keystoreService == null) {
        logger.error("Cannot find keystore service for interface {}", interfaceName);
        return;
    }

    final String clientCertString = String.format("net.interface.%s.config.802-1x.client-cert-name", interfaceName);
    final String caCertString = String.format("net.interface.%s.config.802-1x.802-1x.ca-cert-name", interfaceName);
    final String privateKeyString = String.format("net.interface.%s.config.802-1x.private-key-name", interfaceName);
    final List<String> keyCertStrings = Arrays.asList(clientCertString, caCertString, privateKeyString);

    for (String key : keyCertStrings) {
        if (!modifiedProps.containsKey(key)) { // Early return because I hate too many indentations
            continue;
        }

        Object value = modifiedProps.get(key);
        try {
            String valueString = value.toString();
            if (key.contains("cert")) { // a isCertificate method might be better
                modifiedProps.put(key, decryptCertificate(valueString, keystoreService));
            } else {
                modifiedProps.put(key, decryptPrivateKey(valueString, keystoreService));
            }
        } catch (Exception e) { // I don't super like catching Exception... prefer something more specific if you can
            logger.error("Unable to decode key/certificate {} from keystore.", key, e);
            modifiedProps.put(key, value);
        }
    }
}
```

kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/configuration/NMConfigurationServiceImpl.java
mattdibi: Also, again, don't log the certificate.

```suggestion
                    logger.error("Enable to decode private key {} from keystore.", prop.getKey(), e);
```

kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/configuration/NMSettingsConverter.java
```diff
@@ -171,36 +173,56 @@ private static void create8021xTls(NetworkProperties props, String deviceId, Map
...
         settings.put("identity", new Variant<>(identity));
 
-        String clientCert = props.get(String.class, "net.interface.%s.config.802-1x.client-cert", deviceId);
-        settings.put("client-cert", new Variant<>(clientCert.getBytes(StandardCharsets.UTF_8)));
+        try {
+            Certificate clientCert = props.get(Certificate.class, "net.interface.%s.config.802-1x.client-cert-name",
+                    deviceId);
+            settings.put("client-cert", new Variant<>(clientCert.getEncoded()));
+        } catch (Exception e) {
+            logger.error("Enable to find or decode Client Certificate", e);
```
mattdibi:
```suggestion
            logger.error("Unable to find or decode Client Certificate", e);
```

kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/configuration/NMSettingsConverter.java
```diff
@@ -171,36 +173,56 @@ private static void create8021xTls(NetworkProperties props, String deviceId, Map
...
+        }
 
-        String privateKey = props.get(String.class, "net.interface.%s.config.802-1x.private-key", deviceId);
-        settings.put("private-key", new Variant<>(privateKey.getBytes(StandardCharsets.UTF_8)));
+        try {
+            PrivateKey privateKey = props.get(PrivateKey.class, "net.interface.%s.config.802-1x.private-key-name",
+                    deviceId);
+            settings.put("private-key", new Variant<>(privateKey.getEncoded()));
+        } catch (Exception e) {
+            logger.error("Enable to find or decode Private Key", e);
```
mattdibi:
```suggestion
            logger.error("Unable to find or decode Private Key", e);
```

kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/configuration/NMSettingsConverter.java
```diff
@@ -171,36 +173,56 @@ private static void create8021xTls(NetworkProperties props, String deviceId, Map
...
-            settings.put("ca-cert", new Variant<>(caCert.get().getBytes(StandardCharsets.UTF_8)));
+        anonymousIdentity.ifPresent(value -> {
+            settings.put("anonymous-identity", new Variant<>(value));
+        });
+
+        try {
+            Certificate caCert = props.get(Certificate.class, "net.interface.%s.config.802-1x.ca-cert-name", deviceId);
+            settings.put("ca-cert", new Variant<>(caCert.getEncoded()));
+        } catch (Exception e) {
+            logger.error("Enable to find or decode CA Certificate", e);
```
mattdibi:
```suggestion
            logger.error("Unable to find or decode CA Certificate", e);
```

kura/test/org.eclipse.kura.nm.test/src/test/java/org/eclipse/kura/nm/configuration/NMSettingsConverterTest.java
```diff
@@ -2845,6 +2855,12 @@ public void givenMockConnection() {
 
     }
 
+    public void givenMockConnectionWithNullSettings() {
+        this.mockedConnection = Mockito.mock(Connection.class);
+        Mockito.when(this.mockedConnection.GetSettings()).thenReturn(null);
+
```
mattdibi: Extra newline

```suggestion
```

mattdibi (changes requested):
In addition to the comments I provided I need some clarification about the scope of this PR: the title says "initial implementation of tls enterprise backend" but I see changes also in the frontend. Is this complete at this point?

kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/configuration/NMConfigurationServiceImpl.java
```diff
@@ -295,6 +313,77 @@ private void decryptAndConvertPasswordProperties(Map<String, Object> modifiedPro
         }
     }
 
+    private void decryptAndConvertCertificatesProperties(Map<String, Object> modifiedProps) throws KuraException {
```
mattdibi: Remove the declaration of thrown exception 'org.eclipse.kura.KuraException', as it cannot be thrown from method's body.

```suggestion
    private void decryptAndConvertCertificatesProperties(Map<String, Object> modifiedProps) {
```

kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/configuration/NMConfigurationServiceImpl.java
```diff
@@ -295,6 +313,77 @@ private void decryptAndConvertPasswordProperties(Map<String, Object> modifiedPro
...
+
+        for (Entry<String, Object> prop : modifiedProps.entrySet()) {
+            if (prop.getKey().contains("802-1x.keystore.pid")) {
+                String keystorePid = (String) prop.getValue();
+
+                String interfaceName = prop.getKey().split("\\.")[2];
+                findAndDecodeCertificatesForInterface(interfaceName, modifiedProps, getKeystore(keystorePid));
```
mattdibi, line 323: Final thought: do you really need this? Do you really need to know the interface name to decrypt the certificate? Can't you just iterate over the properties and, if you find one that matches `.config.802-1x.client-cert-name` (regardless of the interface name) decrypt it? The same goes for the other properties.

... or am I missing something?

kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/configuration/NMConfigurationServiceImpl.java
```diff
@@ -295,6 +313,77 @@ private void decryptAndConvertPasswordProperties(Map<String, Object> modifiedPro
         }
     }
 
+    private void decryptAndConvertCertificatesProperties(Map<String, Object> modifiedProps) throws KuraException {
+
+        for (Entry<String, Object> prop : modifiedProps.entrySet()) {
+            if (prop.getKey().contains("802-1x.keystore.pid")) {
+                String keystorePid = (String) prop.getValue();
+
+                String interfaceName = prop.getKey().split("\\.")[2];
```
mattdibi: 🤔  I know that it shouldn't happen* but... what about VLAN interfaces? They have a `.` in their name.

\* i.e. VLAN interfaces should **not** support 802-1x authentication

kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/configuration/NMConfigurationServiceImpl.java
```diff
@@ -295,6 +313,77 @@ private void decryptAndConvertPasswordProperties(Map<String, Object> modifiedPro
...
+
+    private PrivateKey decryptPrivateKey(String privateKeyName, KeystoreService keystoreService) throws KuraException {
+        PrivateKeyEntry key = (PrivateKeyEntry) keystoreService.getEntry(privateKeyName);
+
+        return key.getPrivateKey();
+    }
+
+    private KeystoreService getKeystore(String keystoreServicePid) {
+        return this.keystoreServices.get(keystoreServicePid);
+    }
```
mattdibi: Is this method really necessary?

kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/configuration/NMConfigurationServiceImpl.java
mattdibi: Wouldn't it be better (and more performant) to iterate over the known/configured interfaces and then check if the `modifiedProps` map contains the "[interface-name].802-1x.keystore.pid" key? If you do it this way you also avoid the issue with the VLAN interface names. LMKWYT

GregoryIvo:
these changes also seem to fix the clear text password bug: 
image

kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/configuration/NMConfigurationServiceImpl.java
```diff
@@ -295,6 +314,81 @@ private void decryptAndConvertPasswordProperties(Map<String, Object> modifiedPro
...
+    private void findAndDecodeCertificatesForInterface(String interfaceName, Map<String, Object> modifiedProps,
+            KeystoreService keystoreService) {
+
+        if (keystoreService == null) {
+            logger.error("Cannot find keystore service for interface {}", interfaceName);
+            return;
+        }
+
+        final String clientCertString = String.format("net.interface.%s.config.802-1x.client-cert-name", interfaceName);
+        final String caCertString = String.format("net.interface.%s.config.802-1x.802-1x.ca-cert-name", interfaceName);
```
mattdibi: 🤔 Too many `802-1x`s... is this correct?

```suggestion
        final String caCertString = String.format("net.interface.%s.config.802-1x.ca-cert-name", interfaceName);
```

kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/configuration/NMConfigurationServiceImpl.java
```diff
@@ -295,6 +314,81 @@ private void decryptAndConvertPasswordProperties(Map<String, Object> modifiedPro
...
+            } catch (KuraException e) {
+                logger.error("Unable to decode key/certificate {} from keystore.", key, e);
+                modifiedProps.put(key, value);
+            }
+        }
+    }
+
+    private boolean isCertificate(String key) {
+        return key.contains("cert");
+    }
```
mattdibi: I know it works and it's ok as-is but I would further future proof it by being a little bit more specific, just to be on the safe side:

```suggestion
    private boolean isCertificate(String key) {
        return key.contains("802-1x.client-cert-name") || key.contains("802-1x.ca-cert-name");
    }
```

kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/configuration/NMConfigurationServiceImpl.java
```diff
@@ -295,6 +314,81 @@ private void decryptAndConvertPasswordProperties(Map<String, Object> modifiedPro
...
+            throws KuraException {
+        if (keystoreService.getEntry(certificateName) instanceof TrustedCertificateEntry) {
+            TrustedCertificateEntry cert = (TrustedCertificateEntry) keystoreService.getEntry(certificateName);
+            return cert.getTrustedCertificate();
+        } else if (keystoreService.getEntry(certificateName) instanceof PrivateKeyEntry) {
+            PrivateKeyEntry cert = (PrivateKeyEntry) keystoreService.getEntry(certificateName);
+            return cert.getCertificate();
+        } else {
+            throw new KuraException(KuraErrorCode.CONFIGURATION_ERROR,
+                    String.format("Certificate %s is not expected key type or not found.", certificateName));
```
mattdibi: Wording/formatting. Feel free to correct me if this doesn't sound right.

```suggestion
                    String.format("Certificate \"%s\" is not of the expected key type or not found.", certificateName));
```

kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/configuration/NMConfigurationServiceImpl.java
```diff
@@ -295,6 +314,81 @@ private void decryptAndConvertPasswordProperties(Map<String, Object> modifiedPro
...
+        }
+    }
+
+    private PrivateKey getTrustedPrivateKeyFromKeystore(String privateKeyName, KeystoreService keystoreService) throws KuraException {
+        PrivateKeyEntry key = (PrivateKeyEntry) keystoreService.getEntry(privateKeyName);
+
+        return key.getPrivateKey();
```
mattdibi: Why didn't you add the same safety checks as you did in the `getTrustedCertificateFromKeystore` method? I would expect the same:

```suggestion
        if(!keystoreService.getEntry(privateKeyName) instanceof PrivateKeyEntry) {
             throw new KuraException(KuraErrorCode.CONFIGURATION_ERROR,
                    String.format("Private key \"%s\" is not of the expected key type or not found.", privateKeyName));
        }
        
        PrivateKeyEntry key = (PrivateKeyEntry) keystoreService.getEntry(privateKeyName);
        return key.getPrivateKey();
```

kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/configuration/NMSettingsConverter.java
```diff
@@ -171,36 +173,56 @@ private static void create8021xTls(NetworkProperties props, String deviceId, Map
...
 
-        Optional<String> caCert = props.getOpt(String.class, "net.interface.%s.config.802-1x.ca-cert", deviceId);
-        if (caCert.isPresent()) {
-            settings.put("ca-cert", new Variant<>(caCert.get().getBytes(StandardCharsets.UTF_8)));
+        anonymousIdentity.ifPresent(value -> {
+            settings.put("anonymous-identity", new Variant<>(value));
+        });
+
+        try {
+            Certificate caCert = props.get(Certificate.class, "net.interface.%s.config.802-1x.ca-cert-name", deviceId);
```
mattdibi, line 214: Here we have only one `802-1x`... [here](https://github.com/eclipse/kura/pull/4872#discussion_r1349958415) there's two... is this actually working? 🤔

mattdibi (changes requested):
There's still some minor issues... we're almost there!

kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/configuration/NMConfigurationServiceImpl.java
GregoryIvo: This was because there was only one possible public key type. 
However, You make an excellent point!

kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/configuration/NMSettingsConverter.java
GregoryIvo: Surprisingly yes, since CA cert is an optional field. this was definitely introduced accidentally during some refactoring.

mattdibi:
@GregoryIvo there's a couple of issues reported by Sonar that require fixing. The code smells are all easy and quick, the coverage needs some more work though. Let me know if you need help with that.

kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/configuration/NMSettingsConverter.java
```diff
@@ -171,36 +173,50 @@ private static void create8021xTls(NetworkProperties props, String deviceId, Map
...
         settings.put("identity", new Variant<>(identity));
 
-        String clientCert = props.get(String.class, "net.interface.%s.config.802-1x.client-cert", deviceId);
-        settings.put("client-cert", new Variant<>(clientCert.getBytes(StandardCharsets.UTF_8)));
+        try {
+            Certificate clientCert = props.get(Certificate.class, "net.interface.%s.config.802-1x.client-cert-name",
+                    deviceId);
+            settings.put("client-cert", new Variant<>(clientCert.getEncoded()));
+        } catch (Exception e) {
+            logger.error("Unable to find or decode Client Certificate", e);
```
mattdibi: Just a question: are we sure we won't be leaking any sensitive info by logging this exception?

Also: catching `Exception` is a bad practice. We should always try to catch a specific type of exception.

kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/configuration/NMSettingsConverter.java
```diff
@@ -171,36 +173,50 @@ private static void create8021xTls(NetworkProperties props, String deviceId, Map
...
+        }
 
-        String privateKey = props.get(String.class, "net.interface.%s.config.802-1x.private-key", deviceId);
-        settings.put("private-key", new Variant<>(privateKey.getBytes(StandardCharsets.UTF_8)));
+        try {
+            PrivateKey privateKey = props.get(PrivateKey.class, "net.interface.%s.config.802-1x.private-key-name",
+                    deviceId);
+            settings.put("private-key", new Variant<>(privateKey.getEncoded()));
+        } catch (Exception e) {
+            logger.error("Unable to find or decode Private Key", e);
```
mattdibi: Same here: are we sure we won't be leaking any sensitive info by logging this exception?

kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/configuration/NMSettingsConverter.java
```diff
@@ -171,36 +173,50 @@ private static void create8021xTls(NetworkProperties props, String deviceId, Map
...
 
-        String privateKeyPassword = props
-                .get(Password.class, "net.interface.%s.config.802-1x.private-key-password", deviceId).toString();
-        settings.put("private-key-password", new Variant<>(privateKeyPassword));
+        Optional<Password> privateKeyPassword = props.getOpt(Password.class,
+                "net.interface.%s.config.802-1x.private-key-password", deviceId);
+
+        privateKeyPassword.ifPresent(value -> settings.put("private-key-password", new Variant<>(value.toString())));
+
+        settings.put("private-key-password-flags", new Variant<>(new UInt32(4)));
```
mattdibi: I would use a constant here for clarity. From the [NM documentation I found](https://manpages.ubuntu.com/manpages/trusty/man5/nm-settings.5.html) that:

image

Therefore I would add something like:

```suggestion
        static final UInt32 NM_SECRET_FLAGS_NOT_REQUIRED = new UInt32(4);
        settings.put("private-key-password-flags", new Variant<>(NM_SECRET_FLAGS_NOT_REQUIRED));
```

This is just for illustrative purpose. The constant should be defined where you see fit.

kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/server/net2/configuration/NetworkConfigurationServiceProperties.java
mattdibi, line 1: Whenever possible I would avoid committing formatting changes to files you didn't actually change. It adds to the noise and make reviews more difficult.

Obviously, since the formatting changes are correct, leave it as-is but take this into account for the future.

kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/client/ui/network/TabIp4Ui.java
mattdibi, line 1: Whenever possible I would avoid committing formatting changes to files you didn't actually change. It adds to the noise and make reviews more difficult.

Obviously, since the formatting changes are correct, leave it as-is but take this into account for the future.

kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/configuration/NMSettingsConverter.java
mattdibi: Also: is this correct? I would expect this to be set to `none` as the system should be responsible for providing and storing the secret 🤔 ... this also depends on who is the "user" here 😆

kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/configuration/NMConfigurationServiceImpl.java
```diff
@@ -295,6 +315,86 @@ private void decryptAndConvertPasswordProperties(Map<String, Object> modifiedPro
...
+
+                Object prop = modifiedProps.get(key);
+
+                if (prop instanceof String) {
+                    String keystorePid = (String) prop;
+
+                    findAndDecodeCertificatesForInterface(interfaceName, modifiedProps,
+                            this.keystoreServices.get(keystorePid));
+                }
+            }
```
mattdibi, line 332: You can leave the code as-is but consider using [Early Returns](https://medium.com/swlh/return-early-pattern-3d18a41bba8) in the future. I think the code is much easier to read:

```suggestion
            if (!modifiedProps.containsKey(key)) {
                return; // we cannot "continue" since we're inside a lambda
            }

            Object keystorePid = modifiedProps.get(key);
            if (!keystorePid instanceof String) {
                return;
            }

            findAndDecodeCertificatesForInterface(interfaceName, modifiedProps,
                    this.keystoreServices.get((String) keystorePid));
```

mattdibi:
@GregoryIvo I think the code is OK.

I suggested [a small change](https://github.com/eclipse/kura/pull/4872#discussion_r1356310756) to avoid magic numbers in the code and have a couple of questions ([[1]](https://github.com/eclipse/kura/pull/4872#discussion_r1356248286), [[2]](https://github.com/eclipse/kura/pull/4872#discussion_r1356252746) and [[3]](https://github.com/eclipse/kura/pull/4872#discussion_r1356365971)). Everything else I mentioned is optional and is not required for merging.

I will now move on with testing.

kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/configuration/NMConfigurationServiceImpl.java
```diff
@@ -295,6 +315,86 @@ private void decryptAndConvertPasswordProperties(Map<String, Object> modifiedPro
         }
     }
 
+    private void decryptAndConvertCertificatesProperties(Map<String, Object> modifiedProps, Set<String> interfaces) {
```
mattdibi, line 318: One final thought: this is a [Side Effect Method](https://en.wikipedia.org/wiki/Side_effect_(computer_science)).

> A side effect method is a method which modifies some state variable value/arguments passed having a consequence beyond its scope, that is to say it has an observable effect besides returning a value (the main effect) to the invoker of the operation.

which is a known anti-pattern/bad practice (see Uncle Bob Martin's "Clean Code", there's a dedicated section for that).

The correct way to handle this would be to return a modified version of the passed argument, but we already had the `decryptAndConvertPasswordProperties` that behaved the same way and therefore we're doing the same for consistency.

Again: leave it as-is but be aware of the fact that this is not the best implementation.

Also notice that you've hidden the actual modification of the map down a couple of layer of abstractions (you actually change the values inside `findAndDecodeCertificatesForInterface`), so you've hidden the Side Effect even more than what was done in `decryptAndConvertPasswordProperties`.

Regarding *why* this is considered a bad practice... you experienced it first hand while chasing that bug caused by the modified map 😆

mattdibi:
@GregoryIvo tested on my setup. It works! 🥳 

I used the UI in #4881 and the instruction [here](https://github.com/eclipse/kura/pull/4872#issuecomment-1742198354) and [here](https://github.com/sxiii/fralp) and was able to connect to my WPA Enterprise hotspot.

Here's some screenshots of the setup

image

image

image

image

image

The resulting NM configuration:

<details>

<summary> Expand </summary>

```
connection.id:                          kura-wlan0-connection
connection.uuid:                        10852979-9610-42ec-bcfc-eb609c0860de
connection.stable-id:                   --
connection.type:                        802-11-wireless
connection.interface-name:              wlan0
connection.autoconnect:                 yes
connection.autoconnect-priority:        0
connection.autoconnect-retries:         1
connection.multi-connect:               0 (default)
connection.auth-retries:                -1
connection.timestamp:                   1697105080
connection.read-only:                   no
connection.permissions:                 --
connection.zone:                        --
connection.master:                      --
connection.slave-type:                  --
connection.autoconnect-slaves:          -1 (default)
connection.secondaries:                 --
connection.gateway-ping-timeout:        0
connection.metered:                     unknown
connection.lldp:                        default
connection.mdns:                        -1 (default)
connection.llmnr:                       -1 (default)
connection.dns-over-tls:                -1 (default)
connection.mptcp-flags:                 0x0 (default)
connection.wait-device-timeout:         -1
connection.wait-activation-delay:       -1
802-1x.optional:                        no
802-1x.eap:                             tls
802-1x.identity:                        password
802-1x.anonymous-identity:              --
802-1x.pac-file:                        --
802-1x.ca-cert:                         /etc/NetworkManager/system-connections/10852979-9610-42ec-bcfc-eb609c0860de-ca-cert.der
802-1x.ca-cert-password:                <hidden>
802-1x.ca-cert-password-flags:          0 (none)
802-1x.ca-path:                         --
802-1x.subject-match:                   --
802-1x.altsubject-matches:              --
802-1x.domain-suffix-match:             --
802-1x.domain-match:                    --
802-1x.client-cert:                     /etc/NetworkManager/system-connections/10852979-9610-42ec-bcfc-eb609c0860de-client-cert.der
802-1x.client-cert-password:            <hidden>
802-1x.client-cert-password-flags:      0 (none)
802-1x.phase1-peapver:                  --
802-1x.phase1-peaplabel:                --
802-1x.phase1-fast-provisioning:        --
802-1x.phase1-auth-flags:               0x0 (none)
802-1x.phase2-auth:                     --
802-1x.phase2-autheap:                  --
802-1x.phase2-ca-cert:                  --
802-1x.phase2-ca-cert-password:         <hidden>
802-1x.phase2-ca-cert-password-flags:   0 (none)
802-1x.phase2-ca-path:                  --
802-1x.phase2-subject-match:            --
802-1x.phase2-altsubject-matches:       --
802-1x.phase2-domain-suffix-match:      --
802-1x.phase2-domain-match:             --
802-1x.phase2-client-cert:              --
802-1x.phase2-client-cert-password:     <hidden>
802-1x.phase2-client-cert-password-flags:0 (none)
802-1x.password:                        <hidden>
802-1x.password-flags:                  0 (none)
802-1x.password-raw:                    <hidden>
802-1x.password-raw-flags:              0 (none)
802-1x.private-key:                     /etc/NetworkManager/system-connections/10852979-9610-42ec-bcfc-eb609c0860de-private-key.pem
802-1x.private-key-password:            <hidden>
802-1x.private-key-password-flags:      4 (not required)
802-1x.phase2-private-key:              --
802-1x.phase2-private-key-password:     <hidden>
802-1x.phase2-private-key-password-flags:0 (none)
802-1x.pin:                             <hidden>
802-1x.pin-flags:                       0 (none)
802-1x.system-ca-certs:                 no
802-1x.auth-timeout:                    0
802-11-wireless.ssid:                   GREG_TEST
802-11-wireless.mode:                   infrastructure
802-11-wireless.band:                   a
802-11-wireless.channel:                48
802-11-wireless.bssid:                  --
802-11-wireless.rate:                   0
802-11-wireless.tx-power:               0
802-11-wireless.mac-address:            --
802-11-wireless.cloned-mac-address:     --
802-11-wireless.generate-mac-address-mask:--
802-11-wireless.mac-address-blacklist:  --
802-11-wireless.mac-address-randomization:default
802-11-wireless.mtu:                    auto
802-11-wireless.seen-bssids:            7A:A7:41:8D:5B:1D,DC:A6:32:E0:54:F1
802-11-wireless.hidden:                 no
802-11-wireless.powersave:              0 (default)
802-11-wireless.wake-on-wlan:           0x1 (default)
802-11-wireless.ap-isolation:           -1 (default)
802-11-wireless-security.key-mgmt:      wpa-eap
802-11-wireless-security.wep-tx-keyidx: 0
802-11-wireless-security.auth-alg:      --
802-11-wireless-security.proto:         --
802-11-wireless-security.pairwise:      --
802-11-wireless-security.group:         --
802-11-wireless-security.pmf:           0 (default)
802-11-wireless-security.leap-username: --
802-11-wireless-security.wep-key0:      <hidden>
802-11-wireless-security.wep-key1:      <hidden>
802-11-wireless-security.wep-key2:      <hidden>
802-11-wireless-security.wep-key3:      <hidden>
802-11-wireless-security.wep-key-flags: 0 (none)
802-11-wireless-security.wep-key-type:  unknown
802-11-wireless-security.psk:           <hidden>
802-11-wireless-security.psk-flags:     0 (none)
802-11-wireless-security.leap-password: <hidden>
802-11-wireless-security.leap-password-flags:0 (none)
802-11-wireless-security.wps-method:    0x0 (default)
802-11-wireless-security.fils:          0 (default)
ipv4.method:                            auto
ipv4.dns:                               --
ipv4.dns-search:                        --
ipv4.dns-options:                       --
ipv4.dns-priority:                      0
ipv4.addresses:                         --
ipv4.gateway:                           --
ipv4.routes:                            --
ipv4.route-metric:                      -1
ipv4.route-table:                       0 (unspec)
ipv4.routing-rules:                     --
ipv4.replace-local-rule:                -1 (default)
ipv4.ignore-auto-routes:                no
ipv4.ignore-auto-dns:                   no
ipv4.dhcp-client-id:                    --
ipv4.dhcp-iaid:                         --
ipv4.dhcp-timeout:                      0 (default)
ipv4.dhcp-send-hostname:                yes
ipv4.dhcp-hostname:                     --
ipv4.dhcp-fqdn:                         --
ipv4.dhcp-hostname-flags:               0x0 (none)
ipv4.never-default:                     no
ipv4.may-fail:                          yes
ipv4.required-timeout:                  -1 (default)
ipv4.dad-timeout:                       -1 (default)
ipv4.dhcp-vendor-class-identifier:      --
ipv4.link-local:                        0 (default)
ipv4.dhcp-reject-servers:               --
ipv4.auto-route-ext-gw:                 -1 (default)
ipv6.method:                            disabled
ipv6.dns:                               --
ipv6.dns-search:                        --
ipv6.dns-options:                       --
ipv6.dns-priority:                      0
ipv6.addresses:                         --
ipv6.gateway:                           --
ipv6.routes:                            --
ipv6.route-metric:                      -1
ipv6.route-table:                       0 (unspec)
ipv6.routing-rules:                     --
ipv6.replace-local-rule:                -1 (default)
ipv6.ignore-auto-routes:                no
ipv6.ignore-auto-dns:                   no
ipv6.never-default:                     no
ipv6.may-fail:                          yes
ipv6.required-timeout:                  -1 (default)
ipv6.ip6-privacy:                       -1 (unknown)
ipv6.addr-gen-mode:                     default
ipv6.ra-timeout:                        0 (default)
ipv6.mtu:                               auto
ipv6.dhcp-duid:                         --
ipv6.dhcp-iaid:                         --
ipv6.dhcp-timeout:                      0 (default)
ipv6.dhcp-send-hostname:                yes
ipv6.dhcp-hostname:                     --
ipv6.dhcp-hostname-flags:               0x0 (none)
ipv6.auto-route-ext-gw:                 -1 (default)
ipv6.token:                             --
proxy.method:                           none
proxy.browser-only:                     no
proxy.pac-url:                          --
proxy.pac-script:                       --
GENERAL.NAME:                           kura-wlan0-connection
GENERAL.UUID:                           10852979-9610-42ec-bcfc-eb609c0860de
GENERAL.DEVICES:                        wlan0
GENERAL.IP-IFACE:                       wlan0
GENERAL.STATE:                          activated
GENERAL.DEFAULT:                        no
GENERAL.DEFAULT6:                       no
GENERAL.SPEC-OBJECT:                    /org/freedesktop/NetworkManager/AccessPoint/45
GENERAL.VPN:                            no
GENERAL.DBUS-PATH:                      /org/freedesktop/NetworkManager/ActiveConnection/26
GENERAL.CON-PATH:                       /org/freedesktop/NetworkManager/Settings/2
GENERAL.ZONE:                           --
GENERAL.MASTER-PATH:                    --
IP4.ADDRESS[1]:                         192.168.1.112/24
IP4.GATEWAY:                            192.168.1.1
IP4.ROUTE[1]:                           dst = 0.0.0.0/0, nh = 192.168.1.1, mt = 20600
IP4.ROUTE[2]:                           dst = 192.168.1.0/24, nh = 0.0.0.0, mt = 600
IP4.DNS[1]:                             192.168.1.10
IP4.DOMAIN[1]:                          localdomain
IP4.SEARCHES[1]:                        localdomain
DHCP4.OPTION[1]:                        broadcast_address = 192.168.1.255
DHCP4.OPTION[2]:                        dad_wait_time = 0
DHCP4.OPTION[3]:                        dhcp_lease_time = 86400
DHCP4.OPTION[4]:                        dhcp_message_type = 5
DHCP4.OPTION[5]:                        dhcp_server_identifier = 192.168.1.1
DHCP4.OPTION[6]:                        domain_name = localdomain
DHCP4.OPTION[7]:                        domain_name_servers = 192.168.1.10
DHCP4.OPTION[8]:                        domain_search = localdomain.
DHCP4.OPTION[9]:                        expiry = 1697191513
DHCP4.OPTION[10]:                       ip_address = 192.168.1.112
DHCP4.OPTION[11]:                       network_number = 192.168.1.0
DHCP4.OPTION[12]:                       next_server = 0.0.0.0
DHCP4.OPTION[13]:                       ntp_servers = 162.159.200.1 185.19.184.35
DHCP4.OPTION[14]:                       requested_broadcast_address = 1
DHCP4.OPTION[15]:                       requested_domain_name = 1
DHCP4.OPTION[16]:                       requested_domain_name_servers = 1
DHCP4.OPTION[17]:                       requested_domain_search = 1
DHCP4.OPTION[18]:                       requested_host_name = 1
DHCP4.OPTION[19]:                       requested_interface_mtu = 1
DHCP4.OPTION[20]:                       requested_ms_classless_static_routes = 1
DHCP4.OPTION[21]:                       requested_netbios_name_servers = 1
DHCP4.OPTION[22]:                       requested_netbios_scope = 1
DHCP4.OPTION[23]:                       requested_ntp_servers = 1
DHCP4.OPTION[24]:                       requested_rfc3442_classless_static_routes = 1
DHCP4.OPTION[25]:                       requested_root_path = 1
DHCP4.OPTION[26]:                       requested_routers = 1
DHCP4.OPTION[27]:                       requested_static_routes = 1
DHCP4.OPTION[28]:                       requested_subnet_mask = 1
DHCP4.OPTION[29]:                       requested_time_offset = 1
DHCP4.OPTION[30]:                       requested_wpad = 1
DHCP4.OPTION[31]:                       routers = 192.168.1.1
DHCP4.OPTION[32]:                       subnet_mask = 255.255.255.0
IP6.GATEWAY:                            --
```

</details>

---

## Additional notes:

1. For decrypting the private key I used `openssl rsa -in toshiba.key -out toshiba`
2. For running the RADIUS server I had to use the following:

```
docker run -it -p 1812:1812/udp --restart=always -v $(pwd)/pki:/etc/raddb/certs:ro -e CLIENT_ADDRESS=172.17.0.1 -e CLIENT_SECRET=secret -e PRIVATE_KEY_PASSWORD=password fralp:latest
```

Where:
- `CLIENT_SECRET`: Is what Unifi calls "Shared Secret" in the RADIUS profile
- `PRIVATE_KEY_PASSWORD`: is the password required for decrypting the `toshiba.key` used by the client
- `CLIENT_ADDRESS`: is the address of the client forwarding authentication requests. In my case the router. Issue is, since the RADIUS server is running as a docker container, it was not seeing the correct IP address but the inner Docker representation hence the `172.16.0.1` in my case.

3. I had some trouble in generating and extracting the keys from the easyrsa container. In the end I ran the container in interactive mode overriding the entrypoint using:

```
docker run -it --entrypoint /bin/sh easyrsa
```

ran the commands container in the `entrypoint.sh`, and then issued the command for generating the keys. Once done copied the `pki` folder from the container using:

```
 docker cp <container_id>:/easyrsa/pki ./pki
 ```

mattdibi:
@GregoryIvo one final question: do we really need two different field for passing the client key if they are the same?

image

... or did I do something wrong?

kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/configuration/NMSettingsConverter.java
GregoryIvo: These exceptions just show the stack trace and should be safe.

mattdibi (approved):

PR #4873, opened 2023-09-29
fix: resolved inconsistency for empty string csv input

sfiorani:
This PR fixes a bug occurring in the CSV channels upload in the asset section.

If the CSV file contained an empty value for a field that should be casted in string format, it was treated as a "null" value, so the default one from the metatype was set. The correct behaviour should be that an empty csv value should be considered as an empty string, not a null value.

This change can be very impactful on the behavior of all those drivers/assets/channels that read variables in fields like String. Let’s say the case of a string field `value` which is then casted in the code to an integer with a `Integer.decode`: at the current state of the art, if this channel is load from csv null field, the validator would set this to a default value from the metatype (for example "10") which would then be casted into an integer; with the modification of this PR, the csv would be read as "" and then would fail the decode because you would try to cast an empty value as integer.

An example of this variation is depicted in the image below: the cannels were loaded from a CSV cotaining all the `value` field empty (first image). The driver in reading phase, reads the values set in the `value` channel field and tries to cast it to the `value.type` type. The metatype set the default value to `0`, but passing a null value from the CSV, all values are considered as `""`, so the different casts fail (second image).

**Description of the solution adopted:** The validator will now check which is the field type of the CSV field, and if it's a string of type "", it's admitted and not considered as a null value.

**Screenshots:**
csv_uploaded_channels
csv_uploaded_channels_result

**Manual Tests**: Tested on Raspberry Pi 3b, Kura version: generic-aarch64

### DIFF
--- a/kura/org.eclipse.kura.web2/src/main/resources/org/eclipse/kura/web/client/messages/Messages.properties
+++ b/kura/org.eclipse.kura.web2/src/main/resources/org/eclipse/kura/web/client/messages/Messages.properties
@@ -771,6 +771,8 @@ wiresDownloadChannels=Download Channels
 wiresUploadChannels=Upload Channels
 wiresNewChannel=New Channel
 wiresDeleteChannel=Delete Channel
+wiresEmptyStringCheck=Import empty values as empty strings.
+wiresEmptyStringCheckHelpText=If checked, empty values contained in the csv will be parsed as empty strings, for all those fields whose type is <String>.
 wiresAppendChannelsCheck=Replace current channels.
 wiresAppendChannelsHelpText=If checked, channels will be replaced by the imported ones. Otherwise, existing Asset configuration will be merged with the imported one, parameters values of imported channels will override the existing ones. The change will not be applied until Asset configuration is saved.
 
--- a/kura/org.eclipse.kura.web2/src/main/resources/org/eclipse/kura/web/client/messages/Messages.properties
+++ b/kura/org.eclipse.kura.web2/src/main/resources/org/eclipse/kura/web/client/messages/Messages.properties
@@ -771,10 +771,10 @@ wiresDownloadChannels=Download Channels
 wiresUploadChannels=Upload Channels
 wiresNewChannel=New Channel
 wiresDeleteChannel=Delete Channel
-wiresEmptyStringCheck=Import empty values as empty strings.
+wiresEmptyStringCheck=Force import empty string policy.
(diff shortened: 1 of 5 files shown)


MMaiero:
@sfiorani Do we have a way to load an empty string value from csvs?
Can we add a flag in the import modal to let the user select the preferred way to manage empty csv string values?
That would allow us to keep the consistency with the old behaviour

sfiorani:
To mantain the compatibility with the previous behaviour, the following steps have been implemented:
- Added a checkbox in the csv upload popup that allow choosing if the old or new "empty value management" will be follow in the upload
- Added a short help box for the mentioned new checkbox
- Reorder of items in the popup
So now it is possible to choose how to treat empty values in the csv when the field of interest is of String type

sfiorani:
Tested also on Raspberry Pi, Kura version: raspberrypi-arm64

kura/org.eclipse.kura.web2/src/main/resources/org/eclipse/kura/web/client/messages/Messages.properties
```diff
@@ -771,6 +771,8 @@ wiresDownloadChannels=Download Channels
 wiresUploadChannels=Upload Channels
 wiresNewChannel=New Channel
 wiresDeleteChannel=Delete Channel
+wiresEmptyStringCheck=Import empty values as empty strings.
+wiresEmptyStringCheckHelpText=If checked, empty values contained in the csv will be parsed as empty strings, for all those fields whose type is <String>.
```
MMaiero, line 775: > If checked, empty values contained in the csv will be parsed as empty strings, for all those fields whose type is <String>

Maybe better like this: "If checked, all <String> typed channel fields that have an empty value in the input csv will be parsed as empty strings."

kura/org.eclipse.kura.web2/src/main/resources/org/eclipse/kura/web/client/messages/Messages.properties
```diff
@@ -771,6 +771,8 @@ wiresDownloadChannels=Download Channels
 wiresUploadChannels=Upload Channels
 wiresNewChannel=New Channel
 wiresDeleteChannel=Delete Channel
+wiresEmptyStringCheck=Import empty values as empty strings.
```
MMaiero: > Import empty values as empty strings.

Force import empty string policy

MMaiero:
@sfiorani Did you already submit the updated documentation along with this PR?

MMaiero (approved):

PR #4874, opened 2023-09-29
refactor(css): Improve Admin UI headers 

gnunzi:
This is a small change in the css to improve the look of the Admin UI:

1. Headers of sections "System" and "Services" are blue and bold (before just black), see `.panel-title`.
2. Headers of each page are blue, bold  and 16px (before just black and 14px), see `.panel-default .panel-heading`.
3. The tabs in the security section have white background if selected (same look&file as the tabs on the left side), see `.nav-tabs li.active`.

This is a minor improvement, but it increae the look&feel of the UI.

Example Section Header System:
kura_header3

Examples Page Header:
kura_header2
kura_header1

Example Tabs in Security:
kura_tabs1
kura_tabs2

Tested locally with the docker image.

I tried to avoid the !important, but the css instruction does not work without !important.

### DIFF
--- a/kura/org.eclipse.kura.web2/src/main/resources/org/eclipse/kura/web/css/denali.css
+++ b/kura/org.eclipse.kura.web2/src/main/resources/org/eclipse/kura/web/css/denali.css
@@ -71,6 +71,12 @@ body {
 	margin-bottom: 0.3cm;
 }
 
+.panel-default .panel-heading {
+	color: #2860a6 !important;
+	font-size: 16px;
+	font-weight: bold;
+}
+
 .panel:not(.help-panel):not(.network-interfaces-panel):not(.packages-dropzone-message), .panel-header:not(.help-panel) {
 	border: none !important;
 	border-radius: 0 !important;
@@ -83,6 +89,11 @@ body {
 
 .panel-body {
 	padding: 0 !important;
+}
+
+.panel-title {
+	color: #2860a6;
+	font-weight: bold;
 }
 
 .form-control {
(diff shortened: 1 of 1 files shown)


GregoryIvo:
Just posting the old UI for reference for other commuters to see the changes:

image
image
image
image
image

MMaiero:
Hi @gnunzi, 
thanks for the contribution.
We are looking into it.

Best regards,
Matteo

MMaiero:
The PR is ok for me. The visual changes make sense.
As defined in the PR the !important is a little bit problematic as it forces a behaviour and that can be an issue to adopters that have their own custom CSS.

gnunzi:
>
> As defined in the PR the !important is a little bit problematic as it
> forces a behaviour and that can be an issue to adopters that have their own
> custom CSS.


I totally agree. On the other hand, the code contains a lot of `!important´
already. Defining a custom CSS is already difficult because of the other
existing `!important` clauses. I just continued what seems the existing
approach in the css template.

gnunzi:
>As defined in the PR the !important is a little bit problematic as itforces a behaviour and that can be an issue to adopters that have their own custom CSS.


I totally agree. On the other hand,  [the code](https://github.com/eclipse/kura/blob/c8cb6af77115708c56f8714d2839c0bd66dd5465/kura/org.eclipse.kura.web2/src/main/resources/org/eclipse/kura/web/css/denali.css#L58) contains a lot of `!important` already. Defining a custom CSS is already difficult because of the other existing `!important` clauses. I just continued what seems the existing approach in the css template.

PR #4877, opened 2023-10-02
feat(nm): mtu configuration support in networkmanager and gui

fdizazzo:

This PR adds support for MTU Configuration using Networkmanager, as discussed in #4866

**Description of the solution adopted:** Mtu is new configurable through the new snapshot properties `net.interface.{ifName}.config.ip4.mtu` an `net.interface.{ifName}.config.ip6.mtu`.

**Any side note on the changes made:** Selective configuration of network interface is now possibile through the class `NMVersion`, this becomes needed as ipv6 mtu configuration is only allowed in NetworkManager 1.40 and above


mattdibi:
Hello there @fdizazzo! Welcome back :)

Thanks for your contribution. I'll perform a review ASAP. 

Meanwhile, please notice that we're experiencing some issues with our CI pipeline and the failures you see are not trustworthy. We [opened a ticket](https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/3750) to the Eclipse Foundation for that.

kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/NMVersion.java
```diff
@@ -0,0 +1,111 @@
...
+ * SPDX-License-Identifier: EPL-2.0
+ * 
+ * Contributors:
+ *  Areti
+ *******************************************************************************/
+package org.eclipse.kura.nm;
+
+import java.util.Objects;
+
+public class NMVersion implements Comparable<NMVersion> {
```
mattdibi: 🤔  Instead of implementing it from scratch wouldn't it be better to use [what OSGi provides](https://docs.osgi.org/javadoc/r4v43/core/org/osgi/framework/Version.html)?

We already use it in the codebase (example: https://github.com/eclipse/kura/blob/ec2d052dd3e698d60d8eddac24325f0f32980fd5/kura/org.eclipse.kura.deployment.agent/src/main/java/org/eclipse/kura/deployment/agent/impl/DeploymentAgent.java#L49)

kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/configuration/NMSettingsConverter.java
```diff
@@ -107,7 +109,15 @@ public static Map<String, Map<String, Variant<?>>> buildSettings(NetworkProperti
...
             Map<String, Variant<?>> vlanSettingsMap = buildVlanSettings(properties, deviceId);
+            Map<String, Variant<?>> ethSettingsMap = NMSettingsConverter.buildEthernetSettings(properties, deviceId);
             newConnectionSettings.put("vlan", vlanSettingsMap);
+            newConnectionSettings.put("802-3-ethernet", ethSettingsMap);
+        } else if (deviceType == NMDeviceType.NM_DEVICE_TYPE_ETHERNET) {
+            Map<String, Variant<?>> ethSettingsMap = NMSettingsConverter.buildEthernetSettings(properties, deviceId);
+            newConnectionSettings.put("802-3-ethernet", ethSettingsMap);
+        } else if (deviceType == NMDeviceType.NM_DEVICE_TYPE_LOOPBACK) {
+            Map<String, Variant<?>> loopSettingsMap = NMSettingsConverter.buildLoopbackSettings(properties, deviceId);
+            newConnectionSettings.put("loopback", loopSettingsMap);
```
mattdibi: We currently do not support Loopback interface configuration therefore this can be removed

```suggestion
```

kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/configuration/NMSettingsConverter.java
```diff
@@ -374,6 +385,13 @@ public static Map<String, Variant<?>> buildIpv6Settings(NetworkProperties props,
             logger.warn("Unexpected ip status received: \"{}\". Ignoring", ip6Status);
         }
 
+        Optional<Integer> mtu = props.getOpt(Integer.class, "net.interface.%s.config.ip6.mtu", deviceId);
+        logger.info("REMOVEME-- ip6mtu is {}", mtu);
+        //ipv6.mtu only supported in networkManager 1.40 and above
+        logger.info("--REMOVEME; {} > 1.40? {}", nmVersion, nmVersion.isGreaterEqualThan("1.40"));
+        if (nmVersion.isGreaterEqualThan("1.40")) {
+            mtu.ifPresent(value -> settings.put("mtu", new Variant<>(new UInt32(value))));
+        }
```
mattdibi: For a better UX I would log a warning if and only if the mtu is set and the version is lower than 1.40. Something like:

```suggestion
        Optional<Integer> mtu = props.getOpt(Integer.class, "net.interface.%s.config.ip6.mtu", deviceId);
        if(mtu.isPresent()) {
          if (nmVersion.isGreaterEqualThan("1.40")) {
              settings.put("mtu", new Variant<>(new UInt32(mtu.get())));
          } else {
             logger.warn("NetworkManager version lower than 1.40, 'mtu' parameter ignored");
          }
       }
```

but I'm open to nicer suggestions

kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/configuration/NMSettingsConverter.java
```diff
@@ -724,5 +762,10 @@ private static String connectionTypeConvert(NMDeviceType deviceType) {
                     .format("Unsupported connection type conversion from NMDeviceType \"%s\"", deviceType.toString()));
         }
     }
+    
+    private static Integer getIp4MtuFromProperties(NetworkProperties properties, String deviceId) {
+        Optional<Integer> mtu = properties.getOpt(Integer.class, "net.interface.%s.config.ip4.mtu", deviceId);
+        return mtu.orElse(0);
+    }
```
mattdibi: Do we really need this? I would follow the established convention and, in case `mtu` is not set, simply not put it in the outgoing map.

I think you need this is caused by [this code here](https://github.com/eclipse/kura/blob/fc539f270f41906058e2fff72401479e75bd48ee/kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/configuration/NMSettingsConverter.java#L110-118) where you'll have empty map if no `mtu` is set.

I would check there if the map is empty and avoid setting the map entirely in the higher level 🤔  What do you think?

This is to better separate abstraction levels. In the `buildEthernetSettings` we only care about `802-1-ethernet` settings. It's the abstraction level above that should care about the fact that the Map is empty and therefore not set it.

kura/test/org.eclipse.kura.nm.test/src/test/java/org/eclipse/kura/nm/configuration/NMSettingsConverterTest.java
```diff
@@ -27,6 +27,7 @@
 import java.util.Optional;
```
mattdibi, line 33: Here I'd like too see at least one test checking that, if the NM version is greater than 1.40 the `mtu` parameter is set. If, for the same configuration, the NM version is lower than 1.40 the `mtu` parameter is not set

kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/configuration/NMSettingsConverter.java
```diff
@@ -531,6 +553,22 @@ public static Map<String, Variant<?>> buildVlanSettings(NetworkProperties props,
...
+        return settings;        
+    }
+
+    public static Map<String, Variant<?>> buildLoopbackSettings(NetworkProperties props, String deviceId) {
+        Map<String, Variant<?>> settings = new HashMap<>();
+        Integer mtu = getIp4MtuFromProperties(props, deviceId);
+        logger.info("REMOVEME-- lo mtu {}", mtu);
+        settings.put("mtu", new Variant<>(new UInt32(mtu)));
+        return settings;
+    }
```
mattdibi: As we said, we don't support Loopback configuration, therefore this can be removed.

```suggestion
```

kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/NMDbusConnector.java
```diff
@@ -487,14 +487,15 @@ private void enableInterface(String deviceId, NetworkProperties properties, Opti
...
+        NMVersion nmVersion = this.networkManager.getVersion();
         if (Boolean.FALSE.equals(this.networkManager.isDeviceManaged(device))) {
             this.networkManager.setDeviceManaged(device, true);
         }
         String interfaceName = this.networkManager.getDeviceInterface(device);
 
         Optional<Connection> connection = this.networkManager.getAssociatedConnection(device);
         Map<String, Map<String, Variant<?>>> newConnectionSettings = NMSettingsConverter.buildSettings(properties,
-                connection, deviceId, interfaceName, deviceType);
+                connection, deviceId, interfaceName, deviceType, nmVersion);
```
mattdibi: I would avoid creating a variable just for this (and think that Sonar would complain) therefore just pass `this.networkManager.getVersion()` to the `buildSettings` method

kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/NMDbusConnector.java
```diff
@@ -537,8 +538,9 @@ private void enableInterface(String deviceId, NetworkProperties properties, Devi
     
     private void createVirtualInterface(String deviceId, NetworkProperties properties, NMDeviceType deviceType)
         throws DBusException {
+        NMVersion nmVersion = networkManager.getVersion();
         Map<String, Map<String, Variant<?>>> newConnectionSettings = NMSettingsConverter.buildSettings(properties,
-                Optional.empty(), deviceId, deviceId, deviceType);
+                Optional.empty(), deviceId, deviceId, deviceType, nmVersion);
```
mattdibi: I would avoid creating a variable just for this (and think that Sonar would complain) therefore just pass `this.networkManager.getVersion()` to the `buildSettings` method

mattdibi (changes requested):
Looking good so fare. I suggested a couple of changes LMKWYT

kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/configuration/NMSettingsConverter.java
fdizazzo: This was to let a user delete the mtu from its field and let it reset to 0/auto on its own, but your suggestion is also fine :)
`getIp4MtuFromProperties` will return an Optional (I'd keep the method as its called from 3 different places) and as such will follow the usual map population logic

kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/NMVersion.java
fdizazzo: Didn't notice its existance :)

After a few tests I've found that is breaks on non numeric characters and we'll end up using `.compareTo(Version.parseVersion(x)>=0` for each comparison or wrapping it with the additional logic needed 🤔

mattdibi: 🤔 I don't expect to find alphanumeric characters for NM version, and yeah it's not the best from a readability point of view... but it's something we already use in the codebase and doesn't need additional unit tests.

🤷‍♂️ Up to you... :)

kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/configuration/NMSettingsConverter.java
fdizazzo: > I think you need this is caused by [this code here](https://github.com/eclipse/kura/blob/fc539f270f41906058e2fff72401479e75bd48ee/kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/configuration/NMSettingsConverter.java#L110-118) where you'll have empty map if no mtu is set.
> 
> I would check there if the map is empty and avoid setting the map entirely in the higher level 🤔 What do you think?

Can confirm that Network Manager doesn't care about empty maps, so there's no need for additional checks 😋 
Should the UI bring up a 0 as a default value or just leave an empty field until otherwise configured?

mattdibi: > Should the UI bring up a 0 as a default value or just leave an empty field until otherwise configured?

0 as default. We had a field set as your second option in the past, but decided it was confusing for the user (it was the WAN priority, which behaves similarly to MTU)

mattdibi:
@fdizazzo [CI build should be fixed now](https://github.com/eclipse/kura/pull/4877#issuecomment-1746214633). Please update your branch to integrate the latest changes on `develop`

fdizazzo:
Hi @mattdibi glad to see a proper failure on my build🥳
How did we solve the issue about generic wildcards on `Variant<?>`? Does it requires manual exclusion on each instance?

mattdibi:
> Hi @mattdibi glad to see a proper failure on my build🥳 How did we solve the issue about generic wildcards on `Variant<?>`? Does it requires manual exclusion on each instance?

😆 Hi there, yeah for each

image

we need @MMaiero super-powers unfortunately.

MMaiero:
Can you please add the link reference to the reported issues and the justification?
So that we keep track on the decision process

mattdibi:
> Can you please add the link reference to the reported issues and the justification? So that we keep track on the decision process

https://sonarcloud.io/project/issues?branch=PR-4877&open=AYsd-aDwR_HEYAh0mGOa&id=org.eclipse.kura%3Akura "Won't fix. Required by the underlying library (dbus-java)"

https://sonarcloud.io/project/issues?impactSeverities=HIGH&resolved=false&types=CODE_SMELL&branch=PR-4877&id=org.eclipse.kura%3Akura&open=AYsd-aDwR_HEYAh0mGOb "Won't fix. Required by the underlying library (dbus-java)"

MMaiero:
> > Can you please add the link reference to the reported issues and the justification? So that we keep track on the decision process
> 
> https://sonarcloud.io/project/issues?branch=PR-4877&open=AYsd-aDwR_HEYAh0mGOa&id=org.eclipse.kura%3Akura "Won't fix. Required by the underlying library (dbus-java)"
> 
> https://sonarcloud.io/project/issues?impactSeverities=HIGH&resolved=false&types=CODE_SMELL&branch=PR-4877&id=org.eclipse.kura%3Akura&open=AYsd-aDwR_HEYAh0mGOb "Won't fix. Required by the underlying library (dbus-java)"

Done

mattdibi:
@MMaiero now Sonar [is complaining](https://sonarcloud.io/project/issues?resolved=false&sinceLeakPeriod=true&types=CODE_SMELL&branch=PR-4877&id=org.eclipse.kura%3Akura&open=AYsgNQXQygiQpO2sgOh-) because it wants us to use the newly introduced `NM_SETTINGS_ETHERNET` constant for the device type. I think it is fundamentally wrong since they are two distinct concepts that happen to have the same string representation.

I would propose to add a "Won't fix" here because the constant represent a different concept.

@fdizazzo for everything else I'll perform the review ASAP

mattdibi:
@fdizazzo Hi there,
due to unforeseen circumstances and [us attending the EclipseCon conference](https://github.com/eclipse/kura/discussions/4897) I won't be able to review your PR until the 23rd Oct at the earliest. Thank you for your patience 🙏 

Meanwhile I would ask you to address the 3 smells reported by sonar:
image

fdizazzo:
Hi @mattdibi, given we now have some more time I'd widen the scope of this PR to add:

- Support for promiscuous mode in NetworkManager via snapshot configuration (requires 1.32 and above)
- Display currently installed NetworkManager version in the 'Hardware' tab

GL for Eclipsecon! 🤘

mattdibi:
> Hi @mattdibi, given we now have some more time I'd widen the scope of this PR to add:
> 
> * Support for promiscuous mode in NetworkManager via snapshot configuration (requires 1.32 and above)
> * Display currently installed NetworkManager version in the 'Hardware' tab
> 
> GL for Eclipsecon! 🤘

Hi there @fdizazzo,

thanks! :)

Regarding the scope of the PR: this is great to hear and we'll gladly accept your contribution but we'd prefer to have these two additional features in separate PRs. This simplify review, testing and is more compliant to [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) which we recently adopted.

kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/client/ui/network/TabIp4Ui.java
mattdibi, line 1: The possibility to set MTU is a Net2 (i.e. new networking) feature only. Therefore you should add it to the `initNet2FeaturesOnly` method so that it is only shown in Kura installation running the new networking code

kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/server/net2/configuration/GwtNetInterfaceConfigBuilder.java
```diff
@@ -122,6 +123,9 @@ private void setIpv4Properties() {
         this.gwtConfig.setSubnetMask(this.properties.getIp4Netmask(this.ifName));
         this.gwtConfig.setGateway(this.properties.getIp4Gateway(this.ifName));
         this.gwtConfig.setDnsServers(this.properties.getIp4DnsServers(this.ifName));
+
+        Optional<Integer> mtu = this.properties.getIp4Mtu(this.ifName);
+        this.gwtConfig.setMtu(mtu.isPresent() ? mtu.get() : DEFAULT_MTU);
```
mattdibi: I think this can be simplified to:

```suggestion
        this.gwtConfig.setMtu(mtu.orElse(DEFAULT_MTU));
```

kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/server/net2/configuration/GwtNetInterfaceConfigBuilder.java
```diff
@@ -161,6 +165,9 @@ private void setIpv6Properties() {
         if (privacy.isPresent()) {
             this.gwtConfig.setIpv6Privacy(privacy.get());
         }
+
+        Optional<Integer> mtu = this.properties.getIp6Mtu(this.ifName);
+        this.gwtConfig.setIpv6Mtu(mtu.isPresent() ? mtu.get() : DEFAULT_MTU);
```
mattdibi: Same here:

```suggestion
        this.gwtConfig.setIpv6Mtu(mtu.orElse(DEFAULT_MTU));
```

kura/org.eclipse.kura.web2/src/main/resources/org/eclipse/kura/web/client/messages/Messages.properties
```diff
@@ -336,7 +336,10 @@ netIPv4ToolTipAddress=For static configuration, enter the IP address that will b
 netIPv4ToolTipSubnetMask=For static configuration, enter the subnet mask that will be assigned to the device.
 netIPv4ToolTipGateway=For static configuration, enter the default gateway for the device.
 netIPv4ToolTipRenew=For DHCP configuration, clicking this button will release the current IP address and acquire a new lease from the server.
-netIPv4ToolTipDns=List of DNS servers.<br/>New servers can be added by entering an address and clicking apply.<br/>To remove a server, delete the address from the list and click apply.<br/>Servers MUST be separated by a comma (,).<br/><br/>When configured for DHCP, manually entered DNS entries will replace the ones provided by the DHCP server.  Removing them will restore the DHCP entries.
+netIPv4ToolTipDns=List of DNS servers.<br/>New servemrs can be added by entering an address and clicking apply.<br/>To remove a server, delete the address from the list and click apply.<br/>Servers MUST be separated by a comma (,).<br/><br/>When configured for DHCP, manually entered DNS entries will replace the ones provided by the DHCP server.  Removing them will restore the DHCP entries.
```
mattdibi: Typo

```suggestion
netIPv4ToolTipDns=List of DNS servers.<br/>New servers can be added by entering an address and clicking apply.<br/>To remove a server, delete the address from the list and click apply.<br/>Servers MUST be separated by a comma (,).<br/><br/>When configured for DHCP, manually entered DNS entries will replace the ones provided by the DHCP server.  Removing them will restore the DHCP entries.
```

kura/test/org.eclipse.kura.nm.test/src/test/java/org/eclipse/kura/nm/configuration/NMSettingsConverterTest.java
```diff
@@ -452,6 +454,34 @@ public void buildIpv6SettingsShouldThrowIllegalArgumentExceptionWhenGivenExpecte
...
+    
+    @Test
+    public void buildIpv6SettingsShouldHaveMtuWhenSupported() {
+    	givenMapWith("net.interface.wlan0.config.ip6.address.method", "netIPv6MethodAuto");
+        givenMapWith("net.interface.wlan0.config.ip6.status", "netIPv6StatusEnabledWAN");
+        givenMapWith("net.interface.wlan0.config.ip6.mtu", 2345);
+        givenNetworkPropsCreatedWithTheMap(this.internetNetworkPropertiesInstanciationMap);
+        givenNetworkManagerVersion("1.40.18");
+
+        whenBuildIpv6SettingsIsRunWith(this.networkProperties, "wlan0");
```
mattdibi: Can you please update the `whenBuildIpv6SettingsIsRunWith` signature so that it accepts the version as an argument? This is so that we're consistent throughout the codebase.

```suggestion
        givenNetworkManagerVersion("1.40.18");

        whenBuildIpv6SettingsIsRunWith(this.networkProperties, "wlan0", this.nmVersion);
```

kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/NMVersion.java
mattdibi: I think, at this point, you want to keep this class (which is ok for me). But `NMVersion` doesn't really make much sense since it just represents a Semantic Version not the version for NetworkManager specifically. We could use the same class for representing the ModemManager or WPA Supplicant versions... could that be a better name? `SemanticVersion`?

mattdibi (changes requested):
A couple of minor issues. The most important one is the one regarding the `MTU` form being shown in "old networking" installations UIs. We're almost there!

Thank you for your patience BTW. It took some time to review this XD

mattdibi:
@fdizazzo Also please, if you can, add some unit tests for the `NMVersion` class. It should be pretty easy and helps us understand how it works/should be used in the future :)

kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/NMVersion.java
fdizazzo: Ended up finding a `1.30.0-1` as a NetworkManager version, which would not work on `org.osgi.framework.Version`. Not sure if it's creativity from the OEM or something we can realistically encounter out in the field, but it's still breaking for my use case 👼 
OK on renaming it 👍

kura/test/org.eclipse.kura.nm.test/src/test/java/org/eclipse/kura/nm/SemanticVersionTest.java
```diff
@@ -0,0 +1,179 @@
...
+            assertEquals(0, compare);
+        } catch (Exception e) {
+            this.thrownException = e;
+            this.exceptionHasBeenThrown = true;           
+        }
+    }
+    
+    
+    
+
```
mattdibi: A couple of extra newlines here

```suggestion
```

kura/test/org.eclipse.kura.nm.test/src/test/java/org/eclipse/kura/nm/SemanticVersionTest.java
```diff
@@ -0,0 +1,179 @@
...
+    }
+    
+    private void thenNoExceptionHasBeenThrown() {
+        assertFalse(this.exceptionHasBeenThrown);
+    }
+    
+    private void thenNullPointerExceptionHasBeenThrown() {
+        assertTrue(this.exceptionHasBeenThrown);
+        assertTrue(this.thrownException instanceof NullPointerException);
+    }
```
mattdibi: For easier debugging we recently switched to a new "recipe" just for this

> There are basically two different approaches, the usual try-catch method and the usage of JUnit APIs. We opted for using the try-catch approach since it is easier to understand and allows to easily check for side effects.
>
>The usual pattern will be as follows:
>
> 
> ```java
> private Exception occurredException;
> 
> ...
> 
> @Test
> public void myTestShouldThrowKuraException() {
>     givenX();
>     
>     whenMyMethodIsCalled();
>     
>     thenExceptionOccurred(KuraException.class);
>     thenDirtyFileWasRemoved();
> }
> 
> ...
> 
> private void whenMyMethodIsCalled() {
>     try {
>         this.testedObject.myMethod();
>     } catch (Exception e) {
>         this.occurredException = e;
>     } 
> }
> 
> ...
> 
> private void thenNoExceptionOccurred() {
>     String errorMessage = "Empty message";
>     if (Objects.nonNull(this.occurredException)) {
>         StringWriter sw = new StringWriter();
>         this.occurredException.printStackTrace(new PrintWriter(sw));
> 
>         errorMessage = String.format("No exception expected, \"%s\" found. Caused by: %s",
>                 this.occurredException.getClass().getName(), sw.toString());
>     }
> 
>     assertNull(errorMessage, this.occurredException);
> }
> 
> private <E extends Exception> void thenExceptionOccurred(Class<E> expectedException) {
> 	assertNotNull(this.occurredException);
> 	assertEquals(expectedException.getName(),
> 	this.occurredException.getClass().getName());
> }
> ```
>
> In the example above, we have the then clause that compares for equality of the exceptions class names. Using this approach, when the test fails we have logged the expected output and the actual one directly by JUnit. It is possible to integrate the error message with the method [assertEquals(String message, Object o1, Object o2);](https://junit.org/junit4/javadoc/4.13/org/junit/Assert.html) like, for example, embedding the stack trace.

With this recipe we can log the stacktrace in case an unexpected exception is thrown which allows for a much easier test debugging experience

Please add the two `then` methods reported above and use those in your tests. They should work out-of-the-box

mattdibi (changes requested):
After my two final comment we should be good. I'll perform some testing in the next days and we should be ready to merge 👍

mattdibi:
Tested working on my Raspberry Pi with both old and new networking. After the latest suggested changes we'll be ready for merge! GJ @fdizazzo 👍  :)

mattdibi (approved):

mattdibi:
Thanks again for your contribution @fdizazzo :)

PR #4881, opened 2023-10-06
feat(web2): Wpa enterprise EAP-TLS certificate support frontend

GregoryIvo:
This PR adds front-end support for: https://github.com/eclipse/kura/pull/4872


GregoryIvo:
V1 Web UI
image


I am now testing to see if dropdowns with key stores are possible now.

GregoryIvo:
Almost ready for review:
Just need to add a default password value

mattdibi:
Tested this at d4d9c2e03ba614777d7f0cafc7f364aa18deb320 in #4872. Looking good so far!

@GregoryIvo one question: do we really need two different fields for passing the client key if they are the same?

image

... or did I do something wrong?

GregoryIvo:
This is correct, and expected. I was thinking about that too but am torn because I like the potential flexibility

mattdibi:
> This is correct, and expected. I was thinking about that too but am torn because I like the potential flexibility

As a user I find it confusing... Why should I set the same for both if I added it as a single Pub/Priv key pair? Also I have a hard time imagining a situation in which this flexibility could be taken advantage of? Can you explain to me a scenario where these two separated fields can be useful? 🤔

mattdibi:
Superseded by #4895

PR #4887, opened 2023-10-10
ci(Jenkinsfile): remove build discarder overrides

mattdibi:
We're currently suffering from a pile up in orphaned Jenkins builds with 300+ old builds on our Jenkins instance.

image

This is probably due to the fact that we're overriding the Orphaned Item Strategy in our Jenkinsfile by specifing a different Build Discarder configuration. See: https://plugins.jenkins.io/build-discarder/

By removing the override in the Jenkinsfile the old builds should be removed.


mattdibi:
Looks like there's no need for this Jenkinsfile update since a Repository Scan is all that is needed for starting the pruning process.

https://github.com/jenkinsci/multibranch-action-triggers-plugin/issues/16

An improvement would be to have a periodic scan of the repo so that the pruning process is kicked automatically

PR #4889, opened 2023-10-10
fix(linux.net): Added check for existing file before iptables restore

pierantoniomerlino:

This PR adds a check before launching `iptables-restore` command for testing the file presence.


kura/org.eclipse.kura.linux.net/src/main/java/org/eclipse/kura/linux/net/iptables/IptablesConfig.java
```diff
@@ -623,20 +624,24 @@ private void writeAdditionalRulesToMangleTable(PrintWriter writer) {
...
         List<FilterForwardChainRule> filterForwardChain = new ArrayList<>();
-        try (FileReader fr = new FileReader(getFirewallConfigFileName()); BufferedReader br = new BufferedReader(fr)) {
-            parseIptablesRules(natPreroutingChain, natPostroutingChain, filterForwardChain, br);
-            // ! done parsing !
-            parsePortForwardingRules(natPreroutingChain, natPostroutingChain);
-            parseIpForwardingRules(natPreroutingChain, natPostroutingChain, filterForwardChain);
-        } catch (IOException e) {
-            throw new KuraIOException(e, "restore() :: failed to read configuration file");
+        File iptablesFile = new File(getFirewallConfigFileName());
+        if (iptablesFile.exists()) {
```
MMaiero: Is it better to reduce the cognitive complexity to reverse the check?

```
if (iptablesFile.exists()) {
  //maybe log as it is not the supposed state?
  return;
}
...
```

MMaiero (approved):

PR #4893, opened 2023-10-12
fix: Return placeholder instead of encrypted passwords in REST APIs

nicolatimeus:

Changes the following rest endpoints that currently return encrypted configuration passwords to return the "placeholder" string instead:

* `CONF-V2/configurableComponents/configurations`
* `CONF-V2/configurableComponents/configurations/byPid`
* `WIRE-V1/configs/byPid`


mattdibi (approved):
LGTM

Tested on RPi with

```bash
curl -X GET -k -u $USER:$PASSWORD https://$ADDRESS/services/configuration/v2/configurableComponents/configurations -v
```

```
... "keystore.password":{"value":"placeholder","type":"PASSWORD"} ...
```

PR #4905, opened 2023-10-20
feat(linux.net): Add a Dhcp Server lease file for interface

pierantoniomerlino:

This PR implements a specific lease file for each network interface when using the DHCP server.

**Description of the solution adopted:** Currently if two DHCP servers are running in the system, the DHCP server instances will use the same default lease file in `/var/lib/dhcp/dhcpd.leases`. To avoid this, this PR use the same approach of the DHCP server configuration file and pid file that have a per interface file. 
The specific lease file is added in the configuration file for every supported DHCP server (dhcpd, udhcpd and dnsmasq).

Moreover, a limited refactoring is done in order to simply the creation of the configuration string for the tools used as DHCP servers. A set of converters from the `DhcpServerConfigIP` to String are implemented.

Finally, the `DhcpLeaseTool` class has been updated to take into account the new lease files.


kura/org.eclipse.kura.net.admin/src/main/java/org/eclipse/kura/net/admin/visitor/linux/DhcpConfigWriter.java
```diff
@@ -18,6 +18,7 @@
 import java.io.PrintWriter;
```
marcellorinaldo, line 18: Please update copyright header

kura/org.eclipse.kura.linux.net/src/main/java/org/eclipse/kura/linux/net/dhcp/DhcpServerManager.java
```diff
@@ -87,6 +98,14 @@ public boolean enable(String interfaceName) throws KuraException {
         return false;
     }
 
+    private void createLeasesFile(String interfaceName) {
+        try {
+            FileUtils.touch(new File(DhcpServerManager.getLeasesFilename(interfaceName)));
+        } catch (IOException e) {
+            logger.error("Cannot create DHCP server leases file", e);
```
marcellorinaldo: Shall we throw a `KuraException` here to prevent starting the interface without leases file?

kura/org.eclipse.kura.linux.net/META-INF/MANIFEST.MF
```diff
@@ -54,6 +54,7 @@ Export-Package: org.eclipse.kura.internal.board;version="1.0.0";x-internal:=true
    org.eclipse.kura.net.modem,
    org.eclipse.kura.usb",
  org.eclipse.kura.linux.net.dhcp;version="2.1.0";uses:="org.eclipse.kura.net.dhcp",
+ org.eclipse.kura.linux.net.dhcp.server;version="1.0.0";uses="",
```
marcellorinaldo: Can we remove the empty `uses=` clause?

kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/server/GwtNetworkServiceImplFacade.java
```diff
@@ -144,14 +144,14 @@ public boolean isIEEE80211ACSupported(GwtXSRFToken xsrfToken, String ifaceName)
...
-    public List<String> getDhcpLeases(GwtXSRFToken xsrfToken) throws GwtKuraException {
+    public List<String> getDhcpLeases(GwtXSRFToken xsrfToken, String interfaceName) throws GwtKuraException {
         checkXSRFToken(xsrfToken);
 
         if (isNet2()) {
             // TODO
             return new ArrayList<>();
         } else {
-            return org.eclipse.kura.web.server.net.GwtNetworkServiceImpl.getDhcpLeases();
+            return org.eclipse.kura.web.server.net.GwtNetworkServiceImpl.getDhcpLeases(interfaceName);
```
marcellorinaldo, line 154: Maybe we can call `org.eclipse.kura.web.server.net.GwtNetworkServiceImpl.getDhcpLeases(interfaceName);` also for net2 case.

pierantoniomerlino: We can't since that method uses the `NetworkAdminService` that is not available on net2.

marcellorinaldo (approved):

PR #4906, opened 2023-10-21
ci(automation): fix labels and nr of operations for stale issues bot

mattdibi:
This PR fixes some small issues we found when running the stale bot for the first time:
1. Changed exempt labels to be more consistent with existing ones (`exempt-stale` instead of `ExemptStale`)
2. Higher [`operation-per-run`](https://github.com/actions/stale#operations-per-run) since we were hitting the limit due to the high number of open issues and PRs we have. Please note that this have a chance of hitting the [rate limit](https://docs.github.com/en/rest/overview/resources-in-the-rest-api?apiVersion=2022-11-28#rate-limiting) of Github APIs when running in production mode (we're running the workflow in debug mode for now)


mattdibi:
Merging without review since it's a dry-run and will have no effect on the repository so that we can run this overnight and have the result tomorrow morning.

PR #4920, opened 2023-10-26
feat(distrib): network interface names retrieved without `nmcli`

marcellorinaldo:
This PR changes the installation scripts to not rely on `nmcli` anymore and instead use the information available in `/sys/class/net`.

**Related Issue:** N/A.

**Description of the solution adopted:** `/sys/class/net` contains symlinks to the network interface devices [1]. These links are named after the currently used naming convention (systemd consistent naming, hardcoded, etc.). The general rule [2] [3] for naming interfaces is as follows:

- name starts with `et` or `en` -> ethernet interface
- name starts with `wl` -> wireless lan
- name starts with `ww` -> wireless wan

If consistent network interface naming is applied [3], then:

- `o` -> on-board device (example: `eno0`, `eno1`)
- `s` -> hotplug device (example: `ens1`)
- `p` -> pci device (example: `enp3`, `wlp1`)

To determine the primary network interface the retrieved list is ordered by name, so that on-board devices have precedence over pci, that, in turn, have precedence over hotplug slots. Example: `enp3s0, eno1` -> primary is `eno1`.

**Screenshots:** N/A.

**References:**
- [1] https://www.kernel.org/doc/Documentation/ABI/testing/sysfs-class-net
- [2] https://www.freedesktop.org/wiki/Software/systemd/PredictableNetworkInterfaceNames/
- [3] https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/7/html/networking_guide/sec-understanding_the_predictable_network_interface_device_names

### DIFF
--- a/kura/distrib/src/main/resources/common/find_net_interfaces.py
+++ b/kura/distrib/src/main/resources/common/find_net_interfaces.py
@@ -39,13 +39,6 @@ def get_eth_wlan_interfaces_names():
     wireless_interface_names = list()
 
     interface_names = get_interface_names()
-
-    # en, et = ethernet
-    # wl = wireless lan
-    # ww = wireless wan
-    # eno<index> = on-board ethernet
-    # ens<slot> = hotplug on slot <slot> ethernet
-    # enp<index> = pci ethernet
     
     for ifname in interface_names:
         if ifname.startswith("en") or ifname.startswith("et"):
--- a/kura/distrib/src/main/resources/common/find_net_interfaces.py
+++ b/kura/distrib/src/main/resources/common/find_net_interfaces.py
@@ -11,57 +11,58 @@
 # Contributors:
 #  Eurotech
 #
-import subprocess
 import sys
 import logging
+import os
 
-ETHERNET_TYPES = ['ethernet', 'eth', 'wired']
-WIRELESS_TYPES = ['wifi', 'wireless']
+def get_interface_names():
+    if not os.path.exists("/sys/class/net"):
+        logging.error("'/sys/class/net' does not exists, unable to read interface names. Exiting.")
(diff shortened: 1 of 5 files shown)


kura/distrib/src/main/resources/common/find_net_interfaces.py
```diff
@@ -11,57 +11,58 @@
...
-            ethernet_inteface_names.append(interface_name)
-        
-        if  interface_type in WIRELESS_TYPES:
-            wireless_inteface_names.append(interface_name)
+    # en, et = ethernet
+    # wl = wireless lan
+    # ww = wireless wan
+    # eno<index> = on-board ethernet
+    # ens<slot> = hotplug on slot <slot> ethernet
+    # enp<index> = pci ethernet
```
mattdibi: One small detail: while reading this for a moment I thought you forgot to delete some commented out code. Can you add a small comment just to explain that this is your criteria for differentiating the interfaces types?

Something like:

```suggestion
    # We use the interface name to differentiate their type. Criteria used below:
    # en*, et* -> ethernet
    # wl* -> wireless lan
    # ww* -> wireless wan
    # eno<index> -> on-board ethernet
    # ens<slot> -> hotplug on slot <slot> ethernet
    # enp<index> -> pci ethernet
```

Or what you wrote in the PR description

marcellorinaldo: Maybe the code is clear enough. I think we can remove the comment. What do you think?

mattdibi: Good idea. The code is clear enough that it doesn't require explainatory comment 👍

mattdibi (approved):

PR #4957, opened 2023-11-07
feat(nm): promiscuous mode configuration via snapshot

fdizazzo:
This PR adds promiscuous mode support for ethernet interfaces.
This is achieved by adding a property `net.interface.<interfaceName>.config.promisc` with the possibile values:

- -1 - System default (usually false)
-  0 - Promisc mode disabled
- 1 - Promisc mode enabled

mapping the possibile values of NetworkManager `NMTernary` enumeration [1] which is then mapped to the `accept-all-mac-addresses` property of 802-3-ethernet settings map [2].

This behaviour in only enabled when NetworkManager 1.32 or newer is installed [3].

The following snapshot snippet can be used for testing
```xml
<?xml version="1.0" encoding="UTF-8"?>
<esf:configurations xmlns:esf="http://eurotech.com/esf/2.0" xmlns:ocd="http://www.osgi.org/xmlns/metatype/v1.2.0">
    <esf:configuration pid="org.eclipse.kura.net.admin.NetworkConfigurationService">
        <esf:properties>
            <esf:property array="false" encrypted="false" name="net.interface.ens33.config.promisc" type="Integer">
                <esf:value>1</esf:value>
            </esf:property>
        </esf:properties>
    </esf:configuration>
</esf:configurations>
```
and the modification can be checked from either nmcli `802-3-ethernet.accept-all-mac-addresses` property or ifconfig `PROMISC` flag

[1] https://networkmanager.dev/docs/api/latest/nm-dbus-types.html
[2] https://networkmanager.dev/docs/api/latest/nm-settings-dbus.html
[3] https://networkmanager.dev/blog/networkmanager-1-32/


kura/org.eclipse.kura.net.configuration/src/main/java/org/eclipse/kura/net/configuration/NetworkConfigurationServiceCommon.java
```diff
@@ -457,6 +458,13 @@ private static void addMtuDefinition(Tocd tocd, String ifaceName) {
         tad.setRequired(true);
         tocd.addAD(tad);
     }
+    
+    private static void addPromiscDefinition(Tocd tocd, String ifaceName) {
+        Tad tad = buildAttributeDefinition(String.format(PREFIX + "%s.config.promisc", ifaceName),
+                NetworkConfigurationPropertyNames.CONFIG_PROMISC, Tscalar.INTEGER);
+        tad.setRequired(true);
```
mattdibi, line 465: What happens if it is not provided? There's no default provided either so I expect bad things to happen in the old networking 🤔

mattdibi (commented):
Hello there @fdizazzo

thank you for your contribution and your patience. We're full in the release process of 5.4.0 and it's taking quite a lot of bandwidth.

Your contribution looks great but we have some concerns about setting the OSGi property as "required" without a default. Is this working on the old networking profiles?

Also, can you share the use case for the promiscuous mode?

kura/org.eclipse.kura.net.configuration/src/main/java/org/eclipse/kura/net/configuration/NetworkConfigurationServiceCommon.java
fdizazzo: This could probably be a `.setRequired(false)`. The idea is the Kura should do nothing if the promisc parameter is not populated. And it's actually working fine with required fields being missing 🤔 

Should we actually want it as required we can use `-1` as the default value, as populated by NetworkManager on newly created connections.

fdizazzo:
> We're full in the release process of 5.4.0 and it's taking quite a lot of bandwidth

Sorry I didn't even noticed the feature freeze announcement until yesterday :)

> Also, can you share the use case for the promiscuous mode?

I'm needing this to intercept layer 2 traffic not actually meant for the edge gateway so it can be wrapped in tcp packets

kura/org.eclipse.kura.net.configuration/src/main/java/org/eclipse/kura/net/configuration/NetworkConfigurationServiceCommon.java
mattdibi: 🤔  @pierantoniomerlino thoughts?

pierantoniomerlino: IMO, it is sufficient to add a default value (probably -1 is the right choice).
```
tad.setDefault("-1");
```

pierantoniomerlino (approved):

mattdibi:
Hello there @fdizazzo, sorry for taking so long. We were dealing with the long tail of tasks related to the release process.

I'll now take a look into this and report back ASAP.

Meanwhile, would you consider contributing also the UI part of this PR? Could be a separeted PR if you want.

mattdibi:
Tested working on RPi Bookworm (NetworkManager 1.42.4) Generic ARM64 profile.

Taking the Raspberry Pi profile for a spin just to be on the safe side...

mattdibi:
Raspberry Pi 64 bit profile on RPi Bookworm looks ok. Merging as-is! Thanks @fdizazzo let me know what you think about the UI proposal and if you can take care of the documentation side of things :)

PR #4988, opened 2023-11-16
fix: Fixed web ui access for identity with limited permissions

nicolatimeus:
Currently, after creating a new identity and assigning it a set of specific permissions Kura starts to flicker with an error dialog on screen (see video below).

https://github.com/eclipse/kura/assets/22748355/f292c183-8a2c-467c-a44b-0fd85f606d2c

This was due to the Web UI trying to perform some calls (trying to establish if the profile was using the new networking) even if the current identity does not have the required permissions, resulting in 403 errors.

This PR fixes the issue


mattdibi:
## Test performed

### On `develop` branch

Created an identity `testIdentity` with permission "kura.device".

image

 This results in the behaviour of the video below.

https://github.com/eclipse/kura/assets/22748355/0c06c1da-95cf-4f74-9a7d-e1a430b5b8fb

### On the fix branch

The identity with the same permissions results in no error in the UI.

image

mattdibi (approved):

marcellorinaldo (approved):

PR #5008, opened 2023-11-21
fix(distrib): Added missing rest permissions for rest.security, rest.system and rest.cloudconnection in snapshot_0.xml files.

salvatore-coppola:
Added missing rest permissions for `rest.security`,` rest.system` and `rest.cloudconnection` in snapshot_0.xml files.


kura/org.eclipse.kura.rest.security.provider/src/main/java/org/eclipse/kura/internal/rest/security/provider/SecurityRestService.java
```diff
@@ -86,6 +86,11 @@ public void unbindRequestHandlerRegistry(RequestHandlerRegistry registry) {
     @Path("/security-policy-fingerprint/reload")
     @Produces(MediaType.APPLICATION_JSON)
     public Response reloadSecurityPolicyFingerprint() {
+
+        if (this.security == null) {
```
nicolatimeus: This change is not needed for this PR

nicolatimeus (approved):

PR #5026, opened 2023-11-27
fix(web2): Fixed network radio mode value passing in the old networking.

salvatore-coppola:
Fixed the passing of the radio mode value.
For the old netwokorking we use the radio mode labe again (unfortunately).


marcellorinaldo (approved):

marcellorinaldo:
@salvatore-coppola seems to be working for generic profiles. Tested on Raspbian 64.

PR #5047, opened 2023-12-06
fix(rest.identity): adding specific error messages when user is not found

sfiorani:
This PR adds a more exhaustive response when an user tries to execute the following REST APIs:

- POST services/identity/v1/identities
- DELETE services/identity/v1/identities

with a non-existing `userName` in the request json.

Before this variation, an unspecific message would have been received, whic could be misleading.

**Related Issue:**

**Screenshots:**

**Manual Tests:**

### DIFF
--- a/kura/org.eclipse.kura.rest.identity.provider/src/main/java/org/eclipse/kura/internal/rest/identity/provider/IdentityService.java
+++ b/kura/org.eclipse.kura.rest.identity.provider/src/main/java/org/eclipse/kura/internal/rest/identity/provider/IdentityService.java
@@ -23,12 +23,16 @@ import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
 
+import javax.ws.rs.WebApplicationException;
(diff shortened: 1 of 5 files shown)


kura/org.eclipse.kura.rest.identity.provider/src/main/java/org/eclipse/kura/internal/rest/identity/provider/IdentityService.java
```diff
@@ -77,18 +81,24 @@ public void createUser(UserDTO user) throws KuraException {
         }
     }
 
-    public void deleteUser(String userName) {
-        this.userAdminHelper.deleteUser(userName);
+    public void deleteUser(String userName) throws WebApplicationException {
```
salvatore-coppola: I'd prefer the identityservice doesn't throw `WebApplicationException` . Throw the KuraException and then catch and convert it in WebApplicationException, building it in the `IdentityRestService`

salvatore-coppola (approved):

PR #5055, opened 2023-12-12
fix: 802.1x tab displaying and refresh

MMaiero:

**Related Issue:** N/A

**Description of the solution adopted:** Another attempt to improve the availability and refresh for the 802.1x tab in the wireless section

**Screenshots:** N/A

**Manual Tests**: Tested with ipv4 and ipv6 disabled as well enabling each of them. All the tests done first without and then with the selection of the enterprise security.

**Any side note on the changes made:** N/A

### DIFF
--- a/kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/client/ui/network/Tab8021xUi.java
+++ b/kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/client/ui/network/Tab8021xUi.java
@@ -49,8 +49,6 @@ public class Tab8021xUi extends Composite implements NetworkTab {
     interface Tab8021xUiUiBinder extends UiBinder<Widget, Tab8021xUi> {
     }
 
-    private static final Logger logger = Logger.getLogger(Tab8021xUi.class.getSimpleName());
-
     private final NetworkTabsUi netTabs;
 
     Gwt8021xConfig activeConfig;
@@ -353,11 +351,9 @@ public class Tab8021xUi extends Composite implements NetworkTab {
     }
 
     private void update() {
-        logger.info("update");
         setValues();
         refreshForm();
         this.netTabs.updateTabs();
-        logger.info("update... done");
     }
 
     private void resetValidations() {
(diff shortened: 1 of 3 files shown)


mattdibi:
## Disabled parameters initialization

After first opening the tab the field are disabled as expected

Before this PR:

image

After this PR:

image

**Note**: both screenshot were taken upon first opening the 802.1x tab in Kura

## 802.1x tab enabling

802.1x tab is correctly enabled when either one (or both) between ipv4 and ipv6 is enabled **and** the wireless security is set to WPA Enterprise.

kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/client/ui/network/Tab8021xUi.java
```diff
@@ -348,6 +352,14 @@ private void initPrivateKeyNameTextBox() {
         });
     }
 
+    private void update() {
+        logger.info("update");
+        setValues();
+        refreshForm();
+        this.netTabs.updateTabs();
+        logger.info("update... done");
```
mattdibi: Only thing here is... is this log entry worth the `info` level or would it be better suited for the `debug` level?

```suggestion
        logger.debug("update");
        setValues();
        refreshForm();
        this.netTabs.updateTabs();
        logger.debug("update... done");
```

kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/client/ui/network/Tab8021xUi.java
```diff
@@ -370,6 +382,7 @@ private void refreshForm() {
     }
 
     private void reset() {
+        logger.info("reset");
```
mattdibi: Same here

```suggestion
        logger.debug("reset");
```

kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/client/ui/network/Tab8021xUi.java
```diff
@@ -390,6 +403,8 @@ private void reset() {
         this.keystorePid.setValue("");
         this.caCertName.setValue("");
         this.publicPrivateKeyPairName.setValue("");
+        update();
+        logger.info("reset... done");
```
mattdibi: ... and here

```suggestion
        logger.debug("reset... done");
```

mattdibi (approved):

MMaiero:
@mattdibi I have removed the logs.

mattdibi (approved):

PR #5067, opened 2023-12-13
chore(driver): explain that PreparedRead of a Driver is optional 

gnunzi:
From the javadoc it is not clear if one MUST or CAN provide a PreparedRead in a kura Driver.
One would think that the PrepareRead be optional, but this is not clearly stated.

The BaseAsset in Kura handles both cases: driver with and without PrepareRead. Therefore it should be clear that PrepareRead is optional.

The proposal is to add the following text in the javadoc:

- "If the driver does not implement such optimizations, null should be returned."
- "return The link PreparedRead instance or null if the driver does not implement any protocol specific optimization."

Check BaseAsset.java:

```
if (preparedRead != null) {
                records = preparedRead.execute();
            } else {
                records = conf.getAllReadRecords();
                if (!records.isEmpty()) {
                    state.getDriver().read(records);
                }
            }
```

It looks that PrepareRead was always designed to be optional. It also makes sense. The javadoc is not explicit, therefore the clarification in this PR.

### DIFF
--- a/kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/driver/Driver.java
+++ b/kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/driver/Driver.java
@@ -230,11 +230,14 @@ public interface Driver {
      * If the validation of the channel configuration fails for some channels, the driver must not throw an exception
      * but it is required to return channel records with proper error flags set as a result of the
      * {@link PreparedRead#execute()} call.
+     * 
+     * If the driver does not implement such optimizations, null should be returned.
      *
      * @see PreparedRead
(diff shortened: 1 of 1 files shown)


nicolatimeus:
Hi @gnunzi, sorry for the late reply.
I'm not sure about this change.. as you said correctly the Asset is capable of handling a driver that returns null as preparedRead result, but the Asset is not the only consumer of the Driver APIs. For example an user application could use Driver APIs directly and expect that the PreparedRead is implemented.

mattdibi:
Thanks for your contribution but this would introduce a breaking change and we cannot introduce that ATM. Closing.

PR #5079, opened 2023-12-22
feat(web2): Added Distinguished Names to the certificate table

salvatore-coppola:
Added Distinguished Names to the 
certificate table to better figure out the purpose of the certificate.

immagine


kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/server/GwtCertificatesServiceImpl.java
```diff
@@ -153,6 +154,7 @@ public List<GwtKeystoreEntry> listEntries() throws GwtKuraException {
                     Date validityStartDate = null;
                     Date validityEndDate = null;
 
+                    List<String> dNs = new ArrayList<>();
```
MMaiero:
```suggestion
                    List<String> distinguishedNames = new ArrayList<>();
```

Maybe it's better for readability to have the full name. It will be more clear what it will do or contain

MMaiero (approved):

PR #5099, opened 2024-01-12
fix(nm): 802.1x TLS Private Key encryption

mattdibi:
Currently, when connecting to a WPA Enterprise Access Point using EAP-TLS, we're providing the client private key **in clear text** which is not recommended as per the [NetworkManager documentation.](https://networkmanager.dev/docs/api/latest/settings-802-1x.html).

> WARNING: "private-key" is not a "secret" property, and thus unencrypted private key data using the BLOB scheme may be readable by unprivileged users. Private keys should always be encrypted with a private key password to prevent unauthorized access to unencrypted private key data.

This PR addresses the issue by passing NetworkManager encrypted private keys.

**Implementation**: since the Kura Keystore only supports not password-protected keys, we need to encrypt the key before sending it to NetworkManager. We chose to use the base64-encoded sha-256 hash of the private key as password so that, even when sending the configuration multiple times, the encrypted key and password stays the same.

## Note about key duplication

NetworkManager saves the keys used in the 802.1x TLS authentication on disk inside the `/etc/NetworkManager/system-connection` folder. It appears that it doesn't perform a cleanup after the connection associated with a particular set of keys gets deleted (i.e. the keys stays there).

Kura re-applies/updated network configuration every time it starts/there's a network configuration updates. Since the keys encryption has a random component (i.e. for the same password the encrypted key will differ every time it gets encrypted) I was worried about key duplication on disk.

This appears to be not a problem: NetworkManager is smart enough to **update** the key associated with a particular connection

```
root@device::/etc/NetworkManager/system-connections# sha1sum ce7696b9-440b-4de1-abf5-a4bcbe8039dd-private-key.pem
636513148df942c2317169c6c5c817edbc872d53  ce7696b9-440b-4de1-abf5-a4bcbe8039dd-private-key.pem
root@device::/etc/NetworkManager/system-connections# sha1sum ce7696b9-440b-4de1-abf5-a4bcbe8039dd-private-key.pem
738bbf400eed98bcecb76b421f34d86be0acaa2b  ce7696b9-440b-4de1-abf5-a4bcbe8039dd-private-key.pem
```


mattdibi:
## Tests

Performed connection tests on both platforms (`nss`-based and `gnutls`-based ones). Kura is able to connect to the WPA Enterprise endpoint without issues.

**Additional tests to be performed**: verify keys cleanup/duplication. What happens when Kura reconfigure the network? A new key gets created and saved on disk by NetworkManager?

**Test performed**: a series of test were performed monitoring the `/etc/NetworkManager/system-connection/` folder, looking for duplication of the 802.1x TLS keys. Results are the following:
- [X] Kura restart with `systemctl restart kura`: keys were not duplicated.
- [X] Device reboot: keys were not duplicated.
- [X] 802.1x tab configuration change (added CA certificate): keys were not duplicated. Only the CA cert was added.
- [X] Network tab configuration change (switched eth0 WAN -> LAN): keys were not duplicated.

It appears that NetworkManager just updates the key:

```
root@device::/etc/NetworkManager/system-connections# sha1sum ce7696b9-440b-4de1-abf5-a4bcbe8039dd-private-key.pem
636513148df942c2317169c6c5c817edbc872d53  ce7696b9-440b-4de1-abf5-a4bcbe8039dd-private-key.pem
root@device::/etc/NetworkManager/system-connections# sha1sum ce7696b9-440b-4de1-abf5-a4bcbe8039dd-private-key.pem
738bbf400eed98bcecb76b421f34d86be0acaa2b  ce7696b9-440b-4de1-abf5-a4bcbe8039dd-private-key.pem
```

pierantoniomerlino (approved):

PR #5106, opened 2024-01-24
fix: Network Advanced tab positioning

MMaiero:
Moved the advanced tab before the hardware one.
This would simplify the user interaction as it is expected to be less opened than the other tabs in networking.


mattdibi (approved):
Tested working on RPi ARM64

image

LGTM 👍

PR #5114, opened 2024-02-02
refactor(sparkplug): applied State pattern to SparkplugMqttClient, fixed onConnectionLost status reset

marcellorinaldo:
This PR applies the State design pattern to the `SparkplugMqttClient` class. In addition, it fixes the behavior in which `onConnectionLost` callback was not setting the state to terminated, leaving the client unable to connect again.

**Related Issue:** N/A.

**Screenshots:** N/A.

**Manual Tests**: N/A.

### DIFF
--- a/kura/org.eclipse.kura.cloudconnection.sparkplug.mqtt.provider/src/main/java/org/eclipse/kura/cloudconnection/sparkplug/mqtt/transport/SparkplugMqttClient.java
+++ b/kura/org.eclipse.kura.cloudconnection.sparkplug.mqtt.provider/src/main/java/org/eclipse/kura/cloudconnection/sparkplug/mqtt/transport/SparkplugMqttClient.java
@@ -81,7 +81,7 @@ public class SparkplugMqttClient {
 
         public abstract SessionStatus confirmSession();
 
-        SessionStatus toEstabilishing(boolean shouldConnectClient) throws KuraConnectException {
+        SessionStatus toEstablishing(boolean shouldConnectClient) throws KuraConnectException {
             try {
                 if (shouldConnectClient) {
                     newClientConnection();
@@ -93,14 +93,14 @@ public class SparkplugMqttClient {
                 if (SparkplugMqttClient.this.primaryHostId.isPresent()) {
                     subscribe(SparkplugTopics.getStateTopic(SparkplugMqttClient.this.primaryHostId.get()), 1);
                 } else {
-                    return toEstabilished();
+                    return toEstablished();
                 }
             } catch (MqttException | GeneralSecurityException | IOException e) {
                 SparkplugMqttClient.this.bdSeqCounter = new BdSeqCounter();
                 throw new KuraConnectException(e);
             }
 
-            return new Estabilishing();
+            return new Establishing();
         }
 
         SessionStatus toTerminated(boolean shouldDisconnectClient, long quiesceTimeout) {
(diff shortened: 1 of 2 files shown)


marcellorinaldo:
Relevant Sparkplug TCK tests passing:

- [SessionEstablishment.log](https://github.com/eclipse/kura/files/14139271/SessionEstablishment.log)
- [SessionTermination.log](https://github.com/eclipse/kura/files/14139272/SessionTermination.log)
- [ReceiveCommand.log](https://github.com/eclipse/kura/files/14139273/ReceiveCommand.log)
- [PrimaryHost.log](https://github.com/eclipse/kura/files/14139274/PrimaryHost.log)
- [MultipleMQTTServer.log](https://github.com/eclipse/kura/files/14139275/MultipleMQTTServer.log)

kura/org.eclipse.kura.cloudconnection.sparkplug.mqtt.provider/src/main/java/org/eclipse/kura/cloudconnection/sparkplug/mqtt/transport/SparkplugMqttClient.java
```diff
@@ -67,99 +67,259 @@ public class SparkplugMqttClient {
...
-        this.options = options.getMqttConnectOptions();
-        this.callback = callback;
-        this.listeners = listeners;
+    private abstract class SessionStatus {
 
-        this.groupId = options.getGroupId();
-        this.nodeId = options.getNodeId();
-        this.primaryHostId = options.getPrimaryHostApplicationId();
-        this.connectionTimeoutMs = options.getConnectionTimeoutMs();
+        public abstract SessionStatus estabilishSession(boolean shouldConnectClient) throws KuraConnectException;
```
pierantoniomerlino: Typo? estab**i**lish -> establish

pierantoniomerlino:
@marcellorinaldo

1. Does it make sense to move the States to a separate class?
2. Do we need to add more tests? 
3. I saw in the Sparkplug TCK tests that the [SessionEstablishment.log](https://github.com/eclipse/kura/files/14139271/SessionEstablishment.log) show one test failed. Is it something that we need to care of? Moreover, some other tests are not executed. Is it intended?

marcellorinaldo:
> 3. I saw in the Sparkplug TCK tests that the [SessionEstablishment.log](https://github.com/eclipse/kura/files/14139271/SessionEstablishment.log) show one test failed. Is it something that we need to care of? Moreover, some other tests are not executed. Is it intended?

The device does not support writing to outputs, so the test can fail since the specification states that "If the Device supports writing to outputs..."

marcellorinaldo:
> 2. Do we need to add more tests?

Yes, I will add a test for simulating the `onConnectionLost`

marcellorinaldo:
> 1. Does it make sense to move the States to a separate class?

Don't know... It is a state representation for the client which uses the fields of the client to perform some operations

kura/org.eclipse.kura.cloudconnection.sparkplug.mqtt.provider/src/main/java/org/eclipse/kura/cloudconnection/sparkplug/mqtt/transport/SparkplugMqttClient.java
```diff
@@ -309,7 +309,7 @@ public synchronized void handleConnectionLost() {
 
     public synchronized void estabilishSession(boolean shouldConnectClient) throws KuraConnectException {
```
pierantoniomerlino: Also here... search for enstabililish in the class...

pierantoniomerlino:
> > 1. Does it make sense to move the States to a separate class?
> 
> Don't know... It is a state representation for the client which uses the fields of the client to perform some operations

If it is used only here, probably it is better to keep it as is. Otherwise, move it outside.

pierantoniomerlino (approved):

PR #5126, opened 2024-02-15
feat(examples): add DummyContainerSignatureValidationService

mattdibi:
This PR introduces an example implementation for the APIs introduced in #5125.

The `DummyContainerSignatureValidationService` implements a `ContainerSignatureValidationService` that always returns the same outcome which is set in its configuration. This service can be used as a starting point for a real implementation and for testing container signature validation-related features.

image

PR #5125, opened 2024-02-13
feat(container.signature): add ContainerSignatureValidationService APIs

mattdibi:
The purpose of this PR is to define the APIs for future implementations of a `ContainerSignatureValidationService`, whose goal will be (you guessed it) to verify container image signatures.

Our plan is to provide an API generic enough that all major tools will be compatible:
- [Sigstore's Cosign](https://docs.sigstore.dev/signing/quickstart/)
- [Docker Content Trust (a.k.a. Notary V1)](https://docs.docker.com/engine/security/trust/)
- [Notation (a.k.a. Notary V2)](https://notaryproject.dev/)

The `ContainerSignatureValidationService` implementations will be used by the `ContainerOrchestrationService` to validate the authenticity of the requested container images to further improve the overall security of the platform Kura is running on.

PR #5129, opened 2024-02-15
feat(container.provider): add container image signature validation

mattdibi:
This PR leverages the recently introduced `ContainerSingatureValidationService`(see #5125 and #5136) API to perform a check on the authenticity of the requested container image, using the provided trust anchor.

Currently the `ContainerInstance` only performs a check and logs the results. Future PRs will build upon this one to prevent unvetted container to run on the device (if Kura is configured to do so).

**Details**
This PR introduces two new fields in the `ContainerInstance` UI:
- **Trust Anchor**: Trust anchor (it can be a public key or a X509 certificate) used to verify the container image signature.
- **Verify in transparency log** flag: Sets the transparency log verification. This flag should be enabled when a container image signature has been uploaded to the [transparency log](https://docs.sigstore.dev/logging/overview/).

Under the hood, the `ContainerInstance` registers a list of `ContainerSignatureValidationService` providers that, upon receiving a configuration update, it interrogates to check whether the requested container image is authentic using the provided Trust Anchor.
- If *even only one* service reports the signature as valid, the image signature is considered valid.
- If none of the available services reports the signature as valid, the image signature is considered *not* valid.

This mean that we don't need to prompt the user for selecting the correct `ContainerSignatureValidationService` when performing the check.

If no Trust Anchor is provided the check fails.
If a Trust Anchor is given but no `ContainerSignatureValidationService` is available, the check fails.

image

**Note**: for testing one can use the `DummyContainerSignatureValidationService` introduced in #5126 which implements the `ContainerSingatureValidationService` API and serves as a reference implementation.

**Note**: we already know the signature verification takes a lot of time to be performed (almost 10s on my workstation). In its current implementation, the UI will be blocked while the signature verification is performed. We reserve moving the signature verification in the container state machine in the future.


kura/org.eclipse.kura.container.provider/src/main/java/org/eclipse/kura/container/provider/ContainerInstance.java
```diff
@@ -120,6 +153,50 @@ public void onDisabled() {
...
+
+        if (Objects.isNull(this.availableContainerSignatureValidationService)
+                || this.availableContainerSignatureValidationService.isEmpty()) {
+            logger.warn("No container signature validation service available. Signature validation failed.");
+            return FAILED_VALIDATION;
+        }
+
+        Optional<String> optTrustAnchor = configuration.getSignatureTrustAnchor();
+        if (!optTrustAnchor.isPresent() || optTrustAnchor.get().isEmpty()) {
+            logger.warn("No trust anchor available. Signature validation failed.");
```
marcellorinaldo, line 166: Unreachable code, see `private Optional<String> parseOptionalString` in options class.

kura/org.eclipse.kura.container.provider/src/main/java/org/eclipse/kura/container/provider/ContainerInstance.java
```diff
@@ -120,6 +153,50 @@ public void onDisabled() {
...
+                if (registryCredentials.isPresent()) {
+                    results = validationService.verify(configuration.getContainerImage(),
+                            configuration.getContainerImageTag(), trustAnchor, verifyInTransparencyLog,
+                            registryCredentials.get());
+                } else {
+                    results = validationService.verify(configuration.getContainerImage(),
+                            configuration.getContainerImageTag(), trustAnchor, verifyInTransparencyLog);
+                }
+            } catch (KuraException e) {
+                logger.warn(
```
marcellorinaldo, line 187: This should be error log level in my opinion

mattdibi: I'm usually referencing this

image

Since the process can continue even though the process state is unwanted this is a "warning" level.

kura/org.eclipse.kura.container.provider/src/main/java/org/eclipse/kura/container/provider/ContainerInstance.java
mattdibi: Actually no, this code is reachable: if someone calls the `validateContainerImageSignature` without checking the trust anchor presence beforehand this branch can be traversed.

marcellorinaldo: But the only place where you call it is in the updated method, and it is private

mattdibi: Yes but what if, years down the line, someone else needs to use this method and doesn't know about the requirement? It's a simple sanity check to prevent this method to be misused.

In any case, I can't remove it because otherwise Sonar will scream bloody murder at me:
image

marcellorinaldo: I think that we should change the method signature to have the already unwrapped parameters, leaving the unwrapping of `ContainerInstanceOptions` in the update method. Maybe overloading the method for reusable code.

```java
validateContainerImageSignature(String trustAnchor, boolean verifyInTransparencyLog, String imageName, String imageTag) {
...
```

What do you think?

mattdibi: I kinda disagree.

The purpose of the `validateContainerImageSignature` method is to:
- unwrap the `ContainerInstanceOptions` for the verification process (which involves checking for valid data)
- pass these data to the correct method exposed by the `ContainerSignatureValidationService`

the `updated` method should not know what the signature validation needs. It just calls the high level methods handling the container stuff (like it does with `onConfigurationUpdated`).

Also I don't think removing the `if` statement is needed: it serves as documentation as it makes clear what that method needs to work (in other circumstances I would have used an `assert` statement)

marcellorinaldo: Trying to convince you... I think that the update should call the correct methods based on the configuration provided, hence it should be the place where all the ifs and unwrapping of configurations happen.

`validateContainerImageSignature` should be renamed into `unwrapPropertiesAndValidateContainerImageSignature` if you proceed in that way. Now the method does not serve a single purpose. Moreover, a method called `validateX` should fail when inputs are wrong, not return a failed validation status.

mattdibi: I understand what you're saying but I think you're looking at it from the wrong perspective.

> validateContainerImageSignature should be renamed into unwrapPropertiesAndValidateContainerImageSignature if you proceed in that way.

Going with this principle the `updated` method name would be *way* longer :D ... but looking at it *top-down* it does just that: given the new configuration, it updates the component.

> The purpose of the validateContainerImageSignature method is to:
>
> - unwrap the ContainerInstanceOptions for the verification process (which involves checking for valid data)
> - pass these data to the correct method exposed by the ContainerSignatureValidationService

Here I should have explained myself better: my purpose with the `validateContainerImageSignature` method is to have something **self-contained**. By proceeding you're way we would be splitting the method unnecessarily.

Again, looking at it *top-down*, the `validateContainerImageSignature` method does only one thing: given the container configuration it validates the signature. It hides the complexity inside it because, at the `updated` level of abstraction we don't care what happens inside.

To sum everything up:
- The highlighted code *is reachable*
- The check performed *is necessary* both from the CI perspective (Sonar whines about it) and from the readability perspective (it's clear what the method expects i.e. the preconditions)
- The purpose of the method is to have something *self-contained* and therefore extracting the unwrapping of the option in the `updated` method is working against that principle. Furthermore having the unwrapping at the `updated` level is mixing the abstraction layers.

marcellorinaldo: Ok

marcellorinaldo (approved):

PR #5130, opened 2024-02-15
fix(linux.usb): memory leak on non-usb udev events

marcellorinaldo:
This PR calls `udev_device_unref` at the end of the loop, not only inside the `if` positive branch. It also adds `udev_unref` as per [udev example](https://github.com/robertalks/udev-examples/tree/master). Without these fixes a memory leak occurs on udev events that are not caught by the `dev_parent`.

**Related Issue:** N/A.

**Screenshots:** N/A.

**Manual Tests**: N/A.


marcellorinaldo:
Results on a long running test. Test consists of having a systemd unit using the `mmc` tool every 1ms producing a series of udev events.

image

MMaiero:
Is this an issue affecting also Kura 4?

marcellorinaldo:
> Is this an issue affecting also Kura 4?

Yes

MMaiero:
Can we make a back port for that version? (As it is still supported)

pierantoniomerlino (approved):

PR #5148, opened 2024-02-21
refactor(container.provider.test): refactor ContainerInstance unit tests

mattdibi:
Refactoring of `ContainerInstance` unit tests required by #5129

### DIFF
--- a/kura/org.eclipse.kura.container.provider/src/main/java/org/eclipse/kura/container/provider/ContainerInstance.java
+++ b/kura/org.eclipse.kura.container.provider/src/main/java/org/eclipse/kura/container/provider/ContainerInstance.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
-  * Copyright (c) 2022 Eurotech and/or its affiliates and others
+  * Copyright (c) 2022, 2024 Eurotech and/or its affiliates and others
   *
   * This program and the accompanying materials are made
   * available under the terms of the Eclipse Public License 2.0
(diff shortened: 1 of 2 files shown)


kura/org.eclipse.kura.container.provider/src/main/java/org/eclipse/kura/container/provider/ContainerInstance.java
```diff
@@ -95,6 +101,18 @@ public void deactivate() {
         logger.info("deactivate...done");
     }
 
+    public synchronized ContainerInstanceState getState() {
```
marcellorinaldo: What about returning just `this.state.getClass().getSimpleName()`? In this way adding or removing a state does not require keeping the enum `ContainerInstanceState` up to date. Something similar is done in log at line 140.

mattdibi: I don't like passing around the `State` reference. What if a user attempts to change it/call a method that it shouldn't call? This was my thinking... passing around only the name could be a good idea though...

marcellorinaldo: Yes, I was referring to just the name

mattdibi: Addressed in 4bd4975eb71b45ac01da4ff0b1c42350b2b67339

marcellorinaldo (approved):

PR #5150, opened 2024-02-22
fix: Corrected default container max-size for logging

MMaiero:
Updated the default value with the correct property as stated in the docker documentation. https://docs.docker.com/config/containers/logging/local/

It sets a new default of 10 MB instead of the default of 20MB per container log
Unchanged the number of rotations that default to 5.

image

### DIFF
--- a/kura/org.eclipse.kura.container.provider/OSGI-INF/metatype/org.eclipse.kura.container.provider.ContainerInstance.xml
+++ b/kura/org.eclipse.kura.container.provider/OSGI-INF/metatype/org.eclipse.kura.container.provider.ContainerInstance.xml
@@ -161,7 +161,7 @@
          </AD>
             
        <AD id="container.loggerParameters" name="Logger Parameters"
-            description="Used to pass logger parameters to a container's logging driver. Example: max-size=5m, max-file=2." type="String" cardinality="1"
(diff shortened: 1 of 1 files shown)


MMaiero:
@GregoryIvo Can you please have a check if this is ok for you? Thanks

kura/org.eclipse.kura.container.provider/OSGI-INF/metatype/org.eclipse.kura.container.provider.ContainerInstance.xml
```diff
@@ -161,8 +161,8 @@
          </AD>
             
        <AD id="container.loggerParameters" name="Logger Parameters"
-            description="Used to pass logger parameters to a container's logging driver. Example: max-buffer-size=5m, max-file=2." type="String" cardinality="1"
-            required="false" default="max-buffer-size=5m, max-file=2" />
+            description="Used to pass logger parameters to a container's logging driver. Example: max-size=5m, max-file=2." type="String" cardinality="1"
+            required="false" default="max-size=10m" />
```
mattdibi: Shouldn't we keep the default consistent with the example provided in the description?

```suggestion
            description="Used to pass logger parameters to a container's logging driver. Example: max-size=10m, max-file=2." type="String" cardinality="1"
            required="false" default="max-size=10m" />
```

mattdibi (approved):

GregoryIvo:
LGTM 👍

PR #5157, opened 2024-02-29
fix: Fixed unnecessary column type changes in H2DbWireRecordStoreImpl

nicolatimeus:

Fixes issue that causes H2DbWireRecordStore to incorrectly change the column type and drop existing values when a new TypedValue of one of the following types is inserted:

* `FLOAT`
* `DOUBLE`
* `STRING`
* `BYTE_ARRAY`


marcellorinaldo (approved):
Tested as working. Retested also the deprecated wires implementations.

PR #5162, opened 2024-03-01
feat(container.orchestration.provider): implemented enforcement allowlist

sfiorani:
This PR introduces the Enforcement Allowlist into the Container Orchestration bundle: this adds another security level when managing containers.

The user can enable the `Enforcement` through a new component option and write the list of image digests that are allowed to be run on the device.

If a container is run, the container orchestration will retrieve the container's image digests and compare it/them to the ones in the `Enforcement Allowlist`:

- If the image digest is in the allowlist, the container will be allowed to start
- If not, the container is immediately stopped and deleted

**Related Issue:**

**Screenshots:**

**Manual Tests:**

### DIFF
--- a/kura/org.eclipse.kura.container.orchestration.provider/OSGI-INF/metatype/org.eclipse.kura.container.orchestration.provider.ContainerOrchestrationService.xml
+++ b/kura/org.eclipse.kura.container.orchestration.provider/OSGI-INF/metatype/org.eclipse.kura.container.orchestration.provider.ContainerOrchestrationService.xml
@@ -33,7 +33,7 @@
         <AD
             id="enforcement.allowlist"
             name="Container Image Allowlist "
-            description="List of values containing the image digests. Each entry must be inserted in a new line of the text box|TextArea"
+            description="List of container image digests. Each entry must be separated by a new line of the text box|TextArea"
             type="String"
             cardinality="1"
             required="false"
--- a/kura/org.eclipse.kura.container.orchestration.provider/OSGI-INF/metatype/org.eclipse.kura.container.orchestration.provider.ContainerOrchestrationService.xml
+++ b/kura/org.eclipse.kura.container.orchestration.provider/OSGI-INF/metatype/org.eclipse.kura.container.orchestration.provider.ContainerOrchestrationService.xml
@@ -27,7 +27,7 @@
             cardinality="0" required="true" default="unix:///var/run/docker.sock" />
             
         <AD id="allowlist.enabled" name="Allowlist Enforcement Enabled" type="Boolean" cardinality="0" required="true" default="false"
-            description="Enable/Disable the Allowlist enforcement. If enabled, only containers images whose digest can be found in the allowlist will be allowed to be ran/loaded/created.">
+            description="Enable/Disable the Allowlist enforcement. If enabled, only containers images whose digest can be found in the allowlist will be allowed to be run/loaded/created.">
         </AD>
         
         <AD
--- a/kura/org.eclipse.kura.container.orchestration.provider/OSGI-INF/metatype/org.eclipse.kura.container.orchestration.provider.ContainerOrchestrationService.xml
+++ b/kura/org.eclipse.kura.container.orchestration.provider/OSGI-INF/metatype/org.eclipse.kura.container.orchestration.provider.ContainerOrchestrationService.xml
@@ -26,14 +26,14 @@
             description="Host URL: tcp://localhost:2376 or unix:///var/run/docker.sock" type="String"
             cardinality="0" required="true" default="unix:///var/run/docker.sock" />
             
-        <AD id="allowlist.enabled" name="Allowlist Enforcement Enabled" type="Boolean" cardinality="0" required="true" default="false"
+        <AD id="enforcement.enabled" name="Allowlist Enforcement Enabled" type="Boolean" cardinality="0" required="true" default="false"
             description="Enable/Disable the Allowlist enforcement. If enabled, only containers images whose digest can be found in the allowlist will be allowed to be run/loaded/created.">
         </AD>
         
         <AD
-            id="allowlist.content"
-            name="Container Image Allowlist Content"
-            description="List of comma separated 'sha256' values containing the image digests.|TextArea"
+            id="enforcement.allowlist"
+            name="Container Image Allowlist "
+            description="List of values containing the image digests. Each entry must be inserted in a new line of the text box|TextArea"
             type="String"
             cardinality="1"
             required="false"
--- a/kura/org.eclipse.kura.container.orchestration.provider/OSGI-INF/metatype/org.eclipse.kura.container.orchestration.provider.ContainerOrchestrationService.xml
+++ b/kura/org.eclipse.kura.container.orchestration.provider/OSGI-INF/metatype/org.eclipse.kura.container.orchestration.provider.ContainerOrchestrationService.xml
@@ -31,13 +31,13 @@
         </AD>
         
         <AD
-			id="allowlist.content"
-			name="Container Image Allowlist Content"
-			description="List of comma separated 'sha256' values containing the image digests.|TextArea"
-			type="String"
-			cardinality="1"
-			required="false"
-			default="" />
+            id="allowlist.content"
+            name="Container Image Allowlist Content"
+            description="List of comma separated 'sha256' values containing the image digests.|TextArea"
+            type="String"
+            cardinality="1"
+            required="false"
+            default="" />
         
     </OCD>
 
--- a/kura/org.eclipse.kura.container.orchestration.provider/OSGI-INF/metatype/org.eclipse.kura.container.orchestration.provider.ContainerOrchestrationService.xml
+++ b/kura/org.eclipse.kura.container.orchestration.provider/OSGI-INF/metatype/org.eclipse.kura.container.orchestration.provider.ContainerOrchestrationService.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!--
 
-    Copyright (c) 2022 Eurotech and/or its affiliates and others
+    Copyright (c) 2022, 2024 Eurotech and/or its affiliates and others
   
     This program and the accompanying materials are made
     available under the terms of the Eclipse Public License 2.0
@@ -25,6 +25,19 @@
         <AD id="container.engine.host" name="Container Engine Host URL"
             description="Host URL: tcp://localhost:2376 or unix:///var/run/docker.sock" type="String"
             cardinality="0" required="true" default="unix:///var/run/docker.sock" />
+            
+        <AD id="allowlist.enabled" name="Allowlist Enforcement Enabled" type="Boolean" cardinality="0" required="true" default="false"
+            description="Enable/Disable the Allowlist enforcement. If enabled, only containers images whose digest can be found in the allowlist will be allowed to be ran/loaded/created.">
+        </AD>
+        
+        <AD
+			id="allowlist.content"
+			name="Container Image Allowlist Content"
+			description="List of comma separated 'sha256' values containing the image digests.|TextArea"
+			type="String"
+			cardinality="1"
+			required="false"
+			default="" />
         
     </OCD>
 
--- a/kura/org.eclipse.kura.container.orchestration.provider/src/main/java/org/eclipse/kura/container/orchestration/provider/impl/ContainerOrchestrationServiceImpl.java
+++ b/kura/org.eclipse.kura.container.orchestration.provider/src/main/java/org/eclipse/kura/container/orchestration/provider/impl/ContainerOrchestrationServiceImpl.java
@@ -168,7 +168,7 @@ public class ContainerOrchestrationServiceImpl implements ConfigurableComponent,
             this.allowlistEnforcementMonitor.awaitCompletion(5, TimeUnit.SECONDS);
             this.allowlistEnforcementMonitor = null;
         } catch (InterruptedException ex) {
-            logger.error("Waited to long to close enforcement monitor, stopping it...", ex);
+            logger.error("Waited too long to close enforcement monitor, stopping it...", ex);
             Thread.currentThread().interrupt();
         } catch (IOException ex) {
             logger.error("Failed to close enforcement monitor, stopping it...", ex);
--- a/kura/org.eclipse.kura.container.orchestration.provider/src/main/java/org/eclipse/kura/container/orchestration/provider/impl/ContainerOrchestrationServiceImpl.java
+++ b/kura/org.eclipse.kura.container.orchestration.provider/src/main/java/org/eclipse/kura/container/orchestration/provider/impl/ContainerOrchestrationServiceImpl.java
@@ -163,6 +163,11 @@ public class ContainerOrchestrationServiceImpl implements ConfigurableComponent,
     }
 
     private void closeEnforcementMonitor() {
+
+        if (this.allowlistEnforcementMonitor == null) {
+            return;
+        }
+
         try {
             logger.info("Enforcement monitor closing...");
             this.allowlistEnforcementMonitor.close();
--- a/kura/org.eclipse.kura.container.orchestration.provider/src/main/java/org/eclipse/kura/container/orchestration/provider/impl/ContainerOrchestrationServiceImpl.java
+++ b/kura/org.eclipse.kura.container.orchestration.provider/src/main/java/org/eclipse/kura/container/orchestration/provider/impl/ContainerOrchestrationServiceImpl.java
@@ -144,12 +144,10 @@ public class ContainerOrchestrationServiceImpl implements ConfigurableComponent,
                 try {
                     startEnforcementMonitor();
                 } catch (Exception ex) {
-                    logger.error("Error starting enforcement monitor, disconnecting from docker...", ex);
-                    cleanUpDocker();
+                    logger.error("Error starting enforcement monitor. Due to {}", ex.getMessage());
                     closeEnforcementMonitor();
-                    logger.error("Disconnected from docker");
+                    logger.warn("Enforcement won't be active.");
                 }
-
                 enforceAlreadyRunningContainer();
             }
         }
@@ -158,15 +156,19 @@ public class ContainerOrchestrationServiceImpl implements ConfigurableComponent,
     }
 
     private void startEnforcementMonitor() {
+        logger.info("Enforcement monitor starting...");
         this.allowlistEnforcementMonitor = this.dockerClient.eventsCmd().withEventFilter("start")
                 .exec(new AllowlistEnforcementMonitor(currentConfig.getEnforcementAllowlist(), this));
+        logger.info("Enforcement monitor starting...done.");
     }
 
     private void closeEnforcementMonitor() {
         try {
+            logger.info("Enforcement monitor closing...");
             this.allowlistEnforcementMonitor.close();
             this.allowlistEnforcementMonitor.awaitCompletion(5, TimeUnit.SECONDS);
             this.allowlistEnforcementMonitor = null;
+            logger.info("Enforcement monitor closing...done.");
         } catch (InterruptedException ex) {
             logger.error("Waited too long to close enforcement monitor, stopping it...", ex);
             Thread.currentThread().interrupt();
@@ -176,7 +178,9 @@ public class ContainerOrchestrationServiceImpl implements ConfigurableComponent,
     }
 
     private void enforceAlreadyRunningContainer() {
+        logger.info("Enforcement check on already running containers...");
         this.allowlistEnforcementMonitor.enforceAllowlistFor(listContainerDescriptors());
+        logger.info("Enforcement check on already running containers...done");
     }
 
     @Override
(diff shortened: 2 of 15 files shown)


kura/test/org.eclipse.kura.container.orchestration.provider.test/src/test/java/org/eclipse/kura/container/orchestration/provider/ContainerOrchestrationServiceImplTest.java
mattdibi, line 1: Missing update on copyright header

kura/test/org.eclipse.kura.container.orchestration.provider.test/src/test/java/org/eclipse/kura/container/orchestration/provider/ContainerOrchestrationServiceOptionsTest.java
mattdibi, line 1: Missing update on copyright header

kura/test/org.eclipse.kura.container.orchestration.provider.test/src/test/java/org/eclipse/kura/container/orchestration/provider/EnforcementSecurityTest.java
```diff
@@ -0,0 +1,198 @@
+package org.eclipse.kura.container.orchestration.provider;
```
mattdibi, line 14: Missing copyright header

kura/org.eclipse.kura.container.orchestration.provider/OSGI-INF/metatype/org.eclipse.kura.container.orchestration.provider.ContainerOrchestrationService.xml
```diff
@@ -25,6 +25,19 @@
...
+        </AD>
+        
+        <AD
+			id="allowlist.content"
+			name="Container Image Allowlist Content"
+			description="List of comma separated 'sha256' values containing the image digests.|TextArea"
+			type="String"
+			cardinality="1"
+			required="false"
+			default="" />
```
mattdibi: Here you're using tabs for indenting while the rest of the file uses spaces. Fun fact: if you use Vim you can fix this by properly configuring it and issuing the `:retab` command 😎 

```suggestion
            id="allowlist.content"
            name="Container Image Allowlist Content"
            description="List of comma separated 'sha256' values containing the image digests.|TextArea"
            type="String"
            cardinality="1"
            required="false"
            default="" />
```

kura/org.eclipse.kura.container.orchestration.provider/src/main/java/org/eclipse/kura/container/orchestration/provider/impl/enforcement/AllowlistEnforcementMonitor.java
```diff
@@ -0,0 +1,76 @@
...
+    private static final String ENFORCEMENT_SUCCESS = "Enforcement allowlist contains image digests {}...container {} is starting";
+    private static final String ENFORCEMENT_FAILURE = "Enforcement allowlist doesn't contain image digests...container {} will be stopped";
+    private final List<String> enforcementAllowlistContent;
+    private final ContainerOrchestrationServiceImpl orchestrationServiceImpl;
+
+    public AllowlistEnforcementMonitor(String allowlistContent,
+            ContainerOrchestrationServiceImpl containerOrchestrationService) {
+
+        this.enforcementAllowlistContent = Arrays
+                .asList(allowlistContent.replaceAll("\\s", "").replace("\n", "").trim().split(","));
```
mattdibi: I would suggest to split the list on newlines actually.

When creating the list of digests it's quite hard to read them if they are separated by commas, while, separating them by newlines, you have a much easier time reading the different digests.

There's nothing technically wrong in separating on commas, I also think that we use it more often than newlines as separators for multi-parameter field, but in this particular case I think it's much more readable for the user.

```suggestion
        this.enforcementAllowlistContent = Arrays.asList(allowlistContent.split("\\r?\\n|\\r"));
```

kura/test/org.eclipse.kura.container.orchestration.provider.test/src/test/java/org/eclipse/kura/container/orchestration/provider/ContainerOrchestrationServiceImplTest.java
```diff
@@ -265,30 +274,44 @@ public void testStopContainer() throws KuraException, InterruptedException {
...
+    public void testImageDigestsByContainerName() throws KuraException, InterruptedException {
+        givenFullProperties(true);
+        givenDockerServiceImplSpy();
+        givenDockerClient();
+
+        whenMockForImageDigestsListing();
+
+        whenGetImageDigestsByContainerName(IMAGE_NAME_NGINX);
+
+        thenDigestsListIsNotEmpty();
```
mattdibi: Uh? Why didn't you check for the expected digest to be returned?

kura/test/org.eclipse.kura.container.orchestration.provider.test/src/test/java/org/eclipse/kura/container/orchestration/provider/EnforcementSecurityTest.java
```diff
@@ -0,0 +1,198 @@
...
+    @Test
+    public void shouldNotAllowStartingWithEmptyAllowlistContent() throws KuraException, InterruptedException {
+
+        givenMockedContainerOrchestrationService();
+        givenMockedDockerClient(new String[] { CORRECT_DIGEST });
+        givenAllowlistEnforcement(EMPTY_ALLOWLIST_CONTENT);
+
+        whenOnNext();
+
+        thenContainerDigestIsNotVerifiedAndStopped();
```
mattdibi: Actually the digest has been verified, it's just not valid.

```suggestion
        thenContainerDigestIsNotValidAndStopped();
```

kura/test/org.eclipse.kura.container.orchestration.provider.test/src/test/java/org/eclipse/kura/container/orchestration/provider/EnforcementSecurityTest.java
```diff
@@ -0,0 +1,198 @@
...
+        whenOnNext();
+
+        thenContainerDigestIsVerified();
+    }
+
+    @Test
+    public void shouldAllowStartingWithCorrectAllowlistContentWithSpaces() throws KuraException, InterruptedException {
+
+        givenMockedContainerOrchestrationService();
+        givenMockedDockerClient(new String[] { CORRECT_DIGEST });
```
mattdibi: I think you're backing yourself into a corner here.

For the future I suggest you to have more flexible methods that renders the boundary conditions more explicit.

Example:

```suggestion
        givenMockedContainerOrchestrationService();
        givenMockedContainerOrchestrationServiceWith(CONTAINER_NAME, IMAGE_NAME, IMAGE_TAG, ContainerState.ACTIVE);
        givenMockedDockerClient(new String[] { CORRECT_DIGEST });
```

or something like that. Refer to the recent [ContainerInstance test refactoring](https://github.com/eclipse/kura/pull/5148) for ideas.

kura/org.eclipse.kura.container.orchestration.provider/src/main/java/org/eclipse/kura/container/orchestration/provider/impl/ContainerOrchestrationServiceImpl.java
```diff
@@ -95,6 +98,11 @@ public void setCryptoService(CryptoService cryptoService) {
         this.cryptoService = cryptoService;
     }
 
+    private void startEnforcementMonitor() {
+        this.enforcementEvent = this.dockerClient.eventsCmd().withEventFilter("start")
+                .exec(new AllowlistEnforcementMonitor(currentConfig.getEnforcementAllowlistContent(), this));
+    }
```
mattdibi: Just to be clear: this means that, if a container is started **before** Kura is up or is not running, we won't be able to stop it.

@MMaiero is this acceptable? (We already talked about it but I just wanted to be 100% sure since this is kinda important).

mattdibi (changes requested):

kura/org.eclipse.kura.container.orchestration.provider/src/main/java/org/eclipse/kura/container/orchestration/provider/impl/ContainerOrchestrationServiceImpl.java
mattdibi: @sfiorani I just spoke we @MMaiero and this is not ok... so we need to find a way to block and delete containers that are not allowed.

kura/org.eclipse.kura.container.orchestration.provider/src/main/java/org/eclipse/kura/container/orchestration/provider/impl/enforcement/AllowlistEnforcementMonitor.java
```diff
@@ -0,0 +1,106 @@
...
+
+    private void deleteContainer(String containerId) {
+        try {
+            this.orchestrationServiceImpl.deleteContainer(containerId);
+        } catch (KuraException ex) {
+            logger.error("Error during container deleting process of {}:", containerId, ex);
+        }
+    }
+
+    public void verifyAlreadyRunningContainersDigests(List<ContainerInstanceDescriptor> containerDescriptors) {
```
mattdibi: I think that this name is misleading. It doesn't perform only the verification, it also *acts* on the Docker engine and stops the containers that are not allowed to run.

Would `enforceAlreadyRunningContainersDigests` be a better name?

mattdibi: If not, consider that this is just `implementAllowlistEnforcement` acting on multiple containers. You can have it as a simple overload of the original `implementAllowlistEnforcement`

mattdibi: I would still call both `enforceAllowlistFor(containerInstanceDescriptor)`

kura/org.eclipse.kura.container.orchestration.provider/src/main/java/org/eclipse/kura/container/orchestration/provider/impl/ContainerOrchestrationServiceImpl.java
```diff
@@ -125,19 +128,57 @@ public void updated(Map<String, Object> properties) {
...
+                .exec(new AllowlistEnforcementMonitor(currentConfig.getEnforcementAllowlist(), this));
+    }
+
+    private void closeEnforcementMonitor() {
+        try {
+            this.allowlistEnforcementMonitor.close();
+            this.allowlistEnforcementMonitor.awaitCompletion(5, TimeUnit.SECONDS);
+            this.allowlistEnforcementMonitor = null;
+        } catch (InterruptedException ex) {
+            logger.error("Waited to long to close enforcement monitor, stopping it...", ex);
```
mattdibi:
```suggestion
            logger.error("Waited too long to close enforcement monitor, stopping it...", ex);
```

kura/test/org.eclipse.kura.container.orchestration.provider.test/src/test/java/org/eclipse/kura/container/orchestration/provider/EnforcementSecurityTest.java
```diff
@@ -0,0 +1,254 @@
+/*******************************************************************************
+ * Copyright (c) 2022, 2024 Eurotech and/or its affiliates and others
```
mattdibi: Since this is a new file it should only be 2024

```suggestion
 * Copyright (c) 2024 Eurotech and/or its affiliates and others
```

kura/org.eclipse.kura.container.orchestration.provider/src/main/java/org/eclipse/kura/container/orchestration/provider/impl/enforcement/AllowlistEnforcementMonitor.java
```diff
@@ -0,0 +1,106 @@
+/*******************************************************************************
+ * Copyright (c) 2022, 2024 Eurotech and/or its affiliates and others
```
mattdibi: New file. Only 2024

```suggestion
 * Copyright (c) 2024 Eurotech and/or its affiliates and others
```

kura/test/org.eclipse.kura.container.orchestration.provider.test/src/test/java/org/eclipse/kura/container/orchestration/provider/ContainerOrchestrationServiceImplTest.java
```diff
@@ -265,30 +277,44 @@ public void testStopContainer() throws KuraException, InterruptedException {
...
+    public void testImageDigestsByContainerName() throws KuraException, InterruptedException {
+        givenFullProperties(true);
+        givenDockerServiceImplSpy();
+        givenDockerClient();
+
+        whenMockForImageDigestsListing();
+
+        whenGetImageDigestsByContainerName(IMAGE_NAME_NGINX);
+
+        thenDigestsListEqualsExpectedOne(EXPECTED_DIGESTS_ARRAY);
```
mattdibi, line 316: Readability note: this is somewhat awkward to read since we already have the expectation as a parameter of the method. You could just have:

```suggestion
        thenDigestsListEquals(EXPECTED_DIGESTS_ARRAY);
```

which reads way better.

mattdibi, line 318: Readability note: this is somewhat awkward to read since we already have the expectation as a parameter of the method. You could just have:

```suggestion
        thenDigestsListEquals(EXPECTED_DIGESTS_ARRAY);
```

which reads way better.

kura/test/org.eclipse.kura.container.orchestration.provider.test/src/test/java/org/eclipse/kura/container/orchestration/provider/EnforcementSecurityTest.java
```diff
@@ -0,0 +1,254 @@
...
+
+    private void whenVerifyAlreadyRunningContainersDigests(List<ContainerInstanceDescriptor> containerDescriptors) {
+        this.allowlistEnforcementMonitor.enforceAllowlistFor(containerDescriptors);
+    }
+
+    /*
+     * Then
+     */
+
+    private void thenContainerDigestIsVerified() {
```
mattdibi: Misleading method name. You're not actually checking whether the container digest was verified, you're checking it is actually running.

kura/test/org.eclipse.kura.container.orchestration.provider.test/src/test/java/org/eclipse/kura/container/orchestration/provider/EnforcementSecurityTest.java
```diff
@@ -0,0 +1,254 @@
...
+
+        assertEquals(ContainerState.ACTIVE, this.containerInstanceDescriptor.getContainerState());
+
+    }
```
mattdibi: ... on another note you could simplify these checks a lot by simply passing the expected state

```suggestion
    private void thenContainerIs(ContainerState expectedContainerState) {
        assertEquals(expectedContainerState, this.containerInstanceDescriptor.getContainerState());
    }
```

kura/test/org.eclipse.kura.container.orchestration.provider.test/src/test/java/org/eclipse/kura/container/orchestration/provider/EnforcementSecurityTest.java
```diff
@@ -0,0 +1,254 @@
...
+
+    private void thenContainerDigestIsNotValidAndStoppedAndDeleted() {
+        assertEquals(ContainerState.STOPPING, this.stoppingResult);
+        assertEquals(ContainerState.FAILED, this.containerInstanceDescriptor.getContainerState());
+    }
+
+    private void thenContainerDigestIsNotValidAndOnlyDeleted() {
+        assertEquals(null, this.stoppingResult);
+        assertEquals(ContainerState.FAILED, this.containerInstanceDescriptor.getContainerState());
+    }
```
mattdibi: From the methods name (blabla**AND**bla**AND**bla**AND**bla) you can see that you're checking way too many things at once. Try split the checks up for readability.

kura/org.eclipse.kura.container.orchestration.provider/src/main/java/org/eclipse/kura/container/orchestration/provider/impl/ContainerOrchestrationServiceImpl.java
```diff
@@ -125,19 +128,57 @@ public void updated(Map<String, Object> properties) {
...
             }
-
             logger.info("Connection Successful");
+
+            if (currentConfig.isEnforcementEnabled()) {
+                try {
+                    startEnforcementMonitor();
+                } catch (Exception ex) {
+                    logger.error("Error starting enforcement monitor, disconnecting from docker...", ex);
+                    cleanUpDocker();
```
mattdibi: Question: why are you disconnecting from the Docker daemon when failing to bring up the enforcement monitor?

I'm not saying you're wrong, I'd just like to understand the reasoning behind this choice.

kura/org.eclipse.kura.container.orchestration.provider/OSGI-INF/metatype/org.eclipse.kura.container.orchestration.provider.ContainerOrchestrationService.xml
```diff
@@ -25,6 +25,19 @@
...
             cardinality="0" required="true" default="unix:///var/run/docker.sock" />
+            
+        <AD id="enforcement.enabled" name="Allowlist Enforcement Enabled" type="Boolean" cardinality="0" required="true" default="false"
+            description="Enable/Disable the Allowlist enforcement. If enabled, only containers images whose digest can be found in the allowlist will be allowed to be run/loaded/created.">
+        </AD>
+        
+        <AD
+            id="enforcement.allowlist"
+            name="Container Image Allowlist "
+            description="List of values containing the image digests. Each entry must be inserted in a new line of the text box|TextArea"
```
mattdibi: I think it's a little weirdly worded. This is just a list of container image digests.

```suggestion
            description="List of container image digests. Each entry must be separated by a new line of the text box|TextArea"
```

kura/org.eclipse.kura.container.orchestration.provider/src/main/java/org/eclipse/kura/container/orchestration/provider/impl/enforcement/AllowlistEnforcementMonitor.java
```diff
@@ -0,0 +1,106 @@
...
+        for (ContainerInstanceDescriptor descriptor : containerDescriptors) {
+            enforceAllowlistFor(descriptor.getContainerId());
+        }
+    }
+
+    private String getContainerNameById(String containerId) {
+        return this.orchestrationServiceImpl.listContainerDescriptors().stream()
+                .filter(container -> container.getContainerId().equals(containerId)).findFirst()
+                .map(container -> container.getContainerName()).orElse(null);
+    }
```
mattdibi: I'm a little surprised to find this method here. Wouldn't the `ContainerOrchestrationServiceImpl` be a better fit? I would also consider using `Optional<String>` as the returned type but depends on what we usually do in the Orchestrator...

kura/org.eclipse.kura.container.orchestration.provider/src/main/java/org/eclipse/kura/container/orchestration/provider/impl/enforcement/AllowlistEnforcementMonitor.java
```diff
@@ -0,0 +1,106 @@
...
+
+    @Override
+    public void onNext(Event item) {
+        enforceAllowlistFor(item.getId());
+    }
+
+    private void enforceAllowlistFor(String containerId) {
+
+        List<String> digestsList = this.orchestrationServiceImpl
+                .getImageDigestsByContainerName(getContainerNameById(containerId));
```
mattdibi: ... in addition to my previous comment... why are you performing a double conversion (containerID -> containerName -> digest) instead of simply creating a `getImageDigestsByContainerId` method?

mattdibi (changes requested):
I think we're 90% there. See comments and let's discuss them if needed.

MMaiero:

mattdibi (approved):
LGTM

mattdibi:
Merging without waiting for the CI to complete because we're experiencing *even more issues* with it (lack of executors, Sonar broken). The check introduced in the last commit doesn't change the code behaviour.

PR #5163, opened 2024-03-04
fix(web2): Added better error message for password never set

salvatore-coppola:
Added better error message for password never set for the wifi configuraiton

### DIFF
--- a/kura/org.eclipse.kura.web2/src/main/resources/org/eclipse/kura/web/client/messages/Messages.properties
+++ b/kura/org.eclipse.kura.web2/src/main/resources/org/eclipse/kura/web/client/messages/Messages.properties
@@ -87,6 +87,7 @@ illegalNullArgumentError=Null argument provided. Please review.
 certificateParseFailure=Error while parsing certificate. Please review.
 unauthorizedError=The current identity is not allowed to perform the requested operation.
 sessionExpiredError=The session has expired.
+passwordNeverSetError=The password for this configuration has never been set. Please set it and apply the configuration again.
 
 enable=Enable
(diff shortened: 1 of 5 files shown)


marcellorinaldo (dismissed):

kura/org.eclipse.kura.web2/src/main/resources/org/eclipse/kura/web/client/messages/Messages.properties
```diff
@@ -87,6 +87,7 @@ illegalNullArgumentError=Null argument provided. Please review.
 certificateParseFailure=Error while parsing certificate. Please review.
 unauthorizedError=The current identity is not allowed to perform the requested operation.
 sessionExpiredError=The session has expired.
+passwordNeverSetError=The password for this configuration has never been set. Please set it and apply the configuration again.
```
marcellorinaldo, line 90: Please add JAP version too

marcellorinaldo (changes requested):

kura/org.eclipse.kura.web2/src/main/resources/org/eclipse/kura/web/client/messages/Messages.properties
salvatore-coppola: > Please add JAP version too

well my Japanese is a bit rusty :P

marcellorinaldo: Just use google translate

marcellorinaldo (approved):

PR #5187, opened 2024-03-22
style: Added simple sitemap.xml

salvatore-coppola:
Added sitemap.xml to improve Kura discoverability on search engine (limited to 500 entry on approximately 800


salvatore-coppola:
wrong repo

PR #5184, opened 2024-03-20
docs(container.orchestration): added docs for ContainerInstance enfor…

sfiorani:
This PR add the documentation for the Container Instance Enforcement Digest option added into **STILL NEED TO BE OPENED, WAITING FOR CI RESTORING**.

**Related Issue:**

**Description of the solution adopted:**

**Screenshots:**

**Manual Tests:**

**Any side note on the changes made:**

PR #5197, opened 2024-04-02
feat(container.provider): added container instances enforcement allowlist

sfiorani:
This PR adds an option to the Container Instances to be compliant with the Enforcement feature added to the Container Orchestration Service in #5162 and it's the second step of building the Container Authenticity feature.

This option allows the user to provide an Enforcement Digest for a specific Container Instance: when a container is enabled with the given option filled, the provided digest is added to the Container Orchestration Service enforcement allowlist. In this way the Container Instance is allowed to run even if in the Orchestration Service allowlist the digest is not included, because is the container instance to provide it.

Related docs is present in #5184 .

**Related Issue:**

**Screenshots:**

**Manual Tests:**


kura/org.eclipse.kura.container.provider/OSGI-INF/metatype/org.eclipse.kura.container.provider.ContainerInstance.xml
```diff
@@ -175,6 +175,14 @@
         <AD id="container.restart.onfailure" name="Restart Container On Failure" type="Boolean" cardinality="1" required="true" default="false"
             description="Automatically restart the container when it has failed.">
         </AD>
+        
+        <AD
+            id="enforcement.digest"
+            name="Container Image Enforcement Digest "
+            description="Digest of the image wanted to be run, added to the Enforcement Allowlist of the Container Orchestration Service, if the enforcement monitoring is enabled. If not given, it will be computed by the Signature Verification of the service."
```
mattdibi: I find this statement somewhat confusing... 

> Digest of the image wanted to be run, added to the Enforcement Allowlist of the Container Orchestration Service, if the enforcement monitoring is enabled

Does it mean that I have to put the digest in the Container Orchestration Service allowlist by myself? (I know it doesn't). Does the user need to know about the enforcement allowlist in this instance? I would propose something like:

```suggestion
            description="Digest of the container image allowed to run on the device if the Container Enforcement Monitor is enabled. If not given, it will be computed by the Container Signature Verification service."
```

mattdibi (changes requested):
A reeeeeaaaaaaally small thing. Let me know what you think...

kura/org.eclipse.kura.container.provider/OSGI-INF/metatype/org.eclipse.kura.container.provider.ContainerInstance.xml
sfiorani: Yeah I agree that the user is not required to know which is the underlying process. I commit the change

mattdibi:
As per our offline discussion, among other smaller things, we decided to move the ContainerInstance digest allowlist from the Container Orchestrator to the Allowlist Enforcement Monitor for a better logic separation of the concerns.

kura/org.eclipse.kura.container.orchestration.provider/src/main/java/org/eclipse/kura/container/orchestration/provider/impl/enforcement/AllowlistEnforcementMonitor.java
```diff
@@ -55,11 +55,17 @@ private void enforceAllowlistFor(String containerId) {
...
+                    .filter(digestsList::contains).collect(Collectors.toSet());
+        }
+
 
         if (!digestIntersection.isEmpty()) {
-            logger.info(ENFORCEMENT_SUCCESS, digestIntersection, containerId);
+            logger.info(ENFORCEMENT_CHECK_SUCCESS, digestIntersection, containerId);
         } else {
-            logger.error(ENFORCEMENT_FAILURE, containerId);
+            logger.error(ENFORCEMENT_CHECK_FAILURE, containerId);
```
mattdibi: This should be a `info` level

```suggestion
            logger.info(ENFORCEMENT_CHECK_FAILURE, containerId);
```

mattdibi (approved):
LGTM

mattdibi:
> As per our offline discussion, among other smaller things, we decided to move the ContainerInstance digest allowlist from the Container Orchestrator to the Allowlist Enforcement Monitor for a better logic separation of the concerns.

We discarded this since the only object that knows the digest associated with the Container Instances is the Container Orchestrator... it is natural to have it have the list of the digests.

mattdibi:

Merging as-is for now.

PR #5200, opened 2024-04-05
fix: Fixed error on null password when AP name changed

MMaiero:
1. Create an AP with name "test-ap" with the following configuration:
    
        - range: 172.16.1.10/24 - 172.16.1.11/24, lease 300sec
        - mode: DHCP only, pass DNS
        - security WPA/WPA2, add some password
        - everything else with default values

2. Apply changes and wait for them to be applied
3. Try to change the AP name results in an error dialog, the chages are not applied
4. Try to change the AP name and also the password (to the same set before), the error does not appear

Screenshot 2024-03-04 at 15 45 08


salvatore-coppola (approved):
@MMaiero it is necessary to update the year in the copyright plate as usual

MMaiero:

Merging as-is for now.

PR #5203, opened 2024-04-08
feat: Signature Verification Digest Update of Container Instances

sfiorani:
This PR combines the different features provided by the Container Instance and the new Signature Verification service: indeed, if a Container Instance doesn't provide the Enforcement Digest option (added in #5197 ) but a Trust Anchor is given, Kura will try to calcuate the image digest through the Signature Verification service.

 If no instances of the latter service are available or the verification fails, the Container Instance will continue to provide an empty Enforcement Digest option.

This mechanism is part of the Container Authenticity feature introduced in Kura. Docs will be provided in #5205 

**Related Issue:**

**Screenshots:**

**Manual Tests:**

### DIFF
--- a/kura/org.eclipse.kura.container.provider/src/main/java/org/eclipse/kura/container/provider/ContainerInstance.java
+++ b/kura/org.eclipse.kura.container.provider/src/main/java/org/eclipse/kura/container/provider/ContainerInstance.java
@@ -16,6 +16,7 @@ package org.eclipse.kura.container.provider;
 import static java.util.Objects.isNull;
 
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Map;
 import java.util.Objects;
@@ -49,8 +50,7 @@ public class ContainerInstance implements ConfigurableComponent, ContainerOrches
 
     private ContainerOrchestrationService containerOrchestrationService;
     private Set<ContainerSignatureValidationService> availableContainerSignatureValidationService = new HashSet<>();
-    private String signatureExtractedDigest;
-
+    private ConfigurationService configurationService;
     private State state = new Disabled(new ContainerInstanceOptions(Collections.emptyMap()));
 
     public void setContainerOrchestrationService(final ContainerOrchestrationService containerOrchestrationService) {
@@ -69,6 +69,14 @@ public class ContainerInstance implements ConfigurableComponent, ContainerOrches
         logger.info("Container signature validation service {} removed.",
                 containerSignatureValidationService.getClass());
         this.availableContainerSignatureValidationService.remove(containerSignatureValidationService);
+    }
+
+    public synchronized void setConfigurationService(final ConfigurationService confService) {
+        this.configurationService = confService;
+    }
+
+    public synchronized void unsetConfigurationService() {
+        this.configurationService = null;
     }
 
     // ----------------------------------------------------------------
@@ -92,7 +100,6 @@ public class ContainerInstance implements ConfigurableComponent, ContainerOrches
 
         try {
             ContainerInstanceOptions newProps = new ContainerInstanceOptions(properties);
-            this.signatureExtractedDigest = null;
 
             if (!newProps.getEnforcementDigest().isPresent()) {
 
@@ -101,9 +108,10 @@ public class ContainerInstance implements ConfigurableComponent, ContainerOrches
 
                 if (newProps.getSignatureTrustAnchor().isPresent()) {
                     ValidationResult containerSignatureValidated = validateContainerImageSignature(newProps);
-                    this.signatureExtractedDigest = containerSignatureValidated.imageDigest().orElse("?");
+                    newProps.setEnforcementDigest(containerSignatureValidated.imageDigest());
+                    updateSnapshotWithSignatureDigest(newProps.getEnforcementDigest(), properties);
                     logger.info("Container signature validation result for {}@{}({}) - {}",
-                            newProps.getContainerImage(), this.signatureExtractedDigest,
+                            newProps.getContainerImage(), newProps.getEnforcementDigest().orElse("?"),
                             newProps.getContainerImageTag(),
                             containerSignatureValidated.isSignatureValid() ? "OK" : "FAIL");
                 } else {
@@ -352,9 +360,7 @@ public class ContainerInstance implements ConfigurableComponent, ContainerOrches
             int maxRetries = options.getMaxDownloadRetries();
             int retryInterval = options.getRetryInterval();
 
-            final ContainerConfiguration containerConfiguration = options.getEnforcementDigest().isPresent()
-                    ? options.getContainerConfiguration()
-                    : options.getContainerConfigurationBySignature(signatureExtractedDigest);
+            final ContainerConfiguration containerConfiguration = options.getContainerConfiguration();
 
             int retries = 0;
             while ((unlimitedRetries || retries < maxRetries) && !Thread.currentThread().isInterrupted()) {
@@ -435,6 +441,21 @@ public class ContainerInstance implements ConfigurableComponent, ContainerOrches
             return new Disabled(this.options);
         }
 
+    }
+
+    private void updateSnapshotWithSignatureDigest(Optional<String> enforcementDigest, Map<String, Object> properties) {
+
+        if (enforcementDigest.isPresent()) {
+            try {
+                Map<String, Object> updatedProperties = new HashMap<>(properties);
+                updatedProperties.put("enforcement.digest", enforcementDigest.get());
+                this.configurationService.updateConfiguration(
+                        (String) properties.get(ConfigurationService.KURA_SERVICE_PID), updatedProperties, true);
+            } catch (KuraException ex) {
+                logger.error("Impossible to update snapshot for pid {} due to {}",
+                        properties.get(ConfigurationService.KURA_SERVICE_PID), ex.getMessage());
+            }
+        }
     }
 
 }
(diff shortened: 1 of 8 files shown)


kura/org.eclipse.kura.container.provider/src/main/java/org/eclipse/kura/container/provider/ContainerInstance.java
```diff
@@ -49,8 +50,9 @@ public class ContainerInstance implements ConfigurableComponent, ContainerOrches
 
     private ContainerOrchestrationService containerOrchestrationService;
     private Set<ContainerSignatureValidationService> availableContainerSignatureValidationService = new HashSet<>();
-
+    private ConfigurationService configurationService;
     private State state = new Disabled(new ContainerInstanceOptions(Collections.emptyMap()));
+    private ContainerInstanceOptions actualOptions = null;
```
mattdibi: Mmmmmmh I think you fell for a classic "false friend" blunder here. You probably wanted to say *current* options, didn't you?

```suggestion
    private ContainerInstanceOptions currentOptions = null;
```

kura/org.eclipse.kura.container.provider/src/main/java/org/eclipse/kura/container/provider/ContainerInstance.java
```diff
@@ -90,25 +100,35 @@ public void updated(Map<String, Object> properties) {
...
-                        containerSignatureValidated.isSignatureValid() ? "OK" : "FAIL");
-            } else {
-                logger.info("No trust anchor available. Signature validation skipped.");
-            }
 
-            if (newProps.isEnabled()) {
-                this.containerOrchestrationService.registerListener(this);
-            } else {
-                this.containerOrchestrationService.unregisterListener(this);
+            if (this.actualOptions == null || !this.actualOptions.equals(new ContainerInstanceOptions(properties))) {
```
mattdibi: As always I would suggest you to use early returns here for improved readability

```suggestion
            if (this.actualOptions != null && this.actualOptions.equals(new ContainerInstanceOptions(properties))) {
                return;
            }
```

mattdibi: Additionally, instead of building the `ContainerInstanceOptions` object so many times (which is a costly operation) I would store it in a variable before checking if it equals the `currentOptions`. Doing so will also improve readability:

```suggestion
            ContainerInstanceOptions newOptions = new ContainerInstanceOptions(properties);
            if (this.currentOptions != null && this.currentOptions.equals(newOptions)) {
```

kura/org.eclipse.kura.container.provider/src/main/java/org/eclipse/kura/container/provider/ContainerInstance.java
```diff
@@ -425,4 +445,42 @@ public State onDisabled() {
 
     }
 
-}
\ No newline at end of file
+    private void getDigestFromSignatureVerification(Map<String, Object> properties) {
```
mattdibi: This could be personal taste but I think this method is hiding quite a lot of things... for starters it's not actually only `get`ting the digest, it is also updating the properties **and** the snapshot.

Personally I would make it return the digest and I would perform the properties/snapshot update in the caller method. This would render this method useless though...

At a minimum we need to update the name. Let's discuss this offline though.

kura/test/org.eclipse.kura.container.orchestration.provider.test/src/test/java/org/eclipse/kura/container/orchestration/provider/ContainerOrchestrationServiceImplTest.java
```diff
@@ -663,6 +664,11 @@ private void whenStopContainer() throws KuraException {
         this.dockerService.stopContainer(this.containerId);
     }
 
+    private void whenDeleteContainer() throws KuraException {
+        // deleteContainer
+        this.dockerService.deleteContainer(this.containerId);
```
mattdibi: Is this comment really needed though? 😄 

```suggestion
        this.dockerService.deleteContainer(this.containerId);
```

kura/test/org.eclipse.kura.container.provider.test/src/test/java/org/eclipse/kura/container/provider/ContainerInstanceTest.java
```diff
@@ -502,7 +540,10 @@ private void givenContainerSignatureValidationServiceReturningSuccessFor(String
         // Generate random sha256 string
         String sha256 = "sha256:" + Long.toHexString(Double.doubleToLongBits(Math.random()));
         when(this.mockContainerSignatureValidationService.verify(eq(imageName), eq(imageTag), any(String.class),
-                any(Boolean.class))).thenReturn(new ValidationResult(true, sha256));
+                any(Boolean.class))).thenAnswer(answer -> {
+                    this.properties.put("enforcement.digest", sha256);
```
mattdibi: I don't like this at all... why do you need it?

kura/test/org.eclipse.kura.container.provider.test/src/test/java/org/eclipse/kura/container/provider/ContainerInstanceTest.java
```diff
@@ -609,6 +656,12 @@ private void thenAuthenticatedVerifySignatureWasCalledFor(String imageName, Stri
                 verifyTlog, passwordRegistryCredentials);
     }
 
+    private void thenSignatureVerificationVerifyWasNeverCalledFor(String imageName, String imageTag, String trustAnchor,
+            boolean verifyTlog, PasswordRegistryCredentials passwordRegistryCredentials) throws KuraException {
+        verify(this.mockContainerSignatureValidationService, never()).verify(imageName, imageTag, trustAnchor,
+                verifyTlog, passwordRegistryCredentials);
+    }
+
```
mattdibi: Why did you need this when there's already: https://github.com/eclipse/kura/blob/3401ffe18656115a03500c2a00f0689a83554f6d/kura/test/org.eclipse.kura.container.provider.test/src/test/java/org/eclipse/kura/container/provider/ContainerInstanceTest.java#L640

kura/test/org.eclipse.kura.container.provider.test/src/test/java/org/eclipse/kura/container/provider/ContainerInstanceTest.java
```diff
@@ -365,6 +372,7 @@ public void signatureValidationWorks() throws KuraException, InterruptedExceptio
         givenPropertiesWith(CONTAINER_IMAGE_TAG, "latest");
         givenPropertiesWith(CONTAINER_TRUST_ANCHOR, "aRealTrustAnchor ;)");
         givenPropertiesWith(CONTAINER_VERIFY_TLOG, true);
+        givenPropertiesWith(CONTAINER_ENFORCEMENT_DIGEST, "");
```
mattdibi: If this is needed here I would suspect it was needed for all the other signature-related tests... otherwise they would fail for the wrong reasons... the same goes for the `givenMockedConfigurationService();`

kura/test/org.eclipse.kura.container.provider.test/src/test/java/org/eclipse/kura/container/provider/ContainerInstanceTest.java
```diff
@@ -57,6 +62,7 @@ public class ContainerInstanceTest {
     private static final String CONTAINER_VERIFY_TLOG = "container.signature.verify.transparency.log";
     private static final String CONTAINER_REGISTRY_USERNAME = "registry.username";
     private static final String CONTAINER_REGISTRY_PASSWORD = "registry.password";
+    private static final String CONTAINER_ENFORCEMENT_DIGEST = "enforcement.digest";
```
mattdibi, line 65: I noticed this only now but... for consistency, since it's releated to the other parameters, would it make sense to call it:

```suggestion
    private static final String CONTAINER_ENFORCEMENT_DIGEST = "container.signature.enforcement.digest";
```

Obviously such rename would require a separate PR. I'm just throwing this here in the meantime...

mattdibi (changes requested):
Looks good. Some minor things here and there.

kura/org.eclipse.kura.container.provider/src/main/java/org/eclipse/kura/container/provider/ContainerInstance.java
sfiorani: I did it **110%**

mattdibi (approved):

mattdibi (changes requested):
I just noticed that the "Container Image Enforcement Digest" box is at the bottom of the Container Instance configuration UI. I think its proper place would be under the "Verify In transparency log" radio button as per our initial sketch... let's discuss this tomorrow

image

kura/org.eclipse.kura.container.provider/src/main/java/org/eclipse/kura/container/provider/ContainerInstance.java
```diff
@@ -90,25 +100,51 @@ public void updated(Map<String, Object> properties) {
...
+                logger.info(
+                        "Container configuration doesn't include enforcement digest. Validating with Container Signature Validation service");
+
+                if (this.currentOptions.getSignatureTrustAnchor().isPresent()) {
+
+                    ValidationResult containerSignatureValidated = validateContainerImageSignature(this.currentOptions);
+
+                    logger.info("Container signature validation result for {}@{}({}) - {}",
+                            this.currentOptions.getContainerImage(),
+                            this.currentOptions.getEnforcementDigest().orElse("?"),
```
mattdibi: If we want to display the resulting digest we should log `containerSignatureValidated`... not the current options.

```suggestion
                            containerSignatureValidated.imageDigest().orElse("?"),
```

mattdibi (approved):

mattdibi:

Merging as-is for now.

kura/test/org.eclipse.kura.container.provider.test/src/test/java/org/eclipse/kura/container/provider/ContainerInstanceTest.java
sfiorani: Done in #5208

PR #5218, opened 2024-04-18
fix(net.configuration): Fix network properties default values

pierantoniomerlino:

This PR modifies the `NetworkConfigurationServiceImpl` to return a complete metatype with also the defaults values.

**Description of the solution adopted:** In the current implementation of the `NetworkConfigurationService` (both old and new one), the metatype returned by this `SelfConfiguringComponent` lacks of every default values. The default values are in the classes that define the configuration, so they are spread in the code. This PR tries to centralize the default values putting them in the new `NetworkConfigurationConstants` class. Moreover, the metatype is modified to return also the default values for the properties, when they are available.

Finally, a builder is implemented to create the `AttributeDefinition`s in the `NetworkConfigurationServiceCommon` class.

**Any side note on the changes made:** Since all the default values are present in the metatype, they are stored in the snapshot that is longer than before.


kura/org.eclipse.kura.core.net/META-INF/MANIFEST.MF
```diff
@@ -12,6 +12,7 @@ Import-Package: org.eclipse.kura;version="[1.0,2.0)",
  org.eclipse.kura.core.util;version="[1.0,2.0)",
  org.eclipse.kura.executor;version="[1.0,2.0)",
  org.eclipse.kura.net;version="[2.0,3.0)",
+ org.eclipse.kura.net.configuration;version="[1.0,2.0)",
```
MMaiero: Are we compliant with the 1.0 API or do we need to make it strictly requiring the current net.configuration version?

kura/org.eclipse.kura.core.net/src/main/java/org/eclipse/kura/core/net/WifiConfigurationInterpreter.java
```diff
@@ -121,68 +122,68 @@ private static WifiConfig getWifiConfig(String netIfConfigPrefix, WifiMode mode,
...
+            return new WifiBgscan((String) bgscanObject);
         }
-        return new WifiBgscan(bgscan);
+        return new WifiBgscan(NetworkConfigurationConstants.DEFAULT_WIFI_BGSCAN_VALUE);
     }
 
-    private static String getPairwiseCiphers(Map<String, Object> properties, String prefix) {
-        return (String) properties.get(prefix + WIFI_PAIRWISE_CIPHERS_KEY);
+    private static WifiCiphers getPairwiseCiphers(Map<String, Object> properties, String prefix) {
+        Object pairwireCiphersObject = properties.get(prefix + WIFI_PAIRWISE_CIPHERS_KEY);
```
MMaiero:
```suggestion
        Object pairwiseCiphersObject = properties.get(prefix + WIFI_PAIRWISE_CIPHERS_KEY);
```

kura/org.eclipse.kura.net.configuration/src/main/java/org/eclipse/kura/net/configuration/NetworkConfigurationConstants.java
```diff
@@ -0,0 +1,81 @@
...
+    public static final Integer DEFAULT_MODEM_HEADER_COMPRESSION_VALUE = 0;
+    public static final PdpType DEFAULT_MODEM_PDP_TYPE_VALUE = PdpType.IP;
+    public static final AuthType DEFAULT_MODEM_AUTH_TYPE_VALUE = AuthType.NONE;
+    public static final Integer DEFAULT_MODEM_PPP_NUMBER_VALUE = 0;
+
+    // Wifi properties default values
+    public static final Boolean DEFAULT_WIFI_BROADCAST_VALUE = false;
+    public static final WifiRadioMode DEFAULT_WIFI_RADIO_MODE_VALUE = WifiRadioMode.RADIO_MODE_80211b;
+    public static final WifiSecurity DEFAULT_WIFI_SECURITY_VALUE = WifiSecurity.NONE;
+    public static final String DEFAULT_WIFI_CHANNEL_VALUE = "1";
```
MMaiero, line 49: Does it make sense to have it as a string?

kura/org.eclipse.kura.net.configuration/src/main/java/org/eclipse/kura/net/configuration/NetworkConfigurationConstants.java
```diff
@@ -0,0 +1,81 @@
...
+import org.eclipse.kura.net.wifi.WifiSecurity;
+
+public final class NetworkConfigurationConstants {
+
+    // Modem properties default values
+    public static final Boolean DEFAULT_MODEM_PERSIST_VALUE = true;
+    public static final Integer DEFAULT_MODEM_HOLDOFF_VALUE = 1;
+    public static final Integer DEFAULT_MODEM_MAXFAIL_VALUE = 5;
+    public static final Integer DEFAULT_MODEM_RESET_TIMEOUT_VALUE = 5;
+    public static final Integer DEFAULT_MODEM_IDLE_VALUE = 95;
```
MMaiero: Why not the primitive types?

kura/org.eclipse.kura.net.configuration/src/main/java/org/eclipse/kura/net/configuration/NetworkConfigurationConstants.java
pierantoniomerlino: Yes, it does, since it's a comma-separated list of channels (i.e. 1,12,43). The default value is only one channel.

kura/org.eclipse.kura.core.net/META-INF/MANIFEST.MF
pierantoniomerlino: Mhhh... the metatype returned by the `NetworkConfigurationServiceCommon` class with the `getDefinition` method has the same properties and type as before. The only things changed are the default values: before the defaut values were always an empty string, while now they are different values. So, the old code is compatible, but the behavior is a bit different (even if the default values added by the networking classes are the same). Probably it's better to increase the version of the `org.eclipse.kura.net.configuration` to 1.1. What do you think?

MMaiero: Agreed, better to stay safe

kura/org.eclipse.kura.net.configuration/src/main/java/org/eclipse/kura/net/configuration/NetworkConfigurationConstants.java
MMaiero: ok

kura/org.eclipse.kura.core.net/src/main/java/org/eclipse/kura/core/net/WifiConfigurationInterpreter.java
```diff
@@ -175,7 +175,7 @@ private static WifiBgscan getBgScan(Map<String, Object> properties, String prefi
     }
 
     private static WifiCiphers getPairwiseCiphers(Map<String, Object> properties, String prefix) {
-        Object pairwireCiphersObject = properties.get(prefix + WIFI_PAIRWISE_CIPHERS_KEY);
+        Object pairwiseCiphersObject = properties.get(prefix + WIFI_PAIRWISE_CIPHERS_KEY);
         if (pairwireCiphersObject instanceof String) {
```
MMaiero: The change is not complete, as there were obviously other references to the variable.

pierantoniomerlino: It is a local variable used only in this method.

MMaiero: Yes

MMaiero (approved):

PR #5223, opened 2024-04-29
feat(distrib): Add generic nn profiles

pierantoniomerlino:

This PR adds three new profiles:

- generic ARM32 NN
- generic ARM64 NN
- generic X86_64 NN


kura/distrib/pom.xml
```diff
@@ -1364,6 +1457,99 @@
...
+                                <configuration>
+                                    <target>
+                                        <property name="buildNumber" value="buildNumber" />
+                                        <property name="project.version" value="${project.version}" />
+                                        <property name="project.build.profile" value="${project.build.profile}" />
+                                        <property name="project.build.directory" value="${project.build.directory}" />
+                                        <property name="build.name" value="generic-x86_64-nn" />
+                                        <property name="target.device" value="generic-x86_64-nn" />
+                                        <property name="native.tag" value="x86_64" />
+                                        <property name="kura.os.version" value="raspbian" />
```
marcellorinaldo, line 1508: I think we have the wrong `kura.os.version` on all of the generic profiles, even networking ones...

marcellorinaldo (approved):

kura/distrib/pom.xml
pierantoniomerlino: We should check where that property is used and decide what to set in case of generic...

PR #5229, opened 2024-05-06
fix: Improved parameter validation in IdentityService

nicolatimeus:


kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/shared/validator/NoWhitespaceCharactersValidator.java
```diff
@@ -0,0 +1,36 @@
...
+ *  Eurotech
+ *******************************************************************************/
+package org.eclipse.kura.web.shared.validator;
+
+import java.util.function.Predicate;
+
+public class NoWhitespaceCharactersValidator extends PredicateValidator {
+
+    private static final Predicate<String> PREDICATE = value -> {
+        if (value == null) {
```
salvatore-coppola, line 20: is null a valid non whitespace string ?

salvatore-coppola (approved):

kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/shared/validator/NoWhitespaceCharactersValidator.java
nicolatimeus: Good question.. I accepted a null string so that it will trigger the nonEmptyValidator instead of this one providing the "value cannot be empty" message instead of "value cannot contain whitespace characters"

salvatore-coppola: Does it makes sense change the validators order and perform the nonEmptyValidator before NonWhitespaceCharactersValidator ?

nicolatimeus: It should be already like that https://github.com/eclipse/kura/blob/3cd7aeb21827a773a827dcce770ae06387ad93b4/kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/client/ui/validator/GwtValidators.java#L86, in any case a null or empty string does not contain any whitespace character

salvatore-coppola:
Successfully tested on RaspberryPI 4B+

PR #5230, opened 2024-05-06
feat(rest.security.provider): Rest Security Update

pierantoniomerlino:

This PR adds two methods in the `rest.security.provider` bundle to upload the security policy. Moreover, two new methods are added to the `SecurityService` interface to support the new rest calls.

**Description of the solution adopted:** The following methods are added:

POST methods
Default Production Security Policy load

    Description: This method loads the default production security policy
    Method: POST
    API PATH: services/security/v1/security-policy/load-default-production

Responses

    200 OK status
    500 Internal Server Error

Security Policy upload

    Description: This method allows to upload a security policy
    Method: POST
    API PATH: services/security/v1/security-policy/load
    Body:
           <base64 encoded xml security policy>"

Responses

    200 OK status
    500 Internal Server Error


kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/security/SecurityService.java
```diff
@@ -46,4 +44,24 @@ public interface SecurityService {
...
+    /**
+     * This method allows the load of the provided security policy
+     *
+     * @param securityPolicy
+     *         the security policy to be loaded
+     * @throws KuraException
+     *         in case an error is raised during the application of the security policy.
+     * @since 2.7
+     */
+    public void loadSecurityPolicy(String securityPolicy) throws KuraException;
```
marcellorinaldo: `load` in my opinion is too ambiguos: does it mean get, apply, or reload?

pierantoniomerlino: This method is used to upload the security policy, copy in the right place and reload the fingerprint. So, something like an `apply` method, but in this way is not clear that you have also to upload the policy.
Conversely, the `load-default-production` performs the same operation with the default policy that is it alredy on the system. Maybe `apply-default-production` is more suitable there?

marcellorinaldo: > something like an `apply` method, but in this way is not clear that you have also to upload the policy

Maybe that is clear from the parameters. IMHO apply is better suited for both methods. We could describe the methods better in the javadoc

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/security/SecurityService.java
```diff
@@ -46,4 +44,24 @@ public interface SecurityService {
      */
     public boolean isDebugEnabled();
 
+    /**
+     * This method allows the application of the default production security policy present in the system.
```
MMaiero:
```suggestion
     * This method allows to apply the default production security policy available in the system.
```

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/security/SecurityService.java
```diff
@@ -46,4 +44,24 @@ public interface SecurityService {
...
+     *
+     * @throws KuraException
+     *         in case an error is raised during the application of the default production security policy.
+     * @since 2.7
+     */
+
+    public void applyDefaultProductionSecurityPolicy() throws KuraException;
+
+    /**
+     * This method allows the application of the provided security policy.
```
MMaiero:
```suggestion
     * This method allows to apply the user provided security policy.
```

kura/org.eclipse.kura.rest.security.provider/src/main/java/org/eclipse/kura/internal/rest/security/provider/SecurityRestService.java
```diff
@@ -98,19 +96,57 @@ public Response reloadSecurityPolicyFingerprint() {
...
+    }
+
+    /**
+     * POST method <br /> This method replaces the security policy with the default production one. Then a fingerprint
+     * reload is performed.
+     */
+    @POST
+    @RolesAllowed(REST_ROLE_NAME)
+    @Path("/security-policy/apply-default-production")
+    public Response applyDefaultProductionSecurityPolicy() {
```
nicolatimeus: Please create a security/v2 REST endpoint and SEC-V2 request handler that provides the new methods and the existing ones leaving SEC-V1 in place unmodified

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/security/SecurityService.java
```diff
@@ -46,4 +44,24 @@ public interface SecurityService {
      */
     public boolean isDebugEnabled();
 
+    /**
+     * This method allows to apply the default production security policy available in the system.
```
nicolatimeus: We could also mention that the changes to the security policy are persistent and should remain in effect after a framework restart

kura/org.eclipse.kura.rest.security.provider/src/main/java/org/eclipse/kura/internal/rest/security/provider/SecurityRestServiceV2.java
```diff
@@ -0,0 +1,102 @@
...
+        } catch (Exception e) {
+            throw DefaultExceptionHandler.toWebApplicationException(e);
+        }
+
+        return Response.ok().build();
+    }
+
+    private String readSecurityPolicyString(InputStream securityPolicyInputStream) throws IOException {
+        if (securityPolicyInputStream == null) {
+            throw new IllegalArgumentException("Security Policy cannot be null or empty");
```
nicolatimeus: Throwing an IllegalArgumentException would result in status code 500, maybe a 400 would be more appropriate here, you should be able to do that using this method https://github.com/eclipse/kura/blob/71def61f3775869ba6fa762deb81789a44c0b73f/kura/org.eclipse.kura.request.handler.jaxrs/src/main/java/org/eclipse/kura/request/handler/jaxrs/DefaultExceptionHandler.java#L77. The same applies for the other exceptions below

nicolatimeus (approved):

PR #5231, opened 2024-05-07
feat(configuration): Added PID validation for conf-v1 and v2 endpoint (mqtt and rest). Made configuration update synchronous in mqtt conf-v1

salvatore-coppola:
Added PID validation for conf-v1 and v2 endpoint (mqtt and rest). Made configuration update synchronous in mqtt conf-v1.

Pids are invalid if the components they refer are not tracked or they are not factory components.


salvatore-coppola:
Following the discussion had regarding the impacts of this change it was decided not to merge this PR and to adopt other possible solutions to the problem of partial updates, solutions that will be implemented in the next major version

PR #5240, opened 2024-05-21
fix(container.orchestration): Removed exposedPorts from the state of ContainerOrchestrationServiceImpl object.

salvatore-coppola:
Removed `exposedPorts `from the state of `ContainerOrchestrationServiceImpl` object.

### DIFF
--- a/kura/org.eclipse.kura.container.orchestration.provider/src/main/java/org/eclipse/kura/container/orchestration/provider/impl/ContainerOrchestrationServiceImpl.java
+++ b/kura/org.eclipse.kura.container.orchestration.provider/src/main/java/org/eclipse/kura/container/orchestration/provider/impl/ContainerOrchestrationServiceImpl.java
@@ -703,10 +703,8 @@ public class ContainerOrchestrationServiceImpl implements ConfigurableComponent,
                     try {
                         ipProtocol = InternetProtocol.parse(port.getInternetProtocol().toString());
                     } catch (IllegalArgumentException e) {
-                        logger.error("Invalid internet protocol: {}. Using TCP.", port.getInternetProtocol());
+                        logger.warn("Invalid internet protocol: {}. Using TCP.", port.getInternetProtocol());
                     }
-                } else {
-                    logger.error("Null internet protocol: {}. Using TCP.", port.getInternetProtocol());
                 }
 
                 ExposedPort exposedPort = new ExposedPort(port.getInternalPort(), ipProtocol);
--- a/kura/org.eclipse.kura.container.orchestration.provider/src/main/java/org/eclipse/kura/container/orchestration/provider/impl/ContainerOrchestrationServiceImpl.java
+++ b/kura/org.eclipse.kura.container.orchestration.provider/src/main/java/org/eclipse/kura/container/orchestration/provider/impl/ContainerOrchestrationServiceImpl.java
@@ -683,42 +683,35 @@ public class ContainerOrchestrationServiceImpl implements ConfigurableComponent,
         return configuration;
     }
 
-    private HostConfig containerPortManagementHandler(ContainerConfiguration containerDescription,
+    private HostConfig containerPortManagementHandler(ContainerConfiguration containerConfiguration,
             HostConfig hostConfig, CreateContainerCmd commandBuilder) {
 
-        if (containerDescription.getContainerPorts() != null && !containerDescription.getContainerPorts().isEmpty()) {
-            List<ExposedPort> exposedPortsList = new LinkedList<>();
-            Ports portbindings = new Ports();
+        List<ExposedPort> exposedPorts = new LinkedList<>();
+        Ports portbindings = new Ports();
 
-            for (org.eclipse.kura.container.orchestration.ContainerPort port : containerDescription
-                    .getContainerPorts()) {
+        if (containerConfiguration.getContainerPorts() != null
+                && !containerConfiguration.getContainerPorts().isEmpty()) {
 
-                InternetProtocol ipPro;
+            List<org.eclipse.kura.container.orchestration.ContainerPort> containerPorts = containerConfiguration
+                    .getContainerPorts();
 
-                switch (port.getInternetProtocol()) {
-                case UDP:
-                    ipPro = InternetProtocol.UDP;
-                    break;
-                case SCTP:
-                    ipPro = InternetProtocol.SCTP;
-                    break;
-                default:
-                    ipPro = InternetProtocol.TCP;
-                    break;
-                }
+            for (org.eclipse.kura.container.orchestration.ContainerPort port : containerPorts) {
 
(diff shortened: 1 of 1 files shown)


kura/org.eclipse.kura.container.orchestration.provider/src/main/java/org/eclipse/kura/container/orchestration/provider/impl/ContainerOrchestrationServiceImpl.java
```diff
@@ -684,45 +683,37 @@ private HostConfig containerRuntimeConfigurationHandler(ContainerConfiguration c
...
-                }
-
-                ExposedPort tempExposedPort = new ExposedPort(port.getInternalPort(), ipPro);
-                exposedPortsList.add(tempExposedPort);
-                portbindings.bind(tempExposedPort, Binding.bindPort(port.getExternalPort()));
-            }
+            for (org.eclipse.kura.container.orchestration.ContainerPort port : containerPorts) {
 
-            commandBuilder.withPortBindings(portbindings);
+                InternetProtocol ipProtocol = InternetProtocol.parse(port.getInternetProtocol().toString());
```
nicolatimeus: This changes the behaviour slightly, in the previous implementation `TCP` would be used as a default if port.getInternetProtocol() returns null, now the code will throw a NullPointerException in the same case.

salvatore-coppola: port.getInternetProtocol() cannot be null because is set in ContainerPort constructors and cannot be something different from one item of org.eclipse.kura.container.orchestration.PortInternetProtocol enum

nicolatimeus: nothing stops the user from passing null as third argument here https://github.com/eclipse/kura/blob/6629ddf759d88af03b1da88bc27a3fc82253ce38/kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/container/orchestration/ContainerPort.java#L30

kura/org.eclipse.kura.container.orchestration.provider/src/main/java/org/eclipse/kura/container/orchestration/provider/impl/ContainerOrchestrationServiceImpl.java
```diff
@@ -705,6 +705,8 @@ private HostConfig containerPortManagementHandler(ContainerConfiguration contain
                     } catch (IllegalArgumentException e) {
                         logger.error("Invalid internet protocol: {}. Using TCP.", port.getInternetProtocol());
                     }
+                } else {
+                    logger.error("Null internet protocol: {}. Using TCP.", port.getInternetProtocol());
```
nicolatimeus: If the protocol is null this will log `Null internet protocol: null. Using TCP`, maybe the second `null` is not necessary, we could also switch the log level to warn, since it does not seem to be an unrecoverable error.

nicolatimeus: This comment has not been completely addressed

salvatore-coppola: sry I don't understand your suggestion, can you make a suggestion edit or directly the change? It will be faster

nicolatimeus: Nevermind, I didn't notice that you had removed the log entry altogheter.

kura/org.eclipse.kura.container.orchestration.provider/src/main/java/org/eclipse/kura/container/orchestration/provider/impl/ContainerOrchestrationServiceImpl.java
```diff
@@ -684,45 +683,45 @@ private HostConfig containerRuntimeConfigurationHandler(ContainerConfiguration c
...
-            this.exposedPorts = exposedPortsList;
+                ExposedPort exposedPort = new ExposedPort(port.getInternalPort(), ipProtocol);
+                exposedPorts.add(exposedPort);
+                portbindings.bind(exposedPort, Binding.bindPort(port.getExternalPort()));
+            }
 
-        } else {
-            logger.error("portsExternal and portsInternal must be int[] of the same size or they do not exist: {}",
-                    containerDescription.getContainerName());
+            if (exposedPorts.size() != portbindings.getBindings().size()) {
```
nicolatimeus, line 715: Is it possible for this to be false?

nicolatimeus: Maybe if there are multiple entries for the same port in received configuration https://github.com/docker-java/docker-java/blob/b29527510d9f7c16b30b138ae4a2f6703d88bb01/docker-java-api/src/main/java/com/github/dockerjava/api/model/Ports.java#L51

salvatore-coppola: The description of the metatype says that the size must be the same:

https://github.com/eclipse/kura/blob/6629ddf759d88af03b1da88bc27a3fc82253ce38/kura/org.eclipse.kura.container.provider/OSGI-INF/metatype/org.eclipse.kura.container.provider.ContainerInstance.xml#L91-L99

nicolatimeus: That part is related to the ContainerInstance configurable component, which is a client of the ContainerOrchestrationService API, so it shouldn't be relevant here, since ContainerOrchestrationServiceImpl only considers the result of ContainerConfiguration.getContainerPorts().

Anyway if the user specifies a list of Internal Ports with a different length than External Ports in ContanierInstance, the configuration parsing logic of that component will probably fail https://github.com/eclipse/kura/blob/6629ddf759d88af03b1da88bc27a3fc82253ce38/kura/org.eclipse.kura.container.provider/src/main/java/org/eclipse/kura/container/provider/ContainerInstanceOptions.java#L371

The condition here could probably evaluate to false if the user specifies multiple container ports with the same internal/external port, in the future we should investigate if that configuration is valid and what happens in that case.

nicolatimeus (approved):

PR #5246, opened 2024-05-24
feat(linux.gpio): Added support to gpio symlinks

pierantoniomerlino:

This PR adds the support to symlinks in the `jdk.dio.properties` file.

**Description of the solution adopted:** The current implementation of the `GPIOService` parses the `jdk.dio.properties` and reads the GPIO indexes/numbers. The entry looks like:

```
123 = deviceType: gpio.GPIOPin, pinNumber:123, name:GPIO123
```

This PR allows to set a symlink instead of a pin number in the configuration file:

```
/dev/dio1 = deviceType: gpio.GPIOPin, name:DIGITAL_INPUT1
```

where the `/dev/dio1` points to `/sys/class/gpio/gpioXYZ`. The `GPIOService` implementation will follow the symlink and read the pin number (XYZ) from the path.

### DIFF
--- a/kura/org.eclipse.kura.linux.gpio/src/main/java/org/eclipse/kura/linux/gpio/JdkDioPin.java
+++ b/kura/org.eclipse.kura.linux.gpio/src/main/java/org/eclipse/kura/linux/gpio/JdkDioPin.java
@@ -1,27 +1,16 @@
 /*******************************************************************************
- * Copyright (c) 2011, 2020 Eurotech and/or its affiliates and others
- * 
+ * Copyright (c) 2011, 2024 Eurotech and/or its affiliates and others
+ *
  * This program and the accompanying materials are made
  * available under the terms of the Eclipse Public License 2.0
  * which is available at https://www.eclipse.org/legal/epl-2.0/
- * 
+ *
  * SPDX-License-Identifier: EPL-2.0
- * 
+ *
  * Contributors:
  *  Eurotech
  *******************************************************************************/
 package org.eclipse.kura.linux.gpio;
-
-import java.io.IOException;
-
-import org.eclipse.kura.gpio.KuraClosedDeviceException;
-import org.eclipse.kura.gpio.KuraGPIODeviceException;
-import org.eclipse.kura.gpio.KuraGPIODirection;
-import org.eclipse.kura.gpio.KuraGPIOMode;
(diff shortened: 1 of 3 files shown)


kura/org.eclipse.kura.linux.gpio/src/main/java/org/eclipse/kura/linux/gpio/JdkDioPin.java
```diff
@@ -85,6 +93,24 @@ public static JdkDioPin parseFromProperty(Object key, String property) {
...
+        } catch (NumberFormatException e) {
+            return parseGpioSymbolicLink(key);
+        }
+    }
+
+    private static int parseGpioSymbolicLink(String gpioPath) throws IOException {
+        Path path = Paths.get(gpioPath);
+        if (Files.isSymbolicLink(path)) {
+            Path realPath = path.toRealPath();
+            return Integer.parseInt(realPath.getFileName().toString().substring(4));
```
MMaiero, line 108: Should we catch the NumberFormatException that can be thrown here?

kura/test/org.eclipse.kura.linux.gpio.test/src/test/java/org/eclipse/kura/linux/gpio/GPIOServiceImplTest.java
```diff
@@ -12,48 +12,51 @@
...
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotEquals;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertNull;
+import static org.junit.Assert.assertTrue;
 
 
 public class GPIOServiceImplTest {
 
+    Path jdkProperties;
```
MMaiero: private?

kura/org.eclipse.kura.linux.gpio/src/main/java/org/eclipse/kura/linux/gpio/JdkDioPin.java
pierantoniomerlino: Should it be caught at line 88?

MMaiero: True. Then we are good

MMaiero (approved):

PR #5249, opened 2024-05-29
fix(distrib): kurad installation on docker-alpine-x86_64-nn container

marcellorinaldo:
This PR fixes the installation of `kurad` user when using the `alpine` base image.

It adds the `shadow` package to have all the dependencies for the `manage_kura_users.sh` script. In this way, `kurad` is the owner of the installed files and the `CommandService` can be used in unprivileged mode as in the ubi8 container.

**Related Issue:** N/A.

**Screenshots:** N/A.

**Manual Tests**: N/A.


salvatore-coppola (changes requested):
We should also remove this line: 
https://github.com/eclipse/kura/blob/002b1476b111d168d10da181225ceb8133c1fe63/kura/distrib/src/main/resources/generic-x86_64-nn/kura_install.sh#L84

marcellorinaldo:
At the end we decided to keep the installation for the `kurad` user to ensure retrocompatibility with applications using the Kura ubi8 container.

Adding package `shadow` solved the missing packages on Alpine linux required by the users installation script.

salvatore-coppola (commented):
There is stiil a problem [here](https://github.com/eclipse/kura/pull/5249/files#diff-96b78b3fc6e033607d38fed382556a2e767925db32f02055bc7eb15ea10ce567R45)

This is the log:


```
	Registering '/opt/eclipse/kura_5.5.0-SNAPSHOT_docker-x86_64-nn/kura/packages/de.dentrassi.kura.addons.utils.fileinstall-0.6.0.dp' ...
	/usr/local/bin/dp-extract-name: line 6: jar: command not found
	grep: unrecognized option: P
	BusyBox v1.36.1 (2024-05-21 13:38:37 UTC) multi-call binary.
	
	Usage: grep [-HhnlLoqvsrRiwFE] [-m N] [-A|B|C N] { PATTERN | -e PATTERN... | -f FILE... } [FILE]...
	
	Search for PATTERN in FILEs (or stdin)
	
	-H	Add 'filename:' prefix
	-h	Do not add 'filename:' prefix
	-n	Add 'line_no:' prefix
	-l	Show only names of files that match
	-L	Show only names of files that don't match
	-c	Show only count of matching lines
	-o	Show only the matching part of line
	-q	Quiet. Return 0 if PATTERN is found, 1 otherwise
	-v	Select non-matching lines
	-s	Suppress open and read errors
	-r	Recurse
	-R	Recurse and dereference symlinks
	-i	Ignore case
	-w	Match whole words only
	-x	Match whole lines only
	-F	PATTERN is a literal (not regexp)
	-E	PATTERN is an extended regexp
	-m N	Match up to N times per file
	-A N	Print N lines of trailing context
	-B N	Print N lines of leading context
	-C N	Same as '-A N -B N'
	-e PTRN	Pattern to match
	-f FILE	Read pattern from file
	rm: can't remove 'META-INF/MANIFEST.MF': No such file or directory
	Register '/opt/eclipse/kura_5.5.0-SNAPSHOT_docker-x86_64-nn/kura/packages/de.dentrassi.kura.addons.utils.fileinstall-0.6.0.dp' as ''
```

Same for ubi8

Also need to update jdk in $PATH for both images.

nicolatimeus (approved):

PR #5252, opened 2024-05-29
fix: Allow kurad to shutdown and reboot the system

nicolatimeus:

Uptates polkit rules to allow kurad to perform the following actions:

* org.freedesktop.login1.reboot-multiple-sessions
* org.freedesktop.login1.suspend-multiple-sessions
* org.freedesktop.login1.power-off-multiple-sessions

allowing to suspend, reboot and poweroff the device through systemd even if there are other users logged in.


salvatore-coppola (approved):
Successfully tested on Raspberry PI4B+ with RaspiOS 64bit

PR #5263, opened 2024-06-05
fix(web2): Removed servlet unregistration during update if options are not changed

salvatore-coppola:
Removed servlets unregistration during the update if options don't change.

### DIFF
--- a/kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/Console.java
+++ b/kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/Console.java
@@ -203,14 +203,6 @@ public class Console implements SelfConfiguringComponent, org.eclipse.kura.web.a
         props.put("kura.version", this.systemService.getKuraVersion());
         EventProperties eventProps = new EventProperties(props);
 
-        try {
-            logger.info("initializing useradmin...");
-            this.userManager.update();
-            logger.info("initializing useradmin...done");
-        } catch (final Exception e) {
-            logger.warn("failed to update UserAdmin", e);
-        }
-
         logger.info("postInstalledEvent() :: posting KuraConfigReadyEvent");
 
         this.eventAdmin.postEvent(new Event(KuraConfigReadyEvent.KURA_CONFIG_EVENT_READY_TOPIC, eventProps));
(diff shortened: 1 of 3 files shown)


nicolatimeus:
I have following the following issue that does not seem to occur on RC1:
* Update HttpService configuration adding an new open port
* Observe that the web ui is no longer accessible

nicolatimeus:
Please update copyright headers

kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/Console.java
```diff
@@ -293,54 +274,70 @@ protected void deactivate(BundleContext context) {
...
+            quietUnregisterServlet(reg.path);
+        }
+
+    }
+
+    private void quietUnregisterServlet(String path) {
+        try {
+            this.httpService.unregister(path);
+        } catch (final IllegalArgumentException iae) {
+            // nothing to do
```
nicolatimeus: What is the reason for suppressing all IllegalArgumentExceptions?

salvatore-coppola: the logic has been unified and now the activation and the update call the same method, if the options are changed, a unregistration is made so if the aren't an exception is raised.
There isn't a method to check the registration status.

nicolatimeus: If the problem is avoiding the unregistration when the component is activated, is it possible to check if the options are null and skip the unregistration step in that case?

kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/ConsoleOptions.java
```diff
@@ -271,4 +273,40 @@ private void addAuthenticationMethodProperties() {
...
+
+        if (this == obj) {
+            return true;
+        }
+        if (obj == null || getClass() != obj.getClass()) {
+            return false;
+        }
+        ConsoleOptions other = (ConsoleOptions) obj;
+
+        return Objects.equals(this.appRoot.get(), other.appRoot.get())
```
nicolatimeus, line 295: Should we compare the result of getOptional() instead of get() where possible to stay on the safe side? If for some reason the property value is not set get() will throw

salvatore-coppola: from what I see appRoot has the default value (as do the other fields, except allowedPorts) so it is impossible, in this case, for the get method to throw the exception.

nicolatimeus (approved):

PR #5264, opened 2024-06-06
fix(distrib): add wpa_supplicant dbus permissions only on NN profiles

marcellorinaldo:
This PR adds a check for adding DBUS permissions to `kurad` user for managing WPA supplicant only on networking profiles. Furthermore, it fixes the missing `-nn` flags in no-network installations.

**Related Issue:** N/A.

**Screenshots:** N/A.

**Manual Tests**:

1. `/etc/dbus-1/system.d/wpa_supplicant.conf` exists and contains a policy for `kurad` on networking profiles.
2. `/etc/dbus-1/system.d/wpa_supplicant.conf` is not modified on NN profiles.
3. `/etc/dbus-1/system.d/wpa_supplicant.conf` is restored at uninstallation on networking profiles.
4. Can connect to an access point as client.
5. Can serve as AP.


marcellorinaldo:
Tested `generic-arm32`, `generic-arm32-nn`, Dockers.

pierantoniomerlino (approved):

PR #5266, opened 2024-06-07
fix(nm): Avoid querying wifi ap for info if it is not available anymore

salvatore-coppola:
Sometimes a wifi ap goes down between the instant is detected and the instant it is probed for its data.

With this change if something goes wrong the ap is skipped.

The file was not formatted and cleaned up as our standard, the only meaningful changs are in lines 380...384


kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/status/NMStatusConverter.java
```diff
@@ -376,7 +377,11 @@ private static List<WifiAccessPoint> wifiAccessPointConvert(List<Properties> nmA
         List<WifiAccessPoint> kuraAccessPoints = new ArrayList<>();
 
         for (Properties prop : nmAccessPoints) {
-            kuraAccessPoints.add(wifiAccessPointConvert(prop));
+            try {
+                kuraAccessPoints.add(wifiAccessPointConvert(prop));
+            } catch (Exception e) {
```
pierantoniomerlino, line 382: Does exist a specific exception? Or we cannot predict the error thrown, so it's better to catch them all?

salvatore-coppola: i wuold catch all, in any case wa want to skip that access point

pierantoniomerlino (approved):

PR #5267, opened 2024-06-07
fix(rest.identity): fixed status codes and error messages for malformed requests

sfiorani:
This PR modifies the response status code and error message in case of malformed request body for Identity/Permission REST APIs of IdentityV2.

When the request doesn't contain the `name` key or it's wrong (eg: `nam`), the corresponding DTO (Identity or Permission DTO) will have a null value as name. On Kura side, null/empty checks were missing, so the IdentityService was trying to remove something that it could not found, responding with a 404 status code with message `Identity not found`.

After this change, the null/empty checks are added: if the `name` value is null (missing or wrong `name` field) or empty, a 400 status code is returned with an appropriate message.

**Related Issue:**

**Screenshots:**

**Manual Tests:**

### DIFF
--- a/kura/org.eclipse.kura.rest.identity.provider/src/main/java/org/eclipse/kura/internal/rest/identity/provider/IdentityRestServiceV2.java
+++ b/kura/org.eclipse.kura.rest.identity.provider/src/main/java/org/eclipse/kura/internal/rest/identity/provider/IdentityRestServiceV2.java
@@ -111,6 +111,9 @@ public class IdentityRestServiceV2 {
         logger.debug(DEBUG_MESSAGE, "createIdentity");
 
         try {
+
+            StringUtils.validateField("name", identity.getName());
+
             boolean created = this.identityService.createIdentity(identity.getName());
             if (!created) {
                 throw DefaultExceptionHandler.buildWebApplicationException(Status.CONFLICT, "Identity already exists");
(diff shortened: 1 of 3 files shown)


kura/org.eclipse.kura.rest.identity.provider/src/main/java/org/eclipse/kura/internal/rest/identity/provider/IdentityRestServiceV2.java
```diff
@@ -195,6 +195,12 @@ public IdentityConfigurationDTO getIdentityDefaultByName(
     public Response deleteIdentity(final IdentityDTO identity) {
         logger.debug(DEBUG_MESSAGE, "deleteIdentity");
         try {
+
+            if (identity.getName() == null) {
```
MMaiero:
```suggestion
            if (isNull(identity.getName())) {
```
More readable like this?
If so, you'll need to add a static import

MMaiero (approved):

kura/org.eclipse.kura.rest.identity.provider/src/main/java/org/eclipse/kura/internal/rest/identity/provider/IdentityRestServiceV2.java
```diff
@@ -293,6 +314,9 @@ public Response deletePermission(final PermissionDTO permissionDTO) {
     @Consumes(MediaType.APPLICATION_JSON)
     public Response validateIdentityConfiguration(final IdentityConfigurationDTO identityConfigurationDTO) {
         try {
+
+            StringUtils.validateField("name", identityConfigurationDTO.getIdentity().getName());
```
MMaiero: Would it make sense to define "name" as a constant?

MMaiero (approved):

PR #5275, opened 2024-06-12
fix(distrib): customize-installation.sh to always patch interface names on snapshot_0.xml and iptables

marcellorinaldo:
This PR modifies the`customize-installation.sh` script that is called only on generic profiles to apply the network interface names patch even on devices that have templates for `snapshot_0.xml` and `iptables`. This is to avoid applying a configuration with the wrong interface name when the standard interface enumeration changes.

**Related Issue:** N/A.

**Screenshots:** N/A.

**Manual Tests**: To test the behavior, setup a Raspberry PI and:

1. Enable the consistent interface naming using `raspi-config` -> `Advanced Options` menu and reboot. `eth0` will change to something else.
2. Install Kura.
3. Inspect the installation logs, without this PR files `snapshot_0.xml` and `iptables` are not patched. Those files still have `eth0` as interface name since the template for the raspberry has been picked up.
4. Reboot, without this PR the device will be unreachable due to the interface naming mismatch.

By testing this PR, logs in `/opt/eclipse/kura/log/kura_install*` on a fresh installation should show that the patch has been applied to:

- `/opt/eclipse/kura/user/snapshots/snapshot_0.xml`
- `/opt/eclipse/kura/.data/snapshot_0.xml`
- `/opt/eclipse/kura/.data/iptables`

### DIFF
--- a/kura/distrib/src/main/resources/common/customize-installation.sh
+++ b/kura/distrib/src/main/resources/common/customize-installation.sh
@@ -38,6 +38,38 @@ setup_libudev() {
     fi
 }
 
+customize_kura_properties() {
+    local BOARD=$1
+
+    sed -i "s/device_name/${BOARD}/g" "/opt/eclipse/kura/framework/kura.properties"
+
+    if python3 -V > /dev/null 2>&1
+    then
+        python3 /opt/eclipse/kura/install/find_net_interfaces.py /opt/eclipse/kura/framework/kura.properties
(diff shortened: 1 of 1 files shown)


pierantoniomerlino (approved):

pierantoniomerlino:
Tested on RPI3 with 32-bit OS. Kura `generic-arm32`.

PR #5300, opened 2024-06-27
fix: Fixed Keypair validity display to link only the leaf

MMaiero:
In the current implementation the validity range for key-airs was linked to the CA validity.
Updated also the DN visualisation.

### DIFF
--- a/kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/server/GwtCertificatesServiceImpl.java
+++ b/kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/server/GwtCertificatesServiceImpl.java
@@ -163,7 +163,7 @@ public class GwtCertificatesServiceImpl extends OsgiRemoteServiceServlet impleme
                         Certificate[] chain = pke.getCertificateChain();
 
                         if (chain.length > 0) {
-                            Certificate leaf = chain[chain.length - 1];
+                            Certificate leaf = chain[0];
 
                             if (leaf instanceof X509Certificate) {
                                 validityStartDate = ((X509Certificate) leaf).getNotBefore();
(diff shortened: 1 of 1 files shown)


kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/server/GwtCertificatesServiceImpl.java
```diff
@@ -163,7 +163,7 @@ public List<GwtKeystoreEntry> listEntries() throws GwtKuraException {
                         Certificate[] chain = pke.getCertificateChain();
 
                         if (chain.length > 0) {
-                            Certificate leaf = chain[chain.length - 1];
+                            Certificate leaf = chain[0];
```
LucaDazi, line 166: Just a doubt here: I'm not sure if there might exist a chain with a CA expiring before the Private Key.
If that's the case, instead of choosing the first certificate of the chain, we should loop on the entire chain and show the first expiration date.

MMaiero: In theory, since the leaf is the one used then in the process, it should be right to display its validity period. It is also true that the validity of a CA can be less than the validity of a leaf that it has signed (during its validity time frame). So, not sure what is best. @salvatore-coppola any thoughts?

salvatore-coppola: I believe the most intersting date to see is the leaf, so I would use that date

MMaiero:
Ok to merge, then?

salvatore-coppola:
> Ok to merge, then?

usual copyright year

salvatore-coppola (approved):

MMaiero:
> > Ok to merge, then?
> 
> usual copyright year

Right. I always forget. Should be done now.

PR #5337, opened 2024-08-27
refactor: Removed debian and redhat bundles and their references

pierantoniomerlino:
This PR removes the debian and redhat bundles and optimize the build ant script.

**Description of the solution adopted:** Since the old Kura networking has been removed, the bundles that provides the methods to write/read networking files have been deleted as well. In particular the bundles for Debian and RedHat style network are removed. 

Moreover, a small refactor of the build ant script is performed.

**Manual Tests**: Install Kura, check that the webUI works and that the network services are present and up. 
The following bundles/services are present in the plugins folder:

- linux.net
- systemd.provider
- nm
- net.admin.firewall
- rest.network.status
- rest.network.configuration
- network.threat.manager

The following bundles should not be present:

- debian.provider
- redhat.provider


marcellorinaldo (approved):
Tested working on x86 profile.

NOTE: on `nn` profiles the following bundles are still present:

```
ls -al /opt/eclipse/kura/plugins/ | grep net
-rw------- 1 kurad kurad    88025 Aug 29 09:23 org.eclipse.kura.core.net_2.0.0-SNAPSHOT.jar
-rw------- 1 kurad kurad    31691 Aug 29 09:23 org.eclipse.kura.net.configuration_2.0.0-SNAPSHOT.jar
```

Their removal is not in the scope of this PR though.

PR #5340, opened 2024-08-29
fix: set correct metadata for debian packages

mattdibi:
Currently, in our `.deb` packages, we're not setting the metadata correctly.

### Architecture

Reference: https://www.debian.org/doc/debian-policy/ch-controlfields.html#architecture

> Depending on context and the control file used, the Architecture field can include the following sets of values:
>
> - A unique single word identifying a Debian machine architecture as described in [Architecture specification strings](https://www.debian.org/doc/debian-policy/ch-customized-programs.html#s-arch-spec).
> - An architecture wildcard identifying a set of Debian machine architectures, see [Architecture wildcards](https://www.debian.org/doc/debian-policy/ch-customized-programs.html#s-arch-wildcard-spec). any matches all Debian machine architectures and is the most frequently used.
> - `all`, which indicates an architecture-independent package.
>- `source`, which indicates a source package.

We're currently using `all` but it's not correct since we have architecture dependent code. I updated the `Architecture` fields using the "[Architecture specification strings](https://www.debian.org/doc/debian-policy/ch-customized-programs.html#s-arch-spec)" as per Debian documentation.

### Description

Reference: https://www.debian.org/doc/debian-policy/ch-controlfields.html#description

> In a source template control file or binary control file, the Description field contains a description of the binary package, consisting of two parts, the synopsis or the short description, and the long description. It is a multiline field with the following format:
>```
>Description: single line synopsis
> extended description over several lines
>```

I added the "single line synopsis" in this PR.

### Filename

WIP


mattdibi:
> This pull request is stale because it has been open for 60 days with no activity.

shoo

mattdibi:
Superseded by #5609

PR #5343, opened 2024-09-02
refactor(core)!: Added H2Db provider bundle

pierantoniomerlino:
This PR extracts the H2Db code and services from the `core` bundle and create a new `db.h2db.provider` one. This bundle is included in the standard installer.

As a side note, this update also includes a breaking change, since the `DataService` will track `MessageStoreProvider`s only. The explicit tracking of `H2DbService`s is not more supported.


kura/org.eclipse.kura.db.h2db.provider/build.properties
```diff
@@ -0,0 +1,23 @@
+#
+#  Copyright (c) 2022 Eurotech and/or its affiliates and others
```
MMaiero:
```suggestion
#  Copyright (c) 2024 Eurotech and/or its affiliates and others
```

kura/org.eclipse.kura.db.h2db.provider/src/main/java/org/eclipse/kura/internal/db/h2db/provider/H2DbQueryableWireRecordStoreImpl.java
```diff
@@ -23,6 +21,8 @@
 import org.eclipse.kura.util.jdbc.ConnectionProvider;
 import org.eclipse.kura.util.wire.store.AbstractJdbcQueryableWireRecordStoreImpl;
 
+import static java.util.Objects.isNull;
```
MMaiero: The static import should be at the beginning? Or I am wrong?

MMaiero:
Tested on device and it's ok for me

MMaiero (approved):

PR #5351, opened 2024-09-12
feat: Autogenerate config.ini before framework startup

nicolatimeus:

* Creates the following subdirectories in the `/opt/eclipse/kura/plugins` folder:
    * 1
    * 1s
    * 2
    * 2s
    * 3
    * 3s
    * 4
    * 4s
    * 5
    * 5s
    * 6
    * 6s
  Each directory is in the `${startlevel}${autostartSetting}` form, Kura plugins are placed in the corresponding folder depending on the start level and autostart setting assigned by the current config.ini.

* Added script that generates a config.ini at startup basing on a template (the current config.ini without the `osgi.bundles` property), adding all bundles discovered in the `${startlevel}${autostartSetting}` folders described above setting appropriate start level and autostart setting.

### DIFF
--- a/kura/distrib/src/main/ant/build_equinox_distrib.xml
+++ b/kura/distrib/src/main/ant/build_equinox_distrib.xml
@@ -17,7 +17,7 @@
     basedir="../../../">
 
     <target name="install-plugin">
-        <echo message="Installing ${bundle.file}, strart level: ${start.level}, start: ${start}"/>
+        <echo message="Installing ${bundle.file}, start level: ${start.level}, start: ${start}"/>
         <copy todir="${project.build.directory}/${build.output.name}/plugins/${start.level}${start}" failonerror="false" overwrite="false" file="${project.build.directory}/plugins/${bundle.file}" />
         <copy todir="${project.build.directory}/${build.output.name}/plugins/${start.level}${start}" failonerror="false" overwrite="false" file="../target-definition/common/source/plugins/${bundle.file}" />
         <copy todir="${project.build.directory}/${build.output.name}/plugins/${start.level}${start}" failonerror="false" overwrite="false" file="${kura.osgi.repo}/plugins/${bundle.file}" />
--- a/kura/distrib/src/main/ant/build_equinox_distrib.xml
+++ b/kura/distrib/src/main/ant/build_equinox_distrib.xml
@@ -16,6 +16,20 @@
 <project name="build_equinox_distrib" default="dist-linux"
     basedir="../../../">
 
+    <target name="install-plugin">
+        <echo message="Installing ${bundle.file}, strart level: ${start.level}, start: ${start}"/>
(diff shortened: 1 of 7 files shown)


kura/distrib/src/main/ant/build_equinox_distrib.xml
```diff
@@ -16,6 +16,20 @@
 <project name="build_equinox_distrib" default="dist-linux"
     basedir="../../../">
 
+    <target name="install-plugin">
+        <echo message="Installing ${bundle.file}, strart level: ${start.level}, start: ${start}"/>
```
mattdibi: Typo:

```suggestion
        <echo message="Installing ${bundle.file}, start level: ${start.level}, start: ${start}"/>
```

kura/distrib/src/main/ant/build_equinox_distrib.xml
```diff
@@ -128,110 +155,107 @@
...
-                value=", reference:file:${kura.install.dir}/${kura.symlink}/${plugins.folder}/org.tigris.mtoolkit.iagent.rpc_3.0.0.20110411-0918.jar@5:start" />
-        </propertyfile>
+        <antcall target="gpio-plugins" />
+        <antcall target="emulator-gpio-plugins" />
+        <antcall target="emulator-plugins" />
+        <antcall target="linux-network-plugins" />
+
+        <!-- Add mToolkit to distribution -->
+        <antcall target="install-plugin">
+            <param name="bundle.file" value="org.tigris.mtoolkit.iagent.rpc_3.0.0.20110411-0918.jar" />
```
mattdibi, line 248: 🤔 I don't have the complete picture so this might be a dumb question: is it absolutely necessary to have this hardcoded?

kura/distrib/src/main/resources/common/gen_config_ini.sh
```diff
@@ -0,0 +1,72 @@
+#!/bin/sh
+
+TEMPLATE=$1
+ROOT=$2
+
+usage() {
+    >&2 echo "Usage: gen_config_ini.sh <config.ini template> <plugin root DIR_NAMEectory>"
```
mattdibi: Probably a leftover from a string substitution

```suggestion
    >&2 echo "Usage: gen_config_ini.sh <config.ini template> <plugin root directory>"
```

mattdibi (commented):
LGTM, some minor things here and there that I highlighted in my comments.

The CI failure seems to be unrelated to this PR

kura/distrib/src/main/ant/build_equinox_distrib.xml
nicolatimeus: I wouldn't know either.. the version was hardcoded also before this change I have kept it as is.

mattdibi: Classic! 😄 

We can keep this as-is (unless you have the time/will to investigate this further... but for me the PR is OK)

mattdibi (approved):

PR #5350, opened 2024-09-11
feat(nm): Modem APN as an optional parameter

pierantoniomerlino:
This PR converts the APN parameter in the `Cellular` tab an optional property. In this way, `ModemManager` will manage automatically this value. If an APN is set, this will be passed to `ModemManager`.

**Related Issue:** This PR fixes/closes N/A

**Description of the solution adopted:** In the `nm` bundle the property is optional. The `web2` bundle is updated accordingly.

**Manual Tests**: Setup a device with cellular connection. In the `Cellular` tab leave empty the APN field and check that the connection succeed.

**Any side note on the changes made:** I noticed the following behavior:

1. Setup a connection without an APN value
2. The connection succeed
3. Set the APN property to a wrong one and connect
4. The connection succeed, since ModemManager uses the previous value as fallback.

This behavior is not present if the interface is disabled before setting the wrong APN. This has to be documented.

This PR has to be backported to Kura 5.6.0

PR #5352, opened 2024-09-12
docs(network): Added a note about APN parameter in cellular doc

pierantoniomerlino:
This PR adds a note about the optional APN parameter in the Cellural documentation page. See #5350 for more details.

### DIFF
--- a/docs/gateway-configuration/cellular-configuration.md
+++ b/docs/gateway-configuration/cellular-configuration.md
@@ -25,7 +25,7 @@ The **Cellular** tab contains the following configuration parameters:
     - HSPA modem: atd&ast;99&ast;&ast;&ast;1#
     - EVDO/CDMA modem: atd#777
 
-- **APN**: defines the modem access point name (HSPA modems only).
+- **APN**: defines the modem access point name (HSPA modems only). This parameter is optional.
 
 - **Auth Type**: specifies the authentication type (HSPA modems only).
     - None
@@ -60,3 +60,7 @@ The **GPS** tab allows the user to enable or disable the GPS module provided by 
     - This port may not be used to send AT commands to the modem.
     - The _PositionService_ should be enabled. Serial settings of the _PositionService_ should not be changed; it will be redirected to the modem GPS port automatically.
 
+### Notes about the APN parameter
+The Access Point Name is an optional parameter. If left empty, the value is automatically picked up from the Mobile Broadband Provider the modem is registered to. If a value is filled, the APN value is esplicitely configured and the default value is overridden.
+
+A good practice is to set the interface status to **Disabled** and then **Enable For WAN** when the APN is explicitely set. NetworkManager, indeed, will fallback to the default value if a wrong APN is specified, causing misleading behaviors. This does not happen if the interface is disabled and re-enabled after APN changes.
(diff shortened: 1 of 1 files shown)


docs/gateway-configuration/cellular-configuration.md
```diff
@@ -60,3 +60,7 @@ The **GPS** tab allows the user to enable or disable the GPS module provided by
     - This port may not be used to send AT commands to the modem.
     - The _PositionService_ should be enabled. Serial settings of the _PositionService_ should not be changed; it will be redirected to the modem GPS port automatically.
 
+### Notes about the APN parameter
+The Access Point Name is an optional parameter. If left empty, the value is automatically picked up from the Mobile Broadband Provider the modem is registered to. If a value is filled, the APN value is esplicitely configured and the default value is overridden.
+
+A good practice is to set the interface status to **Disabled** and then **Enable For WAN** when the APN is explicitely set. NetworkManager, indeed, will fallback to the default value if a wrong APN is specified, causing misleading behaviors. This does not happen if the interface is disabled and re-enabled after APN changes.
```
mattdibi: Would adding some styling help the user notice this note better?

```suggestion
!!! note  Notes about the APN parameter

    The Access Point Name is an optional parameter. If left empty, the value is automatically picked up from the 
    Mobile Broadband Provider the modem is registered to. If a value is filled, the APN value is explicitly 
    configured and the default value is overridden.

    A good practice is to set the interface status to **Disabled** and then **Enable For WAN** when the APN is 
    explicitly set. NetworkManager, indeed, will fallback to the default value if a wrong APN is specified, causing 
    misleading behaviors. This does not happen if the interface is disabled and re-enabled after APN changes.
```

docs/gateway-configuration/cellular-configuration.md
```diff
@@ -60,3 +60,7 @@ The **GPS** tab allows the user to enable or disable the GPS module provided by
     - This port may not be used to send AT commands to the modem.
     - The _PositionService_ should be enabled. Serial settings of the _PositionService_ should not be changed; it will be redirected to the modem GPS port automatically.
 
+### Notes about the APN parameter
+The Access Point Name is an optional parameter. If left empty, the value is automatically picked up from the Mobile Broadband Provider the modem is registered to. If a value is filled, the APN value is esplicitely configured and the default value is overridden.
```
mattdibi:
```suggestion
The Access Point Name is an optional parameter. If left empty, the value is automatically picked up from the Mobile Broadband Provider the modem is registered to. If a value is filled, the APN value is explicitly configured and the default value is overridden.
```

docs/gateway-configuration/cellular-configuration.md
mattdibi:
```suggestion
A good practice is to set the interface status to **Disabled** and then **Enable For WAN** when the APN is explicitly set. NetworkManager, indeed, will fallback to the default value if a wrong APN is specified, causing misleading behaviors. This does not happen if the interface is disabled and re-enabled after APN changes.
```

mattdibi (changes requested):
Some minor typos and stylistic suggestions.

mattdibi:
Another note: we should also point out that it will be optional **for generic profiles only**. I would put that in the 5.6 version of the documentation to avoid confusion in the 6.0 docs but I'm not sure it's the easiest way 🤔

mattdibi (approved):

PR #5360, opened 2024-09-19
feat: Modified modem APN tooltip (backport)

pierantoniomerlino:
Backport of #5359


kura/org.eclipse.kura.web2/src/main/resources/org/eclipse/kura/web/client/messages/Messages.properties
```diff
@@ -706,7 +706,7 @@ netModemToolTipModemIndentifier=Enter a unique name for the modem.
...
-netModemToolTipApn=Modem access point name.
+netModemToolTipApn=Modem access point name (Mandatory parameter for standard profiles and optional for generic ones).<br><br>In the case of generic profiles, if left empty this value will be picked up from the Mobile Broadband Provider the modem is registered to.<br><br>In order to avoid misconfiguration, it is strongly suggested to set it manually.
```
MMaiero: Just popping up to my mind. Does it make sense to split and have a tooltip depending on the device profile?
So stating that is mandatory in older profiles
And having another message where it states as optional for generic ones

MMaiero (approved):

PR #5363, opened 2024-09-19
test(db.h2db.provider.test): Check db file presence

pierantoniomerlino:
This PR updates the `testUpdateFile` test in the `org.eclipse.kura.db.h2db.provider.test`. The test now checks if the specific db file is present in the folder.


mattdibi (approved):
LGTM

PR #5375, opened 2024-09-27
feat(asset.provider): Added type for Scale and Offset in Channels

salvatore-coppola:
Added new field for Channel, scaleoffset.type which assign a type (Double, Float, Integer or Long) to scale and offset value used in the channel.


kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/channel/Channel.java
```diff
@@ -142,20 +200,62 @@ public boolean isEnabled() {
...
+     * 
+     * @throws IllegalArgumentException
      *
      * @since 2.3
+     * 
+     * @deprecated Use {@link #getValueOffsetAsTypedValue()}
      */
+    @Deprecated
     public double getValueOffset() {
+        if (this.valueOffset.getType() != DataType.DOUBLE) {
```
MMaiero: Is it better to convert to double instead of throwing an exception?

nicolatimeus:
I would limit the API changes to what is strictly necessary for implementing the new functionality (if any). For example adding the new `DataType.getNumericDataType`, `TypedValues.newNumericTypedValue` and `TypedValues.parseNumericTypedValue` APIs does not seem necessary to me since they don't add new functionalities, look specific for this use case and can be implemented with minimal effort using the already available APIs.

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/type/ScaleOffsetType.java
```diff
@@ -0,0 +1,64 @@
...
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ * 
+ * SPDX-License-Identifier: EPL-2.0
+ * 
+ * Contributors:
+ *  Eurotech
+ ******************************************************************************/
+package org.eclipse.kura.type;
```
nicolatimeus: Is it better to put this into the `org.eclipse.kura.channel` package?

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/channel/Channel.java
```diff
@@ -126,6 +185,15 @@ public DataType getValueType() {
         return this.valueType;
     }
 
+    /**
+     * Gets the Scale/Offset type as expected for operations.
+     *
```
nicolatimeus, line 189: Please add `@since 2.8`

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/channel/Channel.java
```diff
@@ -142,20 +210,60 @@ public boolean isEnabled() {
      *
      * @return a double that represents the scale factor to be applied to the read value
      *
+     * @throws IllegalArgumentException
+     * 
      * @since 2.3
+     * 
+     * @deprecated Use {@link #getValueScaleAsTypedValue()}
```
nicolatimeus:
```suggestion
     * @deprecated Use {@link #getValueScaleAsNumber()}
```

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/channel/Channel.java
```diff
@@ -142,20 +210,60 @@ public boolean isEnabled() {
...
      */
+    @Deprecated
     public double getValueScale() {
+        if (this.scaleOffsetType != ScaleOffsetType.DEFINED_BY_VALUE_TYPE) {
```
nicolatimeus: As per previous comment, is it better to return `this.valueScale.doubleValue()` instead of throwing?

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/channel/Channel.java
```diff
@@ -142,20 +210,60 @@ public boolean isEnabled() {
...
+     * 
+     * @throws IllegalArgumentException
      *
      * @since 2.3
+     * 
+     * @deprecated Use {@link #getValueOffsetAsTypedValue()}
      */
+    @Deprecated
     public double getValueOffset() {
+        if (this.scaleOffsetType != ScaleOffsetType.DEFINED_BY_VALUE_TYPE) {
```
nicolatimeus: Same as above for the scale

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/channel/Channel.java
```diff
@@ -205,6 +313,20 @@ public void setValueType(final DataType valueType) {
         this.valueType = valueType;
     }
 
+    /**
+     * Set the type of the scale/offset.
+     *
```
nicolatimeus, line 309: Please add `@since 2.8`

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/type/ScaleOffsetType.java
salvatore-coppola: don't know, i put there for simmetry with DataType

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/channel/Channel.java
salvatore-coppola: done

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/channel/Channel.java
salvatore-coppola: done

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/channel/Channel.java
salvatore-coppola: it's an illegal state of the object, i would throw the exception  because using the method results in an unintentional casting

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/channel/Channel.java
salvatore-coppola: it's an illegal state of the object, i would throw the exception because using the method results in an unintentional casting

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/channel/Channel.java
salvatore-coppola: done

salvatore-coppola:
> I would limit the API changes to what is strictly necessary for implementing the new functionality (if any). For example adding the new `DataType.getNumericDataType`, `TypedValues.newNumericTypedValue` and `TypedValues.parseNumericTypedValue` APIs does not seem necessary to me since they don't add new functionalities, look specific for this use case and can be implemented with minimal effort using the already available APIs.

They have been removed

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/type/ScaleOffsetType.java
nicolatimeus: The new enum seems only related to Asset channel configuration and scale and offset to me and not general like DataType and TypedValue. All of the scale and offset related APIs are in the `org.eclipse.kura.channel` package.

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/channel/Channel.java
salvatore-coppola: i'm reconsidering this, an user needs to change the api in a client application to not break it loading a configuraiton without the scaleoffset value property, i'll remove the exception throwing

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/channel/Channel.java
salvatore-coppola: i'm reconsidering this, an user needs to change the api in a client application to not break it loading a configuraiton without the scaleoffset value property, i'll remove the exception throwing

salvatore-coppola:
Replaced by #5418

PR #5377, opened 2024-09-27
docs(connect-field-devices): Added scaleoffset.type and rewritten scale and offset

salvatore-coppola:
Added scaleoffset.type and rewritten scale and offset


nicolatimeus:
This PR needs update after the latest changes. Please also expand it a bit if possible, describing more in detail how scale and offset is computed depending on the value of the new scale offset type parameter with some examples.

salvatore-coppola:
> This PR needs update after the latest changes. Please also expand it a bit if possible, describing more in detail how scale and offset is computed depending on the value of the new scale offset type parameter with some examples.

Udpated, what do you think about ?

nicolatimeus (approved):

PR #5396, opened 2024-10-01
fix(distrib): Added dhcpcd dependency to debian installer

pierantoniomerlino:
This PR adds `dhcpcd` as dependency to the Debian installer. This is needed for devices/OS that doesn't provide `isc-dhcp-client` anymore.

### DIFF
--- a/kura/distrib/src/main/resources/generic-aarch64/deb/control/control
+++ b/kura/distrib/src/main/resources/generic-aarch64/deb/control/control
@@ -19,7 +19,7 @@ Depends: openjdk-17-jre-headless | temurin-17-jdk | openjdk-8-jre-headless | tem
  polkit | policykit-1, ssh | openssh, openssl, busybox, openvpn,
  bluez | bluez5, bluez-hcidump | bluez5-noinst-tools,
  ntpdate, chrony, chronyc | chrony, cron | cronie,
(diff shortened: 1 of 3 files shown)


kura/distrib/src/main/resources/generic-aarch64/deb/control/control
```diff
@@ -19,7 +19,7 @@ Depends: openjdk-17-jre-headless | temurin-17-jdk | openjdk-8-jre-headless | tem
  polkit | policykit-1, ssh | openssh, openssl, busybox, openvpn,
  bluez | bluez5, bluez-hcidump | bluez5-noinst-tools,
  ntpdate, chrony, chronyc | chrony, cron | cronie,
- network-manager | networkmanager, bind9 | bind, dnsmasq | isc-dhcp-server, isc-dhcp-client, iw, iptables, modemmanager,
+ network-manager | networkmanager, bind9 | bind, dnsmasq | isc-dhcp-server, dhcpcd | isc-dhcp-client, iw, iptables, modemmanager,
```
MMaiero:
```suggestion
 network-manager | networkmanager, bind9 | bind, dnsmasq | isc-dhcp-server, isc-dhcp-client | dhcpcd, iw, iptables, modemmanager,
```

I would put it in this way since dhcpcd is for now a dependency that is less important than ism. If disc is available we want that installed

MMaiero (approved):

PR #5406, opened 2024-10-03
feat(nm): add managed GPS mode

mattdibi:
This PR adds a new mode for handling the modem's GPS, allowing ModemManager to take control of the serial port and thus open the way for us to create a new PositionProvider that retrieves the position information from ModemManager.

This new mode leverages the following two ModemManager location sources:
- [MM_MODEM_LOCATION_SOURCE_GPS_RAW](https://www.freedesktop.org/software/ModemManager/doc/latest/ModemManager/ModemManager-Flags-and-Enumerations.html#MM-MODEM-LOCATION-SOURCE-GPS-RAW:CAPS)
- [MM_MODEM_LOCATION_SOURCE_GPS_NMEA](https://www.freedesktop.org/software/ModemManager/doc/latest/ModemManager/ModemManager-Flags-and-Enumerations.html#MM-MODEM-LOCATION-SOURCE-GPS-NMEA:CAPS)

Once these two are enabled ModemManager will take control of the GPS communication and start parsing the data, allowing clients to retrieve the informations via the "[GetLocation](https://www.freedesktop.org/software/ModemManager/doc/latest/ModemManager/gdbus-org.freedesktop.ModemManager1.Modem.Location.html#gdbus-method-org-freedesktop-ModemManager1-Modem-Location.GetLocation)" DBus method.

Before this PR (see #4633) we were only using the `MM_MODEM_LOCATION_SOURCE_GPS_UNMANAGED` mode, which took care of setting up the GPS for the modem but left the serial connection free for other services to connect to (eg `gpsd`).

**Note**: to avoid breaking changes in the snapshot, instead of substituting the `net.interface.[deviceId].config.gpsEnabled` parameter with an enum, I left it untouched and added the `net.interface.[deviceId].config.gpsMode` parameter. It's a little bit more clunky to use but doesn't break snapshot compatibility with older versions. The `gpsMode` parameter is *optional*, when not specified it defaults to the `UNMANAGED` mode, preserving the current behaviour.

Since we added this new parameter, the [`isGpsSupported()` method](https://github.com/eclipse-kura/kura/blob/51fcd57437af67f22809b24ccc116b52c347c336/kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/net/modem/CellularModem.java#L208) in Kura's API **doesn't contain all the required informations anymore**. We'll need to add to the APIs a new `getSupportedGPSModes()` method to know what mode we can set for the modem at hand. This new API will be introduced in a future separate PR.

### DIFF
--- a/kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/KuraModemGPSMode.java
+++ b/kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/KuraModemGPSMode.java
@@ -1,0 +1,51 @@
+package org.eclipse.kura.nm;
+
+import java.util.EnumSet;
+import java.util.Optional;
+import java.util.Set;
+
+import org.eclipse.kura.nm.enums.MMModemLocationSource;
+
+public enum KuraModemGPSMode {
+
+    UNMANAGED,
+    MANAGED_GPS;
+
+    public static KuraModemGPSMode fromString(String mode) {
+        switch (mode) {
+        case "unmanaged":
+            return KuraModemGPSMode.UNMANAGED;
+        case "managed-gps":
+            return KuraModemGPSMode.MANAGED_GPS;
+        default:
+            throw new IllegalArgumentException(String.format("Unsupported modem GPS mode value: \"%s\"", mode));
+        }
+    }
+
+    public static Optional<KuraModemGPSMode> fromString(Optional<String> mode) {
+        if (mode.isPresent()) {
+            switch (mode.get()) {
+            case "unmanaged":
+                return Optional.of(KuraModemGPSMode.UNMANAGED);
+            case "managed-gps":
+                return Optional.of(KuraModemGPSMode.MANAGED_GPS);
+            default:
+                return Optional.empty();
+            }
+        } else {
+            return Optional.empty();
+        }
+    }
+
+    public static Set<MMModemLocationSource> toMMModemLocationSources(KuraModemGPSMode mode) {
+        switch (mode) {
+        case UNMANAGED:
+            return EnumSet.of(MMModemLocationSource.MM_MODEM_LOCATION_SOURCE_GPS_UNMANAGED);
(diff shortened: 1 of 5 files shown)


kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/KuraModemGPSMode.java
```diff
@@ -0,0 +1,71 @@
...
+import java.util.EnumSet;
+import java.util.Set;
+
+import org.eclipse.kura.nm.enums.MMModemLocationSource;
+
+public enum KuraModemGPSMode {
+
+    KURA_MODEM_GPS_MODE_UNMANAGED("kuraModemGpsModeUnmanaged"),
+    KURA_MODEM_GPS_MODE_MANAGED_GPS("kuraModemGpsModeManagedGps");
+
```
pierantoniomerlino, line 24: Does it make sense to call it `KURA_MODEM_GPS_MODE_MANAGED`?

mattdibi: Legitimate question.

I decided to call it this way since ModemManager also supports the ["Assisted-GPS" mode](https://www.freedesktop.org/software/ModemManager/doc/latest/ModemManager/ModemManager-Flags-and-Enumerations.html#MMModemLocationSource) (i.e. `MM_MODEM_LOCATION_SOURCE_AGPS_MSA` and `MM_MODEM_LOCATION_SOURCE_AGPS_MSB`).

If, in the future, we want to add the support for this additional mode there's no clash with the existing name and we can simply add something like `KURA_MODEM_GPS_MODE_MANAGED_AGPS_MSA` and `KURA_MODEM_GPS_MODE_MANAGED_AGPS_MSB`.

That was my overall idea. Feel free to suggest better naming conventions.

pierantoniomerlino (approved):

PR #5381, opened 2024-09-27
feat(api): added gnss type data to PositionService

sfiorani:
This PR introduces a new API to the PositionService: its aim is to provide information about the GNSS Type used to retrieve the position data

**Related Issue:**

**Description of the solution adopted:**

**Screenshots:**

**Manual Tests:**

**Any side note on the changes made:**

PR #5414, opened 2024-10-04
feat(api): add getSupportedGpsModes() to modem status API

mattdibi:
Following #5406 this PR adds a way for consumers to request, from the modem status API, the supported GPS modes for the modem.

This PR adds the following:
1. The `ModemGpsMode` enum, which represent the supported GPS modes for the modem
2. The `getSupportegGpsModes` method in the `ModemStatusInterface`

Furthermore it updates the NMDbusWrapper to correctly populate this information.

---

**Why there's no `getEnabledGpsModes()` method?**: After a discussion with @pierantoniomerlino, we agree that that is part of the _configuration_ not the _status_, and thus is already addressed in #5406. In our design, if something is part of the configuration, we assume that the desired status is correctly configured.

Example: static IP address. It's part of the configuration not the status. Being what we configured we assumed it matches its status.


kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/net/status/modem/ModemGpsMode.java
```diff
@@ -0,0 +1,30 @@
...
+    /*
+     * Unmanaged GPS mode. In this mode the GPS device of the modem will be setup but not directly managed, therefore
+     * freeing the serial port for other services to use.
+     */
+    UNMANAGED,
+    /*
+     * Managed GPS mode. In this mode the GPS device of the modem will be setup and directly managed (typically by
+     * ModemManager) therefore the serial port won't be available for other services to use.
+     */
+    MANAGED_GPS
```
pierantoniomerlino, line 31: Does it make sense to use simply `MANAGED`?

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/net/status/modem/ModemGpsMode.java
```diff
@@ -0,0 +1,30 @@
...
+ *
+ * Contributors:
+ *  Eurotech
+ ******************************************************************************/
+
+package org.eclipse.kura.net.status.modem;
+
+/*
+ * The GPS mode supported by the modem.
+ */
```
pierantoniomerlino, line 20: Please add `@since 3.0` annotation.

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/net/status/modem/ModemInterfaceStatus.java
```diff
@@ -147,6 +149,13 @@ public Boolean isGpsSupported() {
         return this.gpsSupported;
     }
 
+    /*
+     * @since 2.8
```
pierantoniomerlino, line 153: Have we upticked the api version to 3.0, isn't it? :confused:

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/net/status/modem/ModemGpsMode.java
mattdibi: See my answer on [this comment](https://github.com/eclipse-kura/kura/pull/5406#discussion_r1787625222)

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/net/status/modem/ModemInterfaceStatus.java
mattdibi: I think the motivation behind this is given by the fact that this PR will be backported to 5.6 and will first be available there. I'm simply relying on the previous work done by @sfiorani under the suggestion of @nicolatimeus for #5381

Feel free to correct me on this obviously.

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/net/status/modem/ModemGpsMode.java
mattdibi: I'll add it as soon as we clarify things on https://github.com/eclipse-kura/kura/pull/5414#discussion_r1787630255

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/net/status/modem/ModemInterfaceStatus.java
pierantoniomerlino: If this code will be contributed to 5.6, the 2.8 is correct.

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/net/status/modem/ModemGpsMode.java
mattdibi: Addressed in 5576993084626dded6d1652c873310e7f10f1d9c

pierantoniomerlino (approved):

PR #5419, opened 2024-10-04
feat(distrib): Network configuration customization during Kura install

pierantoniomerlino:
Currently the default IPv4 network configuration for all devices is as follows:

| ETH0 | WLAN0 |
|-------------|------------------------------------|
| WAN DHCP | AccessPoint, DHCP&NAT, passDNS |

This PR updates the Kura installation scripts in order to detect the interfaces present in the device and adapt the network configuration as follows:

| ETH0 |
|-------------|
| WAN DHCP |

| ETH0 | ETH1 |
|-------------|-------|
| LAN, DHCP&NAT, passDNS | WAN DHCP |

| ETH0 | ETH1 | WLAN0 |
|-------------|-------|-----|
| LAN, DHCP&NAT, passDNS | WAN DHCP |AccessPoint, DHCP&NAT, passDNS|

In the case of more ethernet or wireless interfaces, they will be disabled. The installation scripts manage the `snapshot_0.xml` file for the network and firewall configuration, the `iptables` file for the firewall configuration and the `kura.properties` file for the device name and the main network interface. 
The script detects the available interfaces on the system, sort them in lexicographic order and pick up the first ones.
When a known device is detected during installation (i.e. a RaspberryPi), the specific configuration is applied.

Regarding the IPv6 configuration, the interfaces will be disabled by default.

When a NN profile is built, the `snapshot_0.xml` is not modified (a NN version is installed instead) and the `iptables` file is not copied.


kura/distrib/src/main/resources/common/customize_iptables.py
```diff
@@ -0,0 +1,60 @@
...
+                iptables_content += line.replace('ETH_INTERFACE_0', eth_names[0])
+    else:
+        iptables_content = iptables.read()
+        iptables_content = iptables_content.replace('ETH_INTERFACE_0', eth_names[0])
+        iptables_content = iptables_content.replace('WIFI_INTERFACE_0', wlan_names[0])
+    
+    iptables.seek(0)
+    iptables.truncate()
+    iptables.write(iptables_content)
+    iptables.close()
```
mattdibi: Prefer the use of the `with` statement when working with files. See: https://www.freecodecamp.org/news/with-open-in-python-with-statement-syntax-example/. In case of exceptions being thrown the file gets closed for you automatically.

Also you can simplify the code a little to improve readability

```suggestion
    with open(iptables_filename, 'r', encoding='utf-8') as iptables:
        iptables_content = iptables.read()
    
    if wlan_number == 0:
        for line in iptables_content.split("\n"):
            if 'WIFI_INTERFACE_0' not in line:
                iptables_content += line.replace('ETH_INTERFACE_0', eth_names[0])
    else:
        iptables_content = iptables_content.replace('ETH_INTERFACE_0', eth_names[0])
        iptables_content = iptables_content.replace('WIFI_INTERFACE_0', wlan_names[0])
    
    with open(iptables_filename, 'w', encoding='utf-8') as iptables:
        iptables.write(iptables_content)
```

kura/distrib/src/main/resources/common/customize_iptables.py
```diff
@@ -0,0 +1,60 @@
...
+    logging.basicConfig(
+        format='[customize_iptables.py] %(asctime)s %(levelname)s %(message)s',
+        level=logging.INFO,
+        datefmt='%Y-%m-%d %H:%M:%S',
+        handlers=[
+            logging.StreamHandler()
+        ]
+    )
+    
+    iptables_filename = "/opt/eclipse/kura/.data/iptables"
```
mattdibi: Use uppercase letters for constants.

Python doesn’t support constants or non-reassignable names. Using uppercase letters is just a convention, and it doesn’t prevent developers from assigning new values to your constant... but it's common enough to be a good idea to do it.

See https://peps.python.org/pep-0008/#constants

```suggestion
    IPTABLES_FILENAME = "/opt/eclipse/kura/.data/iptables"
```

kura/distrib/src/main/resources/common/customize_kura_properties.py
```diff
@@ -0,0 +1,58 @@
...
+    logging.basicConfig(
+        format='[customize_kura_properties.py] %(asctime)s %(levelname)s %(message)s',
+        level=logging.INFO,
+        datefmt='%Y-%m-%d %H:%M:%S',
+        handlers=[
+            logging.StreamHandler()
+        ]
+    )
+    
+    board_name = sys.argv[1:] # remove script name from args
```
mattdibi: This is somewhat misleading. `board_name` here is not a string as the name might suggest but a slice of the original `sys.argv` array.

I would either change its name to `cli_args` (or something similar) or directly extract the board_name (risky):

```suggestion
    board_name = sys.argv[1:][0] # remove script name from args
```

mattdibi: It might be overkill for this particular case but you could also use something like [argparse](https://docs.python.org/3/library/argparse.html) which comes built-in into Python.

kura/distrib/src/main/resources/common/customize_snapshot.py
```diff
@@ -0,0 +1,95 @@
...
+    network_template_content = network_template.read()
+    firewall_template_content = firewall_template.read()
+    snapshot_content = snapshot_content.replace('NETWORK_CONFIGURATION', network_template_content)
+    snapshot_content = snapshot_content.replace('FIREWALL_CONFIGURATION', firewall_template_content)
+    
+    interfaces_list = "lo"
+
+    for i, eth_name in enumerate(eth_names):
+        if i > 1:
+            break
```
mattdibi: *If* we're guaranteed to receive the names in order, we can slice the array to avoid the if-break

```suggestion
    for i, eth_name in enumerate(eth_names[:2]):
```

here the if-break might be more explicit in what we're trying to achieve though.

kura/distrib/src/main/resources/common/customize_iptables.py
mattdibi: In case you don't want the to `split` the `iptables_content` you can read the file and put its content in an array of strings like so:

```python
with open(iptables_filename, 'r', encoding='utf-8') as iptables:
   iptables_content = [line for line in iptables]
```

but then you won't be able to perform the file-wise replace in the `else` branch. Probably the `split` option is the best one for this particular case.

kura/distrib/src/main/resources/common/customize_kura_properties.py
pierantoniomerlino: Cool! But I agree that it's a bit overkill in this specific case.

pierantoniomerlino:
The modifications have been successfully tested on a Raspberry PI with several network interfaces attached (eth0, eth1 and wlan0).

MMaiero (dismissed):

pierantoniomerlino:
The last modifications remove the `snapshot_0.xml` and `iptables` files for the raspberry profiles. The idea is to base the installation on the generic resources and customize them even if a known device is recognized.

kura/distrib/src/main/resources/common/customize_snapshot.py
```diff
@@ -0,0 +1,111 @@
...
+    
+    SNAPSHOT_FILENAME = "/opt/eclipse/kura/user/snapshots/snapshot_0.xml"
+    
+    args = sys.argv[1:]
+    if len(args) < 1:
+        logging.info("ERROR: invalid arguments length")
+        exit(1)
+    is_networking_profile = args[0]
+    
+    if is_networking_profile == 'false':
```
mattdibi: Comparing a string containing a boolean really irks me. I would start considering the use of `argparse` here:

```suggestion
    parser = argparse.ArgumentParser(
            description="Customize snapshot application")
    parser.add_argument('--networking_profile', action='store_true')
    parser.set_defaults(networking_profile=False)
    
    args = parser.parse_args()
    
    if not args.networking_profile:
```

With `argparse` you also get some niceties like `help` messages and such. Let me know what you think

pierantoniomerlino: I think that it a bit overkill for this script that it is intended to be used only internally. However, I agree that the boolean parameter management here is not the best :-). I'll implement the `argparse`, so if we'll expand the script in future, it will be alreadt there.

mattdibi (approved):

PR #5429, opened 2024-10-10
feat: added gnssType to position rest service

sfiorani:
This PR updates the Position Rest Service to include also the GNSS Type introduced in #5381 . Now the `getPosition` GET call returns also the GNSS Type if available.

**Related Issue:**

**Screenshots:**

**Manual Tests:**


kura/org.eclipse.kura.rest.position.provider/src/main/java/org/eclipse/kura/internal/rest/position/PositionRestService.java
```diff
@@ -84,7 +84,7 @@ public void unsetRequestHandlerRegistry(final RequestHandlerRegistry registry) {
     @Produces(MediaType.APPLICATION_JSON)
     public PositionDTO getPosition() {
         if (positionServiceImpl.isLocked()) {
-            return new PositionDTO(positionServiceImpl.getPosition());
+            return new PositionDTO(positionServiceImpl.getPosition(), positionServiceImpl.getGnssType());
```
pierantoniomerlino, line 87: I know that it isn't related to this PR and it is a detail, but I'd call this getGnssType**s** or getGnssTypeSet instead of getGnssType since it returns a set not a single value.

kura/org.eclipse.kura.rest.position.provider/src/main/java/org/eclipse/kura/rest/position/api/PositionDTO.java
```diff
@@ -78,4 +90,11 @@ public Double getSpeed() {
     public Double getTrack() {
         return track;
     }
+
+    /*
+     * Returns the GNSS Type used to retrieve position information
+     */
+    public Set<String> getGnssType() {
```
pierantoniomerlino, line 97: The same here.

pierantoniomerlino (approved):

PR #5432, opened 2024-10-11
feat(asset.provider): Now the DataType sent to the driver depends on the DataType of the channel and theDataType of the scaleOffsetDataType field and a new ScaleOffsetDataType, LONG, has been added

salvatore-coppola:


kura/org.eclipse.kura.asset.provider/src/main/java/org/eclipse/kura/internal/asset/provider/BaseAssetConfiguration.java
```diff
@@ -32,6 +32,7 @@
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Objects;
+import java.util.logging.Logger;
```
marcellorinaldo: Please import `org.slf4j.Logger`

marcellorinaldo (changes requested):

marcellorinaldo (approved):

PR #5434, opened 2024-10-11
feat(position): added datetime and gnsstype to position service static provider

sfiorani:
This PR updates the PositionService implementation adding a dropdown option to set the GNSS Type of the Static position. Also, the LocalDateTime is generated as the moment in which the PositionService is asked for the position information.

If we consider the Static provider as a generator of position information, we can assume that the data is generated each time a consumer ask for it: for this reason the LocalDateTime will be set as the request datetime.

**Related Issue:**

**Screenshots:**

**Manual Tests**


kura/org.eclipse.kura.linux.position/src/main/java/org/eclipse/kura/linux/position/PositionServiceImpl.java
```diff
@@ -194,16 +199,34 @@ public String getNmeaDate() {
 
     @Override
     public LocalDateTime getDateTime() {
-        if (this.currentProvider != null) {
-            return this.currentProvider.getDateTime();
+
+        if (this.options.isEnabled()) {
+
+            if (!this.options.isStatic()) {
```
pierantoniomerlino: What happens if the currentProvider is null? Should we return a null as in the previous implementation?

kura/org.eclipse.kura.linux.position/src/main/java/org/eclipse/kura/linux/position/PositionServiceImpl.java
```diff
@@ -194,16 +199,34 @@ public String getNmeaDate() {
...
             return null;
         }
     }
 
     @Override
     public Set<GNSSType> getGnssType() {
-        return this.currentProvider.getGnssType();
+        if (this.options.isEnabled()) {
+
+            if (!this.options.isStatic()) {
```
pierantoniomerlino: Also in this case, what will happen if the provider is null?

pierantoniomerlino:
@sfiorani Please update the copyright where needed.

pierantoniomerlino (approved):

PR #5452, opened 2024-10-16
feat: Added  new KeystoreService.getKeyManagers() method that allows provider selection

nicolatimeus:

Adds a new `KeystoreService.getKeyManagers()` API method that allows to select a specific provider in addition to the algorithm in the same way as [KeyManagerFactory.getInstance(String, String)](https://docs.oracle.com/javase/8/docs/api/javax/net/ssl/KeyManagerFactory.html#getInstance-java.lang.String-java.lang.String-)


kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/security/keystore/KeystoreService.java
```diff
@@ -104,6 +104,24 @@ public interface KeystoreService {
      */
     public List<KeyManager> getKeyManagers(String algorithm) throws KuraException;
 
+    /**
+     * Returns one key manager for each type of key material using the Java Security API Provider with the given name.
```
MMaiero:
```suggestion
     * Returns one key manager for each type of key material using the Java Security API Provider matching the given name.
```

MMaiero (approved):

PR #5463, opened 2024-10-20
chore(poms): uptick version of properties maven plugin

gnunzi:
I noticed we are using an old alpha version of this plugin, so this is an uptick.

The properties maven plugin is used to read the file from target-platform/config/kura.target-platform.build.properties. It should be harmless to uptick.
On the other side, it is wise to move away from the alpha version.

References:
https://mvnrepository.com/artifact/org.codehaus.mojo/properties-maven-plugin/1.2.1

https://www.mojohaus.org/properties-maven-plugin/  (see top right for the version)

### DIFF
--- a/kura/distrib/pom.xml
+++ b/kura/distrib/pom.xml
@@ -43,6 +43,8 @@
         <kura.skip.web>false</kura.skip.web>
         <kura.skip.examples>true</kura.skip.examples>
 
+        <properties-maven-plugin.version>1.2.1</properties-maven-plugin.version>
+
         <!-- used for RPM uninstallation -->
         <kura.install.link>/opt/eclipse/kura</kura.install.link>
     </properties>
@@ -93,7 +95,7 @@
             <plugin>
                 <groupId>org.codehaus.mojo</groupId>
                 <artifactId>properties-maven-plugin</artifactId>
-                <version>1.0-alpha-1</version>
+                <version>${properties-maven-plugin.version}/version>
                 <executions>
                     <execution>
                         <phase>initialize</phase>
@@ -1022,7 +1024,7 @@
                                             properties-maven-plugin
                                         </artifactId>
                                         <versionRange>
-                                            [1.0-alpha-1,)
+                                            [1.2.1,)
                                         </versionRange>
                                         <goals>
                                             <goal>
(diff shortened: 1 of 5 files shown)


marcellorinaldo (commented):
Thanks for the contribution! Could you please also update the other references and do a rebase with develop?

- https://github.com/eclipse-kura/kura/blob/b2035dfae52469ed7c000e169e768ca72dac3a0a/kura/distrib/pom.xml#L96
- https://github.com/eclipse-kura/kura/blob/b2035dfae52469ed7c000e169e768ca72dac3a0a/kura/distrib/src/main/resources/common/kura-bundles-pom.xml#L22

gnunzi:
Ok, I missed those.
Should I introduce the property `<properties-maven-plugin.version>1.2.1</properties-maven-plugin.version>` like in the other poms, or just uptick the version number in the plugin section?
Tiny detail I know :)

marcellorinaldo:
> Should I introduce the property `<properties-maven-plugin.version>1.2.1</properties-maven-plugin.version>` like in the other poms

That would be really nice! I think that way it will be more maintainable

kura/distrib/pom.xml
```diff
@@ -93,7 +95,7 @@
             <plugin>
                 <groupId>org.codehaus.mojo</groupId>
                 <artifactId>properties-maven-plugin</artifactId>
-                <version>1.0-alpha-1</version>
+                <version>${properties-maven-plugin.version}/version>
```
pierantoniomerlino:
```suggestion
                <version>${properties-maven-plugin.version}</version>
```

pierantoniomerlino (commented):
Please fix the typo

gnunzi:
@marcellorinaldo the `kura-bundles-pom.xml` looks like a truncated xml (or?). Maybe it is not in use anymore.


> * https://github.com/eclipse-kura/kura/blob/b2035dfae52469ed7c000e169e768ca72dac3a0a/kura/distrib/src/main/resources/common/kura-bundles-pom.xml#L22

marcellorinaldo:
> @marcellorinaldo the `kura-bundles-pom.xml` looks like a truncated xml (or?). Maybe it is not in use anymore.
> 
> > * https://github.com/eclipse-kura/kura/blob/b2035dfae52469ed7c000e169e768ca72dac3a0a/kura/distrib/src/main/resources/common/kura-bundles-pom.xml#L22

yes, i don't think it is used anymore

marcellorinaldo (approved):

mattdibi:
@gnunzi thanks for your contribution 🙏

PR #5479, opened 2024-10-23
feat: Added support for emitting connection errors in WireAsset

nicolatimeus:

Adds support for emitting an envelope in case of ConnectionException during read operations in WireAsset. The envelope will contain a property named `assetError` containing  the error message associated with the exception and a timestamp.


pierantoniomerlino (approved):

pierantoniomerlino:
Succesfully tested with a RPI and the Modbus Driver.

PR #5481, opened 2024-10-24
feat(nm): Added SIM operator ID to Network Status

pierantoniomerlino:
This PR adds the SIM `operatorIdentifier` to the `NetworkStatusService`.

**Description of the solution adopted:** The new field `operatorIdentifier` has been added to the SIM properties. This value can be obtained using Rest calls.


mattdibi (approved):
LGTM

mattdibi:
I tested this on my device using a Quectel modem. Two things to note:
- The operator identified is only available when the modem is *registered* on the network
- The DBus property containing the information about the operator identifier that we are using (see: https://www.freedesktop.org/software/ModemManager/doc/latest/ModemManager/gdbus-org.freedesktop.ModemManager1.Sim.html#gdbus-property-org-freedesktop-ModemManager1-Sim.OperatorIdentifier) seems to be subject of a caching mechanism. After registering the modem the property was still missing and only after a modem reset the info were given.

PR #5522, opened 2024-11-07
fix(distrib): Updated installer script to write all interface names to snapshot file

pierantoniomerlino:
This PR updates the installer script to add all the interface names detected on the system during installation to the `net.interfaces` property of the snapshot_0 file.

### DIFF
--- a/kura/distrib/src/main/resources/common/customize_snapshot.py
+++ b/kura/distrib/src/main/resources/common/customize_snapshot.py
@@ -90,12 +90,12 @@ def main():
         
         interfaces_list = "lo"
     
-        for i, eth_name in enumerate(eth_names[:2]):
+        for i, eth_name in enumerate(eth_names):
             snapshot_content = snapshot_content.replace('ETH_INTERFACE_' + str(i), eth_name)
             interfaces_list += "," + eth_name
             logging.info("%s : replaced ETH_INTERFACE_%s with %s", SNAPSHOT_FILENAME, str(i), eth_name)
     
-        for i, wlan_name in enumerate(wlan_names[:1]):
+        for i, wlan_name in enumerate(wlan_names):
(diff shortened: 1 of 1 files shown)


kura/distrib/src/main/resources/common/customize_snapshot.py
```diff
@@ -90,12 +90,12 @@ def main():
         
         interfaces_list = "lo"
     
-        for i, eth_name in enumerate(eth_names[:2]):
+        for i, eth_name in enumerate(eth_names):
             snapshot_content = snapshot_content.replace('ETH_INTERFACE_' + str(i), eth_name)
```
mattdibi, line 94: I would try to keep this more explicit since, in the future, we don't know whether the template will change. Therefore I would suggest something like:

```suggestion
        for i, eth_name in enumerate(eth_names):
            if i < 2: # only replace the template content with the first two network interfaces
                snapshot_content = snapshot_content.replace('ETH_INTERFACE_' + str(i), eth_name)
```

kura/distrib/src/main/resources/common/customize_snapshot.py
```diff
@@ -90,12 +90,12 @@ def main():
...
             interfaces_list += "," + eth_name
             logging.info("%s : replaced ETH_INTERFACE_%s with %s", SNAPSHOT_FILENAME, str(i), eth_name)
     
-        for i, wlan_name in enumerate(wlan_names[:1]):
+        for i, wlan_name in enumerate(wlan_names):
             snapshot_content = snapshot_content.replace('WIFI_INTERFACE_' + str(i), wlan_name)
```
mattdibi, line 99: Same thing here:

```suggestion
        for i, wlan_name in enumerate(wlan_names):
            if i < 1: # only replace the template content with the first network interface
                snapshot_content = snapshot_content.replace('WIFI_INTERFACE_' + str(i), wlan_name)
```

mattdibi (approved):

PR #5529, opened 2024-11-07
fix: Use dynamic policy for PositionProvider reference in PositionService

nicolatimeus:

Changes the policy of the `PositionProvider` reference in the `PositionService` from static to dynamic.
Having the policy set as static causes the PositionService and all services that depend on it with a static reference (e.g. the `CloudService`) to be deactivated and reactivated whenever a `PositionProvider` is registered or unregistered, causing unnecessary service reactivations on startup and shutdown.
The fact that the CloudService is reactivated may cause the pubblication of spurious lifecycle messages.

Also fixes possible NPE thrown by `getPosition()`, `getNmeaPosition()` and `isLocked()` if called when the configured provider is not bound.

### DIFF
--- a/kura/org.eclipse.kura.linux.position/OSGI-INF/position.xml
+++ b/kura/org.eclipse.kura.linux.position/OSGI-INF/position.xml
@@ -20,8 +20,8 @@
       <provide interface="org.eclipse.kura.configuration.ConfigurableComponent"/>
    </service>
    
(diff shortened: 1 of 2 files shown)


mattdibi (approved):

mattdibi:
Tested in combination with: #5544. Startup time with active cloud connection went from 140 seconds to 53 seconds

PR #5531, opened 2024-11-08
fix(nm): Removed check of net.interfaces property

pierantoniomerlino:
This PR removes the check of the content of the `net.interfaces` property when a new configuration is applied. Moreover, the webUI behavior has been changed to take into account the fact that a network interface cannot be listed in that property.


mattdibi:
Since this PR is quite invasive (removing a snapshot property that's being there for... as long as the project exists) and we're in the middle of the release process, we decided to opt for a less invasive fix of the issue for the time being.

The current issue is given by the fact that, due to how the Kura snapshot works, we cannot configure a network interface via Rest if it is not already in the snapshot (while via UI we can).

This PR addresses the issue at snapshot level, we'll create a PR that addresses the issue at Rest API level. After this release we'll resurrect this PR.

To work on the new PR we'll need to revert the already contributed changes in #5522

PR #5532, opened 2024-11-08
fix: Ignore modem state when getting position from MM

nicolatimeus:

The ModemManager position provider currently tries to get the location only from modems that are in the MM_MODEM_STATE_CONNECTED or MM_MODEM_STATE_REGISTERED state. This prevents to be able to get the location at least in the following cases:

* Modem GPS is enabled in Kura web UI but IPv4 and IPv6 status is set to disabled
* The modem is enabled for wan but it is not registered to a cellular network (e.g. the cellular antenna is not connected or cellular signal level is low) 

This PR removes the filtering based on modem state to decouple the ability to get the location information from the state of the cellular connection.

### DIFF
--- a/kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/NMDbusConnector.java
+++ b/kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/NMDbusConnector.java
@@ -33,7 +33,6 @@ import org.eclipse.kura.net.status.NetworkInterfaceStatus;
 import org.eclipse.kura.net.wifi.WifiChannel;
 import org.eclipse.kura.nm.configuration.NMSettingsConverter;
 import org.eclipse.kura.nm.enums.MMModemLocationSource;
-import org.eclipse.kura.nm.enums.MMModemState;
 import org.eclipse.kura.nm.enums.NMDeviceState;
 import org.eclipse.kura.nm.enums.NMDeviceType;
 import org.eclipse.kura.nm.signal.handlers.DeviceCreationLock;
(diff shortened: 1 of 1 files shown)


mattdibi:
Tested on DG1014

Enabled GPS mode "managed GPS" via ESF UI resulting in:

image

```
root@dynagate-10-14:/home/guest# mmcli -m 0 --location-status
  ------------------------
  Location | capabilities: 3gpp-lac-ci, gps-raw, gps-nmea, agps-msa, agps-msb
           |      enabled: gps-raw, gps-nmea
           |      signals: no
  ------------------------
  GPS      | refresh rate: 30 seconds
```

```
root@dynagate-10-14:/home/guest# mmcli -m 0 --location-get
  ----------------
  GPS |      nmea: $GPGSA,[REDACTED]
      |            $GPRMC,[REDACTED]
      |            $GPGSV,[REDACTED]
      |            $GPGSV,[REDACTED]
      |            $GPGSV,[REDACTED]
      |            $GPGSV,[REDACTED]
      |            $GPVTG,[REDACTED]
      |            $GPGGA,[REDACTED]
      |       utc: 152158.00
      | longitude: [REDACTED]
      |  latitude: [REDACTED]
      |  altitude: [REDACTED]
```

Once enabled PositionService using `modemManager` provider I can correctly query position via Rest API:

image

```
curl -k -u admin:eurotech https:/172.16.0.1/services/position/v1/position | jq .
{
  "longitude": [REDACTED],
  "latitude": [REDACTED],
  "altitude": [REDACTED],
  "speed": 0.0,
  "track": 260.1,
  "gnssType": [
    "Gps"
  ]
}
```

During all of this the modem connection was disabled:

image

image

---

Before this PR, the same call would result in:

```
curl -k -u admin:eurotech https:/172.16.0.1/services/position/v1/position
{"message":"Service unavailable. Position is not locked."}%  
```

mattdibi (approved):

PR #5535, opened 2024-11-08
fix(rest.network.configuration): Fix net interfaces property management again

pierantoniomerlino:
Since the #5531 has been temporary dropped, we decided to adopt a easier approach to fix the issue about network rest configuration.

The fix is limited to the `org.eclipse.kura.internal.rest.network.configuration.NetworkConfigurationRestService` class.
Now, when an update request is received, the code parses the property names and extracts the network interface names.
They are compared with the ones in the `net.interfaces` property from the current configuration and, it they are not listed there, they are added.

In this way the framework is aware if the new network interface and it can handle it correctly.

**Any side note on the changes made:** The `net.interface.<interface-name>.type` property is now mandatory, otherwise the value on the snapshot will be `UNKNOWN` and this will cause troubles to the webUI.


kura/org.eclipse.kura.rest.network.configuration.provider/src/main/java/org/eclipse/kura/internal/rest/network/configuration/NetworkConfigurationRestService.java
```diff
@@ -273,4 +276,32 @@ private boolean isNetworkConfigurationPid(String pid) {
...
+                    .getConfigurationProperties();
+            String netInterfaces = (String) properties.get("net.interfaces");
+            List<String> netInterfacesList = Arrays.asList(netInterfaces.split(","));
+            for (String name : interfaceNames) {
+                if (!netInterfacesList.contains(name)) {
+                    netInterfaces += "," + name;
+                }
+            }
+            configurationProperties.put("net.interfaces", netInterfaces);
+        }
```
mattdibi, line 298: You know I feel about early returns 😆 

```suggestion
        if (!pid.equals(NETWORK_CONFIGURATION_SERVICE_PID)) {
            return;
        }
        Set<String> interfaceNames = new HashSet<>();
        configurationProperties.keySet().forEach(key -> {
            Optional<String> interfaceName = parseInterfaceName(key);
            interfaceName.ifPresent(interfaceNames::add);
        });
        Map<String, Object> properties = this.configurationService.getComponentConfiguration(pid)
                .getConfigurationProperties();
        String netInterfaces = (String) properties.get("net.interfaces");
        List<String> netInterfacesList = Arrays.asList(netInterfaces.split(","));
        for (String name : interfaceNames) {
            if (!netInterfacesList.contains(name)) {
                netInterfaces += "," + name;
            }
        }
        configurationProperties.put("net.interfaces", netInterfaces);
```

... as always this suggestion might be subjective and therefore is not a blocker for merging.

mattdibi: Or... you could bring that out to the outer caller:

```java
if (pid.equals(NETWORK_CONFIGURATION_SERVICE_PID)) {
    updateNetInterfaces(configurationProperties);
}
```

but that `for` is showing already too many indentation levels 🤔

kura/org.eclipse.kura.rest.network.configuration.provider/src/main/java/org/eclipse/kura/internal/rest/network/configuration/NetworkConfigurationRestService.java
```diff
@@ -273,4 +276,32 @@ private boolean isNetworkConfigurationPid(String pid) {
...
+    }
+
+    private Optional<String> parseInterfaceName(String key) {
+        Optional<String> interfaceName = Optional.empty();
+        if (key.startsWith("net.interface.") && key.contains(".config.")) {
+            interfaceName = Optional.of(key.substring(14, key.indexOf(".config.")));
+        }
```
mattdibi: Since we already use Apache Common's StringUtils in our code base, would [`StringUtils.substringBetween`](https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/StringUtils.html#substringBetween(java.lang.String,java.lang.String,java.lang.String)) be better for readability?

mattdibi:
Test performed on DG1014 in combination with #5533 (actually https://github.com/eurotech/kura_eth/pull/2349 in this case) on a freshly installed system.

At startup we have the following:

```
Nov 11 09:32:24 dynagate-10-14 ESF[866]: Update NetworkConfigurationService...
Nov 11 09:32:24 dynagate-10-14 ESF[866]: Applying configuration using NetworkManager Dbus connector
Nov 11 09:32:24 dynagate-10-14 ESF[866]: Device "lo" of type "NM_DEVICE_TYPE_LOOPBACK" with status "UNMANAGED"/"UNMANAGED" currently not supported
Nov 11 09:32:24 dynagate-10-14 ESF[866]: Settings iface "eth0":NM_DEVICE_TYPE_ETHERNET
Nov 11 09:32:25 dynagate-10-14 ESF[866]: Settings iface "eth1":NM_DEVICE_TYPE_ETHERNET
Nov 11 09:32:30 dynagate-10-14 ESF[866]: Device "p2p0" of type "NM_DEVICE_TYPE_WIFI" not configured. Disabling...
Nov 11 09:32:30 dynagate-10-14 ESF[866]: Device "wlan0" of type "NM_DEVICE_TYPE_WIFI" not configured. Disabling...
Nov 11 09:32:30 dynagate-10-14 ESF[866]: Device "3-1.3" of type "NM_DEVICE_TYPE_MODEM" not configured. Disabling...
```

Which is expected since `3-1.3` doesn't have a configuration in the snapshot0.

Then, when running the following RestAPI request:

```bash
curl -k -u admin:[REDACTED] -X PUT \
    --location https://172.16.0.1/services/networkConfiguration/v1/configurableComponents/configurations/_update \
    --header "Content-Type: application/json" \
    --header "Accept: application/json" \
    --data "{
      \"configs\": [
      {
            \"pid\": \"org.eclipse.kura.net.admin.NetworkConfigurationService\",
            \"properties\": {
                \"net.interface.3-1.3.config.ip6.status\": {
                    \"value\": \"netIPv6StatusDisabled\",
                    \"type\": \"STRING\"
                },
                \"net.interface.3-1.3.config.ip4.status\": {
                    \"value\": \"netIPv4StatusEnabledWAN\",
                    \"type\": \"STRING\"
                },
                \"net.interface.3-1.3.config.dhcpClient4.enabled\": {
                    \"value\": true,
                    \"type\": \"BOOLEAN\"
                },
                \"net.interface.3-1.3.config.resetTimeout\": {
                    \"value\": 5,
                    \"type\": \"INTEGER\"
                }
            }
        }
      ]
    }"
```

the configuration is correctly applied:

```
Applying configuration using NetworkManager Dbus connector
Device "lo" of type "NM_DEVICE_TYPE_LOOPBACK" with status "UNMANAGED"/"UNMANAGED" currently not supported
Settings iface "eth0":NM_DEVICE_TYPE_ETHERNET
Ignoring parameter accept-all-mac-addresses [-1]: NetworkManager 1.32 or above is required
Settings iface "eth1":NM_DEVICE_TYPE_ETHERNET
Ignoring parameter accept-all-mac-addresses [-1]: NetworkManager 1.32 or above is required
Device "p2p0" of type "NM_DEVICE_TYPE_WIFI" not configured. Disabling...
Device "wlan0" of type "NM_DEVICE_TYPE_WIFI" not configured. Disabling...
Settings iface "3-1.3":NM_DEVICE_TYPE_MODEM
```

image

---

Performing the same test on our current Kura version will result in:

```
Applying configuration using NetworkManager Dbus connector
Device "lo" of type "NM_DEVICE_TYPE_LOOPBACK" with status "UNMANAGED"/"UNMANAGED" currently not supported
Settings iface "eth0":NM_DEVICE_TYPE_ETHERNET
Ignoring parameter accept-all-mac-addresses [-1]: NetworkManager 1.32 or above is required
Settings iface "eth1":NM_DEVICE_TYPE_ETHERNET
Ignoring parameter accept-all-mac-addresses [-1]: NetworkManager 1.32 or above is required
Device "p2p0" of type "NM_DEVICE_TYPE_WIFI" not configured. Disabling...
Settings iface "wlan0":NM_DEVICE_TYPE_WIFI
Device "3-1.3" of type "NM_DEVICE_TYPE_MODEM" not configured. Disabling...
```

image

kura/org.eclipse.kura.rest.network.configuration.provider/src/main/java/org/eclipse/kura/internal/rest/network/configuration/NetworkConfigurationRestService.java
```diff
@@ -273,4 +277,32 @@ private boolean isNetworkConfigurationPid(String pid) {
...
+            }
+            configurationProperties.put("net.interfaces", netInterfaces);
+        }
+    }
+
+    private Optional<String> parseInterfaceName(String key) {
+        Optional<String> interfaceName = Optional.empty();
+        if (key.startsWith("net.interface.") && key.contains(".config.")) {
+            interfaceName = Optional.of(StringUtils.substringBetween(key, "net.interface.", ".config."));
+        }
```
mattdibi: I think you can avoid the `if` statement by using the `Optional.ofNullable`

```suggestion
        interfaceName = Optional.ofNullable(StringUtils.substringBetween(key, "net.interface.", ".config."));
```

kura/org.eclipse.kura.rest.network.configuration.provider/src/main/java/org/eclipse/kura/internal/rest/network/configuration/NetworkConfigurationRestService.java
```diff
@@ -273,4 +277,30 @@ private boolean isNetworkConfigurationPid(String pid) {
...
+            }
+            configurationProperties.put("net.interfaces", netInterfaces);
+        }
+    }
+
+    private Optional<String> parseInterfaceName(String key) {
+        Optional<String> interfaceName = Optional.empty();
+        interfaceName = Optional.ofNullable(StringUtils.substringBetween(key, "net.interface.", ".config."));
+        return interfaceName;
+    }
```
mattdibi: Ok I'm being annoying right now but... the variable assignement is not needed anymore at this point:

```suggestion
    private Optional<String> parseInterfaceName(String key) {
        return Optional.ofNullable(StringUtils.substringBetween(key, "net.interface.", ".config."));
    }
```

mattdibi (approved):

PR #5536, opened 2024-11-08
feat: pure maven build

chrisdutz:


mattdibi:
As per our discussion in #5521

> All the dependencies are specified in the target platform, there is no need to duplicate them in the projects. Moreover, some internal package exports will not be found on maven central. This is how OSGi works and this is the reason why you should compile with maven plugins like Eclipse Tycho and bndtools.

Closing

PR #5578, opened 2024-11-21
feat(linux.net): DHCP server selection method

pierantoniomerlino:
This PR updates the method used to check if `dnsmasq` is available on the system.

**Description of the solution adopted:** Since the DHCP server manager uses the Systemd unit for `dnsmasq`, the method used to check the presence of the tool on the system has to check the existence of the unit and the command. This PR adds this check.

The check runs this command: `systemctl status dnsmasq`. Then the exit code is checked, based on the [official documentation](https://www.man7.org/linux/man-pages/man1/systemctl.1.html#EXIT_STATUS). Since the `DhcpServerManager` class uses a static method to check the tools, the method in `LinuxNetworkUtil` has to be static too.
This implies that the `CommandExecutorService` cannot be used since we haven't a reference to it. So, the method uses the standard `ProcessBuilder` class to run the command. 

I don't like it.

In addition to the test for this specific feature, this PR adds few test on the `getTool` method.

### DIFF
--- a/kura/org.eclipse.kura.linux.net/src/main/java/org/eclipse/kura/linux/net/dhcp/server/DnsmasqTool.java
+++ b/kura/org.eclipse.kura.linux.net/src/main/java/org/eclipse/kura/linux/net/dhcp/server/DnsmasqTool.java
@@ -42,10 +42,10 @@ public class DnsmasqTool implements DhcpLinuxTool {
     private String globalConfigFilename = "/etc/dnsmasq.d/dnsmasq-globals.conf";
     private static final String GLOBAL_CONFIGURATION = "port=0\nbind-dynamic\n";
 
-    static final Command IS_ACTIVE_COMMAND = new Command(new String[] { "systemctl", "is-active", "--quiet",
-            DhcpServerTool.DNSMASQ.getValue() });
-    static final Command RESTART_COMMAND = new Command(new String[] { "systemctl", "restart",
-            DhcpServerTool.DNSMASQ.getValue() });
+    static final Command IS_ACTIVE_COMMAND = new Command(
+            new String[] { "systemctl", "is-active", "--quiet", DhcpServerTool.DNSMASQ.getValue() });
+    static final Command RESTART_COMMAND = new Command(
+            new String[] { "systemctl", "restart", DhcpServerTool.DNSMASQ.getValue() });
 
     private CommandExecutorService executorService;
     private Map<String, byte[]> configsLastHash = Collections.synchronizedMap(new HashMap<>());
@@ -88,6 +88,7 @@ public class DnsmasqTool implements DhcpLinuxTool {
         CommandStatus restartStatus = this.executorService.execute(RESTART_COMMAND);
 
         if (!restartStatus.getExitStatus().isSuccessful()) {
+            logger.error("dnsmasq Systemd unit startup failed. Is dnsmasq package installed on the system?");
             removeInterfaceConfig(interfaceName);
             restartStatus = this.executorService.execute(RESTART_COMMAND);
         }
--- a/kura/org.eclipse.kura.linux.net/src/main/java/org/eclipse/kura/linux/net/dhcp/server/DnsmasqTool.java
+++ b/kura/org.eclipse.kura.linux.net/src/main/java/org/eclipse/kura/linux/net/dhcp/server/DnsmasqTool.java
@@ -12,6 +12,7 @@
  *******************************************************************************/
 package org.eclipse.kura.linux.net.dhcp.server;
 
+import java.io.ByteArrayOutputStream;
 import java.io.File;
 import java.io.IOException;
 import java.nio.charset.StandardCharsets;
@@ -42,10 +43,10 @@ public class DnsmasqTool implements DhcpLinuxTool {
     private String globalConfigFilename = "/etc/dnsmasq.d/dnsmasq-globals.conf";
     private static final String GLOBAL_CONFIGURATION = "port=0\nbind-dynamic\n";
 
-    static final Command IS_ACTIVE_COMMAND = new Command(
-            new String[] { "systemctl", "is-active", "--quiet", DhcpServerTool.DNSMASQ.getValue() });
-    static final Command RESTART_COMMAND = new Command(
-            new String[] { "systemctl", "restart", DhcpServerTool.DNSMASQ.getValue() });
+    static final String[] IS_ACTIVE_COMMANDLINE = new String[] { "systemctl", "is-active", "--quiet",
+            DhcpServerTool.DNSMASQ.getValue() };
+    static final String[] RESTART_COMMANDLINE = new String[] { "systemctl", "restart",
+            DhcpServerTool.DNSMASQ.getValue() };
 
     private CommandExecutorService executorService;
     private Map<String, byte[]> configsLastHash = Collections.synchronizedMap(new HashMap<>());
@@ -57,7 +58,7 @@ public class DnsmasqTool implements DhcpLinuxTool {
 
     @Override
     public boolean isRunning(String interfaceName) throws KuraProcessExecutionErrorException {
-        CommandStatus status = this.executorService.execute(IS_ACTIVE_COMMAND);
+        CommandStatus status = this.executorService.execute(new Command(IS_ACTIVE_COMMANDLINE));
 
         boolean isRunning;
         try {
@@ -85,12 +86,19 @@ public class DnsmasqTool implements DhcpLinuxTool {
                     "Failed to start DHCP server for interface: " + interfaceName);
         }
 
-        CommandStatus restartStatus = this.executorService.execute(RESTART_COMMAND);
+        Command restartCommand = new Command(RESTART_COMMANDLINE);
+        ByteArrayOutputStream out = new ByteArrayOutputStream();
+        ByteArrayOutputStream err = new ByteArrayOutputStream();
+        restartCommand.setErrorStream(err);
+        restartCommand.setOutputStream(out);
+        CommandStatus restartStatus = this.executorService.execute(restartCommand);
 
         if (!restartStatus.getExitStatus().isSuccessful()) {
-            logger.error("dnsmasq Systemd unit startup failed. Is dnsmasq package installed on the system?");
+            logger.error("dnsmasq Systemd unit startup failed. Check if the tool is properly installed on the system.");
+            logger.error("dnsmasq stderr {}", new String(err.toByteArray(), StandardCharsets.UTF_8));
+            logger.error("dnsmasq stdout {}", new String(out.toByteArray(), StandardCharsets.UTF_8));
             removeInterfaceConfig(interfaceName);
-            restartStatus = this.executorService.execute(RESTART_COMMAND);
+            restartStatus = this.executorService.execute(restartCommand);
         }
 
         return restartStatus;
@@ -101,7 +109,7 @@ public class DnsmasqTool implements DhcpLinuxTool {
         boolean isInterfaceDisabled = true;
 
         if (removeInterfaceConfig(interfaceName)) {
-            CommandStatus status = this.executorService.execute(RESTART_COMMAND);
+            CommandStatus status = this.executorService.execute(new Command(RESTART_COMMANDLINE));
             isInterfaceDisabled = status.getExitStatus().isSuccessful();
         }
 
--- a/kura/org.eclipse.kura.linux.net/src/main/java/org/eclipse/kura/linux/net/util/LinuxNetworkUtil.java
+++ b/kura/org.eclipse.kura.linux.net/src/main/java/org/eclipse/kura/linux/net/util/LinuxNetworkUtil.java
@@ -273,8 +273,8 @@ public class LinuxNetworkUtil {
             return true;
         } else {
             for (String folder : DEFAULT_PATH) {
-                File fUnit = new File(folder + tool);
-                if (fUnit.exists()) {
+                File fTool = new File(folder + tool);
+                if (fTool.exists()) {
                     TOOLS.add(tool);
                     return true;
                 }
--- a/kura/org.eclipse.kura.linux.net/src/main/java/org/eclipse/kura/linux/net/util/LinuxNetworkUtil.java
+++ b/kura/org.eclipse.kura.linux.net/src/main/java/org/eclipse/kura/linux/net/util/LinuxNetworkUtil.java
@@ -17,6 +17,8 @@ import java.io.ByteArrayOutputStream;
 import java.io.File;
 import java.io.FileReader;
 import java.io.IOException;
+import java.io.InputStream;
+import java.io.InputStreamReader;
 import java.net.UnknownHostException;
 import java.util.ArrayList;
 import java.util.Collection;
@@ -26,6 +28,13 @@ import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
 import java.util.StringTokenizer;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+import java.util.function.Consumer;
 
 import org.apache.commons.io.Charsets;
 import org.eclipse.kura.KuraErrorCode;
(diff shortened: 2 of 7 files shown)


kura/org.eclipse.kura.linux.net/src/main/java/org/eclipse/kura/linux/net/util/LinuxNetworkUtil.java
```diff
@@ -1124,7 +1152,7 @@ public void createApNetworkInterface(String ifaceName, String dedicatedApInterfa
             return;
         }
 
-        CommandStatus status = this.executeCommand(formIwDevIfaceInterfaceAddAp(ifaceName, dedicatedApInterface));
+        CommandStatus status = executeCommand(formIwDevIfaceInterfaceAddAp(ifaceName, dedicatedApInterface));
```
mattdibi, line 1160: Why did you remove `this`? Shouldn't our Kura cleanup profile enforce the use of `this` for member methods? Does it change behaviour in case of static classes?

kura/org.eclipse.kura.linux.net/src/main/java/org/eclipse/kura/linux/net/util/LinuxNetworkUtil.java
```diff
@@ -65,13 +71,12 @@ public class LinuxNetworkUtil {
...
     private static final String IP = "ip";
-
     private static final String LINE_MSG = "line: {}";
-
     private static final String ERR_EXECUTING_CMD_MSG = "error executing command --- {} --- exit value={}";
-
     private static final String FAKE_MAC_ADDRESS = "12:34:56:78:ab:cd";
-
+    private static final String ETHTOOL_COMMAND = "ethtool";
+    private static final String[] DEFAULT_PATH = new String[] { "/sbin/", "/usr/sbin/", "/bin/", "/usr/bin/" };
```
mattdibi, line 69: This is probably out of scope therefore feel free to ignore this comment.

Wouldn't it be better to retrieve these paths with:

```suggestion
    private static final String[] DEFAULT_PATH = System.getenv("PATH").split(File.pathSeparator);
```

kura/org.eclipse.kura.linux.net/src/main/java/org/eclipse/kura/linux/net/util/LinuxNetworkUtil.java
```diff
@@ -264,22 +269,45 @@ private boolean isWifiLinkUpInternal(String ifaceName) throws KuraException {
...
         if (TOOLS.contains(tool)) {
-            ret = true;
+            return true;
         } else {
-            for (String folder : searchFolders) {
-                File fTool = new File(folder + tool);
-                if (fTool.exists()) {
+            for (String folder : DEFAULT_PATH) {
+                File fUnit = new File(folder + tool);
+                if (fUnit.exists()) {
```
mattdibi: Are these `Unit` or `Binaries`?

kura/org.eclipse.kura.linux.net/src/main/java/org/eclipse/kura/linux/net/util/LinuxNetworkUtil.java
pierantoniomerlino: AFAIK, the cleanup enforces the `this.` prefix only for variables, not for methods. Maybe the cleanup fix it?

kura/org.eclipse.kura.linux.net/src/main/java/org/eclipse/kura/linux/net/util/LinuxNetworkUtil.java
```diff
@@ -264,22 +269,45 @@ private boolean isWifiLinkUpInternal(String ifaceName) throws KuraException {
...
+     * Checks if the given Systemd system unit is installed.
+     * The result is based on the exit code of "systemctl status <unitName>"
+     * as presented in https://www.man7.org/linux/man-pages/man1/systemctl.1.html#EXIT_STATUS
+     *
+     * @param unitName
+     *            the name of the Systemd system unit
+     * @return true if the unit is installed
+     */
+    public static boolean systemdSystemUnitExists(String unitName) {
+        PROCESS_BUILDER.command("sh", "-c", "systemctl status " + unitName);
```
mattdibi: I already know there's a reason for these but, just so we are all on the same page and I can learn something, can you explain why can't we use the instance's `CommandExecutorService`?

kura/org.eclipse.kura.linux.net/src/main/java/org/eclipse/kura/linux/net/util/LinuxNetworkUtil.java
pierantoniomerlino: > This is probably out of scope therefore feel free to ignore this comment.

I agree. I don't want to dive in this motification in this PR. Thanks for the comment.

kura/org.eclipse.kura.linux.net/src/main/java/org/eclipse/kura/linux/net/util/LinuxNetworkUtil.java
pierantoniomerlino: Copy-and-paste error. I'll fix it.

kura/org.eclipse.kura.linux.net/src/main/java/org/eclipse/kura/linux/net/util/LinuxNetworkUtil.java
pierantoniomerlino: The [CommandExecutorService](https://github.com/eclipse-kura/kura/blob/develop/kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/executor/CommandExecutorService.java) is an interface that is extended by the [PrivilegedExecutorService](https://github.com/eclipse-kura/kura/blob/develop/kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/executor/PrivilegedExecutorService.java) and the [UnprivilegedExecutorService](https://github.com/eclipse-kura/kura/blob/develop/kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/executor/UnprivilegedExecutorService.java) as marker interfaces. 

They are implemented by the [PrivilegedExecutorServiceImpl](https://github.com/eclipse-kura/kura/blob/develop/kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/linux/executor/privileged/PrivilegedExecutorServiceImpl.java) and [UnprivilegedExecutorServiceImpl](https://github.com/eclipse-kura/kura/blob/develop/kura/org.eclipse.kura.core/src/main/java/org/eclipse/kura/core/linux/executor/unprivileged/UnprivilegedExecutorServiceImpl.java).

The idea behind this classes is to provide a way to run system command using privileged (e.g. kurad) or unprivileged (e.g. kura) linux users. They are both OSGI services and are activated during startup.

The `LinuxNetworkUtil` class is a standard Java class and it is not registered in the OSGI framework. It cannot be bounded or use other services. Indeed, the `CommandExecutorService` is passed as parameter to the constructor, since some methods need it. The static methods cannot use it since the passed reference to the service is not static. Moreover, the service cannot be passed to the static methods of this class, since the caller (e.g. in the `DhcpServermanager`) is static too and it haven't access to the `CommandExecutorService` instantiated in that class.

One way to overcome this is to directly instantiate a `PrivilegedCommandExecutorServiceImpl`, but I refuse to do it.

kura/org.eclipse.kura.linux.net/src/main/java/org/eclipse/kura/linux/net/util/LinuxNetworkUtil.java
mattdibi: > Maybe the cleanup fix it?

That's probably it 👍

kura/org.eclipse.kura.linux.net/src/main/java/org/eclipse/kura/linux/net/util/LinuxNetworkUtil.java
mattdibi: Thanks for taking the time to explain this! 👍

kura/org.eclipse.kura.linux.net/src/main/java/org/eclipse/kura/linux/net/util/LinuxNetworkUtil.java
mattdibi: Addressed in 168582f200f8b7919d89d40c1ea3052aff5f29f0

kura/org.eclipse.kura.linux.net/src/main/java/org/eclipse/kura/linux/net/util/LinuxNetworkUtil.java
```diff
@@ -264,22 +269,66 @@ private boolean isWifiLinkUpInternal(String ifaceName) throws KuraException {
...
                 }
             }
         }
-        return ret;
+        return Optional.empty();
+    }
+
+    private static Optional<String> getCachedTool(String tool) {
+        return TOOLS.stream().filter(item -> {
+            return item.endsWith(tool);
```
marcellorinaldo: We can replace this with:
```java
return TOOLS.stream().filter(item -> item.endsWith(tool)).findFirst();
```

kura/org.eclipse.kura.linux.net/src/main/java/org/eclipse/kura/linux/net/util/LinuxNetworkUtil.java
```diff
@@ -264,22 +269,66 @@ private boolean isWifiLinkUpInternal(String ifaceName) throws KuraException {
...
+            process = PROCESS_BUILDER.start();
+            StreamGobbler streamGobbler = new StreamGobbler(process.getInputStream(), logger::debug);
+            ExecutorService executor = Executors.newSingleThreadExecutor();
+            Future<?> future = executor.submit(streamGobbler);
+
+            int exitCode = process.waitFor();
+            future.get(10, TimeUnit.SECONDS);
+            return exitCode < 4;
+        } catch (IOException | ExecutionException | TimeoutException e) {
+            logger.error(String.format("Cannot check %s unit existence", unitName), e);
```
marcellorinaldo: In general we should avoid computing logging arguments like `String.format`, it is inefficient since they are computed even if that particular log level is disabled.

You can safely write:

```java
logger.error("Cannot check {} unit existence", unitName, e);
```

As of log4j 1.6 (we are on 1.7+), we can parametrize the arguments in the presence of an exception, assuming the exception is the last parameter

https://www.slf4j.org/faq.html#paramException
https://www.slf4j.org/apidocs/org/slf4j/Logger.html

kura/org.eclipse.kura.linux.net/src/main/java/org/eclipse/kura/linux/net/util/LinuxNetworkUtil.java
```diff
@@ -264,22 +269,66 @@ private boolean isWifiLinkUpInternal(String ifaceName) throws KuraException {
...
+        if (!optionalSystemctlPath.isPresent()) {
+            logger.debug("Systemctl command not found in default paths");
+            return false;
+        }
+        PROCESS_BUILDER.command(optionalSystemctlPath.get(), "status", unitName);
+        Process process;
+        try {
+            process = PROCESS_BUILDER.start();
+            StreamGobbler streamGobbler = new StreamGobbler(process.getInputStream(), logger::debug);
+            ExecutorService executor = Executors.newSingleThreadExecutor();
```
marcellorinaldo: We should shutdown this executor once no more used. From [ExecutorService javadoc](https://docs.oracle.com/javase/8/docs/api/index.html?java/util/concurrent/ExecutorService.html):

*An unused ExecutorService should be shut down to allow reclamation of its resources.*

We can remove the need for the Future:
```java
executor.shutdown();
if (!executor.awaitTermination(10, TimeUnit.SECONDS)) {
    executor.shutdownNow();
}
```

kura/org.eclipse.kura.linux.net/src/main/java/org/eclipse/kura/linux/net/util/LinuxNetworkUtil.java
```diff
@@ -264,22 +269,66 @@ private boolean isWifiLinkUpInternal(String ifaceName) throws KuraException {
...
+        Optional<String> optionalSystemctlPath = getToolPath("systemctl");
+        if (!optionalSystemctlPath.isPresent()) {
+            logger.debug("Systemctl command not found in default paths");
+            return false;
+        }
+        PROCESS_BUILDER.command(optionalSystemctlPath.get(), "status", unitName);
+        Process process;
+        try {
+            process = PROCESS_BUILDER.start();
+            StreamGobbler streamGobbler = new StreamGobbler(process.getInputStream(), logger::debug);
```
marcellorinaldo: I think we should call this only if debug level is active

mattdibi:
Tested on RPi and found the following issue.

Starting point: in this setup `isc-dhcp-server`, `dnsmasq` and `dnsmasq-base` are all installed on the system.

#### When `dnsmasq` is removed with `apt remove`

Kura still thinks the unit is "functional" and "available".

image

In Kura logs:

```
2024-11-22T10:48:51,302 [Start Level: Equinox Container: 831d6796-ff4e-4b32-b8e2-7ea0f15b04f7] INFO  o.e.k.l.n.d.DhcpServerManager - Using dnsmasq as DHCP server.
```

which results in client not being able to correctly get an IP address from Kura's access point.

#### When `dnsmasq` is removed with `apt purge`

image

In Kura logs:

```
2024-11-22T11:08:25,565 [Start Level: Equinox Container: 76f5aed4-bdae-40e0-8249-9cac2fb13f8a] INFO  o.e.k.l.n.d.DhcpServerManager - Using dhcpd as DHCP server.
```

#### Conclusion

`systemctl status` return value is not entirely reliable to evaluate whether the systemd unit is installed and functional.

marcellorinaldo:
Tested working on 3 different scenarios:

1. `dnsmasq-base` installed, `dnsmasq` not installed, `/usr/sbin/dnsmasq` binary present, `isc-dhcp-server` installed
2. `dnsmasq-base` installed, `dnsmasq` installed, `/usr/sbin/dnsmasq` binary present, `isc-dhcp-server` installed
3. `dnsmasq-base` installed, `dnsmasq` installed, `/usr/sbin/dnsmasq` binary present, `isc-dhcp-server` not installed

In scenario 1, Kura picked up `dhcpd` as DHCP server tool. In scenarios 2 and 3 `dnsmasq` was selected correctly. In all 3 scenarios the following basic test passes:

- Configure an interface as WAN, configure another one as DHCP server with NAT with passDNS=true
- Connect a client to the DHCP server interface and ping google

marcellorinaldo:
> #### Conclusion
> `systemctl status` return value is not entirely reliable to evaluate whether the systemd unit is installed and functional.

Exit code is 3 when `apt remove`. When purging, exit code is 4. Currently we check that `systemctl status` returns an exit code < 4, we might solve this by changing just that

mattdibi:
> 1. `dnsmasq-base` installed, `dnsmasq` not installed, `/usr/sbin/dnsmasq` binary present, `isc-dhcp-server` installed

3 is returned when `dnsmasq` is installed but stopped. That's why this is not reliable.

kura/org.eclipse.kura.linux.net/src/main/java/org/eclipse/kura/linux/net/util/LinuxNetworkUtil.java
pierantoniomerlino: Indeed, I don't care of the output of the command. We check only the exit code.

mattdibi (dismissed):
Tested with the latest changes: when the condition described in https://github.com/eclipse-kura/kura/pull/5578#issuecomment-2493389565 verifies, we get the following error in the log:

```
2024-11-22T14:18:08,345 [pool-22-thread-1] ERROR o.e.k.l.n.d.s.DnsmasqTool - dnsmasq Systemd unit startup failed. Is dnsmasq package installed on the system?
```

which gives enough information for the user to fix this issue.

Therefore, given that the situation in which
1. The user has `isc-dhcp-server`, `dnsmasq` and `dnsmasq-base` installed.
2. They remove `dnsmasq` without purging
3. Install kura

is quite unlikely and programmatically handling this situation brings diminishing returns, we accept current solution albeit not completely reliable.

We're planning to remove the support for `isc-dhcp-server` in the near future and thus this issue should be even less relevant.

kura/org.eclipse.kura.linux.net/src/main/java/org/eclipse/kura/linux/net/dhcp/server/DnsmasqTool.java
```diff
@@ -88,6 +88,7 @@ public CommandStatus startInterface(String interfaceName) throws KuraProcessExec
         CommandStatus restartStatus = this.executorService.execute(RESTART_COMMAND);
 
         if (!restartStatus.getExitStatus().isSuccessful()) {
+            logger.error("dnsmasq Systemd unit startup failed. Is dnsmasq package installed on the system?");
```
marcellorinaldo: The error here might also be due to another config file that uses an inconsitent setup (like defining the `bind-interfaces` property). I would remove this `Is dnsmasq package installed on the system?` and print the error stream instead

kura/org.eclipse.kura.linux.net/src/main/java/org/eclipse/kura/linux/net/util/LinuxNetworkUtil.java
```diff
@@ -1178,4 +1214,20 @@ private CommandStatus executeCommand(String[] commandString) {
         return this.executorService.execute(command);
     }
 
+    private static class StreamGobbler implements Runnable {
```
marcellorinaldo: We can remove this if it isn't used anymore

kura/org.eclipse.kura.linux.net/src/main/java/org/eclipse/kura/linux/net/dhcp/server/DnsmasqTool.java
pierantoniomerlino: IMO, the message is a good hint to the user about `dnsmasq`. However, it is a good idea to print also the exception.

marcellorinaldo (approved):

mattdibi (approved):

PR #5584, opened 2024-11-25
fix(asset.provider): Added integer and float as possible Scale and Offset type

salvatore-coppola:
Added integer and float as possible Scale and Offset type


kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/channel/Channel.java
```diff
@@ -121,30 +91,28 @@ public Channel(final String name, final ChannelType type, final DataType valueTy
...
      * @param config
      *            the configuration
      * @throws NullPointerException
      *             if any of the arguments is null
      * @throws IllegalArgumentException
      *             if any of the valueScale and valueOffset have different types
      *
-     * @since 2.8
+     * @since 3.0
      */
```
pierantoniomerlino: Since this class has been released with Kura 5.6.0, I don't think it is possible to change now this method. Probabily we need to deprecate it and create a new one.

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/channel/Channel.java
```diff
@@ -215,55 +183,27 @@ public boolean isEnabled() {
...
-
     /**
      * Returns a {@link Number} that represents the scale factor to be applied to the read
      * value
      *
      * @return a {@link Number} that represents the scale factor to be applied to the read value
      *
-     * @since 2.8
+     * @since 3.0
      */
```
pierantoniomerlino: Same here.

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/channel/Channel.java
```diff
@@ -215,55 +183,27 @@ public boolean isEnabled() {
...
     /**
      * Returns a {@link TypedValue} that represents the offset factor to be applied to the read
      * value
      *
      * @return a {@link TypedValue} that represents the offset factor to be applied to the read value
      *
-     * @since 2.8
+     * @since 3.0
      */
-    public Number getValueOffsetAsNumber() {
```
pierantoniomerlino: Same here.

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/channel/Channel.java
salvatore-coppola: we target kura 6, we are allowed to do that right ?

nicolatimeus: We are allowed to do that only for the methods that have been deprecated in a previous release

nicolatimeus: In my opinion we could return 1 as scale and 0 as offset instead of null, if they are not specified so that it is not necessary to change the APIs. Not specifyng the scale or offset is equivalent to specifing 1 and 0 for them right?

salvatore-coppola: i can restore the previous api and deprecate them, i think have null value is better (not much better but more correct)

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/channel/Channel.java
```diff
@@ -339,17 +280,6 @@ public void setEnabled(boolean isEnabled) {
...
-     * Specifies the scale to be applied to the channel value
-     *
-     * @param scale
-     *            a double value that specifies the scale to be applied to the channel value
-     * @since 2.3
-     */
-    public void setScale(double scale) {
-        this.valueScale = scale;
-    }
-
```
pierantoniomerlino: I don't think you can delete this method without a deprecation.

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/channel/Channel.java
```diff
@@ -361,17 +291,6 @@ public void setScale(Number scale) {
...
-     * Specifies the offset to be applied to the channel value
-     *
-     * @param offset
-     *            a double value that specifies the offset to be applied to the channel value
-     * @since 2.3
-     */
-    public void setOffset(double offset) {
-        this.valueOffset = offset;
-    }
-
```
pierantoniomerlino: I don't think you can delete this method without a deprecation.

kura/org.eclipse.kura.asset.provider/src/main/java/org/eclipse/kura/internal/asset/provider/BaseAssetConfiguration.java
```diff
@@ -363,7 +363,8 @@ private static String getUnit(final Map<String, Object> properties) {
 
         private static boolean isEnabled(final Map<String, Object> properties) {
             try {
-                return Boolean.parseBoolean(properties.get(ENABLED.value()).toString());
+                Object enableProp = properties.get(ENABLED.value());
+                return enableProp == null || Boolean.parseBoolean(enableProp.toString());
```
pierantoniomerlino: Just to be sure: is it correct that if the `enableProp` is null, the method has to return true? The old code returns false in this case.

salvatore-coppola: The old code returned true, because the exception was caught and then there is a return true

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/channel/Channel.java
salvatore-coppola: I forgot to deprecate them in the past release, we can mantain them but i have to throw an exception in case scaleoffset is different type

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/channel/Channel.java
salvatore-coppola: I forgot to deprecate them in the past release, we can mantain them but i have to throw an exception in case scaleoffset is different type

salvatore-coppola: I changed my mind, since the class is not immutable a client could change first the scale and offset and then the value type and the scale/offset values. So I only restored methods and deprecated them.

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/channel/Channel.java
salvatore-coppola: I changed my mind, since the class is not immutable a client could change first the scale and offset and then the value type and the scale/offset values. So I only restored methods and deprecated them.

pierantoniomerlino (approved):

PR #5586, opened 2024-11-26
feat: WPA3 personal support

pierantoniomerlino:
This PR adds the **experimental** support for the WPA Personal security in WiFi connections, both for Station and AP mode.

**Description of the solution adopted:** The PR removes the filtering of WPA3 hotspots, so the user can perform a scan from the webUI and select such types of networks. The supported securities are WPA3 and WPA2/WPA3 Personal.
As usual, after the network selection, the Wireless tab is filled accordingly.
The AP mode is supported as well. So, the user can configure Kura to implement an Access Point with WPA3 Personal security.
The code configured the proper properties to NetworkManager using the DBus APIs.

The feature is marked as **experimental** (and the tooltip on the webUI warns the user about it) because on several devices (like the RPI) it is not fully supported [1].

WPA3 allows passwords of any length, instead of WPA2 that limits to 8-63 chars. Since this is dependant on the underlying implementation, I preferred to leave this as is. For example, it seems that the [Linux driver](https://stackoverflow.com/questions/72311297/whats-the-minimum-and-maximum-length-of-wpa3-personal-password) limit the password length to 128bytes.

[1] On the RPI5, with the latest wifi module firmware, the station mode will probably work. The Access Point mode is not currently supported.


MMaiero:
@pierantoniomerlino Is this ready for review or still a draft?

MMaiero (dismissed):
Tested with Raspberry PI 3. Direct wpa3 AP not possible, but compatibility mode tested.

kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/KuraWifiSecurityType.java
```diff
@@ -18,29 +18,35 @@ public enum KuraWifiSecurityType {
...
+            case "SECURITY_WPA3":
+                return KuraWifiSecurityType.SECURITY_WPA3;
+            case "SECURITY_WPA2_WPA3_ENTERPRISE":
+                return KuraWifiSecurityType.SECURITY_WPA2_WPA3_ENTERPRISE;
+            case "SECURITY_WPA_WPA2":
+                return KuraWifiSecurityType.SECURITY_WPA_WPA2;
+            case "SECURITY_WPA2_WPA3":
+                return KuraWifiSecurityType.SECURITY_WPA2_WPA3;
+            default:
+                throw new IllegalArgumentException("Invalid security type: " + securityType);
```
mattdibi, line 49: Indentation

kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/shared/model/GwtWifiSecurity.java
```diff
@@ -19,7 +19,9 @@ public enum GwtWifiSecurity {
     netWifiSecurityWPA("WPA"),
     netWifiSecurityWPA2("WPA2"),
     netWifiSecurityWPA2WPA3Enterprise("WPA2/WPA3-Enterprise"),
```
mattdibi, line 21: Probably out of scope for this PR but we should probably not call this WPA2/WPA3 enterprise....

```suggestion
    netWifiSecurityWPA2Enterprise("WPA2-Enterprise"),
```

kura/org.eclipse.kura.web2/src/main/resources/org/eclipse/kura/web/client/messages/Messages.properties
```diff
@@ -621,7 +621,7 @@ netWifiToolTipWirelessModeAdhoc=Configure the wireless interface in adhoc mode.
...
-netWifiToolTipSecurity=Select the appropriate security protocol for this wireless network.<br>WPA2 is the most secure encryption method available for wireless networks - we recommend using WPA2 with the CCMP cipher whenever possible. WPA2 with CCMP is the only option permitted for high throughput 802.11n transmissions.<br>If you need to accommodate legacy devices with an SSID, enable WPA encryption with the TKIP cipher. 
+netWifiToolTipSecurity=Select the appropriate security protocol for this wireless network.<br>WPA2 is the most secure encryption method available for wireless networks - we recommend using WPA2 with the CCMP cipher whenever possible. WPA2 with CCMP is the only option permitted for high throughput 802.11n transmissions.<br>If you need to accommodate legacy devices with an SSID, enable WPA encryption with the TKIP cipher.<br><br>The WPA3-SAE security protocol is experimental. Please check if the device, driver and firmware are compliant with the new standard.
```
mattdibi: I would rephrase this as follows. Let me know what you think

```suggestion
netWifiToolTipSecurity=Select the appropriate security protocol for this wireless network.<br>WPA2 is the most secure encryption method available for wireless networks - we recommend using WPA2 with the CCMP cipher whenever possible. WPA2 with CCMP is the only option permitted for high throughput 802.11n transmissions.<br>If you need to accommodate legacy devices with an SSID, enable WPA encryption with the TKIP cipher.<br><br>The WPA3-SAE security protocol support is experimental. Please check whether the new standard is supported by the Wifi module device, driver and firmware. 
```

kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/shared/model/GwtWifiSecurity.java
pierantoniomerlino: Yeah... but this will break a lot of things... So, I decided not to do this modification for now.

mattdibi:
Tested on RPi4. I created an ad-hoc WPA3 Personal endpoint for testing. The RPi4 was not able to connect even when issuing the command via `nmcli`...

```bash
root@raspberrypi:/home/pi# nmcli dev wifi connect WPA3Testing password passwordForTesting ifname wlan0
Error: Connection activation failed: Secrets were required, but not provided.
```

after power cycling the RPi4 it was finally able to connect via command line.

```bash
root@raspberrypi:/home/pi# nmcli dev wifi connect WPA3Testing password passwordForTesting ifname wlan0
Device 'wlan0' successfully activated with '98aeec5c-4d04-4d37-88e0-38052140ee99'.
```

Unfortunately Kura doesn't seem to be happy about what's going on under the hood

image

Even after a reinstall, when trying to connect to a WPA3 endpoint Kura gets stuck in this "Roll-back changes" loop

mattdibi:
@MMaiero @pierantoniomerlino to me the code is ok... everything else outside our control isn't... are we sure we want to merge this?

pierantoniomerlino:
@mattdibi Thanks for the feedback. With @MMaiero we decided to contribute the feature, but to hide the WPA3 entry in the webUI. It will be available only setting a property in the `kura.properties` file. This will be documented, so a user will be able to test the WPA3 connection on his device. When the support for WPA3 will be more reliable, we'll remove the property and make the feature available by default.

mattdibi:
> @mattdibi Thanks for the feedback. With @MMaiero we decided to contribute the feature, but to hide the WPA3 entry in the webUI. It will be available only setting a property in the `kura.properties` file. This will be documented, so a user will be able to test the WPA3 connection on his device. When the support for WPA3 will be more reliable, we'll remove the property and make the feature available by default.

Oh ok... this is a good compromise 👍

mattdibi (approved):

PR #5478, opened 2024-10-22
build!: Updated to Equinox SDK 4.33, Jetty 12.0.14, Jersey 3.1.9, added jakartarest-osgi 1.2.3

salvatore-coppola:
- Updated equinox sdk to 4.33
- Updated jetty to 12.0.14
- Moved to jetty.ee10
- Moved to Jersey 3.1.9
- Switched to Felix implementation of _Http Service Specification_,  _Http Whiteboard Specification_ and _Whiteboard Specification for Jakarta™ Servlet_ OSGi specs
- Replaced com.eclipsesource:osgi-jaxrs-connector with org.eclipse.osgi-technology.rest, implementing the _Whiteboard Specification for Jakarta™ RESTful Web Services_ OSGi spec
- Removed KuraJettyCustomizer
- Added new Jetty launcher
- Moved from javax namespace to jakarta
- Moved to whiteboard pattern to register servlets and restful web services in WebConsole
- Removed Equinox SDK zip download step
- Removed mToolkit

PR #5602, opened 2024-12-03
ci: switch to Java 17 for CI builds

MMaiero:
This PR switches to Java17 for CI builds

This PR additionally performs the following:
1. Adds the Surefire rerun policy to our CI build.
2. Disable failing tests on Java17. See list below for details.
3. Removes the DP packages from the `buildAll` target
4. Updates our main README to only refer to Java 17

#### Details
 
This PR started off as a simple update to the `manve-surefire-plugin` rerun policy. We later found out that we need to switch to Java 17 for it to work properly and decided to do so since we needed it anyway (see: #5478)

Then we found out that some tests fails in the Java17 profile which prompted us to disable them and address the issue at a later date (see: https://eurotech.atlassian.net/browse/ECESF-8549)

Finally, we removed all `.dp` from the `buildAll` target since we cannot generate `.dp` packages with Java17 and the `distrib` build would fail since it was looking for them (see https://eurotech.atlassian.net/browse/ECESF-3770)

#### Disabled tests

1. DeploymentAgentTest
7. LinuxNetworkUtilTest
8. ~NMDbusConnectorTest~ fixed in 4481dd471abea19ef7252e737887d8723dbc24e5
9. PackagesRestServiceTest

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -38,7 +38,7 @@ node {
     stage('Build') {
         timeout(time: 2, unit: 'HOURS') {
             dir("kura") {
-                withMaven(jdk: 'adoptopenjdk-hotspot-jdk8-latest', maven: 'apache-maven-3.9.6') {
+                withMaven(jdk: 'temurin-jdk17-latest', maven: 'apache-maven-3.9.6') {
                     sh "touch /tmp/isJenkins.txt"
                     sh "mvn -f target-platform/pom.xml clean install -Pno-mirror -Pcheck-exists-plugin"
                     sh "mvn -f kura/pom.xml -Dsurefire.rerunFailingTestsCount=3 clean install -Pcheck-exists-plugin"
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -25,6 +25,7 @@ node {
     stage('Preparation') {
         dir("kura") {
             checkout scm
+            sh "touch /tmp/isJenkins.txt"
         }
     }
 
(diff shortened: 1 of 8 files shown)


mattdibi (changes requested):
This doesn't seem to be working 🤔 

For some reason, even on `develop`, the test reports are duplicated... but this is a separate issue.

Looking at the test reports I can't see the failing tests being re-run (apart from the duplication mentioned above).

mattdibi:
Tested locally on a modified **unit** test and it seems to be working correctly:

image

image

It doesn't seem to be working correctly for **integration** tests... some further investigation is needed

MMaiero:
There it could be on the java version used that dictate the tycho plugin used.
If building java 8, it is using an old version of the plugin that it may have no support for such property

mattdibi:
@MMaiero 🤔  in the CI we're building with Java8, see: https://github.com/eclipse-kura/kura/blob/27b8caa7a9f2551be6a8fb1601e7e3c06ec6f14b/Jenkinsfile#L21

mattdibi:
@MMaiero I tested it locally on a Java17 installation. It does work simply by adding `-Dsurefire.rerunFailingTestsCount` in the test run command.

How do we want to proceed now?
1. Personally I would prefer enabling the `rerun` in the CLI command rather than modifying the POM. During development I want my tests to fail fast and not be re-ran since they might simply be broken because we're developing them. Having the `rerun` in the CLI command is much more flexible. What do you think?
2. Should we switch to Java17 for CI builds? It'll become a necessary shortly with #5478 right?

MMaiero:
Yes to both. Le't's use the command line instead of the configuration in pom files.
Let's start the switch to java 17 as baseline

mattdibi:
Good news is that it's working:

image

Bad news is that there are broken tests since we are not running them regularly and we'll need to fix them...

mattdibi:
For documentation purposes: we decided to disable the failing tests for now and fix them at a later date.

mattdibi (dismissed):

mattdibi (approved):

marcellorinaldo:
Do not merge yet, have to check something...

marcellorinaldo (approved):
Looks good to me

mattdibi:
@salvatore-coppola if you don't see any issues I would proceed with merging this

PR #5603, opened 2024-12-04
feat: added MTU limits in webUI

sfiorani:
This PR updates the checks made in the web UI for the MTU values set in the Advanced Tab of a network interface.

The old limits just checked that the input was a valid integer (0 to MaxInteger), while now they limit the range from the minimum valid MTU value (576 for IPv4 and 1280 for IPv6) to the max value accepted which is 1500. The `0` value is still accepted and translated into the automatic retrieve of the MTU from the physical interface.

Actually the MTU can be higher than 1500, if the physical link supports the Jumbo frame: we could rise the upper limit of `9000`, but it would create more cons than pros.

The main one is that the NM docs tells that the MTU setting is considered only if it's in higher or equal the minimum value (always granted) and lower or equal the maximum value **allowed by the physical link**, otherwise **it's ignored and the automatica value is used**, meaning that the user is not always sure of what configuration it's applying.

Moreover, to avoid problems in the network (bad performances, missing data, and so on), all the devices should be correctly set with a proper MTU: if, for example, a user set by mistake a high value of MTU, the device may result slow or unable to comunicate with other devices correctly.

For all this reasons we decided to limit the range to a value of 1500, which is the maximum standard for **ALL* the routers, switches and so on.

**Related Issue:**

**Screenshots:**

**Manual Tests:**

### DIFF
--- a/kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/client/ui/network/TabAdvancedUi.java
+++ b/kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/client/ui/network/TabAdvancedUi.java
@@ -44,6 +44,10 @@ public class TabAdvancedUi extends Composite implements NetworkTab {
     private static final String PROMISC_ENABLED = "netAdvPromiscEnabled";
     private static final String PROMISC_DISABLED = "netAdvPromiscDisabled";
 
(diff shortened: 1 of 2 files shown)


mattdibi:
@sfiorani @MMaiero should we backport this to 5.6?

mattdibi (approved):
Tested on RPi. LGTM

PR #5606, opened 2024-12-05
ci: skip build if only documentation files were changed

mattdibi:
WIP


mattdibi:
Closing this since the approach is not what we want

PR #5617, opened 2024-12-13
feat(web2): partial snapshot download

sfiorani:
This PR introduces the possibility to download a partial version of the snapshot, contatining only a subset of the available pid on the system.

The new snapshot download popup presents now a scrollable list of checkboxes representing the available service configurations that can be downloaded: the user can chose one or more pids, having also the possibility to select/remove all the entries in one shot with a specific anchor button.

At least one of the items must be selected: if no pid is picked, an error message will be displayed and you will not allow the download.

The Wiregraph download popup doesn't change, except for the update of the popup title and hint that are now specific for the wirgraph download (before this PR the titles and hint were shared between the Snapshot and Wiregraph download popup).

image
image

**Related Issue:**

**Screenshots:**

**Manual Tests:**


MMaiero:
I would move the "Choose the format in which the file will be downloaded" to the top.
We don't have it in that position anywhere and I believe it makes it more confusing.

MMaiero:
The description label on the top could be something like:

"Select the PIDs for the service configurations that will be downloaded locally.
The configurations can be exported either as JSON or XML format."

@nicolatimeus what do you think?

kura/org.eclipse.kura.web2/src/main/resources/org/eclipse/kura/web/client/messages/Messages.properties
```diff
@@ -309,9 +309,14 @@ deviceSnapshotsNone=No Snapshots Available
 deviceSnapshotRollbackConfirm=Are you sure you want to rollback to a previous configuration? During the rollback operation, the device will disconnect and reconnect. If a timeout is encountered during the reload of the device configuration, please refresh it manually. 
 deviceSnapshotDownloadModalTitle=Snapshot download
 deviceWiregraphDownloadModalTitle=Wiregraph snapshot download
-deviceSnapshotDownloadModalHint=Select pids of the service configurations that will be included in the downloaded snapshot.
+deviceSnapshotDownloadModalHint=Select the PIDs for the service configurations that will be downloaded locally.
+deviceWiregraphDownloadModalHint=Download partial snapshot containing the configuration of the current wiregraph.
```
MMaiero:
```suggestion
deviceWiregraphDownloadModalHint=A partial snapshot containing the current wiregraph configuration will be generated and downloaded.
```

MMaiero (approved):

PR #5627, opened 2024-12-19
feat: added pid filtering searchbox and selection counter in snapshot download popup

sfiorani:
This PR updates what was contributed in #5617 with the following updates:

- Added a searchbox in the snapshot download popup that allows filtering the pids in the checboxes list to easily find the desired ones
- Updated the anchor that select/remove all selections, so when the pid list is filtered with the search box, the anchor select/remove all **the filtered and visible** checkboxes
- Added a label at the right that indicates the number of pids selected over the total number
- Imrpved the backend that filters the pids checkboxes

image
image


kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/client/ui/settings/PartialSnapshotCheckboxStatus.java
```diff
@@ -0,0 +1,26 @@
+package org.eclipse.kura.web.client.ui.settings;
```
MMaiero, line 14: Add copyright header

kura/org.eclipse.kura.web2/src/main/resources/org/eclipse/kura/web/client/messages/Messages.properties
```diff
@@ -316,7 +316,10 @@ downloadSnapshotXmlButton=Download as XML
 downloadSnapshotJsonButton=Download as JSON
 downloadSnapshotError=Please select at least one pid from the list
 selectAllAnchorText=Select All Pids
+selectAllVisibleAnchorText=Select All Visible Pids
```
MMaiero: I believe we need to get consistent on how we display PIDs. In some places is PIDs, in others is Pids

I believe we need to converge to PIDs

pids

MMaiero:
A potential corner case that I've found is related to when the session is ended. You can still select the snapshot and perform your actions to try to download. But after clicking on the download as button, you get no output or feedback.
Is it possible to intercept this case?
Maybe displaying an error or such. It's also strange that there is no redirect to the login page as per other sections of the UI. Maybe we need a bit of alignment there...

kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/client/ui/settings/PartialSnapshotCheckboxStatus.java
MMaiero: I think we can already put 2025 at this point.

sfiorani:
> A potential corner case that I've found is related to when the session is ended. You can still select the snapshot and perform your actions to try to download. But after clicking on the download as button, you get no output or feedback. Is it possible to intercept this case? Maybe displaying an error or such. It's also strange that there is no redirect to the login page as per other sections of the UI. Maybe we need a bit of alignment there...

Yes I can reproduce this wrong behaviour. However, I have already a PR which follows this one and migrates all the snapshot download requests from GET to POST ones. This new mechanism will generate a new token for each download, preventing this situation and triggering the "Session expiration" popup. This will align the snapshot download failure to what happens in the rest of the framework

kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/client/ui/settings/SnapshotDownloadModal.java
```diff
@@ -180,16 +208,36 @@ private void initWiregraphDownloadButtons() {
...
+            });
+        }
+
+        if (this.noPidSelectedError.isVisible()) {
+            this.noPidSelectedError.setVisible(false);
+        }
+
+        updateSelectOrRemoveAllText(checkPidsCheckboxStates());
+    }
+
```
MMaiero:
```suggestion
    private void onSearchBoxEvent(KeyUpEvent event) {
        TextBox searchBox = (TextBox) event.getSource();
        String searchedPid = searchBox.getValue();

        if (searchedPid == null || searchedPid.isEmpty() || searchedPid.equals("")) {
            this.pidPanel.iterator().forEachRemaining(widget -> widget.setVisible(true));
        } else {
            this.pidPanel.iterator().forEachRemaining(widget -> widget
                    .setVisible(isMatchingSearch(widget, searchedPid)));
        }

        if (this.noPidSelectedError.isVisible()) {
            this.noPidSelectedError.setVisible(false);
        }

        updateSelectOrRemoveAllText(checkPidsCheckboxStates());
    }
    
    private boolean isMatchingSearch(Widget widget, String searchedPid) {
        return ((CheckBox) widget).getText().toLowerCase().contains(searchedPid.toLowerCase());
    }

```

Could it be better like this?

MMaiero (approved):

PR #5633, opened 2025-01-07
feat: partial snapshot download requestes turned into POST

sfiorani:
This PR follows #5627 improving the snapshot download mechanism.

The Snapshot download request is now turned into a more robus POST request, removing the link to the native javascript code from the [DownloadHelper](https://github.com/eclipse-kura/kura/blob/9ccb148e3940938a8f8aaece0676753f3c664805/kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/client/util/DownloadHelper.java#L25) class used before.

This PR also avoids also the problem that occurs when the session expires during the selection of the PIDs to be downloaded: indeed, if the session expired after the snapshot download popup spawned, clicking on one of the download button led to a sort of undefined state of the webUI that would stop working. With this PR, if the session expires the token generation fails and the correct error popup is shown correctly.


kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/client/ui/settings/SnapshotDownloadModal.java
```diff
@@ -98,34 +121,51 @@ public SnapshotDownloadModal() {
...
+        this.downloadModal.show();
     }
 
-    public void show(Consumer<SnapshotDownloadOptions> consumer, List<String> availablePids) {
-        this.snapshotDownloadConsumer = consumer;
+    private void initWiregraphDownloadButtons() {
+
+        this.downloadJson.addClickHandler(e -> {
+            this.downloadModal.hide();
+            this.wiregraphDownloadConsumer.accept("JSON");
```
MMaiero: Would it make sense to externalise the JSON and XML in constants?
It should help in keeping the alignment in the code in case of changes/refactorings

kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/client/ui/settings/SnapshotDownloadModal.java
```diff
@@ -275,36 +341,52 @@ private void onSelectOrRemoveAllSelection(ClickEvent handler) {
...
-        List<String> selectedPids = new ArrayList<>();
+    private String getSelectedPids() {
+        StringBuilder selectedPidsBuilder = new StringBuilder("SelectedPids: ");
         this.pidPanel.iterator().forEachRemaining(pid -> {
             CheckBox checkBox = (CheckBox) pid;
-            if (checkBox.getValue().booleanValue() && !checkBox.getText().equals(MSGS.selectAllAnchorText())
-                    && !checkBox.getText().equals(MSGS.removeAllAnchorText())) {
-                selectedPids.add(checkBox.getText());
+            if (checkBox.getValue().booleanValue()) {
+                selectedPidsBuilder.append(checkBox.getText() + ",");
```
MMaiero, line 355: is the trailing comma an issue for the last entry of this builder?

kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/client/ui/settings/SnapshotDownloadModal.java
```diff
@@ -275,36 +341,52 @@ private void onSelectOrRemoveAllSelection(ClickEvent handler) {
...
 
-    private boolean isOnePidSelected() {
-        boolean result = false;
-        Iterator<Widget> pidPanelIterator = this.pidPanel.iterator();
-        while (pidPanelIterator.hasNext()) {
-            CheckBox box = (CheckBox) pidPanelIterator.next();
+    private void downloadEntireSnapshot(Long snapshotId, String format) {
+        RequestQueue.submit(context -> this.gwtXSRFService.generateSecurityToken(context.callback(token -> {
+            xsrfTokenField.setValue(token.getToken());
+            pidsListField.setValue("EntireSnapshot");
```
MMaiero: Would it make more sense to download the entire snapshot if no pidsList is specified?
I am kind of scared of those magic values

sfiorani: I agree with you. Indeed, I started with the idea "EmptyValue --> Download All", but then I changed my mind to avoid empty values...but thinking about it, it's probably more secure and robust. I'll change it

kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/client/ui/settings/SnapshotDownloadModal.java
sfiorani: The servlet that finally downloads the snapshot splits the string using the comma as separator, so the last comma just disappear without any effect. If you prefare I can add a final check that remove the last comma, let me know

MMaiero: up to you. just wanted to make sure we are aware and is covered for any potential corner case

sfiorani: I added a line that removes the last comma. The payload is more clear and it's more maintanable

MMaiero (approved):

PR #5641, opened 2025-01-09
build(nm): updated autogenerated ModemManager code to 1.22.0

sfiorani:
This PR updates the autogenerated code to interface with ModemManager to the lates version available at the time of creation of this pull request, namely 1.22.0.

Code generated using dbus-java api 3.2.2.

### DIFF
--- a/kura/org.eclipse.kura.nm/src/main/java/org/freedesktop/modemmanager1/PropertyConnectionErrorStruct.java
+++ b/kura/org.eclipse.kura.nm/src/main/java/org/freedesktop/modemmanager1/PropertyConnectionErrorStruct.java
@@ -31,11 +31,11 @@ public class PropertyConnectionErrorStruct extends Struct {
     }
 
     public String getMember0() {
-        return member0;
+        return this.member0;
     }
 
     public String getMember1() {
-        return member1;
+        return this.member1;
     }
 
 }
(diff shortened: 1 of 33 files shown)


kura/org.eclipse.kura.nm/src/main/java/org/freedesktop/modemmanager1/PropertyConnectionErrorStruct.java
```diff
@@ -0,0 +1,41 @@
...
+        this.member1 = member1;
+    }
+
+    public String getMember0() {
+        return this.member0;
+    }
+
+    public String getMember1() {
+        return this.member1;
+    }
```
mattdibi, line 39: As you might have noticed the autogenerated code is not the best. Here `member0` and `member1` is not really descriptive of the content of the struct. Looking at the documentation we can understand more:

image

In my original contribution, I left these as is since we wanted to limit the changes to the autogenerated code as much as possible. This allows us to update the autogenerated code almost painlessly at the cost of some readability.

Let's keep them as `memberX` for consistency but be aware of the trade-off we're making.

kura/org.eclipse.kura.nm/src/main/java/org/freedesktop/modemmanager1/modem/Messaging.java
```diff
@@ -43,17 +43,17 @@ public interface Messaging extends DBusInterface {
 
     public static class Added extends DBusSignal {
 
-        private final DBusPath dbusPath;
+        private final DBusPath path;
```
mattdibi: If possible I would avoid this rename

kura/org.eclipse.kura.nm/src/main/java/org/freedesktop/modemmanager1/modem/Voice.java
```diff
@@ -53,30 +53,30 @@ public interface Voice extends DBusInterface {
 
     public static class CallAdded extends DBusSignal {
 
-        private final DBusPath dbusPath;
+        private final DBusPath path;
```
mattdibi: Please avoid this unnecessary rename

mattdibi (changes requested):

mattdibi (approved):
LGTM

PR #5644, opened 2025-01-10
feat(nm): Added modem reset timer when connection is failed

pierantoniomerlino:
This PR adds a modem reset timer when the connection fails.

**Description of the solution adopted:** When a cellular connection fails at the first try (most likely because of an error on the sim), the modem stays in `FAILED` state and NM/MM refuse to start a connection. As a result, no dbus signals are reported and the usual modem reset timer is never triggered.
This PR adds a reset timer when the connection fails. If the connection is recovered before the timeout of the timer, the task is terminated. If the connection is down, instead, a modem reset is requested.

**Manual Tests**: To trigger this behavior, power on a device without a sim inserted. Then try to setup a cellular connection. Observe in the logs that after `Modem Reset Timeout` minutes the modem is reset. Insert the sim and observe that after a further modem reset, the connection can be established.

### DIFF
--- a/kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/ModemManagerDbusWrapper.java
+++ b/kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/ModemManagerDbusWrapper.java
@@ -297,10 +297,10 @@ public class ModemManagerDbusWrapper {
 
         Modem mmModemDevice = this.dbusConnection.getRemoteObject(MM_BUS_NAME, modemManagerDbusPath.get(), Modem.class);
 
-        MMFailedModemResetTimer resetHandler = new MMFailedModemResetTimer(mmModemDevice, delayMinutes);
-        resetHandler.schedule();
+        MMFailedModemResetTimer resetTimer = new MMFailedModemResetTimer(mmModemDevice, delayMinutes);
+        resetTimer.schedule();
 
-        this.failedModemHandlers.put(deviceId, resetHandler);
+        this.failedModemHandlers.put(deviceId, resetTimer);
     }
 
     protected void failedModemResetTimerCancel() {
--- a/kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/ModemManagerDbusWrapper.java
+++ b/kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/ModemManagerDbusWrapper.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2023, 2024 Eurotech and/or its affiliates and others
+ * Copyright (c) 2023, 2025 Eurotech and/or its affiliates and others
  *
  * This program and the accompanying materials are made
  * available under the terms of the Eclipse Public License 2.0
@@ -20,10 +20,12 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
+import java.util.Timer;
 
 import org.eclipse.kura.nm.enums.MMModemLocationSource;
 import org.eclipse.kura.nm.enums.MMModemState;
 import org.eclipse.kura.nm.signal.handlers.NMModemResetHandler;
+import org.eclipse.kura.nm.signal.handlers.NMModemResetTimerTask;
 import org.eclipse.kura.nm.status.SimProperties;
 import org.freedesktop.dbus.DBusPath;
 import org.freedesktop.dbus.connections.impl.DBusConnection;
@@ -49,6 +51,7 @@ public class ModemManagerDbusWrapper {
     private final DBusConnection dbusConnection;
 
     private final Map<String, NMModemResetHandler> modemHandlers = new HashMap<>();
+    private final Map<String, MMFailedModemResetHandler> failedModemHandlers = new HashMap<>();
 
     public ModemManagerDbusWrapper(DBusConnection dbusConnection) {
         this.dbusConnection = dbusConnection;
@@ -140,6 +143,15 @@ public class ModemManagerDbusWrapper {
 
     public MMModemState getMMModemState(Properties modemProperties) {
         return MMModemState.toMMModemState(modemProperties.Get(MM_MODEM_NAME, MM_MODEM_PROPERTY_STATE));
+    }
+
+    public MMModemState getMMModemState(String modemPath) throws DBusException {
+        Optional<Properties> properties = getModemProperties(modemPath);
+        if (properties.isPresent()) {
+            return getMMModemState(properties.get());
+        } else {
+            return MMModemState.MM_MODEM_STATE_UNKNOWN;
+        }
     }
 
     public Optional<Properties> getModemProperties(String modemPath) throws DBusException {
@@ -273,6 +285,86 @@ public class ModemManagerDbusWrapper {
                 logger.warn("Couldn't remove signal handler for: {}. Caused by:", handler.getNMDevicePath(), e);
             }
             this.modemHandlers.remove(deviceId);
+        }
+    }
+
+    protected void failedResetHandlerEnable(String deviceId, Optional<String> modemManagerDbusPath, int delayMinutes)
+            throws DBusException {
+        if (!modemManagerDbusPath.isPresent()) {
+            logger.warn("Cannot retrieve modem device for {}. Skipping modem reset monitor setup.", deviceId);
+            return;
+        }
+
+        Modem mmModemDevice = this.dbusConnection.getRemoteObject(MM_BUS_NAME, modemManagerDbusPath.get(), Modem.class);
+
+        MMFailedModemResetHandler resetHandler = new MMFailedModemResetHandler(mmModemDevice,
+                modemManagerDbusPath.get(), delayMinutes);
+        resetHandler.schedule();
+
+        this.failedModemHandlers.put(deviceId, resetHandler);
+    }
+
+    protected void failedResetHandlersDisable() {
+        for (String deviceId : this.failedModemHandlers.keySet()) {
+            failedResetHandlersDisable(deviceId);
+        }
+        this.modemHandlers.clear();
+    }
+
+    protected void failedResetHandlersDisable(String deviceId) {
+        if (this.failedModemHandlers.containsKey(deviceId)) {
+            MMFailedModemResetHandler handler = this.failedModemHandlers.get(deviceId);
+            handler.cancel();
+        }
+    }
+
+    protected boolean isMMFailedModemResetTimerArmed(String deviceId) {
+        return this.failedModemHandlers.containsKey(deviceId);
+    }
+
+    private class MMFailedModemResetTimerTask extends NMModemResetTimerTask {
+
+        private final String modemManagerDbusPath;
+
+        public MMFailedModemResetTimerTask(Modem modem, String modemManagerDbusPath) {
+            super(modem);
+            this.modemManagerDbusPath = modemManagerDbusPath;
+        }
+
+        @Override
+        public void run() {
+            try {
+                MMModemState modemState = getMMModemState(modemManagerDbusPath);
+                if (MMModemState.MM_MODEM_STATE_FAILED.equals(modemState)) {
+                    super.run();
+                }
+            } catch (DBusException e) {
+                ModemManagerDbusWrapper.logger.warn("Couldn't get state of modem interface, caused by:", e);
+            }
+        }
+
+    }
+
+    private class MMFailedModemResetHandler {
+
+        private final Timer timer = new Timer("FailedModemResetTimer");
+        private final MMFailedModemResetTimerTask task;
+        private final long delay;
+
+        public MMFailedModemResetHandler(Modem modem, String modemManagerDbusPath, long delayMinutes) {
+            this.delay = delayMinutes * 60L * 1000L;
+            this.task = new MMFailedModemResetTimerTask(modem, modemManagerDbusPath);
+        }
+
+        public void schedule() {
+            this.timer.schedule(this.task, this.delay);
+        }
+
+        public void cancel() {
+            if (this.task != null) {
+                this.task.cancel();
+            }
+            this.timer.cancel();
         }
     }
 }
--- a/kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/ModemManagerDbusWrapper.java
+++ b/kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/ModemManagerDbusWrapper.java
@@ -334,10 +334,10 @@ public class ModemManagerDbusWrapper {
                 if (MMModemState.MM_MODEM_STATE_FAILED.equals(modemState)) {
                     super.run();
                 } else {
-                    logger.info("Modem state changed. Reset skipped.");
+                    NMModemResetTimerTask.logger.info("Modem state changed. Reset skipped.");
                 }
             } catch (DBusException e) {
-                logger.warn("Couldn't get state of modem interface, caused by:", e);
+                NMModemResetTimerTask.logger.warn("Couldn't get state of modem interface, caused by:", e);
             }
         }
 
--- a/kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/ModemManagerDbusWrapper.java
+++ b/kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/ModemManagerDbusWrapper.java
@@ -51,7 +51,7 @@ public class ModemManagerDbusWrapper {
     private final DBusConnection dbusConnection;
 
     private final Map<String, NMModemResetHandler> modemHandlers = new HashMap<>();
-    private final Map<String, MMFailedModemResetTimer> failedModemHandlers = new HashMap<>();
+    private final Map<String, MMFailedModemResetTimer> failedModemResetTimers = new HashMap<>();
 
     public ModemManagerDbusWrapper(DBusConnection dbusConnection) {
         this.dbusConnection = dbusConnection;
@@ -300,25 +300,25 @@ public class ModemManagerDbusWrapper {
         MMFailedModemResetTimer resetTimer = new MMFailedModemResetTimer(mmModemDevice, delayMinutes);
         resetTimer.schedule();
 
-        this.failedModemHandlers.put(deviceId, resetTimer);
+        this.failedModemResetTimers.put(deviceId, resetTimer);
     }
 
     protected void failedModemResetTimerCancel() {
-        for (String deviceId : this.failedModemHandlers.keySet()) {
+        for (String deviceId : this.failedModemResetTimers.keySet()) {
             failedModemResetTimerCancel(deviceId);
         }
         this.modemHandlers.clear();
     }
 
     protected void failedModemResetTimerCancel(String deviceId) {
-        if (this.failedModemHandlers.containsKey(deviceId)) {
-            MMFailedModemResetTimer timer = this.failedModemHandlers.get(deviceId);
+        if (this.failedModemResetTimers.containsKey(deviceId)) {
+            MMFailedModemResetTimer timer = this.failedModemResetTimers.get(deviceId);
             timer.cancel();
         }
     }
 
     protected boolean isMMFailedModemResetTimerArmed(String deviceId) {
-        return this.failedModemHandlers.containsKey(deviceId);
+        return this.failedModemResetTimers.containsKey(deviceId);
     }
 
     private class MMFailedModemResetTimerTask extends NMModemResetTimerTask {
--- a/kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/ModemManagerDbusWrapper.java
+++ b/kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/ModemManagerDbusWrapper.java
@@ -51,7 +51,7 @@ public class ModemManagerDbusWrapper {
     private final DBusConnection dbusConnection;
 
     private final Map<String, NMModemResetHandler> modemHandlers = new HashMap<>();
-    private final Map<String, MMFailedModemResetHandler> failedModemHandlers = new HashMap<>();
+    private final Map<String, MMFailedModemResetTimer> failedModemHandlers = new HashMap<>();
 
     public ModemManagerDbusWrapper(DBusConnection dbusConnection) {
         this.dbusConnection = dbusConnection;
@@ -288,7 +288,7 @@ public class ModemManagerDbusWrapper {
         }
     }
 
-    protected void failedResetHandlerEnable(String deviceId, Optional<String> modemManagerDbusPath, int delayMinutes)
+    protected void failedModemResetTimerEnable(String deviceId, Optional<String> modemManagerDbusPath, int delayMinutes)
             throws DBusException {
         if (!modemManagerDbusPath.isPresent()) {
             logger.warn("Cannot retrieve modem device for {}. Skipping modem reset monitor setup.", deviceId);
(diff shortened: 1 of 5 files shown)


kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/ModemManagerDbusWrapper.java
```diff
@@ -272,4 +284,84 @@ protected void resetHandlersDisable(String deviceId) {
...
+
+    }
+
+    private class MMFailedModemResetHandler {
+
+        private final Timer timer = new Timer("FailedModemResetTimer");
+        private final MMFailedModemResetTimerTask task;
+        private final long delay;
+
+        public MMFailedModemResetHandler(Modem modem, String modemManagerDbusPath, long delayMinutes) {
```
mattdibi: This might be a little confusing. Is `modemManagerDbusPath` the path of the modem DBus object or the path to the ModemManager DBus instance? I think it's the former. If that's the case I would rename this variable (and all other references in this file) to `modemPath` as you did for `getMMModemState(String modemPath)`

kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/ModemManagerDbusWrapper.java
```diff
@@ -272,4 +284,84 @@ protected void resetHandlersDisable(String deviceId) {
...
+                    super.run();
+                }
+            } catch (DBusException e) {
+                ModemManagerDbusWrapper.logger.warn("Couldn't get state of modem interface, caused by:", e);
+            }
+        }
+
+    }
+
+    private class MMFailedModemResetHandler {
```
mattdibi: When I developed the feature I used the "Handler" suffix for classes that implemented a DBus *Signal* Handler. This is not really handling signals, it just manages the timer task. I would change its name to something more meaningful/less confusing... `MMFailedModemResetTaskManager`?

kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/ModemManagerDbusWrapper.java
```diff
@@ -272,4 +284,84 @@ protected void resetHandlersDisable(String deviceId) {
...
+                MMModemState modemState = getMMModemState(modemManagerDbusPath);
+                if (MMModemState.MM_MODEM_STATE_FAILED.equals(modemState)) {
+                    super.run();
+                }
+            } catch (DBusException e) {
+                ModemManagerDbusWrapper.logger.warn("Couldn't get state of modem interface, caused by:", e);
+            }
+        }
+
+    }
```
mattdibi: I would probably move where the `NMModemResetTimerTask` is stored (`signal/handlers`... or should we move both somewhere else? Extending it here is quite confusing IMO)

kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/ModemManagerDbusWrapper.java
```diff
@@ -272,4 +284,84 @@ protected void resetHandlersDisable(String deviceId) {
...
+        return this.failedModemHandlers.containsKey(deviceId);
+    }
+
+    private class MMFailedModemResetTimerTask extends NMModemResetTimerTask {
+
+        private final String modemManagerDbusPath;
+
+        public MMFailedModemResetTimerTask(Modem modem, String modemManagerDbusPath) {
+            super(modem);
+            this.modemManagerDbusPath = modemManagerDbusPath;
```
mattdibi: Is this really needed? You already have the `modem` instance

kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/ModemManagerDbusWrapper.java
```diff
@@ -272,4 +284,84 @@ protected void resetHandlersDisable(String deviceId) {
...
+        }
+
+        @Override
+        public void run() {
+            try {
+                MMModemState modemState = getMMModemState(modemManagerDbusPath);
```
mattdibi: Why not use the original variable?

```suggestion
                MMModemState modemState = getMMModemState(this.modem.getObjectPath());
```

kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/ModemManagerDbusWrapper.java
```diff
@@ -272,4 +284,84 @@ protected void resetHandlersDisable(String deviceId) {
...
+                if (MMModemState.MM_MODEM_STATE_FAILED.equals(modemState)) {
+                    super.run();
+                }
```
mattdibi: in the else branch I would write something to ease debugging

```suggestion
                if (MMModemState.MM_MODEM_STATE_FAILED.equals(modemState)) {
                    super.run();
                } else {
                    logger.info("Modem state changed. Reset skipped");
                }
```

kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/ModemManagerDbusWrapper.java
```diff
@@ -272,4 +284,84 @@ protected void resetHandlersDisable(String deviceId) {
...
+            } catch (DBusException e) {
+                ModemManagerDbusWrapper.logger.warn("Couldn't get state of modem interface, caused by:", e);
```
mattdibi: I don't super like this. Why not use the logger within the super class? See:  https://github.com/eclipse-kura/kura/blob/e6c27549de30c95e2c3926e0cecbf7c9253c52ba/kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/signal/handlers/NMModemResetTimerTask.java#L25

kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/ModemManagerDbusWrapper.java
mattdibi: Ah I see... you need it here because you're leveraging the `getMMModemState` method. Still not super sure about it 🤔

kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/NMDbusConnector.java
```diff
@@ -582,6 +602,18 @@ private void enableInterface(String deviceId, NetworkProperties properties, Devi
...
     }
 
+    private boolean isModemFailed(Device device, NMDeviceType deviceType) throws DBusException {
+        if (deviceType != NMDeviceType.NM_DEVICE_TYPE_MODEM) {
+            return false;
+        }
+        Optional<String> mmDbusPath = this.networkManager.getModemManagerDbusPath(device.getObjectPath());
+        if (!mmDbusPath.isPresent()) {
+            return false;
+        }
```
mattdibi: I do not 100% agree on this.

This essentially means:
1. When Kura tried to activate a NM connection on a device, it threw an exception
2. while handling the exception Kura tries to retrieve the DBus path of the device that thew the exception
3. if it doesn't find the dbus path of the device it assumes the device is not in a "FAILED" state

It might be too optimistic. I would throw an exception in this case or log an error since we're in an unrecoverable state. What do you think?

kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/NMDbusConnector.java
```diff
@@ -557,10 +562,25 @@ private void enableInterface(String deviceId, NetworkProperties properties, Devi
...
+
+                if (delayMinutes != 0) {
+                    Optional<String> mmDbusPath = this.networkManager.getModemManagerDbusPath(device.getObjectPath());
+
+                    logger.info("Modem {} in failed state or unavailable. Scheduling modem reset in {} minutes ...",
+                            device.getObjectPath(), delayMinutes);
+
+                    this.modemManager.failedResetHandlerEnable(deviceId, mmDbusPath, delayMinutes);
+                }
+            }
```
mattdibi: Here we have a lot of stuff that could throw an exception and is specific for modem devices.

Would it be better to have a `activationFailed` flag on which we can structure our code outside of this `catch` block? Bonus points if we can manage this in the https://github.com/eclipse-kura/kura/blob/e6c27549de30c95e2c3926e0cecbf7c9253c52ba/kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/NMDbusConnector.java#L594 block.

```suggestion
            activationFailed = true;
```

kura/test/org.eclipse.kura.nm.test/src/test/java/org/eclipse/kura/nm/NMDbusConnectorTest.java
```diff
@@ -1813,7 +1853,7 @@ private void thenModemStatusHasCorrectValues(boolean hasBearers, boolean hasSims
         assertTrue(modemStatus.isGpsSupported());
         assertEquals(EnumSet.of(ModemGpsMode.UNMANAGED, ModemGpsMode.MANAGED_GPS), modemStatus.getSupporteGpsModes());
         assertFalse(modemStatus.isSimLocked());
-        assertEquals(ModemConnectionStatus.REGISTERED, modemStatus.getConnectionStatus());
+        assertEquals(ModemConnectionStatus.FAILED, modemStatus.getConnectionStatus());
```
mattdibi, line 1856: Personally I don't think we should change existing test expectations when we introduce a new feature.

All the old tests still need to pass as before, we just need to test what we added. Is there an easy way to avoid changing existing tests here? If the solution is worse than what is currently implemented ignore this comment.

kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/ModemManagerDbusWrapper.java
```diff
@@ -272,4 +284,84 @@ protected void resetHandlersDisable(String deviceId) {
             }
         }
     }
+
+    protected void failedResetHandlerEnable(String deviceId, Optional<String> modemManagerDbusPath, int delayMinutes)
```
mattdibi: I would rethink this method name. This is not a `ResetHandlerEnable`... this is just scheduling the reset.

kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/ModemManagerDbusWrapper.java
```diff
@@ -272,4 +284,84 @@ protected void resetHandlersDisable(String deviceId) {
...
+    }
+
+    protected void failedResetHandlersDisable() {
+        for (String deviceId : this.failedModemHandlers.keySet()) {
+            failedResetHandlersDisable(deviceId);
+        }
+        this.modemHandlers.clear();
+    }
+
+    protected void failedResetHandlersDisable(String deviceId) {
```
mattdibi: Same here, these are not `ResetHandlerDisable` but simple task cancel

mattdibi (changes requested):

kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/ModemManagerDbusWrapper.java
pierantoniomerlino: Since it is a Timer, I prefer `MMFailedModemResetTimer` and `MMFailedModemResetTimerTask`

kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/ModemManagerDbusWrapper.java
pierantoniomerlino: I'd create a new class instead of using the `NMModemResetTimerTask` one, only because I need to check the state of the modem before reset. I'd not put it in the `signal/handler` folder since it doesn't manage any signal, as you correclty pointed out.
Finally, since we use these classes only here for a specific task (that is we'll not use them anywhere...), I prefer to make them private.

What do you think?

kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/ModemManagerDbusWrapper.java
pierantoniomerlino: Correct. My fault.

kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/ModemManagerDbusWrapper.java
pierantoniomerlino: I cannot use `this.modem` since this is a private field. However, `this.getModemDbusPath` is available.

kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/ModemManagerDbusWrapper.java
pierantoniomerlino: I think that I cannot use it since it is private.

kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/NMDbusConnector.java
pierantoniomerlino: Indeed, I was unsure about this behavior. If the Dbus path is not present, we cannot say anything on the state of the modem. I returned a `false` to avoid the start of the timer. Probably, it is better to throw an exception and log. But we need to do the same when we start the `NMModemResetHandler` few lines below.

kura/test/org.eclipse.kura.nm.test/src/test/java/org/eclipse/kura/nm/NMDbusConnectorTest.java
pierantoniomerlino: Yes, I agree. However, I chose this solution for two reasons:

1. add the modem status as a parameter requires an huge refactoring of the tests
2. the modem status is not really used in the other tests.

I'd prefer not to change this.

kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/NMDbusConnector.java
pierantoniomerlino: I like it. By the way, it simplifies the management of the missing modem dbus path and makes the code more clear.

pierantoniomerlino:
@mattdibi I updated this PR, but I've not tested it yet. So, please take a look, but don't merge it.

kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/ModemManagerDbusWrapper.java
mattdibi: Sorry, I implicitly assumed you would need to change it to `protected`. I think the change in visibility might make sense here. Let me know what you think.

kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/ModemManagerDbusWrapper.java
```diff
@@ -275,4 +287,82 @@ protected void resetHandlersDisable(String deviceId) {
             this.modemHandlers.remove(deviceId);
         }
     }
+
+    protected void failedModemResetTimerEnable(String deviceId, Optional<String> modemManagerDbusPath, int delayMinutes)
```
mattdibi: Nitpicking here. This probably is a `Schedule` more than an `Enable`. Can be left as-is.

```suggestion
    protected void scheduleFailedModemResetTimer(String deviceId, Optional<String> modemManagerDbusPath, int delayMinutes)
```

kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/ModemManagerDbusWrapper.java
```diff
@@ -275,4 +287,82 @@ protected void resetHandlersDisable(String deviceId) {
...
+
+        Modem mmModemDevice = this.dbusConnection.getRemoteObject(MM_BUS_NAME, modemManagerDbusPath.get(), Modem.class);
+
+        MMFailedModemResetTimer resetHandler = new MMFailedModemResetTimer(mmModemDevice, delayMinutes);
+        resetHandler.schedule();
+
+        this.failedModemHandlers.put(deviceId, resetHandler);
+    }
+
+    protected void failedModemResetTimerDisable() {
```
mattdibi: Same thing here:

```suggestion
    protected void cancelFailedModemResetTimer() {
```

kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/ModemManagerDbusWrapper.java
pierantoniomerlino: Yeah... indeed I chose that names to be consistent with the other Timer/Handler since they are related in some way. Let's me think about it.

kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/ModemManagerDbusWrapper.java
```diff
@@ -49,6 +51,7 @@ public class ModemManagerDbusWrapper {
     private final DBusConnection dbusConnection;
 
     private final Map<String, NMModemResetHandler> modemHandlers = new HashMap<>();
+    private final Map<String, MMFailedModemResetTimer> failedModemHandlers = new HashMap<>();
```
mattdibi: These are not "handlers" anymore at this point

```suggestion
    private final Map<String, MMFailedModemResetTimer> failedModemResetTimers = new HashMap<>();
```

mattdibi (changes requested):
LGTM there's only a rename that you missed

kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/ModemManagerDbusWrapper.java
pierantoniomerlino: Argh! :weary:

mattdibi:
## Testing

Running current PR build on a Dg1014. I configured the modem to connect without installing a SIM. In the log I can see the following:

```
Jan 15 08:50:47 dynagate-10-14 ESF[866]: Settings iface "3-1.3":NM_DEVICE_TYPE_MODEM
Jan 15 08:50:47 dynagate-10-14 ESF[866]: Couldn't complete activation of 3-1.3 interface, caused by:
Jan 15 08:50:47 dynagate-10-14 ESF[866]: Modem /org/freedesktop/NetworkManager/Devices/11 in failed state or unavailable. Scheduling modem reset in 5 minutes ...
Jan 15 08:50:47 dynagate-10-14 ESF[866]: Modem /org/freedesktop/ModemManager1/Modem/0 not enabled. Enabling modem...
Jan 15 08:50:47 dynagate-10-14 ESF[866]: Unable to apply configuration to the device 3-1.3
```

I then installed a SIM in the slot and, after 5 minutes, I can see the following:

```
Jan 15 08:55:45 dynagate-10-14 ESF[866]: Cannot get bearers for modem /org/freedesktop/ModemManager1/Modem/0
Jan 15 08:55:45 dynagate-10-14 ESF[866]: 3gpp properties not found.
Jan 15 08:55:45 dynagate-10-14 ESF[866]: failed to get status for interface: 3-1.3
Jan 15 08:55:46 dynagate-10-14 ESF[866]: Cannot get bearers for modem /org/freedesktop/ModemManager1/Modem/0
Jan 15 08:55:46 dynagate-10-14 ESF[866]: 3gpp properties not found.
Jan 15 08:55:46 dynagate-10-14 ESF[866]: failed to parse last tamper event timestamp
Jan 15 08:55:46 dynagate-10-14 ESF[866]: Alert {"message":"\"CPU Utilization\" is below thresholds","code":"cpu","category":"performance","severity":"INFO","createdOn":1736931345686,"value":5,"threshold":85}
Jan 15 08:55:46 dynagate-10-14 ESF[866]: Storing message on topic: #account-name/#client-id/ALERT, priority: 4
Jan 15 08:55:46 dynagate-10-14 ESF[866]: Stored message on topic: #account-name/#client-id/ALERT, priority: 4
Jan 15 08:55:46 dynagate-10-14 ESF[866]: DataPublisherService not connected
Jan 15 08:55:46 dynagate-10-14 ESF[866]: Published alert message: org.eclipse.kura.message.KuraPayload@49b2107a
Jan 15 08:55:46 dynagate-10-14 ESF[866]: Alert {"message":"\"Process CPU utilization for kura\" is below thresholds","code":"cpu","category":"performance","severity":"INFO","createdOn":1736931345688,"value":18,"threshold":85}
Jan 15 08:55:46 dynagate-10-14 ESF[866]: Storing message on topic: #account-name/#client-id/ALERT, priority: 4
Jan 15 08:55:46 dynagate-10-14 ESF[866]: Stored message on topic: #account-name/#client-id/ALERT, priority: 4
Jan 15 08:55:46 dynagate-10-14 ESF[866]: Published alert message: org.eclipse.kura.message.KuraPayload@67261c69
Jan 15 08:55:46 dynagate-10-14 ESF[866]: DataPublisherService not connected
Jan 15 08:55:46 dynagate-10-14 ESF[866]: Alert {"message":"\"Cellular Signal Strength for 3-1.3\" is below critical threshold","code":"cellular","category":"connectivity","severity":"CRITICAL","createdOn":1736931346216,"value":-113,"threshold":-110}
Jan 15 08:55:46 dynagate-10-14 ESF[866]: Storing message on topic: #account-name/#client-id/ALERT, priority: 4
Jan 15 08:55:46 dynagate-10-14 ESF[866]: Stored message on topic: #account-name/#client-id/ALERT, priority: 4
Jan 15 08:55:46 dynagate-10-14 ESF[866]: Published alert message: org.eclipse.kura.message.KuraPayload@29dd9959
Jan 15 08:55:46 dynagate-10-14 ESF[866]: DataPublisherService not connected
Jan 15 08:55:47 dynagate-10-14 ESF[866]: Modem reset timer expired. Resetting modem /org/freedesktop/ModemManager1/Modem/0 ...
Jan 15 08:55:47 dynagate-10-14 ESF[866]: Could not perform modem reset for /org/freedesktop/ModemManager1/Modem/0 because:
org.freedesktop.dbus.exceptions.DBusExecutionException: Cannot reset the modem: operation not supported
	at jdk.internal.reflect.GeneratedConstructorAccessor68.newInstance(Unknown Source)
	at java.base/jdk.internal.reflect.DelegatingConstructorAccessorImpl.newInstance(Unknown Source)
	at java.base/java.lang.reflect.Constructor.newInstanceWithCaller(Unknown Source)
	at java.base/java.lang.reflect.Constructor.newInstance(Unknown Source)
	at org.freedesktop.dbus.errors.Error.getException(Error.java:145)
	at org.freedesktop.dbus.errors.Error.throwException(Error.java:175)
	at org.freedesktop.dbus.RemoteInvocationHandler.executeRemoteMethod(RemoteInvocationHandler.java:152)
	at org.freedesktop.dbus.RemoteInvocationHandler.invoke(RemoteInvocationHandler.java:216)
	at jdk.proxy8/jdk.proxy8.$Proxy71.Reset(Unknown Source)
	at org.eclipse.kura.nm.signal.handlers.NMModemResetTimerTask.run(NMModemResetTimerTask.java:43)
	at org.eclipse.kura.nm.ModemManagerDbusWrapper$MMFailedModemResetTimerTask.run(ModemManagerDbusWrapper.java:335)
	at java.base/java.util.TimerThread.mainLoop(Unknown Source)
	at java.base/java.util.TimerThread.run(Unknown Source)
```

The timer does work but the current installed version of MM (1.16.2) doesn't properly support the installed modem (Quectel EM05-G)

For me this constitutes as a pass. I'll try and update the Dg1014.

**Update**: after updating the OS and repeating the test I get the following:

```
Jan 15 10:59:27 dynagate-10-14 ESF[1138]: Modem reset timer expired. Resetting modem /org/freedesktop/ModemManager1/Modem/0 ...
Jan 15 11:00:02 dynagate-10-14 ESF[1138]: New network device connected at /org/freedesktop/NetworkManager/Devices/12
Jan 15 11:00:02 dynagate-10-14 ESF[1138]: Settings iface "3-1.3":NM_DEVICE_TYPE_MODEM
Jan 15 11:00:08 dynagate-10-14 ESF[1138]: Timeout elapsed. Exiting anyway
```

```
root@dynagate-10-14:/home/guest# nmcli
eth1: connected to eth1
        "eth1"
        ethernet (imx-dwmac), 00:E0:C7:0A:E5:E5, hw, mtu 1500
        ip4 default
        inet4 192.168.1.172/24
        route4 192.168.1.0/24 metric 105
        route4 default via 192.168.1.1 metric 105

eth0: connected to eth0
        "eth0"
        ethernet (fec), 00:E0:C7:0A:E5:E6, hw, mtu 1500
        inet4 172.16.0.1/24
        route4 172.16.0.0/24 metric 104

cdc-wdm0: connecting (prepare) to kura-cdc-wdm0-connection
        "cdc-wdm0"
        gsm (qmi_wwan, option), hw
```

mattdibi (approved):

PR #5645, opened 2025-01-13
fix(distrib): kura_install.sh is broken

reda-alaoui:
The Kura install cannot work as is. This pr fixes the issue.

### DIFF
--- a/kura/distrib/src/main/resources/x86_64-nn/kura_install.sh
+++ b/kura/distrib/src/main/resources/x86_64-nn/kura_install.sh
@@ -57,7 +57,7 @@ cp ${INSTALL_DIR}/kura/install/manage_kura_users.sh ${INSTALL_DIR}/kura/.data/ma
 chmod 700 ${INSTALL_DIR}/kura/.data/manage_kura_users.sh
 ${INSTALL_DIR}/kura/.data/manage_kura_users.sh -i -nn
 
-bash "${INSTALL_DIR}/kura/install/customize-installation.sh"${IS_NETWORKING_PROFILE}
+bash "${INSTALL_DIR}/kura/install/customize-installation.sh" ${IS_NETWORKING_PROFILE}
 
 # copy snapshot_0.xml
 cp ${INSTALL_DIR}/kura/user/snapshots/snapshot_0.xml ${INSTALL_DIR}/kura/.data/snapshot_0.xml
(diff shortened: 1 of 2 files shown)


mattdibi:
Hello there @reda-alaoui!

Thanks for your contribution and spotting this. @reda-alaoui if you don't mind I'd like to ask you a couple of questions: Was the installer broken for you? What hardware platform did you use for testing/finding out the bug?

reda-alaoui:
Yes deb packages (default and nn) from 5.6.0 were broken for me on a debian x86_64.
The deb install *appears* to work, but ends up with non bootable kura. I had to check the install logs to understand the cause.
Using deb packages from 5.5.0 was ok. So this looks like a recent regression.

mattdibi:
> Yes deb packages (default and nn) from 5.6.0 were broken for me on a debian x86_64. The deb install _appears_ to work, but ends up with non bootable kura. I had to check the install logs to understand the cause. Using deb packages from 5.5.0 was ok. So this looks like a recent regression.

Thanks for the quick feedback.

On our reference platform for the x86-64 architecture (i.e. Intel UP2) the installer **seems** to work fine. That's why this slip up through our QAs. Something weird is going on...

image

**Update**: I forgot that the Intel UP2 had its own installer. The bug is only present on the x86-64 **generic** profile.

Going forward our reference platform for the x86-64 Kura installer will be the Zimaboard as discussed in #5226.

pierantoniomerlino (approved):

mattdibi:
Again @reda-alaoui thanks for spotting this. Unfortunately it slipped through our QA. We'll tighten our procedure to prevent this from happening again.

reda-alaoui:
> Again @reda-alaoui thanks for spotting this. Unfortunately it slipped through our QA. We'll tighten our procedure to prevent this from happening again.

Thank you :)

AceSlash:
Hello,

Is it possible to have a bugfix release with this fix? Currently we are stuck on the 5.5 version because of this.

In any case, thank you for the great work with kura!

mattdibi:
Hi there @AceSlash, we're currently discussing how to proceed with this issue. Stay tuned!

mattdibi:
Hi there @AceSlash, #5648 should answer your question. Feel free to provide feedback there. Cheers!

AceSlash:
@mattdibi thank you, it works for us :)

PR #5649, opened 2025-01-16
build: add copyright header checks using maven checkstyle plugin

mattdibi:
This PR introduces the check for copyright headers consistency leveraging the maven checkstyle plugin.

**Details**: with this PR we're now leveraging the [RegexpHeader](https://checkstyle.sourceforge.io/checks/header/regexpheader.html) check implemented by the maven checkstyle plugin. The checkstyle plugin now checks whether the header of the file matches the regex contained in the newly introduced `header.txt` file in the root directory.

Please be aware that the check won't notify the developer of the need to update the copyright year when the file is edited. This type of check is not available in the checkstyle plugin (we would need to know which files with respect to the develop branch, probably a CI of sorts is better suited for the task).

---

While working on this check I found out a file that was not matching our current license. The file is the following:

https://github.com/eclipse-kura/kura/blob/5d524669cff80b1b4dffd5ae26c4a4c1a2ba4ea9/kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/bluetooth/le/beacon/AdvertisingReportPhy.java#L5

introduced in https://github.com/eclipse-kura/kura/pull/3377 by @srware.

I took the chance to update the licence to EPL2.0 for consistency with the rest of the project. and [in accordance to Eclipse Foundation guidelines](https://www.eclipse.org/legal/epl-2.0/faq/) (see: a29bec75ec1c615a3ecf23228872a57a52b11774)

Referring [to the FAQ](https://www.eclipse.org/legal/epl-2.0/faq/#h.cktcxkzh8ks4) I would like to ask @srware whether they're OK with us updating the license of the above-mentioned file.

### DIFF
--- a/checkstyle_checks.xml
+++ b/checkstyle_checks.xml
@@ -48,8 +48,8 @@
 
     <!-- Checks for Headers                                -->
     <!-- See https://checkstyle.org/config_header.html   -->
-    <module name="Header">
-    <!--   <property name="headerFile" value="${checkstyle.header.file}"/> -->
+    <module name="RegexpHeader">
+        <property name="headerFile" value="header.txt"/>
         <property name="fileExtensions" value="java"/>
     </module>
 
--- a/checkstyle_checks.xml
+++ b/checkstyle_checks.xml
@@ -20,23 +20,23 @@
         <property name="fileExtensions" value="java"/>
         <property name="message" value="Do not use Windows line endings"/>
     </module>
-    
+
     <!-- Checks whether files end with a new line.                        -->
     <!-- See https://checkstyle.org/config_misc.html#NewlineAtEndOfFile -->
     <!--<module name="NewlineAtEndOfFile"/>-->
-    
+
     <!-- Checks that property files contain the same keys.         -->
     <!-- See https://checkstyle.org/config_misc.html#Translation -->
     <module name="Translation"/>
-    
+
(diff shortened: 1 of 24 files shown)


srware:
@mattdibi , I am happy for this update to go ahead.

mattdibi:
@srware Thanks for the quick feedback, this is greatly appreciated! 🙏

MMaiero (approved):

salvatore-coppola:
@mattdibi there's a side issue, since the [checkstyle_checks.xml ](https://github.com/eclipse-kura/kura/blob/ca95dd7a2e92ae50999c89bef3847193da2c83f0/checkstyle_checks.xml) file refers to the https://github.com/eclipse-kura/kura/blob/ca95dd7a2e92ae50999c89bef3847193da2c83f0/header.txt file that is in the root directory, the execution in sub-modules and in the target platform results in an error (the plugin doesn't find the header file).

The problem is very annoying because now it's impossible to build the target platform or a single bundle inside the related folder.
I tried to fix it quickly but I failed, do you have any idea ?

mattdibi:
@salvatore-coppola I'm going to revert the checkstyle change ASAP. The current approach is not working as intended, has low coverage and doesn't scale with all our projects. A better/different approach is required.

PR #5651, opened 2025-01-17
ci(Jenkinsfile): add copyright year check

mattdibi:
This PR introduces a new stage in our Jenkins pipeline build that checks whether the copyright header of the updated Java files is set to the current year.

**Details**: given the introduction of the RegexpHeader check in #5649 we know that (almost) all Java files will contain a copyright header in the format described in the `header.txt` file. It's then trivial to retrieve the year of the copyright header.

This PR leverages the structure of our copyright header to enforce the update of the year, only for files that were updated in the PR. I couldn't implement this type of check in #5649 since the checkstyle plugin doesn't have the concept of "updated files", therefore the need of implementing it in Jenkins. Finally, if the Jenkins job is ran on `develop` branch (i.e. no `CHANGE_TARGET`), the copyright header year check is skipped.

Extending this check to `.xml` file should be trivial we just need to grab the correct line (third in the xml case) and accept `.xml` file types.

**Notes**: This stage runs _before_ the code build and will exit in case of failure. An alternative implementation of the same check can be performed via Github Actions which would allow the code build to run independently from the copyright headers. Would that be better? Let me know what you think.


mattdibi:
As per #5665 we're dropping this approach. Closing as obsolete.

PR #5663, opened 2025-01-20
feat(nm): L2Only as alias of Unmanaged mode

mattdibi:
This PR fixes the behaviour of the `netIPv4StatusL2Only`/`netIPv6StatusL2Only` parameter.

Up until now `netIPvXStatusL2Only` was not correctly supported in our codebase in our generic profiles. See: 

https://github.com/eclipse-kura/kura/blob/ca95dd7a2e92ae50999c89bef3847193da2c83f0/kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/NMDbusConnector.java#L87-L88

This meant that, when configured as `netIPvXStatusL2Only`, Kura would quit configuration without touching the interface. See 

https://github.com/eclipse-kura/kura/blob/ca95dd7a2e92ae50999c89bef3847193da2c83f0/kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/NMDbusConnector.java#L467-L472

This behaviour technically still obtains the desired configuration (interface is up and the user is free to configure it) but is not quite correct.

This PR simply promotes `netIPv4StatusL2Only` to an alias of `netIPv4StatusUnmanaged` thus obtaining the expected behaviour.

**Note**: To preserve the compatibility with older Kura versions and the old-networking profiles, specifying only IPv4 as "L2Only" will suffice to set the whole interface as "Unmanaged". This is due to the fact that originally, there was no IPv6 support.

For "Unmanaged" we require users to set both IPv4 and IPv6 as "Unmanaged" otherwise we will throw an error. This is because "Unmanaged" is a property of the **whole network interface** not only the IPv4/6 part of the settings. The same applies for L2Only **but**, requiring the IPv4 and IPv6 status to be both set to L2Only would break compatibility with old snapshots.

@pierantoniomerlino I have only one doubt: currently, if the IPv6 status is `UNKNOWN` the method won't throw an exception. The original idea was to throw an exception when any of the statuses was `UNKNOWN` but, if we do with `L2ONLY`, we might risk breaking existing snapshots? What do you think?

You can use the introduced tests to understand what's going on.

### DIFF
--- a/kura/test/org.eclipse.kura.nm.test/src/test/java/org/eclipse/kura/nm/KuraInterfaceStatusTest.java
+++ b/kura/test/org.eclipse.kura.nm.test/src/test/java/org/eclipse/kura/nm/KuraInterfaceStatusTest.java
@@ -1,3 +1,15 @@
+/*******************************************************************************
+ * Copyright (c) 2023 Eurotech and/or its affiliates and others
+ *
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *  Eurotech
+ *******************************************************************************/
 package org.eclipse.kura.nm;
 
 import static org.junit.Assert.assertEquals;
--- a/kura/test/org.eclipse.kura.nm.test/src/test/java/org/eclipse/kura/nm/KuraInterfaceStatusTest.java
+++ b/kura/test/org.eclipse.kura.nm.test/src/test/java/org/eclipse/kura/nm/KuraInterfaceStatusTest.java
@@ -32,6 +32,10 @@ public class KuraInterfaceStatusTest {
     public static Iterable<Object[]> data() {
         return Arrays.asList(new Object[][] { //
                 { KuraIpStatus.DISABLED, KuraIpStatus.DISABLED, KuraInterfaceStatus.DISABLED, null }, //
+                { KuraIpStatus.ENABLEDLAN, KuraIpStatus.ENABLEDLAN, KuraInterfaceStatus.ENABLED, null }, //
+                { KuraIpStatus.ENABLEDWAN, KuraIpStatus.ENABLEDWAN, KuraInterfaceStatus.ENABLED, null }, //
+                { KuraIpStatus.ENABLEDLAN, KuraIpStatus.ENABLEDWAN, KuraInterfaceStatus.ENABLED, null }, //
+                { KuraIpStatus.ENABLEDWAN, KuraIpStatus.ENABLEDLAN, KuraInterfaceStatus.ENABLED, null }, //
                 { KuraIpStatus.ENABLEDLAN, KuraIpStatus.DISABLED, KuraInterfaceStatus.ENABLED, null }, //
                 { KuraIpStatus.ENABLEDWAN, KuraIpStatus.DISABLED, KuraInterfaceStatus.ENABLED, null }, //
                 { KuraIpStatus.DISABLED, KuraIpStatus.ENABLEDLAN, KuraInterfaceStatus.ENABLED, null }, //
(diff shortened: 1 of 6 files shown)


pierantoniomerlino:
The code looks ok to me. However, I've some doubts about the meaning of this feature. At the end, the `L2Only` is only another label to mean `Unmanaged`. So, IMO it's more clear and less confusing to the end user use the latter label instead of "L2Only" and document it.
Moreover, the combinations of IPv4 and IPv6 statuses should be clearly documented to avoid confusion.

I understand that in this way the old configurations could not work anymore, but I think that with Kura 6 we can accept a breaking change.

@mattdibi @MMaiero what do you think?

pierantoniomerlino:
> I have only one doubt: currently, if the IPv6 status is UNKNOWN the method won't throw an exception. The original idea was to throw an exception when any of the statuses was UNKNOWN but, if we do with L2ONLY, we might risk breaking existing snapshots? What do you think?

@mattdibi 
Mhh.... `UNKNOWN` as a status means that something went wrong with the network interface. So, IMO is it correct to throw an exception. If one sets an IPv4 interface as `L2ONLY`, the IPv6 will be disabled by default, isn't it?

pierantoniomerlino (dismissed):

mattdibi:
> If one sets an IPv4 interface as L2ONLY, the IPv6 will be disabled by default, isn't it?

Nopes. Kura will just not touch the configuration it finds.

Also note that there's a little quirk with L2ONLY: it is a property **of the interface** and doesn't (or better "shouldn't") have any concept of layer 3 stuff.

MMaiero:
> The code looks ok to me. However, I've some doubts about the meaning of this feature. At the end, the `L2Only` is only another label to mean `Unmanaged`. So, IMO it's more clear and less confusing to the end user use the latter label instead of "L2Only" and document it. Moreover, the combinations of IPv4 and IPv6 statuses should be clearly documented to avoid confusion.
> 
> I understand that in this way the old configurations could not work anymore, but I think that with Kura 6 we can accept a breaking change.
> 
> @mattdibi @MMaiero what do you think?

I agree, I believe we should only have one entry (most likely the unmanaged one) in the UI. The backend should be able to manage the old l2 property and maybe convert it to unmanaged. 
But the user at this point is not were of the difference between the two

mattdibi:
Ok, I will revert back to the 39d7aba0cc1e8a11b54ad1a1effe328584fe0b11 implementation.

This should work because, if IPv6 is missing (like it should in old specific-profiles Kura versions), we have the following bit of logic that fills it in:

https://github.com/eclipse-kura/kura/blob/c0feb0b045ebd76fde1993e8ca2145494a8d156b/kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/NMDbusConnector.java#L459-L463

If, when IPv4 is set to L2ONLY, IPv6 is present in the snapshot **it should match IPv4 (i.e. UNMANAGED) otherwise Kura will throw an exception** (as per our offline discussion with @pierantoniomerlino). Does this look ok for everybody?

After this I will remove the "L2ONLY" entry from the UI.

mattdibi:
After testing I found out that my assumption in https://github.com/eclipse-kura/kura/pull/5663#issuecomment-2616339027 were wrong: if IPv6 is missing from the snapshot it gets automatically `DISABLED`. Therefore I reverted back to the original implementation where setting the IPv4 status as `L2Only` will override IPv6 settings and be interpreted as `UNMANAGED`.

This should not break compatibility with older snapshot.

kura/test/org.eclipse.kura.nm.test/src/test/java/org/eclipse/kura/nm/KuraInterfaceStatusTest.java
```diff
@@ -0,0 +1,75 @@
+package org.eclipse.kura.nm;
```
pierantoniomerlino, line 13: Add the copyright header.

mattdibi: Aaaaaaaaaaaaaaaaaaaaaaaaaa I need to setup that automation ASAP

mattdibi: Addressed in 4a93f050db92c5bd9193913c61ecbf03a775a2d4

pierantoniomerlino (approved):

PR #5666, opened 2025-01-24
feat(core.configuration): Improved snapshot rollback

pierantoniomerlino:
This PR updates the `rollback` method in the `ConfigurationServiceImpl` to make it simpler and avoid configuration updates if not really needed.

**Description of the solution adopted:** The current implementation of the `rollback` method in the `ConfigurationServiceImpl` performs these operations:

- remove all existing factory configurations
- create all factory configurations in snapshot
- rollback configuration for all other components
- apply default configuration for components whose configuration is not in the snapshot

In order to simplify and speed up the process, this PR modifies the `rollback` method to perform these operations:

- if it's a factory component:
    - if it's present, but it is **not** in the snapshot to be applied, delete it
    - if it's **not** present, but it is in the snapshot to be applied, create it
    - if it's present and it is in the snapshot to be applied, check if the properties are different and updated it if needed
- if it's **not** a factory component:
    - if it's present, but it is **not** in the snapshot to be applied, apply the default configuration
    - if it's present and it is in the snapshot to be applied, check if the properties are different and updated it if needed

Moreover, several refactors have been done to the class following the suggestions by sonar.

### DIFF
--- a/kura/test/org.eclipse.kura.core.configuration.test/src/test/java/org/eclipse/kura/core/configuration/util/ComponentUtilTest.java
+++ b/kura/test/org.eclipse.kura.core.configuration.test/src/test/java/org/eclipse/kura/core/configuration/util/ComponentUtilTest.java
@@ -1,3 +1,15 @@
+/*******************************************************************************
+ * Copyright (c) 2025 Eurotech and/or its affiliates and others
+ *
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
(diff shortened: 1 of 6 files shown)


nicolatimeus (dismissed):

kura/test/org.eclipse.kura.core.configuration.test/src/test/java/org/eclipse/kura/core/configuration/util/ComponentUtilTest.java
```diff
@@ -0,0 +1,143 @@
+package org.eclipse.kura.core.configuration.util;
```
MMaiero, line 13: Copyright header

MMaiero (approved):

PR #5670, opened 2025-01-29
docs(6.0): l2only mode is now an alias for unmanaged mode

mattdibi:
This PR documents the changes introduced in #5663


pierantoniomerlino:
@mattdibi does it make sense to add a note about the L2 only mode? Something that alerts the user that if he want to configure an interface in L2 only mode has to set it as unmanaged.

mattdibi:
> @mattdibi does it make sense to add a note about the L2 only mode? Something that alerts the user that if he want to configure an interface in L2 only mode has to set it as unmanaged.

🤔 To that purpose I wrote the following line in the footnote:

> Users still using this mode should migrate to `netIPv4StatusUnmanaged`/`netIPv6StatusUnmanaged`

Do you think we need something more?

pierantoniomerlino (approved):

PR #5701, opened 2025-02-04
perf(http.server.manager): Start Jetty in a separate thread

pierantoniomerlino:
In order to speed up the Kura start, this PR adds an `Executor` to start the Jetty HTTP Server.

### DIFF
--- a/kura/org.eclipse.kura.http.server.manager/src/main/java/org/eclipse/kura/http/server/manager/HttpService.java
+++ b/kura/org.eclipse.kura.http.server.manager/src/main/java/org/eclipse/kura/http/server/manager/HttpService.java
@@ -15,6 +15,9 @@ package org.eclipse.kura.http.server.manager;
 import java.util.EventListener;
 import java.util.Map;
 import java.util.Optional;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.TimeUnit;
 
 import org.eclipse.kura.configuration.ConfigurableComponent;
 import org.eclipse.kura.configuration.ConfigurationService;
@@ -42,6 +45,7 @@ public class HttpService implements ConfigurableComponent, EventHandler {
     private String keystoreServicePid;
 
     private JettyServerHolder jettyServerHolder;
+    private ExecutorService executorService = Executors.newSingleThreadExecutor();
 
     public void setSystemService(SystemService systemService) {
         this.systemService = systemService;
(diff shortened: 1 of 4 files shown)


kura/org.eclipse.kura.http.server.manager/src/main/java/org/eclipse/kura/http/server/manager/HttpService.java
```diff
@@ -97,22 +101,33 @@ private synchronized void restartHttpService() {
...
-            this.jettyServerHolder = new JettyServerHolder(this.options, Optional.ofNullable(this.keystoreService),
-                    this.dispatcherServlet, this.eventListener);
-            logger.info("starting Jetty instance...done");
-        } catch (final Exception e) {
-            logger.error("Could not start Jetty Web server", e);
-        }
+        this.executorService.submit(() -> {
+            try {
+                logger.info("starting Jetty instance...");
+                this.jettyServerHolder = new JettyServerHolder(this.options, Optional.ofNullable(this.keystoreService),
```
nicolatimeus: To avoid race conditions we can try passing an immutable copy of the current options instead of referencing this.options directly. The activateHttpService method can be modified to take the options as argument and having the activate() and update() method passing the new options to this one.

kura/org.eclipse.kura.http.server.manager/src/main/java/org/eclipse/kura/http/server/manager/HttpService.java
```diff
@@ -97,22 +101,33 @@ private synchronized void restartHttpService() {
...
+            }
+        });
     }
 
     private synchronized void deactivateHttpService() {
         try {
             logger.info("stopping Jetty instance...");
-            this.jettyServerHolder.stop();
+            if (this.jettyServerHolder != null) {
+                this.jettyServerHolder.stop();
```
nicolatimeus: We should run this in the executor thread as well

kura/org.eclipse.kura.http.server.manager/src/main/java/org/eclipse/kura/http/server/manager/HttpService.java
```diff
@@ -97,22 +101,33 @@ private synchronized void restartHttpService() {
...
+            }
+            this.executorService.shutdown();
```
nicolatimeus, line 126: We should probably shutdown the executor and wait its termination in the deactivate() method instead of here.

kura/org.eclipse.kura.http.server.manager/src/main/java/org/eclipse/kura/http/server/manager/HttpService.java
pierantoniomerlino: Do we need to stop the server in a separate thread? I mean, maybe if we wait a bit longer for the deactivation is not a big issue...

nicolatimeus (approved):

PR #5703, opened 2025-02-05
feat: Add support for SCR and metatype annotations

nicolatimeus:

* Adds `org.osgi.service.component.annotations` and  `org.osgi.service.metatype.annotations` to the target platform, these bundles provide compile time annotations and should not be added to the distribution.
* Adds the `bnd-maven-plugin` `bnd-process` execution to plugin management, that will create declarative services and metatype xml files basing on the annotations
* Updates example publisher and org.eclipse.kura.core bundle to use annotations

The execution of `bnd-maven-plugin` is not enabled by default for now, in order to be able to use annotations in a specific bundle, perform the following steps:

* Add the following to the bundle pom file

```xml
<build>
    <plugins>
        <plugin>
            <groupId>biz.aQute.bnd</groupId>
            <artifactId>bnd-maven-plugin</artifactId>
        </plugin>
     </plugins>
</build>
```

* Add org.osgi.service.component.annotations` and/or `org.osgi.service.metatype.annotations` as `additional.bundles` in build.properties

```
additional.bundles = org.osgi.service.component.annotations,\
                     org.osgi.service.metatype.annotations
```

SCR and metatype annotations are compile time only, the annotation packages should not be referenced in the MANIFEST.MF files.


target-platform/p2-repo-common/pom.xml
```diff
@@ -854,6 +854,16 @@
                                     <artifactId>mimepull</artifactId>
                                     <version>${mimepull.version}</version>
                                 </artifactItem>
+                                <artifactItem>
+                                    <groupId>org.osgi</groupId>
+                                    <artifactId>org.osgi.service.component.annotations</artifactId>
+                                    <version>${org.osgi.service.component.annotations.version}</version>
+                                </artifactItem>
+                                <dependency>
```
marcellorinaldo: `dependency` -> `artifactItem `

salvatore-coppola (approved):

marcellorinaldo (approved):

PR #5706, opened 2025-02-07
refactor(archetype): modified archetype to use SCR annotations

marcellorinaldo:
This PR modifies the archetype used to generate Kura addons to compile with SCR annotations.

**Related Issue:** N/A.

**Screenshots:** N/A.

**Manual Tests**: N/A.


kura/tools/kura-addon-archetype/src/main/resources/archetype-resources/bundles/__artifactId__.bundle/src/main/java/ExampleComponent.java
```diff
@@ -15,13 +15,26 @@
...
+import org.osgi.service.component.annotations.ConfigurationPolicy;
+import org.osgi.service.component.annotations.Deactivate;
+import org.osgi.service.component.annotations.Modified;
+import org.osgi.service.metatype.annotations.Designate;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+@Component(immediate = true, //
+    configurationPolicy = ConfigurationPolicy.REQUIRE, //
+    service = ConfigurableComponent.class, //
```
gnunzi: It is not necesseary to specify `service`. The services are taken from the `implements`.

I find it also intuitive to just let all implemented interfaces be osgi services, so my suggestion would be to leave it from the archetype 😃 

See Javadoc of Component
> If not specified, the service types for this Component are all the <i>directly</i> implemented interfaces of the class being annotated.

... btw looking at the javadoc the value should be an array.

kura/tools/kura-addon-archetype/src/main/resources/archetype-resources/bundles/__artifactId__.bundle/src/main/java/ExampleComponent.java
```diff
@@ -15,13 +15,26 @@
...
+    property = {"kura.service.pid=${package}.ExampleComponent"} //
+)
+@Designate(ocd = ExampleComponentOCD.class, factory = false)
 public class ExampleComponent implements ConfigurableComponent {
 
     private static final Logger logger = LoggerFactory.getLogger(ExampleComponent.class);
 
+    @Activate
     public void activate(final Map<String, Object> properties) {
```
gnunzi, line 49: I find it very useful to pass the ocd as well.
```
//you can use properties and configFromUI or properties only or configFromUI only
protected void activate (final Map<String, Object> properties, ExampleComponentOCD  configFromUI)
```

I would actually suggest to pass the ComponentContext as well.
```
//all parameters in activate, modified, deactivate are optional and can be removed if not needed
protected void activate (ComponentContext componentContext, final Map<String, Object> properties, final ExampleComponentOCD  configFromUI)
```
Imo this extra-info can give some more insights to new-comers using the archetype 😃

kura/tools/kura-addon-archetype/src/main/resources/archetype-resources/bundles/__artifactId__.bundle/src/main/java/ExampleComponent.java
```diff
@@ -15,13 +15,26 @@
...
         logger.info("Activating");
```
gnunzi, line 51: A suggestion is to add the top-level catch:

```
try {//be sure to catch all top-level errors in the component
  //logic of the component
} catch (Exception e) {
  logging.error("Error during activation: "+e.getClass());
}
```

marcellorinaldo: Thanks for the suggestions, those are really valuable!

I wouldn't insert this since one can rely on the fact that exceptions thrown in the hook methods can be managed by the OSGi framework

kura/tools/kura-addon-archetype/src/main/resources/archetype-resources/bundles/__artifactId__.bundle/src/main/java/ExampleComponent.java
marcellorinaldo: Great, thanks!

kura/tools/kura-addon-archetype/src/main/resources/archetype-resources/bundles/__artifactId__.bundle/src/main/java/ExampleComponent.java
marcellorinaldo: I'm adding a comment about this to explain it better

marcellorinaldo:
Hi @gnunzi, thanks for the valuable suggestions! Are you able to import the project generated in the archetype into Eclipse? I am currently struggling to get the "Update Maven Projects" IDE feature to work, every time I use it the plugin dependencies are not recognized

gnunzi:
@marcellorinaldo I am afriad I am not using Eclipse, so it is difficult to reproduce. Can you please tell me me command to create a bundle from the archetype?

marcellorinaldo:
> @marcellorinaldo I am afriad I am not using Eclipse, so it is difficult to reproduce. Can you please tell me me command to create a bundle from the archetype?

Sure, the command is:

```
mvn archetype:generate \
-DarchetypeArtifactId=kura-addon-archetype \
-DarchetypeGroupId=org.eclipse.kura \
-DarchetypeVersion="6.0.0-SNAPSHOT"
```

Since we are using the latest kura snapshot artifacts in the target definition, you might need to run first a kura build to install the artifacts in your local .m2 repository before compiling the generated project.

I have added some documentation here on how tu run the archetype: https://eclipse-kura.github.io/kura/docs-develop/java-application-development/kura-addon-archetype/

pierantoniomerlino:
> 
> Sure, the command is:
> 
> ```
> mvn archetype:generate \
> -DarchetypeArtifactId=kura-addon-archetype \
> -DarchetypeGroupId=org.eclipse.kura \
> -DarchetypeVersion="8.0.0-SNAPSHOT"
> ```
> 

The correct version is `6.0.0-SNAPSHOT`

pierantoniomerlino (dismissed):

pierantoniomerlino (approved):

PR #5711, opened 2025-02-11
fix(container.provider): removed useless unset method

sfiorani:
This PR removes the useless method to unbind the configurationService from a ContainerInstance, when the latter is deleted from the webUI.

That method was wrong, as the signature didn't contain the required parameter (eg: `final ConfigurationService confService`):

`public synchronized void unsetConfigurationService() {`
`    this.configurationService = null;`
`}`

This caused an error, visible also in the logger:
>   deactivate...done
 !ENTRY org.eclipse.kura.container.provider 4 0 2025-02-11 12:47:54.264
 !MESSAGE bundle org.eclipse.kura.container.provider:2.0.0.202502050814 (169)[org.eclipse.kura.container.provider.ContainerInstance(120)] : unbind method [unsetConfigurationService] not found; Component will fail
 !ENTRY org.eclipse.kura.container.provider 4 0 2025-02-11 12:47:54.265
 !MESSAGE bundle org.eclipse.kura.container.provider:2.0.0.202502050814 (169)[org.eclipse.kura.container.provider.ContainerInstance(120)] : unbind method [unsetConfigurationService] not found
 Container signature validation service class com.eurotech.framework.container.signature.validation.cosign.provider.CosignContainerValidationService removed.

Adding the parameter would already fix the problem, but being that method pretty useless, it is better to just remove it.


sfiorani:
Tested on Raspberry Pi 5, Kura aarch64

mattdibi (approved):

PR #5713, opened 2025-02-11
docs: Fix download URL to reference Github releases

MMaiero:
Fixes https://github.com/eclipse-kura/kura/issues/5712

### DIFF
--- a/docs/getting-started/install-kura.md
+++ b/docs/getting-started/install-kura.md
@@ -1,6 +1,6 @@
 # Install Kura
 
-Eclipse Kura&trade; is provided using a Debian Linux package. Visit the [Kura download page](https://github.com/eclipse-kura/kura/releases/latest) to find the correct installation file for the target system.
+Eclipse Kura&trade; is provided using a Debian Linux package. Visit the [Kura download page](https://github.com/eclipse-kura/kura/releases) to find the correct installation file for the target system.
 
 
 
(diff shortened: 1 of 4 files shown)


docs/getting-started/install-kura.md
```diff
@@ -1,6 +1,6 @@
 # Install Kura
 
-Eclipse Kura&trade; is provided using a Debian Linux package. Visit the [Kura download page](https://www.eclipse.org/kura/downloads.php) to find the correct installation file for the target system.
+Eclipse Kura&trade; is provided using a Debian Linux package. Visit the [Kura download page](https://github.com/eclipse-kura/kura/releases/latest) to find the correct installation file for the target system.
```
mattdibi: Since we want to backport this, is it correct to alway point the URL to the **latest** release? Shouldn't older releases point to [https://github.com/eclipse-kura/kura/releases/](https://github.com/eclipse-kura/kura/releases/) instead?

MMaiero: Yes, thought at that. But I also thought that we will forget to update the reference if we don't have a some sort of automation in place

mattdibi: That's why I proposed the use of the `/releases` URL. It will provide all the releases and the user will be able to pick the one he/she wants. No need to update the URL. That page also highlights the "latest" release.

If this still doesn't work for you `latest` is ok ofc

mattdibi (approved):

PR #5725, opened 2025-02-13
build: Migration of compiler target and source to java17 and wire.script.filter and the conditional component removal

salvatore-coppola:
Migration of compiler target and source to java17.
Updated also Require-capability in `META-INF.MF` files.


kura/examples/scenarios/org.eclipse.kura.example.ai/snapshots/graph-anomaly-detector.xml
```diff
@@ -399,66 +402,6 @@
             </esf:property>
         </esf:properties>
     </esf:configuration>
-    <esf:configuration pid="script">
```
marcellorinaldo, line 402: we should replace this with the wire.script.tools implementation

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/net/IPAddress.java
```diff
@@ -81,6 +83,10 @@ public static IPAddress getByAddress(int addr) throws UnknownHostException {
      * @return
      */
     public static IPAddress parseHostAddress(String hostAddress) throws UnknownHostException {
+        if (!hostAddress.isEmpty() && !InetAddresses.isInetAddress(hostAddress)) {
```
marcellorinaldo, line 86: Why is this needed?

kura/pom.xml
```diff
@@ -204,14 +205,15 @@
                     -Dkura.home=${project.basedir}/target/kura
                     -Dkura.plugins=${project.basedir}/target/kura/plugins
                     -Dkura.packages=${project.basedir}/target/kura/data/packages
-                    -Dkura.data=${project.basedir}/target/kura
+                    -Dkura.data=${project.basedir}/target/kura/data
```
marcellorinaldo: I think java 8 profile can be removed here, and the Java 17 settings can be set as default configuration in the plugin

kura/tools/kura-addon-archetype/src/main/resources/archetype-resources/bundles/pom.xml
```diff
@@ -35,8 +37,7 @@
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <tycho-version>4.0.10</tycho-version>
 
-        <!-- Use this BND version as it will generate v1.2.0 DS descriptions suitable for Equinox 3.11.1 -->
-        <bnd-version>4.1.0</bnd-version>
+        <bnd-version>7.1.0</bnd-version>
```
marcellorinaldo: revert this, it is touched in https://github.com/eclipse-kura/kura/pull/5706

target-platform/pom.xml
```diff
@@ -28,8 +28,8 @@
         <kura.addons.url>https://raw.github.com/eurotech/kura_addons/mvn-repo/</kura.addons.url>
         <skip.libsocket-can>true</skip.libsocket-can>
 
-        <maven.compiler.source>1.8</maven.compiler.source>
-        <maven.compiler.target>1.8</maven.compiler.target>
+        <maven.compiler.source>17</maven.compiler.source>
+        <maven.compiler.target>17</maven.compiler.target>
```
marcellorinaldo, line 32: We should remove Java 8 profile from this pom and set the java17 one as default settings of the plugin

marcellorinaldo (commented):
I would remove the wire.script.filter and the conditional component in separate PRs so that the change is more visible in the release notes

kura/distrib/pom.xml
```diff
@@ -32,7 +29,7 @@
     <name>distrib</name>
     <url>http://maven.apache.org</url>
     <properties>
-        <tycho-version>1.6.0</tycho-version>
+        <tycho-version>3.0.5</tycho-version>
```
marcellorinaldo, line 33: Please remove the java8 profile down in this pom. `skip.js.nashorn.engine.components` should not be needed anymore

kura/test/pom.xml
```diff
@@ -66,18 +66,6 @@
                 </pluginManagement>
             </build>
         </profile>
-
```
marcellorinaldo, line 69: Remove also the java8 profile

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/net/IPAddress.java
salvatore-coppola: This may fail in certain host configurations.

kura/examples/scenarios/org.eclipse.kura.example.ai/snapshots/graph-anomaly-detector.xml
salvatore-coppola: the test pass without require it

kura/test/pom.xml
salvatore-coppola: where is the profile ?

salvatore-coppola:
> I would remove the wire.script.filter and the conditional component in separate PRs so that the change is more visible in the release notes

i'll add a message in the title

kura/examples/scenarios/org.eclipse.kura.example.ai/snapshots/graph-anomaly-detector.xml
salvatore-coppola: ok but it's out of scope of this PR. I'll item an item in the backlog

kura/test/pom.xml
marcellorinaldo: Sorry, wrong file

marcellorinaldo (commented):
In `kura/distrib/pom.xml` there are still references to `skip.js.nashorn.engine.components`

marcellorinaldo (approved):

PR #5733, opened 2025-02-20
fix(api): Add nullcheck to IpAddress##parseHostAddress

salvatore-coppola:
Add nullcheck to IpAddress##parseHostAddress


marcellorinaldo (approved):
It works, null pointer is gone

PR #5737, opened 2025-02-21
fix(web2): Fixed whiteboard services unregistration and context managing

salvatore-coppola:
This PR:
- fixes the process of whiteboard services unregistration and add some debug logs.
- add the suggested Scope to resources and servlets as described [here](https://vogella.com/blog/build-rest-services-with-osgi-jakarta-rs-whiteboard/#interlude-prototype-scope)
- change the `ServletContextHelper` naming to force context destruction.


nicolatimeus:
What is the purpose of setting the [service.scope](https://docs.osgi.org/specification/osgi.core/7.0.0/framework.api.html#org.osgi.framework.Constants.SERVICE_SCOPE) property? In my opinion in this case it has no effect since the services are registered manually and there will be a single instance of them for each resource, it would have some effect if the services implement PrototypeServiceFactory or ServiceFactory (which is not the case) or if the services were managed by DS

salvatore-coppola:
> What is the purpose of setting the [service.scope](https://docs.osgi.org/specification/osgi.core/7.0.0/framework.api.html#org.osgi.framework.Constants.SERVICE_SCOPE) property? In my opinion in this case it has no effect since the services are registered manually and there will be a single instance of them for each resource, it would have some effect if the services implement PrototypeServiceFactory or ServiceFactory (which is not the case) or if the services were managed by DS

It certainly has no effect but as the article says it is not wrong specify that a servlet has a prototype scope aligned with the Jakarta-RS design.
BTW i'll remove it.

nicolatimeus (approved):

PR #5742, opened 2025-02-24
style(demo.bnd.heater): converted file to unix End-of-line

salvatore-coppola:
Converted file to unix End-of-line


MMaiero:
Forcing merge as the CI will not be influenced by this change.

mattdibi:
To avoid this issue in the future, could it be a good idea to add the `.gitattributes` configuration file with the proper eol?

See: https://stackoverflow.com/questions/10418975/how-to-change-line-ending-settings

```.gitattributes
* text eol=lf
```

mattdibi:
Update: it already exists... it's not clear to me how this could have happened at this point.

PR #5743, opened 2025-02-24
docs: Updated README.md

salvatore-coppola:
Updated Maven requirement. Tycho 4 needs Maven 3.9.x


mattdibi (changes requested):
I would update the bash command at line 100

https://github.com/eclipse-kura/kura/blob/aad72dc128a6a0262d3879a485315b260a3e80eb/README.md?plain=1#L100

mattdibi (approved):

PR #5747, opened 2025-02-24
style(distrib): fixed warning on regexp during kura startup

salvatore-coppola:
Removed unsupported `^`

### DIFF
--- a/kura/distrib/src/main/resources/common/gen_config_ini.sh
+++ b/kura/distrib/src/main/resources/common/gen_config_ini.sh
@@ -41,7 +41,7 @@ do
         continue
     fi
 
-    if ! expr "${DIR_NAME}" : '^[0-9]\{1,\}s\{0,1\}$' > /dev/null
+    if ! expr "${DIR_NAME}" : '[0-9]\{1,\}s\{0,1\}$' > /dev/null
     then
         continue
     fi


kura/distrib/src/main/resources/common/gen_config_ini.sh
```diff
@@ -41,7 +41,7 @@ do
         continue
     fi
 
-    if ! expr "${DIR_NAME}" : '^[0-9]\{1,\}s\{0,1\}$' > /dev/null
+    if ! expr "${DIR_NAME}" : '[0-9]\{1,\}s\{0,1\}$' > /dev/null
```
mattdibi, line 44: I don't fully agree with this change since it changed the meaning of the regex.

`^[0-9]\{1,\}s\{0,1\}$`:
- `5s` would match
- `5` would match
- `pippo5s` would not

`[0-9]\{1,\}s\{0,1\}$`
- `5s` would match
- `5` would match
- `pippo5s` would also match

Why is Java throwing a fit over `^`? It's a common anchor... am I missing something?

mattdibi: @nicolatimeus what do you think?

salvatore-coppola: It's not java complaining but expr, it says it is not portable and it will be ignored.

However `info expr` states:

_info expr (the GNU version) states that "The arguments are converted to strings and the second is considered to be a (basic, a la GNU ‘grep’) regular expression, with a ‘^’ implicitly prepended"._

So we can remove it safely.

```
2025-02-24 12:23:40 expr: warning: '^[0-9]\{1,\}s\{0,1\}$': using '^' as the first character
2025-02-24 12:23:40 of a basic regular expression is not portable; it is ignored
2025-02-24 12:23:40 expr: warning: '^[0-9]\{1,\}s\{0,1\}$': using '^' as the first character
2025-02-24 12:23:40 of a basic regular expression is not portable; it is ignored
2025-02-24 12:23:40 expr: warning: '^[0-9]\{1,\}s\{0,1\}$': using '^' as the first character
2025-02-24 12:23:40 of a basic regular expression is not portable; it is ignored
2025-02-24 12:23:40 expr: warning: '^[0-9]\{1,\}s\{0,1\}$': using '^' as the first character
2025-02-24 12:23:40 of a basic regular expression is not portable; it is ignored
2025-02-24 12:23:40 expr: warning: '^[0-9]\{1,\}s\{0,1\}$': using '^' as the first character
2025-02-24 12:23:40 of a basic regular expression is not portable; it is ignored
2025-02-24 12:23:40 expr: warning: '^[0-9]\{1,\}s\{0,1\}$': using '^' as the first character
2025-02-24 12:23:40 of a basic regular expression is not portable; it is ignored
2025-02-24 12:23:40 + '[' -f /opt/eclipse/kura/user/security/httpskeystore.ks ']'
```

mattdibi: > ‘^’ implicitly prepended

Now I understand. Thanks for the explanation @salvatore-coppola. Therefore LGTM

mattdibi (approved):

salvatore-coppola:
@MMaiero if you can force the merge of this, a full build is useless

PR #5748, opened 2025-02-26
build: removed Apache Maven Eclipse Plugin

salvatore-coppola:
Removed Apache Maven Eclipse Plugin


marcellorinaldo (changes requested):
Build fails because of missing `bin.includes` in some examples project:

```
org.eclipse.kura.example.publisher/build.properties: bin.includes value(s) [OSGI-INF/] do not match any files.
```

marcellorinaldo (approved):

PR #5778, opened 2025-04-03
refactor: refactor of addon archetype

sfiorani:
This PR works on the Kura Addon Archetype doing the following:

- fix of wrong path in kura/tools/kura-addon-archetype/pom.xml
- reorganization of the target directory of generetate output of DEB/RPM installers: they are grouped in specific folders in target (`target/deb` for DEB, `target/rpm` for RPM)
- reorganization of the target directory of source code jars copied and renamed from `bundles` to `distrib`: they're now copied in a specific folder `target/input_file`
- addition of required unit in the tests pom


pierantoniomerlino (dismissed):

salvatore-coppola (dismissed):

kura/tools/kura-addon-archetype/src/main/resources/archetype-resources/distrib/pom.xml
```diff
@@ -77,7 +79,7 @@
                         </goals>
                         <configuration>
                             <sourceFile>${project.basedir}/../bundles/${jar.name}/target/${jar.name}-${project.version}.jar</sourceFile>
-                            <destinationFile>target/${jar.name}_${project.version}.jar</destinationFile>
+                            <destinationFile>target/input_files/${jar.name}_${project.version}.jar</destinationFile>
```
marcellorinaldo, line 82: Instad of specifying each JAR, can we use an option for specifying the source directory? So everything that resides in input_files is picked up (like if you have multiple jars produced in bundles)

marcellorinaldo (changes requested):

kura/tools/kura-addon-archetype/src/main/resources/archetype-resources/distrib/pom.xml
sfiorani: I have separated executions because we're also changing the name of the output jar

marcellorinaldo: Yes, you may have multiple executions of the `copy-rename-maven-plugin`, one for each bundle, that copy the correct JARs in `input_files`. Then you don't need to define a `<data>` entry for each bundle, you can just pick up the whole `input_files` folder. In this way, with multiple bundles you will just need to add executions of `copy-rename-maven-plugin`

See documentation for `jdeb`, you could use `<type>directory</type>`. Something similar should be available also for `rpm-maven-plugin`.

marcellorinaldo (approved):

PR #5779, opened 2025-04-03
chore: Use gpsd provider by default in PositionService

nicolatimeus:


MMaiero:
Do we need to modify the firewall to open port 2947? https://github.com/eclipse-kura/kura/blob/cf2ba93d9f5d0226b7a9cdda9439c1f6341ee654/kura/distrib/src/main/resources/common/snapshots/template_firewall_eth#L4

nicolatimeus:
> Do we need to modify the firewall to open port 2947?
> 
> https://github.com/eclipse-kura/kura/blob/cf2ba93d9f5d0226b7a9cdda9439c1f6341ee654/kura/distrib/src/main/resources/common/snapshots/template_firewall_eth#L4

There should be no need for that if gpsd is running locally

MMaiero (approved):

PR #5780, opened 2025-04-04
chore: Upgraded protobuf to 3.25.6

MMaiero:


mattdibi:
@MMaiero the CI build keeps failing due to an OOM error, from what I can tell. We might need to take a look into it.

PR #5787, opened 2025-04-16
chore: added target-definition artifact

sfiorani:
This PR aims to create a new target-definition artifact to be used in the Kura enviroment, removing the local link to files and plugins.

The new artifact can be easily updated modifying the .properties file in the project, and it could be used also outside of the Kura project, for example in the addons.


sfiorani:
Huge number of lines are due to a wrong formatter. In a future commit i'll restore the formatting so only the added lines will be shown in the diff

sfiorani:
Restored formatting and style. Putting the PR in ready for review.
Still a high number of line, due to the addition of an entire .target file.

kura/distrib/kura-target-definition/pom.xml
```diff
@@ -0,0 +1,56 @@
...
+
+	<parent>
+		<groupId>org.eclipse.kura</groupId>
+		<version>6.0.0-SNAPSHOT</version>
+		<artifactId>distrib</artifactId>
+	</parent>
+
+	<artifactId>kura-target-definition.target</artifactId>
+	<packaging>eclipse-target-definition</packaging>
+	<name>Target Definition for Kura Enviroment</name>
```
MMaiero:
```suggestion
	<name>Target Definition for Eclipse Kura Environment</name>
```

kura/distrib/pom.xml
```diff
@@ -1950,5 +1961,72 @@
             </build>
         </profile>
 
+        <profile>
+            <properties>
+                <kura.skip.if.target.definition.profile>true</kura.skip.if.target.definition.profile>
```
MMaiero: why we need to skip this and not include in the build all?

I would also expect the artifacts produced to be in a target folder, isn't it?

MMaiero (approved):

MMaiero (approved):

MMaiero:
Forcing merge. Currently the Github infrastructure seems down as the Validate PR title is not responding. It should have no effect on the result.

PR #5789, opened 2025-04-23
fix: Made web2 deps on internal networking, wires, asset and driver p…

nicolatimeus:
…ackages optional

* Web ui dependencies on the `org.eclipse.kura.asset.provider`, `org.eclipse.kura.core.net`, `org.eclipse.kura.core.net.util` and `org.eclipse.kura.internal.wire.asset` are now optional.
* Driver and Assets section will be shown only if the `DriverService` and `DriverHelperService` services are available
* Wires section will be shown only if the `WireGraphService` and `WireComponentDefinitionService` services are available
* Asset management features (asset creation, channel configuration, read/write channels) will only be enabled if the following packages are available: `org.eclipse.kura.asset.provider`, `org.eclipse.kura.internal.wire.asset `


MMaiero:
Tested successfully in my environment.
Currently the Eclipse build infrastructure is broken.
Forcing merge for now to unblock other activities. Will be fully tested and validated afterwards

PR #5820, opened 2025-05-22
chore: Moved bundles in test-util

pierantoniomerlino:
This PR moves the following bundles in `test-util`:

- `org.eclipse.kura.core.testutil`
- `org.eclipse.kura.test`
- `org.eclipse.kura.util.test.driver`

In this way they will be deployed and available to different repos.


pierantoniomerlino:
The copyright check fails since the year is not set to 2025 in several files. However, I think this is a false positive because we only moved files.

nicolatimeus (approved):

PR #5831, opened 2025-05-29
feat(nm): Implemented DNS priority

pierantoniomerlino:
This PR expands the Kura networking capabilities, adding the possibility to set the DNS servers priority. It exploits the `ipv4.dns-priority` and `ipv6.dns-priority` properties provided by NetworkManager (see [here](https://networkmanager.dev/docs/api/latest/settings-ipv4.html) for further details).

The `WAN Priority` property provided by Kura is used to configure both the `route-metric` and the `dns-priority`, so NM will use the DNS provided by the connection with higher priority. The following tables summarize the values assigned to the two parameters according to the `WAN Priority`:

| WAN Priority | route-metric | dns-priority ||
|--------------|--------------|-------------|-|
|-1|-1|0|Disabled|
|0|1|1|Minimal value|
|1|1|1|Minimal value|
|>0|>0|>0|Enabled|
|<0|||Not allowed|

This behavior is needed to keep the consistency: for example setting 0 for the `dns-priority` in IPv6 means a value of 1024.
Moreover, even if negative values are allowed for the `dns-priority` property, it is not supported.

Finally, the behavior of this feature depends on the tool used by NM to configure the DNS servers. See [here](https://docs.redhat.com/en/documentation/red_hat_enterprise_linux/8/html/configuring_and_managing_networking/configuring-the-order-of-dns-servers_configuring-and-managing-networking#default-values-of-dns-priority-parameters_configuring-the-order-of-dns-servers) for details.


mattdibi (approved):

kura/org.eclipse.kura.nm/src/main/java/org/eclipse/kura/nm/configuration/NMSettingsConverter.java
```diff
@@ -305,17 +305,30 @@ public static Map<String, Variant<?>> buildIpv4Settings(NetworkProperties props,
...
         return settings;
     }
 
+    private static void setWanPriority(Map<String, Variant<?>> settings, Integer value) {
+        if (value == -1) {
+            settings.put("route-metric", new Variant<>((long) -1));
+            settings.put("dns-priority", new Variant<>((long) 0));
+        } else if (value == 0) {
+            settings.put("route-metric", new Variant<>((long) 1));
+            settings.put("dns-priority", new Variant<>((long) 1));
```
mattdibi: What happens if a user sets an interface to `0` and another to `1`? NM will see the same values. This is a surprising behaviour.

mattdibi (changes requested):

pierantoniomerlino:
As discussed, I implemented a new method for assigning the `route-metric` and `dns-priority`.

kura/org.eclipse.kura.web2/src/main/resources/org/eclipse/kura/web/client/messages/Messages.properties
```diff
@@ -360,7 +360,7 @@ netIPv4ToolTipStatusUnmanaged=Interface is not managed by Kura
...
 netIPv4ToolTipStatusEnabledWAN=Interface is enabled for use as a gateway to an external network. Use this to connect to a router, wireless access point, or cellular network.
-netIPv4ToolTipPriority=Priority used to determine which interface select as primary WAN.<br><br>In case the primary WAN interface loses connection, then the next highest priority interface is selected.<br><br>Lower values correspond to higher priority.<br><br>If this field is set to an empty value or to -1, it means that the metric is chosen automatically based on the device type.
+netIPv4ToolTipPriority=Priority used to determine which interface select as primary WAN.<br><br>In case the primary WAN interface loses connection, then the next highest priority interface is selected.<br><br>Lower values correspond to higher priority.<br><br>If this field is set to an empty value or to -1, it means that the metric is chosen automatically based on the device type.<br><br>Note that when an interface is selected as WAN according to its priority, also the corresponding DNS servers will be used.
```
mattdibi: I have to admit that this is quite hard to understand... May I suggest something like:

```suggestion
netIPv4ToolTipPriority=Priority used to determine which interface select as primary WAN.<br><br>In case the primary WAN interface loses connection, then the next highest priority interface is selected.<br><br>Lower values correspond to higher priority.<br><br>If this field is set to an empty value or to -1, it means that the metric is chosen automatically based on the device type.<br><br>Note that WAN priority also affects DNS servers selection. Therefore, when an interface is selected according to its WAN priority, the associated DNS server will be used.
```

mattdibi (commented):
Code LGTM. I suggested just a minor clarification in the UI description. Let me know what you think.

mattdibi (approved):

PR #5844, opened 2025-06-06
feat(cloudconnection): Made NetworkService optional for cloudconnections

pierantoniomerlino:
This PR makes the dependency from the `NetworkService` optional for the following cloudconnections:

- `org.eclipse.kura.cloudconnection.kapua.mqtt.provider`
- `org.eclipse.kura.cloudconnection.eclipseiot.mqtt.provider`


MMaiero (approved):

MMaiero:
This change applies to old and existing code without modifications to the current behaviour.
Accepting for now.

PR #5850, opened 2025-06-11
build: use BOM POMs in build

marcellorinaldo:
- Create BOM projects for the target platform and kura builds that declare which dependencies and bundles will be available when used in a target definition
- Modified Kura target definition to reference the new BOMs
- Modified distrib to use the new BOMs: removed the convention to have `_` as separator between bundle name and version, now we are using `-` as Maven standard. It simplifies the build projects in `distrib`
- Removed unecessary profiles
- Removed `kura-target-definition` build from `distrib`
- Tested aarch64 and docker ubi8 profiles


salvatore-coppola (dismissed):
LGTM

salvatore-coppola (dismissed):

kura/distrib/src/main/ant/build_equinox_distrib.xml
```diff
@@ -19,13 +19,7 @@
...
             <union>
                 <fileset dir="${project.build.directory}/plugins">
-                    <include name="${bundle.name}_${bundle.version}*.jar" />
-                </fileset>
-                <fileset dir="../target-definition/common/source/plugins">
-                    <include name="${bundle.name}-${bundle.version}*.jar" />
-                </fileset>
-                <fileset dir="${kura.osgi.repo}/plugins/">
-                    <include name="${bundle.name}-${bundle.version}*.jar" />
+                    <include name="${bundle.name}*.jar" />
```
nicolatimeus: This could match multiple bundles if their name has a common prefix, we could at least add '-' or '_' (the one that is correct) after the name and make the build fail if the fileset contains more than one entry

marcellorinaldo: Addressed in [408a2fc](https://github.com/eclipse-kura/kura/pull/5850/commits/408a2fcd284f63f12ed25c96a67b80f5ae680f65)

nicolatimeus (approved):

Issue #5851, opened 2025-06-12
Endless loop in modem configuration


mattdibi:
### Discussed in https://github.com/eclipse-kura/kura/discussions/5849

<div type='discussions-op-text'>

<sup>Originally posted by **fdizazzo** June 11, 2025</sup>
> 

> Hi everyone! :)

I got a weird issue with a modem on a RaspberryPi running Kura 5.6.0: sometimes when starting up, network configuration gets stuck endlessly rollbacking my network configuration. 

Log repeats as follows (debug only enabled on org.eclipse.kura.nm)
```
2025-06-06T20:02:53,226` [DBus Worker Thread-1] DEBUG o.e.k.n.ModemManagerDbusWrapper - Modem location setup [MM_MODEM_LOCATION_SOURCE_GPS_RAW, MM_MODEM_LOCATION_SOURCE_GPS_NMEA] for modem /org/freedesktop/ModemManager1/Modem/0
2025-06-06T20:02:53,228 [DBus Worker Thread-1] DEBUG o.e.k.n.NMDbusConnector - Network configuration enforcement set to true (Expected: true)
2025-06-06T20:02:55,978 [DBus Worker Thread-4] DEBUG o.e.k.n.s.h.NMConfigurationEnforcementHandler - Device state change detected: NM_DEVICE_STATE_PREPARE -> NM_DEVICE_STATE_CONFIG (reason: NM_DEVICE_STATE_REASON_NONE), for device /org/freedesktop/NetworkManager/Devices/4
2025-06-06T20:02:55,979 [DBus Worker Thread-4] INFO  o.e.k.n.s.h.NMConfigurationEnforcementHandler - Network change detected on interface /org/freedesktop/NetworkManager/Devices/4. Roll-back to cached configuration
2025-06-06T20:02:55,978 [DBus Worker Thread-3] DEBUG o.e.k.n.s.h.NMModemResetHandler - Modem state change detected: NM_DEVICE_STATE_PREPARE -> NM_DEVICE_STATE_CONFIG, for device /org/freedesktop/NetworkManager/Devices/4
2025-06-06T20:02:55,980 [DBus Worker Thread-2] DEBUG o.e.k.n.s.h.NMModemResetHandler - Modem state change detected: NM_DEVICE_STATE_CONFIG -> NM_DEVICE_STATE_IP_CONFIG, for device /org/freedesktop/NetworkManager/Devices/4
[...]

PR #5858, opened 2025-06-19
feat(nm): Made network configuration timeout configurable

pierantoniomerlino:
To address the issue #5851, we decided to create a new property for setting the timeout for the application of a new network configuration. SInce it is an advanced property (a user tipically shouldn't modify it), it is put in the `kura.properties` file.
The default value is 30s.

A side effect of this is a potentially long wait time after a configuration change. Indeed, the methods used for the application of the network configuration are synchronous, so the user has to wait the timeout if the connection fails. 
So, this PR implements some asynchronous methods that are used from the NetworkConfigurationService.

### DIFF
--- a/kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/system/SystemService.java
+++ b/kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/system/SystemService.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2011, 2024 Eurotech and/or its affiliates and others
+ * Copyright (c) 2011, 2025 Eurotech and/or its affiliates and others
  * 
  * This program and the accompanying materials are made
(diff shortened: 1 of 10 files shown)


kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/system/SystemService.java
```diff
@@ -633,4 +638,12 @@ public interface SystemService {
      */
     public boolean isWPA3WifiSecurityEnabled();
 
+    /**
+     * Returns the timeout in seconds used for applying a network configuration.
+     *
+     * @since 3.0
+     * @return the timeput value in seconds
```
mattdibi: Typo

```suggestion
     * @return the timeout value in seconds
```

mattdibi (approved):

PR #5870, opened 2025-06-30
build(target-platform): generate CycloneDX SBOM (exclusion method)

mattdibi:
As discussed in https://github.com/CycloneDX/cyclonedx-maven-plugin/issues/398 we need to exclude transitive dependencies to not include obsolete/unused dependencies in our SBOM.

Example: [artemis-core-client@2.6.4](https://mvnrepository.com/artifact/org.apache.activemq/artemis-core-client/2.6.4) uses `netty-buffer@4.1.24.Final`... but Kura target platform already provides `netty-buffer@4.1.121.Final` which is actually used at runtime. As discussed in https://github.com/CycloneDX/cyclonedx-maven-plugin/issues/398 the best approach seems to be to exclude the unnecessary dependencies (as this PR does).

[WIP]


mattdibi:
Superseded by #5886

PR #5889, opened 2025-07-17
fix(distrib): Fixed http.server.manager start level

salvatore-coppola:
Fixed `http.server.manager` start level


pierantoniomerlino (approved):

salvatore-coppola:
@MMaiero can we force merge this ? Just changed one bundle start level

PR #5892, opened 2025-07-21
fix: Removed H2DbHelper service

nicolatimeus:

The H2DbHelper component creates the "default" H2 db service instance with pid `org.eclipse.kura.core.db.H2DbService` if it does not exist. This has been added when we switched from HSQL implementation to H2 to ensure that the framework always has at least one db instance available (in the HSQL impl db was implemented as a singleton component and H2 as a factory, which allowes for zero or more instances).

This removes the H2DbHelper component to allow using an alternative db implementation whitout having the default H2 instance created on startup.


salvatore-coppola (commented):
is the change to snapshot_0.xml files related to this fix?

salvatore-coppola (approved):

PR #5896, opened 2025-07-21
feat(core.status): GPIO service as optional dependency

pierantoniomerlino:
This pull request introduces several changes to improve the handling of the `GPIOService` dependency in the `CloudConnectionStatusServiceImpl` class and its associated tests. The updates include transitioning to an `Optional` for better null safety, modifying the service's cardinality and policy in the OSGi configuration, and enhancing test coverage to handle scenarios where the `GPIOService` is unavailable.

### Updates to `CloudConnectionStatusServiceImpl`:

* Transitioned `gpioService` to use `Optional` for safer null handling. Updated setter and unsetter methods accordingly. [[1]](diffhunk://#diff-43e442a4780110194cf2d6edf88dc27a5568e4ebd747dc40acbfbc3aa4fc684cL43-R44) [[2]](diffhunk://#diff-43e442a4780110194cf2d6edf88dc27a5568e4ebd747dc40acbfbc3aa4fc684cL78-R83)
* Added a null check in `getGpioStatusWorker` to handle cases where `gpioService` is not present, ensuring the method returns `null` if the service is unavailable.

### Changes to OSGi Configuration:

* Updated the `GPIOService` reference in `status.xml` to have a cardinality of `0..1` and a dynamic policy, allowing the service to be optional and dynamically bound.

### Enhancements to Unit Tests:

* Expanded test cases in `CloudConnectionStatusServiceImplTest` to include scenarios where `GPIOService` is unavailable, ensuring the application behaves correctly in such cases. [[1]](diffhunk://#diff-d2d641078279c727d30c6bf9c8eb906a05e3fe6730fcd0908923af8abd6b0000R222-R257) [[2]](diffhunk://#diff-d2d641078279c727d30c6bf9c8eb906a05e3fe6730fcd0908923af8abd6b0000R280-R282)
* Added helper methods to verify the correct selection of `StatusRunnables` based on the presence or absence of `GPIOService`.

### Minor Changes:

* Updated copyright year in `CloudConnectionStatusServiceImpl` to reflect 2025.
* Added missing imports in `CloudConnectionStatusServiceImplTest` for better readability and functionality.

### DIFF
--- a/kura/org.eclipse.kura.core.status/src/main/java/org/eclipse/kura/core/status/CloudConnectionStatusServiceImpl.java
+++ b/kura/org.eclipse.kura.core.status/src/main/java/org/eclipse/kura/core/status/CloudConnectionStatusServiceImpl.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2011, 2020 Eurotech and/or its affiliates and others
+ * Copyright (c) 2011, 2025 Eurotech and/or its affiliates and others
  * 
  * This program and the accompanying materials are made
  * available under the terms of the Eclipse Public License 2.0
@@ -14,6 +14,7 @@ package org.eclipse.kura.core.status;
 
 import java.io.File;
 import java.util.HashSet;
+import java.util.Optional;
 import java.util.Properties;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
@@ -40,7 +41,7 @@ public class CloudConnectionStatusServiceImpl implements CloudConnectionStatusSe
     private static final Logger logger = LoggerFactory.getLogger(CloudConnectionStatusServiceImpl.class);
 
     private SystemService systemService;
-    private GPIOService gpioService;
+    private Optional<GPIOService> gpioService = Optional.empty();
 
     private final ExecutorService notificationExecutor;
     private Future<?> notificationWorker;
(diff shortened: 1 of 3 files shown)


kura/org.eclipse.kura.core.status/src/main/java/org/eclipse/kura/core/status/CloudConnectionStatusServiceImpl.java
```diff
@@ -230,9 +231,12 @@ private StatusRunnable getLinuxStatusWorker(CloudConnectionStatusEnum status) {
     }
 
     private StatusRunnable getGpioStatusWorker(CloudConnectionStatusEnum status) {
+        if (!this.gpioService.isPresent()) {
```
marcellorinaldo, line 234: As improvement we could enhance the code that is calling this method to not rely on nullable values. If you don't want to, we can keep this code as is.

marcellorinaldo (approved):

kura/org.eclipse.kura.core.status/src/main/java/org/eclipse/kura/core/status/CloudConnectionStatusServiceImpl.java
pierantoniomerlino: Yeah, I don't like the use of `null` in this class. But I preferred to keep it as-is to not rewrite all.

PR #5909, opened 2025-07-28
refactor: Improved deb packages build

nicolatimeus:

Kura Debian packages currently install a zip file in `/tmp` which contains all files that are needed by Kura. The zip is extracted by the postinst script, which copies the files to their final location.

With this change the zip file is removed and all Kura files are now tracked by the debian package. File ownership and permissions are still not tracked properly, since they are adjusted by the manage-kura-users script (we should review and see if it is possible to remove that file).

This PR changes the distrib build in the following way:
 * The `build_equinox_distrib` ant script has been removed
 * Resources files are managed by the `maven-resources-plugin`
 * The `src/main/resources` directories of the `distrib` project and its children has been reorganised to have the following structure:
   * It contains two toplevel folders, `filtered` and `unfiltered`. The files in the first folder will be [filtered](https://maven.apache.org/plugins/maven-resources-plugin/examples/filter.html) by` maven-resources-plugin` before being placed in their final destination. The files in the second one will be copied unchanged.
   * The content of the `filtered` and `unfiltered` directories is organised in the following subdirectories:
     * `deb`: Contains debian control files and install scriptlets
     * `pkg`: Contains Kura files, the content is organised in the same way as the `/opt/eclipse/kura` dir.
 * Files in child projects overwrite the ones in parent project
 * Plugins are copied directly to the `target/pkg/plugins/startLevel` directory by the `maven-dependency-plugin` using the `copy-dependencies` goal, with an execution per start level. Artifacts to be copied are selected using the `includeArtifactIds` parameter, group id and version is taken from the BOM pom.
    *  Child projects can install new plugins on top of the ones installed by the parent

Property replacement in filtered resources is done using the Maven properties set in the distrib pom, the properties are inherited and can be overridden by child modules.

The distrib child projects are the following

|   |   |  
|---|---|
|aarch64-core | kura-core for aarch64  |  
|x86_64-core | kura-core for x86_64  |  
|  docker-base |  debian package for Docker (build is currently disabled) |  
|  distrib-extra |  distributions with additional content (to be removed) |  
|  distrib-extra/aarch64-nn | old no network profile aarch64 |  
|  distrib-extra/x86_64-nn |  old no network profile x86_64 |  
|  distrib-extra/distrib-nm | old profiles with network management |  
|  distrib-extra/distrib-nm/aarch64 | old kura aarch64 |  
|  distrib-extra/distrib-nm/x86_64 | old kura x86_64 | 
|  docker-alpine | Docker alpine (build is currently disabled) |  
|  docker-ubi8 |  Docker ubi8 (build is currently disabled) | 
|  core-dp |  old core-dp profile (build is currently disabled) |


kura/distrib/pom.xml
```diff
@@ -33,40 +33,74 @@
...
+        <kura.serialNumber></kura.serialNumber>
+        <deb.name>kura-core</deb.name>
+        <deb.dependencies>
+            setserial, zip, gzip, unzip, procps, usbutils, socat, gawk, sed, inetutils-telnet,
+            polkit | policykit-1, ssh | openssh, openssl, busybox, cron, logrotate, libpam-modules,
+            python3, openjdk-17-jre-headless | temurin-17-jdk, dos2unix, libtirpc3, chrony
+        </deb.dependencies>
+        <deb.recommendations></deb.recommendations>
+        <deb.provides></deb.provides>
+        <deb.architecture>unset</deb.architecture>
```
marcellorinaldo, line 96: Will the build fail if someone forgets to override this? If not, maybe it is better to use `all`

marcellorinaldo (dismissed):

kura/distrib/pom.xml
nicolatimeus: It will not fail, the produced package will have "unset" as architecture. Such package probably won't install anywhere so the user will be forced to to override.

marcellorinaldo (approved):

marcellorinaldo (dismissed):

marcellorinaldo (approved):

PR #5924, opened 2025-08-06
refactor!:  Removed networking bundles

salvatore-coppola:
Removed networking bundles


marcellorinaldo (changes requested):
In `distrib` resources we have some files and scripts that configure the networking. Do we need to remove also these?

salvatore-coppola:
> In `distrib` resources we have some files and scripts that configure the networking. Do we need to remove also these?

which files ?

salvatore-coppola:
> In `distrib` resources we have some files and scripts that configure the networking. Do we need to remove also these?

no we need them for building nm profiles, required until the creation of kura metapackages with networking profiles.

salvatore-coppola:
> In `distrib` resources we have some files and scripts that configure the networking. Do we need to remove also these?

removed everything

marcellorinaldo (dismissed):

marcellorinaldo (approved):

pierantoniomerlino (approved):

MMaiero (approved):

PR #5928, opened 2025-08-07
refactor(test): Removed rest.command.provider.test bundle moved in kura-command

salvatore-coppola:
Removed `rest.command.provider.test` bundle moved in kura-command.

Leftover from previous PR. The bundle was not the test reactor build.


salvatore-coppola:
deleted wrong bundle

PR #5935, opened 2025-08-23
fix(core.configuration): write snapshots with 600 permissions

MMaiero:
Set snapshot file permissions to owner read/write only (600) instead of the default permissions (644) to enhance security. Configuration snapshots may contain sensitive data and should not be readable by group or other users.

The change:
- Adds POSIX file permissions support to writeSnapshot method
- Sets permissions to rw------- after file creation
- Gracefully handles non-POSIX filesystems with warning log

### DIFF
--- a/kura/org.eclipse.kura.core.configuration/src/main/java/org/eclipse/kura/core/configuration/ConfigurationServiceImpl.java
+++ b/kura/org.eclipse.kura.core.configuration/src/main/java/org/eclipse/kura/core/configuration/ConfigurationServiceImpl.java
@@ -1011,14 +1011,37 @@ public class ConfigurationServiceImpl implements ConfigurationService, OCDServic
             throw new KuraException(KuraErrorCode.CONFIGURATION_SNAPSHOT_NOT_FOUND);
         }
 
-        File tempSnapshotFile;
-        try {
-            tempSnapshotFile = File.createTempFile(fSnapshot.getName(), null, new File(fSnapshot.getParent()));
-            tempSnapshotFile.deleteOnExit();
-        } catch (IOException ex) {
-            throw new KuraIOException(ex);
-        }
+        File tempSnapshotFile = getTempSnapshotFile(fSnapshot);
 
+        storeSnapshotData(conf, fSnapshot, tempSnapshotFile);
+
+        finalizeSnapshotWrite(fSnapshot, tempSnapshotFile);
+
+    }
+
+    private void finalizeSnapshotWrite(File fSnapshot, File tempSnapshotFile) throws KuraIOException {
+        try {
+            // Consolidate snapshot writing
+            Files.move(tempSnapshotFile.toPath(), fSnapshot.toPath(), StandardCopyOption.REPLACE_EXISTING,
+                    StandardCopyOption.ATOMIC_MOVE);
+
+            Set<PosixFilePermission> perms = PosixFilePermissions.fromString("rw-------");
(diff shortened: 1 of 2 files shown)


kura/org.eclipse.kura.core.configuration/src/main/java/org/eclipse/kura/core/configuration/ConfigurationServiceImpl.java
```diff
@@ -1032,21 +1057,17 @@ private void writeSnapshot(long sid, XmlComponentConfigurations conf) throws Kur
...
-
-        } catch (IOException e) {
-            try {
-                Files.delete(tempSnapshotFile.toPath());
-            } catch (IOException e1) {
-                throw new KuraIOException(e1);
-            }
-            throw new KuraIOException(e);
+            tempSnapshotFile = File.createTempFile(fSnapshot.getName(), null, new File(fSnapshot.getParent()));
+            tempSnapshotFile.deleteOnExit();
```
nicolatimeus: Is this really necessary? it will add a new exit hook for each snapshot leaking memory. Is it better to delete the temporary file if it exists in a finally block instead?

MMaiero: I think we always had it: https://github.com/eclipse-kura/kura/blob/5c2f432e10e1feef54ec37c134de08094f24951d/kura/org.eclipse.kura.core.configuration/src/main/java/org/eclipse/kura/core/configuration/ConfigurationServiceImpl.java#L1014
I have just moved it around to refactor in separate methods to improve readability.

I'll have a look based on your suggestion

MMaiero: actually it was contributed not long ago by @sfiorani https://github.com/eclipse-kura/kura/commit/4f1bcc54bf3fc3926759348b09ab57ae088d94fe

MMaiero: Made a change

kura/org.eclipse.kura.core.configuration/src/main/java/org/eclipse/kura/core/configuration/ConfigurationServiceImpl.java
```diff
@@ -1009,14 +1011,46 @@ private void writeSnapshot(long sid, XmlComponentConfigurations conf) throws Kur
...
+                    StandardCopyOption.ATOMIC_MOVE);
+
+            Set<PosixFilePermission> perms = PosixFilePermissions.fromString("rw-------");
+            Files.setPosixFilePermissions(fSnapshot.toPath(), perms);
+        } catch (UnsupportedOperationException e1) {
+            // POSIX permissions not supported on this file system, log and continue
+            logger.warn("Unable to set POSIX file permissions for snapshot file: {}", fSnapshot.getAbsolutePath(), e1);
+        } catch (IOException e) {
+            try {
+                Files.delete(tempSnapshotFile.toPath());
```
nicolatimeus: I think you can remove this after the last change

MMaiero: Right!

nicolatimeus (approved):

PR #5949, opened 2025-08-26
feat(core.identity): Added new method for verify password against the identity name

salvatore-coppola:
Added new method for verify password against the identity name.

### DIFF
--- a/kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/identity/PasswordStrengthVerificationService.java
+++ b/kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/identity/PasswordStrengthVerificationService.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2024 Eurotech and/or its affiliates and others
+ * Copyright (c) 2024, 2025 Eurotech and/or its affiliates and others
  * 
  * This program and the accompanying materials are made
  * available under the terms of the Eclipse Public License 2.0
@@ -29,19 +29,37 @@ public interface PasswordStrengthVerificationService {
      * Checks whether the provided password satisfies the password strength
      * requirements currently configured on the system.
      * 
-     * @param password the password to be verified.
-     * @throws KuraException if the password does not satisfy the current password
-     *                       strength requirements.
+     * @param password
+     *            the password to be verified.
+     * @throws KuraException
+     *             if the password does not satisfy the current password
+     *             strength requirements.
      */
     public void checkPasswordStrength(final char[] password) throws KuraException;
+
+    /**
+     * Perform the same checks done by {@link #checkPasswordStrength(char[])} and in addition verify that the password
+     * does not match the identityName
+     * 
+     * @param identityName
+     *            the name of the identity
+     * @param password
+     *            the password to be verified.
+     * @throws KuraException
+     *             if the password does not satisfy the current password
+     *             strength requirements.
+     */
+    public void checkPasswordStrength(String identityName, final char[] password) throws KuraException;
 
     /**
      * Returns the password strength requirements that the framework should enforce
      * for new passwords.
      * 
+     * @since 3.0
      * @return the password strength requirements.
-     * @throws KuraException if a failure occurs while retrieving the password
-     *                       strength requirements.
+     * @throws KuraException
+     *             if a failure occurs while retrieving the password
+     *             strength requirements.
      */
     public PasswordStrengthRequirements getPasswordStrengthRequirements() throws KuraException;
 }
--- a/kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/identity/PasswordStrengthVerificationService.java
+++ b/kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/identity/PasswordStrengthVerificationService.java
@@ -38,7 +38,7 @@ public interface PasswordStrengthVerificationService {
     public void checkPasswordStrength(final char[] password) throws KuraException;
 
     /**
-     * Perform the same checks done by {@link #checkPasswordStrength(char[])} and in addition verify that the password
+     * Similar to {@link #checkPasswordStrength(char[])} checks whether the provided password satisfies the password strength requirements currently configured on the system and in addition verifies that the password
      * does not match the identityName
      * 
      * @param identityName
--- a/kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/identity/PasswordStrengthVerificationService.java
+++ b/kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/identity/PasswordStrengthVerificationService.java
@@ -45,6 +45,7 @@ public interface PasswordStrengthVerificationService {
      *            the name of the identity
      * @param password
      *            the password to be verified.
+     * @since 3.0
      * @throws KuraException
      *             if the password does not satisfy the current password
      *             strength requirements.
@@ -55,7 +56,6 @@ public interface PasswordStrengthVerificationService {
      * Returns the password strength requirements that the framework should enforce
      * for new passwords.
      * 
-     * @since 3.0
      * @return the password strength requirements.
      * @throws KuraException
      *             if a failure occurs while retrieving the password
--- a/kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/identity/PasswordStrengthVerificationService.java
+++ b/kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/identity/PasswordStrengthVerificationService.java
@@ -38,10 +38,10 @@ public interface PasswordStrengthVerificationService {
     public void checkPasswordStrength(final char[] password) throws KuraException;
 
     /**
-     * Similar to {@link #checkPasswordStrength(char[])} checks whether 
-     * the provided password satisfies the password strength requirements currently configured 
+     * Similar to {@link #checkPasswordStrength(char[])} checks whether
+     * the provided password satisfies the password strength requirements currently configured
      * on the system and in addition verifies that the password
-     * does not match the identityName
+     * does not match the identityName ignoring the characters case.
      * 
      * @param identityName
      *            the name of the identity
--- a/kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/identity/PasswordStrengthVerificationService.java
+++ b/kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/identity/PasswordStrengthVerificationService.java
@@ -1,12 +1,12 @@
 /*******************************************************************************
  * Copyright (c) 2024, 2025 Eurotech and/or its affiliates and others
- * 
+ *
  * This program and the accompanying materials are made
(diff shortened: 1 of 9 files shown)


kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/identity/PasswordStrengthVerificationService.java
```diff
@@ -29,19 +29,37 @@ public interface PasswordStrengthVerificationService {
...
+     *             if the password does not satisfy the current password
+     *             strength requirements.
+     */
+    public void checkPasswordStrength(String identityName, final char[] password) throws KuraException;
+
     /**
      * Returns the password strength requirements that the framework should enforce
      * for new passwords.
      * 
+     * @since 3.0
```
MMaiero: Shouldn't this be set on the previous method since it is new?

salvatore-coppola: > Shouldn't this be set on the previous method since it is new?

obviously...

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/identity/PasswordStrengthVerificationService.java
```diff
@@ -29,19 +29,37 @@ public interface PasswordStrengthVerificationService {
...
+     * @param password
+     *            the password to be verified.
+     * @throws KuraException
+     *             if the password does not satisfy the current password
+     *             strength requirements.
      */
     public void checkPasswordStrength(final char[] password) throws KuraException;
 
+    /**
+     * Perform the same checks done by {@link #checkPasswordStrength(char[])} and in addition verify that the password
```
MMaiero:
```suggestion
     * Similar to {@link #checkPasswordStrength(char[])} checks whether the provided password satisfies the password strength requirements currently configured on the system and in addition verifies that the password
```

kura/org.eclipse.kura.core.identity/src/main/java/org/eclipse/kura/core/identity/PasswordStrengthVerificationServiceImpl.java
```diff
@@ -57,14 +57,31 @@ public PasswordStrengthRequirements getPasswordStrengthRequirements() {
...
+    private List<Validator<String>> buildValidators(String identityName) {
+        PasswordStrengthRequirements currentRequirements = getPasswordStrengthRequirements();
+        ValidatorOptions validatorOptions = new ValidatorOptions(currentRequirements.getPasswordMinimumLength(), //
+                currentRequirements.digitsRequired(), //
+                currentRequirements.bothCasesRequired(), //
                 currentRequirements.specialCharactersRequired());
 
-        final List<Validator<String>> validators = PasswordStrengthValidators.fromConfig(validatorOptions);
+        List<Validator<String>> validators = new ArrayList<>(PasswordStrengthValidators.fromConfig(validatorOptions));
+        if (identityName != null) {
```
MMaiero: Can an empty value create an issue here?

salvatore-coppola: not really but let's add a check

kura/org.eclipse.kura.core.identity/src/main/java/org/eclipse/kura/core/identity/ValidationUtil.java
```diff
@@ -46,6 +46,21 @@ public static void validateNewPassword(final char[] password,
         passwordStrengthVerificationService.checkPasswordStrength(password);
     }
 
+    public static void validateNewPassword(String identityName, final char[] password,
+            final PasswordStrengthVerificationService passwordStrengthVerificationService) throws KuraException {
+        if (password.length == 0) {
+            throw new KuraException(KuraErrorCode.INVALID_PARAMETER, "New password cannot be empty");
+        }
+
+        final String asString = new String(password);
```
MMaiero: Isn't it better, for security reasons to use char[] instead of Strings for passwords?

https://sentry.io/answers/why-is-char-preferred-over-string-for-passwords/#:~:text=According%20to%20the%20official%20Java,safely%20and%20securely%20store%20passwords.

kura/org.eclipse.kura.util/src/main/java/org/eclipse/kura/util/validation/PasswordStrengthValidators.java
```diff
@@ -80,6 +80,11 @@ private static Validator<String> containsBothCases(final Messages messages) {
         };
     }
 
+    public static Validator<String> requireDifferentNameAndPassword(final String identityName) {
+
+        return new PredicateValidator(v -> !v.equals(identityName), new DefaultMessages().pwdNotEqualsUsername());
```
MMaiero: Would it make sense to check ignoring case?

salvatore-coppola: well... it's a choiche, Admin is not the same of AdMIn

kura/org.eclipse.kura.core.identity/src/main/java/org/eclipse/kura/core/identity/ValidationUtil.java
salvatore-coppola: I didn't touch this method but I can change it

kura/org.eclipse.kura.util/src/main/java/org/eclipse/kura/util/validation/PasswordStrengthValidators.java
MMaiero: True. Maybe let's check if there is a best practice for example in NIST 800 standards

salvatore-coppola: ok I checked, NIST 800 says that derivative password must be rejected, included same name in different case, I'll modify it. 

https://pages.nist.gov/800-63-3/sp800-63b.html#sec5 5.1.1.2

kura/org.eclipse.kura.core.identity/src/main/java/org/eclipse/kura/core/identity/ValidationUtil.java
```diff
@@ -79,6 +92,14 @@ private static void requireMaximumLength(final String parameterName, final Strin
...
 
+    private static void requireMaximumLength(final String parameterName, final char[] value, final int length)
+            throws KuraException {
+        if (value.length > length) {
+            throw new KuraException(KuraErrorCode.INVALID_PARAMETER,
+                    parameterName + " must be at most " + length + " characters long");
+        }
+    }
+
     private static void requireNoWhitespaceCharacters(final String parameterName, final String value)
```
MMaiero: Since we have now the new method in line 113 that seems more secure, maybe we can remove this?

kura/org.eclipse.kura.core.identity/src/main/java/org/eclipse/kura/core/identity/PasswordStrengthVerificationServiceImpl.java
```diff
@@ -57,14 +57,31 @@ public PasswordStrengthRequirements getPasswordStrengthRequirements() {
...
+    private List<Validator<String>> buildValidators(String identityName) {
+        PasswordStrengthRequirements currentRequirements = getPasswordStrengthRequirements();
+        ValidatorOptions validatorOptions = new ValidatorOptions(currentRequirements.getPasswordMinimumLength(), //
+                currentRequirements.digitsRequired(), //
+                currentRequirements.bothCasesRequired(), //
                 currentRequirements.specialCharactersRequired());
 
-        final List<Validator<String>> validators = PasswordStrengthValidators.fromConfig(validatorOptions);
+        List<Validator<String>> validators = new ArrayList<>(PasswordStrengthValidators.fromConfig(validatorOptions));
+        if (identityName != null && !identityName.isEmpty()) {
```
MMaiero:
```suggestion
        if (identityName != null && !identityName.trim().isEmpty()) {
```

kura/test/org.eclipse.kura.core.identity.test/src/main/java/org/eclipse/kura/core/identity/test/PasswordStrengthVerificationServiceImplTest.java
```diff
@@ -56,6 +56,26 @@ public void shouldRejectPasswordWithoutDigits() {
         thenExceptionIsThrown(KuraException.class);
     }
 
+    @Test
+    public void shouldRejectPasswordEqualsToIdentityName() {
+        givenPasswordStrenghtVerificationOptions("new.password.min.length", 3, "new.password.require.digits", false,
```
MMaiero:
```suggestion
        givenPasswordStrengthVerificationOptions("new.password.min.length", 3, "new.password.require.digits", false,
```

kura/org.eclipse.kura.core.identity/src/main/java/org/eclipse/kura/core/identity/ValidationUtil.java
```diff
@@ -89,6 +110,16 @@ private static void requireNoWhitespaceCharacters(final String parameterName, fi
         }
     }
 
+    private static void requireNoWhitespaceCharacters(final String parameterName, final char[] value)
```
MMaiero, line 101: value can be null

salvatore-coppola: added check on invokers validateNewPassword

kura/org.eclipse.kura.core.identity/src/main/java/org/eclipse/kura/core/identity/ValidationUtil.java
```diff
@@ -33,19 +33,30 @@ private ValidationUtil() {
 
     public static void validateNewPassword(final char[] password,
             final PasswordStrengthVerificationService passwordStrengthVerificationService) throws KuraException {
-        if (password.length == 0) {
+        if (password == null || password.length == 0) {
```
MMaiero, line 36: Would it make sense to call the new validateNewPassword from here to reduce code duplication and possible errors?

salvatore-coppola: No because they invoke differents PasswordStrengthVerificationService  methods. I could create a private validateNewPassword method with a null check on identityName to choose which method invoke but it's the same amount of code except for avoiding the duplication of 
```java
requireMaximumLength(NEW_PASSWORD, password, 255)
requireNoWhitespaceCharacters(NEW_PASSWORD, password) 
```

not an unmissable improvement IMHO

MMaiero (approved):

mattdibi (approved):

PR #5959, opened 2025-08-30
feat: add Swagger OpenAPI to system REST bundle

MMaiero:
Added OpenAPI mapping to improve user experience when working with Eclipse Kura REST apis


MMaiero:
@salvatore-coppola Does this make sense as an initial effort (potentially to extend)?

PR #5965, opened 2025-09-08
feat: container temporary credentials

MMaiero:


MMaiero:
To verify and validate, via Kura create a new container with the feature for passing credentials enabled and the right permissions set. For example the rest.inventory. Then try to interact with Kura REST via the container just started for example using: 
```
sudo docker exec -it 828037e1b6c3 sh -lc 'curl -k -sS -i -H "Authorization: Kura-Token $KURA_IDENTITY_TOKEN" "https://172.17.0.1:443/services/inventory/v1/inventory"'
```

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/identity/TemporaryIdentityService.java
```diff
@@ -0,0 +1,70 @@
...
+    /**
+     * Creates a temporary identity with the given name and permissions.
+     * The identity will not be persisted and will exist only in memory.
+     * 
+     * @param identityName the name of the temporary identity to be created.
+     * @param permissions the set of permissions to be assigned to this temporary identity.
+     * @return a temporary authentication token that can be used to authenticate as this identity.
+     * @throws KuraException if a failure occurs in creating the temporary identity.
+     */
+    public String createTemporaryIdentity(final String identityName, final Set<Permission> permissions) throws KuraException;
```
salvatore-coppola: I would remove 'Temporary' from the names of all methods. 'Temporary' is implied by the service name.

nicolatimeus:
This PR introduces a new concept of temporary identity and a new token based authentication mechanism, in my opinion the two topics are orthogonal.

In my opinion the only difference between regular and temporary identities should be that the latter are not persisted in the snapshot and have a fixed lifetime period, but I would expect them to behave like the regular ones for all other aspects (for example being able to use the existing auth methods for them). We could for example create a separate in-memory store for them that contains the same informations that we store in the UserAdmin that is used by the IdentityService in addition to the UserAdmin.

A blocker for implementing this approach is the fact that the REST service is currently using the UserAdmin directly for authentication, and has not ported to use the new IdentityService yet.

We could also implement a new token based authentication method that works the same both for regular and temporary identities, here we could use a standard solution like JWT.

I would proceed with the following steps, maybe with separate contributions that can be probably done in parallel:

* Remove the dependency from the REST service to the UserAdmin and use the new IdentityService instead
* Introduce the concept of temporary identities that behave as similar as possible to regular ones
* Add support for a token-based authentication service that works both for regular and temporary identities

MMaiero:
> This PR introduces a new concept of temporary identity and a new token based authentication mechanism, in my opinion the two topics are orthogonal.
> 
> In my opinion the only difference between regular and temporary identities should be that the latter are not persisted in the snapshot and have a fixed lifetime period, but I would expect them to behave like the regular ones for all other aspects (for example being able to use the existing auth methods for them). We could for example create a separate in-memory store for them that contains the same informations that we store in the UserAdmin that is used by the IdentityService in addition to the UserAdmin.
> 
> A blocker for implementing this approach is the fact that the REST service is currently using the UserAdmin directly for authentication, and has not ported to use the new IdentityService yet.
> 
> We could also implement a new token based authentication method that works the same both for regular and temporary identities, here we could use a standard solution like JWT.
> 
> I would proceed with the following steps, maybe with separate contributions that can be probably done in parallel:
> 
> * Remove the dependency from the REST service to the UserAdmin and use the new IdentityService instead
> * Introduce the concept of temporary identities that behave as similar as possible to regular ones
> * Add support for a token-based authentication service that works both for regular and temporary identities

But would make sense then to use password based auth or certificate auth from the containers?

nicolatimeus:
Password authentication currently does not work in the case when the user provides the password using the PasswordConfiguration.newPassword field https://github.com/eclipse-kura/kura/blob/56899bb13aecd90ff1421391c652fe11588c9231/kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/identity/PasswordConfiguration.java#L57. 

In this case the identity service should compute the hash of the given password, and store that one in the passwordHash field, the newPassword fileld should not be populated in the `PasswordConfiguration`s in the temporary store.

Part of the logic performed by the `updateIdentityConfigurationInternal` method (e.g. new password processing) should be done also in the case of a temporary identity.

Consider also further unifying the identity configuration validation logic

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/identity/IdentityService.java
```diff
@@ -189,13 +192,53 @@ public void updateIdentityConfiguration(final IdentityConfiguration identityConf
...
+     *
+     * @param identityName the name of the temporary identity to be created.
+     * @param configuration the identity configuration (passwords, certificates, tokens, permissions, etc.)
+     * @param lifetime the duration before automatic expiration. The identity will be automatically
+     *                 removed after this period.
+     * @throws KuraException if a failure occurs in creating the temporary identity or if an identity
+     *                       with the given name already exists (either regular or temporary).
+     * @since 2.8.0
+     */
+    public void createTemporaryIdentity(final String identityName, final IdentityConfiguration configuration,
```
nicolatimeus: The identityName argument is redundant since it is included in IdentityConfiguration, for symmetry with createIdentity() we could remove the configuration argument

nicolatimeus:
The implementation of the IdentityService is becoming more and more complicated since it contains different code paths for handling reguar and temporary identities. A possible approach for simplifying it could be creating a `IdentityStore` interface in the `org.eclipse.kura.core.identity` package with one implementation for temporary identities (can be based on the currently existing TemporaryIdentityStore) and another for the useradmin. The identity service could delegate the storage aspects of identity management to these implementation trying to keep a single code path for configuration validation and update.

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/identity/IdentityService.java
```diff
@@ -189,13 +192,56 @@ public void updateIdentityConfiguration(final IdentityConfiguration identityConf
...
+     * Deletes a temporary identity.
+     *
+     * @param identityName the name of the temporary identity to delete.
+     * @return {@code true} if the identity was deleted as part of the method call,
+     *         {@code false} if the identity does not exist.
+     * @throws KuraException if the identity is not temporary (is a regular persisted identity)
+     *                       or if deletion fails.
+     * @since 2.8.0
+     */
+    public boolean deleteTemporaryIdentity(final String identityName) throws KuraException;
```
nicolatimeus: Is it necessary to have a dedicated `deleteTemporaryIdentity` method or is it enough to extend the existing `deleteIdentity` method?

kura/org.eclipse.kura.container.provider/OSGI-INF/metatype/org.eclipse.kura.container.provider.ContainerInstance.xml
```diff
@@ -189,6 +189,14 @@
         <AD id="container.restart.onfailure" name="Restart Container On Failure" type="Boolean" cardinality="1" required="true" default="false"
             description="Automatically restart the container when it has failed. Default: false">
         </AD>
+
+        <AD id="container.identity.enabled" name="Enable Identity Integration" type="Boolean" cardinality="1" required="true" default="false"
+            description="Enable integration with Kura Identity Service to provide temporary credentials to the container. Default: false">
+        </AD>
+        
+        <AD id="container.permissions" name="Container Permissions"
+            description="Comma-separated list of permissions to grant to the container when identity integration is enabled. Example: rest.asset,rest.configuration"
```
nicolatimeus:
```suggestion
            description="Comma-separated list of permissions to grant to the container when identity integration is enabled. Example: rest.assets,rest.configuration" 
```

nicolatimeus (dismissed):

nicolatimeus:
There are still some adjustments that we can do but in case we can address with separate PRs:
* Improve temporary identity name generation logic to avoid creating invalid names
* Client containers will not be able to connect to Kura with default firewall configuration, we should fix and/or document this.
* The `createTemporaryIdentity()` API method is not aligned with `createIdentity()`, because the first allows to specify an identity configuration and the second doesn't. We should align by creating a new createIdentity() method that accepts an identity configuration and a createTemporaryIdentity() that doesn't.

kura/org.eclipse.kura.container.orchestration.provider/src/main/java/org/eclipse/kura/container/orchestration/provider/impl/ContainerOrchestrationServiceImpl.java
```diff
@@ -401,6 +403,8 @@ public void stopContainer(String id) throws KuraException {
                 this.dockerClient.stopContainerCmd(id).exec();
             }
 
+        } catch (NotModifiedException e) {
+            logger.debug("Container {} already stopped", id);
```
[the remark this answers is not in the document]
MMaiero: Intended use

kura/org.eclipse.kura.container.provider/OSGI-INF/containerinstance.xml
```diff
@@ -1,16 +1,17 @@
...
 
  	SPDX-License-Identifier: EPL-2.0
 
- 	Contributors:
- 	 Eurotech
+	Contributors:
+	 Eurotech
  -->
-<scr:component xmlns:scr="http://www.osgi.org/xmlns/scr/v1.1.0" enabled="true" activate="activate" configuration-policy="require" deactivate="deactivate" modified="updated" name="org.eclipse.kura.container.provider.ContainerInstance">
+<scr:component xmlns:scr="http://www.osgi.org/xmlns/scr/v1.1.0" enabled="true" immediate="true" activate="activate"
```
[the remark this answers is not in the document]
MMaiero: Intended use

kura/org.eclipse.kura.container.provider/src/main/java/org/eclipse/kura/container/provider/ContainerInstance.java
```diff
@@ -329,12 +376,166 @@ public Starting(final ContainerInstanceOptions options) {
...
+                final ContainerInstanceOptions options) {
+            ContainerConfiguration baseConfig = options.getContainerConfiguration();
+
+            final String identityName = ContainerInstance.this.currentTemporaryIdentityName.get();
+            final char[] password = ContainerInstance.this.currentTemporaryPassword.get();
+
+            if (options.isIdentityIntegrationEnabled() && password != null && identityName != null) {
+                final List<String> envVars = new ArrayList<>(baseConfig.getContainerEnvVars());
+                envVars.add("KURA_IDENTITY_NAME=" + identityName);
+                envVars.add("KURA_IDENTITY_PASSWORD=" + new String(password));
```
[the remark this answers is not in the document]
MMaiero: Already in the plans for feature evolution

nicolatimeus (approved):

PR #5981, opened 2025-09-18
docs: update log-related documentation following latest changes

mattdibi:
Update documentation related to Kura logs following #5974 and #5975

### DIFF
--- a/docs/getting-started/install-kura.md
+++ b/docs/getting-started/install-kura.md
@@ -95,9 +95,8 @@ To have all the Kura features working, the following dependencies are required:
 - General: `setserial`, `zip`, `gzip`, `unzip`, `procps`, `usbutils`, `socat`, `gawk`, `sed`, `inetutils-telnet`.
 - Security: `polkit` or `policykit-1`, `ssh` or `openssh`, `openssl`, `busybox`, `openvpn`.
 - Bluetooth: `bluez` or `bluez5`, `bluez-hcidump` or `bluez5-noinst-tools`.
-- Time: `chrony`, `chronyc`, `cron` or `cronie`.
+- Time: `chrony`, `chronyc`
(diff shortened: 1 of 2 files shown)


docs/getting-started/install-kura.md
```diff
@@ -95,9 +95,8 @@ To have all the Kura features working, the following dependencies are required:
 - General: `setserial`, `zip`, `gzip`, `unzip`, `procps`, `usbutils`, `socat`, `gawk`, `sed`, `inetutils-telnet`.
 - Security: `polkit` or `policykit-1`, `ssh` or `openssh`, `openssl`, `busybox`, `openvpn`.
 - Bluetooth: `bluez` or `bluez5`, `bluez-hcidump` or `bluez5-noinst-tools`.
-- Time: `chrony`, `chronyc`, `cron` or `cronie`.
+- Time: `chrony`, `chronyc`
```
MMaiero, line 98: Are the changes back ported a breaking change?

mattdibi: Why would they? 🤔  The dependency on `cron` was first introduced in #3916 for the purpose of rotating logs via cronjob. Given that the only way we used `cron` was to set up that specific job we are free to remove the dependency.

mattdibi: Are you suggesting that, by removing the `cron` dependency from our package, we might break compatibility for a user that is currently leveraging `cron` in their bundle?

MMaiero: yes

mattdibi: I don't think there's any specific/explicit contract covering our debian dependencies with respect to users' expectations.

MMaiero (approved):

PR #5984, opened 2025-09-22
fix: Fixed KeystoreChangedEvent not posted in some scenarios

nicolatimeus:

The `BaseKeystoreService` currently has some logic that prevents sending `KeystoreChangedEvent`s when adding/removing certificates if CRL cache is enabled and a certificate that references a new CRL distribution point is added/removed.

If the CRL cache needs update after adding or removing a certificate, the `KeystoreChangedEvent` will not be sent when the certificate add/remove operation completes but it will be deferred to the instant when the cache update finishes (the referenced CRL is downloaded), this may cause excessive delays if the framework is not able to download the CRL, for example due to a network error.

This PR removes the suppression mechanism, this may cause additional event to be sent in a short timeframe in the scenario described above, for this reason the HttpService has been updated to defer server restart after a `KeystoreChangedEvent`.


marcellorinaldo (approved):

marcellorinaldo:
Do we backport this to 5.6.0?

PR #5992, opened 2025-09-25
fix: snapshot POSIX file permissions set before moving

MMaiero:
Aligned with changes in 5.6.0 and setting first the permission to the temporary snapshot before moving to destination


nicolatimeus (dismissed):

MMaiero:
Modified again the logic because now moving the file move at the end could cause, if the FS does not support the setting of the file visibility, the corresponding exception would not allow the file move.
Extracted then the posix permission logic in a separate method to catch the exception while keeping the try with catch of the IOException for rethrow as KuraIOException

nicolatimeus (approved):

PR #5996, opened 2025-09-29
ci(Jenkinsfile): fix `sonar.exclusions` parameter

mattdibi:
Fix `sonar.exclusions` parameter as per https://docs.sonarsource.com/sonarqube-server/10.8/project-administration/analysis-scope


mattdibi:
Not working as expected

PR #6009, opened 2025-10-03
build: Updated h2 version to 2.4.240 from 2.1.214

salvatore-coppola:
Updated h2 version to `2.4.240` from `2.1.214` .

Ran several test, the new version is 100% compatible with previous one.


MMaiero:
@salvatore-coppola Since the upgrade to the newest version seems working and resolves CVEs. Do you think it makes sense to backport it to 5.6.0?

salvatore-coppola:
> @salvatore-coppola Since the upgrade to the newest version seems working and resolves CVEs. Do you think it makes sense to backport it to 5.6.0?

yes we could, i prepare a PR

MMaiero:
Do we need to update the [kura/org.eclipse.kura.db.h2db.provider/META-INF/MANIFEST.MF](https://github.com/eclipse-kura/kura/blob/07578fefcaa98ca28e0b4d5429d288384534b2e7/kura/org.eclipse.kura.db.h2db.provider/META-INF/MANIFEST.MF#L27) file?

I think that also the notice file needs to be updated, isn't it?

salvatore-coppola:
> Do we need to update the [kura/org.eclipse.kura.db.h2db.provider/META-INF/MANIFEST.MF](https://github.com/eclipse-kura/kura/blob/07578fefcaa98ca28e0b4d5429d288384534b2e7/kura/org.eclipse.kura.db.h2db.provider/META-INF/MANIFEST.MF#L27) file?
> 
> I think that also the notice file needs to be updated, isn't it?

I'll update but from an OSGi perspective that version is right, with that package version the system still working.

MMaiero:
Could be, but better safe than sorry :-D

MMaiero (approved):

PR #6051, opened 2025-10-21
feat: CryptoService systemd credentials

MMaiero:
Adds support for Systems-credentials to pass the encryption key to the framework.
Keeps compatibility with the Java system property and the well known password by the crypto service


kura/test/org.eclipse.kura.core.crypto.test/src/main/java/org/eclipse/kura/core/crypto/EncryptDecryptTest.java
```diff
@@ -20,19 +20,31 @@
...
+        final SystemService systemService = Mockito.mock(SystemService.class);
+        try {
+            Mockito.when(systemService.getKuraDataDirectory()).thenReturn(Files.createTempDirectory(null).toString());
+        } catch (IOException e) {
+            throw new IllegalStateException("cannot create temporary directory");
+        }
+
         this.cryptoService = new CryptoServiceImpl();
+        this.cryptoService.setSystemService(systemService);
+        this.cryptoService.activate();
```
mattdibi: Not a blocker: To make it easier for people reading the tests I would call the mock SystemService `mockSystemService`


```suggestion
        final SystemService mockSystemService = Mockito.mock(SystemService.class);
        try {
            Mockito.when(mockSystemService.getKuraDataDirectory()).thenReturn(Files.createTempDirectory(null).toString());
        } catch (IOException e) {
            throw new IllegalStateException("cannot create temporary directory");
        }

this.cryptoService = new CryptoServiceImpl();
        this.cryptoService.setSystemService(mockSystemService);
        this.cryptoService.activate();
```

kura/org.eclipse.kura.core.crypto/src/main/java/org/eclipse/kura/core/crypto/SystemdCredentialLoader.java
```diff
@@ -0,0 +1,56 @@
...
+                .map(SystemdCredentialLoader::new);
+    }
+
+    public Optional<byte[]> loadCredential(final String name) throws IOException {
+
+        final File credentialFile = new File(this.credentialRoot, name);
+
+        if (!credentialFile.exists()) {
+            return Optional.empty();
+        }
```
mattdibi, line 42: For additional security best practices: would it be a good idea to have a check on file permissions here like OpenSSH does? Something like `400` given that the file should only be read by the user and read-only.

I currently do not understand the UX implications of this type of change though (if the file doesn't have the correct permissions do I need to reinstall Kura? Might be too much of a pain)

mattdibi (commented):
Will Kura containers be able to leverage this via SystemD? SystemD is not running in our containers. We'll just inject this as an env var?

kura/distrib/src/main/resources/unfiltered/deb/control/postinst
```diff
@@ -82,6 +82,10 @@ function postInstall {
     
     echo ""
     echo "Finished. KURA has been installed to ${INSTALL_DIR}/kura and will start automatically after a reboot"
+    echo ""
+    echo "****************************************************************"
+    echo "WARNING: Set the crypto service encryption key as explained in Eclipse Kura documentation before rebooting to have snapshot encryption enabled."
```
mattdibi, line 87: Correct me if I'm wrong but snapshot encryption is still available even if no encryption key is provided given that we use the default (albeit well known) one.

I would rephrase this as:


```suggestion
    echo "****************************************************************"
    echo "WARNING: Set the crypto service encryption key as explained in Eclipse Kura documentation before rebooting for additional security."
```

but feel free to disagree

mattdibi (commented):
Biggest issue here IMO is the documentation. Looking at #6048:
1. The documentation instruct the user to set the secret key as a Java environment variable, but there's no mention of our start script where they can set it.
2. Here we mention Systemd-credentials but there's no mention of that in the documentation nor a link to the systemd documentation. An example on how to set up the secret so that we can delegate it to systemd-credentials would explain a lot.

kura/org.eclipse.kura.core.crypto/src/main/java/org/eclipse/kura/core/crypto/SystemdCredentialLoader.java
nicolatimeus: The credential files are set up by systemd when the unit is started, I couldn't find documentation describing what are the file permissions that it will apply to them.

nicolatimeus:
> Will Kura containers be able to leverage this via SystemD? SystemD is not running in our containers. We'll just inject this as an env var?

This PR adds support for systemd-credentials along with the system property that is currently documented, that one can be used in containers on in the case when systemd does not support systemd credentials.

nicolatimeus:
> Biggest issue here IMO is the documentation. Looking at #6048:
> 
> 1. The documentation instruct the user to set the secret key as a Java environment variable, but there's no mention of our start script where they can set it.
> 2. Here we mention Systemd-credentials but there's no mention of that in the documentation nor a link to the systemd documentation. An example on how to set up the secret so that we can delegate it to systemd-credentials would explain a lot.

The plan to add support to systemd-credentials in Kura 6 only, updating the documentation mentionining it. We will backport to Kura 5.6 (the target of #6048) all other changes from this PR (start script changes, log messages etc).

kura/distrib/src/main/resources/unfiltered/deb/control/postinst
nicolatimeus: Yes the message is not technically correct but maybe like this it is more effective, in the end having the snapshot encrypted with a well known key is not much better than having in plaintext.

kura/org.eclipse.kura.core.crypto/src/main/java/org/eclipse/kura/core/crypto/SystemdCredentialLoader.java
mattdibi: Ok then, let's leave it as is 👍

kura/distrib/src/main/resources/unfiltered/deb/control/postinst
mattdibi: I thought this might be your original goal: scare the user a little bit. I'm fine with it

mattdibi (dismissed):

mattdibi (approved):

PR #6053, opened 2025-10-21
fix: Prominent alert on using default password for CryptoServiceImpl [backport release-5.6.0]

nicolatimeus:

Backport of #6051 without systemd-credentials support


kura/test/org.eclipse.kura.core.crypto.test/src/main/java/org/eclipse/kura/core/crypto/CryptoServiceImplCustomKeyTest.java
```diff
@@ -0,0 +1,233 @@
...
+        } catch (KuraException e) {
+            fail("Decryption failed: " + e.getMessage());
+        }
+    }
+
+    private void thenServiceIsActivated() {
+        assertNotNull(this.cryptoService);
+    }
+
+    private void thenServiceFallsBackToDefaultKey() {
```
marcellorinaldo: How is this condition different from the one above? We could try to encrypt using the crypto service and then decrypt with the default key

kura/distrib/src/main/resources/docker-x86_64-nn/deb/control/postinst
```diff
@@ -73,6 +73,10 @@ function postInstall {
     
     echo ""
     echo "Finished. KURA has been installed to ${INSTALL_DIR}/kura and will start automatically after a reboot"
+    echo ""
+    echo "****************************************************************"
+    echo "WARNING: Set the crypto service encryption key as explained in Eclipse Kura documentation before rebooting to have snapshot encryption enabled."
```
marcellorinaldo, line 78: Some line breaks would be nice

marcellorinaldo (dismissed):

marcellorinaldo (approved):

PR #6054, opened 2025-10-21
fix(web2): generic error when setting access point with no password [backport release-5.6.0]

sfiorani:
To avoid showing passwords in the Network tab, we replace the value present in the snapshot with the word `Placeholder`. However, we do the same even when the password field is missing in the configuration: this means we do not distinguish between the case where a configuration is present in the snapshot or not.

This behavior can lead to corner cases where the user is not fully aware of what they should or should not do to correctly configure the interface.

A similar problem is present also with the `StationMode + WPA Enterprise` settings combination: also this case was resolved in this PR.

**Example of an edge case:**

There is no WLAN configuration in the snapshot, and from the UI the user tries to switch that interface from `Disabled` to `Access Point`: when clicking on the WLAN tab, the password field does not show any error even though nothing is actually set:

- the validator sees the password as non-empty (the UI field was filled with `Placeholder`)

- the validator does not validate the password against the constraints (the check is triggered on a keyboard event)

Since no errors are shown, the user might click Apply, triggering the following actions:

- the form is validated because it contains no errors

- the configuration is passed to GwtConfig, which removes Placeholder to insert the real password value (which is actually empty)

- the passwordStrengthService tries to enforce password rules but the password is empty → EXCEPTION THROWN

The logger explains the issue clearly, but the UI has already lost control of the exception and will show a generic error popup that does not explain the real problem to the user.

This new approach forced us also to avoid the password evaluation when the Security is set to None: when switching from another security to `None` with the text form empty, lead to a similar bug than the one described before. This was avoided because usually the `Placeholder` value was set as password and the evaluation was skipped. Now the empty password is of course evaluated as non-valid. So we directly skip the evaluation if the security is None.

The fix in this PR does the following steps to avoid the problem:

- adds a check on the value retrieved from the snapshot, and the value `Placeholder` will only be used if the password is present in the snapshot; otherwise, the field will be left empty.
- removes the password validation in the backend if the wireless security is set to None in Access Point mode
- removed the password validation in the backend if the wireless security is set to WPA2/WPA3-Enterprise in Station Mode

In this way, when the tab of the corresponding interface is opened, the password field will be empty and therefore marked as invalid, forcing the user to fill it in before being able to click Apply. Meanwhile, the None Wireless Security option is not validated even when the password is empty.


kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/client/ui/network/TabWirelessUi.java
nicolatimeus: We can try like this, the other cases should be NONE and WPA 

```suggestion
        } else {
            this.password.setValidators();
        }
```

nicolatimeus (approved):

MMaiero:
Forcing merge. Failing tests are not related to this PR.
The peer validation approves the changes. And we are going to start a full QA process.

PR #6070, opened 2025-10-24
fix(distrib): added EQUINOX_DEBUG_OPTS

salvatore-coppola:
Now equinox debug options are only used in development mode.


marcellorinaldo (commented):
Can we include also this flag?

```
-Dosgi.debug=true \
```

salvatore-coppola:
> Can we include also this flag?
> 
> ```
> -Dosgi.debug=true \
> ```

Are you sure that it's the right option ? I found org.eclipse.osgi/debug=true as the option, Let's leave as it is, that is a standard configuration for minimal debug, users can always add its own flags

marcellorinaldo (approved):

PR #6074, opened 2025-10-28
refactor(distrib): remove redundant else clause in find_dbus_config_path

MMaiero:


kura/distrib/src/main/resources/common/manage_kura_users.sh
```diff
@@ -22,8 +22,6 @@ function find_dbus_config_path {
         echo "${default_path}"
     elif [ -f "${fallback_path}" ]; then
         echo "${fallback_path}"
-    else
-        echo "${default_path}"
     fi
```
pierantoniomerlino, line 27: If neither default_path or fallback_path exist, this method will return nothing. So, the bluetooth and wpa_supplicant files will not be changed, but no errors will be reported in the logs. Is it better to make the update fail, to al least signal something to the user? Or, we need to log that no path are found in the system?

pierantoniomerlino (approved):

PR #6077, opened 2025-11-01
docs: contribute agent guide

MMaiero:
Created a new Agents.md file to provide ruling on human and, especially, AI agents contributions to the project.


AGENTS.md
```diff
@@ -0,0 +1,55 @@
...
+- **Checkstyle**: Runs automatically during `process-sources` phase with config in `checkstyle_checks.xml`
+
+## Code Style
+- **Java Version**: Java 17 (source/target)
+- **Formatting**: Use profiles in `kura/setup/formatting/` (KuraFormatter.xml, KuraCleanupProfile.xml)
+- **Line Length**: Max 150 characters
+- **Imports**: No star imports, remove unused imports
+- **Line Endings**: Unix (LF) only, no Windows CRLF
+- **Comments**: DO NOT ADD COMMENTS unless explicitly requested
+- **Copyright**: Always update copyright year to 2025 if editing files; add `// Content with portions generated by generative AI platform` after copyright header
```
pierantoniomerlino: Happy new year! Update it to 2026 :-)

mattdibi: Wouldn't it be better to updated it to "the current year"? @MMaiero

pierantoniomerlino: I agree, so we can avoid to change it every year.

MMaiero: I am not sure if the model knows what is the current year, especially if someone uses a model that is completely offline

pierantoniomerlino (approved):

AGENTS.md
mattdibi: Even if completely offline I would expect the model to try and inquire the OS about the current date (via MCP or simply via CLI). In a year I'm sure we'll forget to update the AGENTS.md file.

MMaiero: > In a year I'm sure we'll forget to update the AGENTS.md file.

For sure

MMaiero: It seems to be still a problem the current date awareness https://github.com/anthropics/claude-code/issues/7537

PR #6131, opened 2025-12-16
fix(container.orchestration.provider): Fixed build path and classpath

salvatore-coppola:
Fixed build path and classpath, adding missing lib in build.properties and MANIFEST.MF

### DIFF
--- a/kura/org.eclipse.kura.container.orchestration.provider/META-INF/MANIFEST.MF
+++ b/kura/org.eclipse.kura.container.orchestration.provider/META-INF/MANIFEST.MF
@@ -38,6 +38,7 @@ Service-Component: OSGI-INF/component.xml
 Bundle-ClassPath: .,
  lib/commons-codec.jar,
  lib/commons-compress.jar,
+ lib/docker-java.jar,
  lib/docker-java-api.jar,
  lib/docker-java-core.jar,
  lib/docker-java-transport.jar,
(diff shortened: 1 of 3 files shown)


kura/org.eclipse.kura.container.orchestration.provider/META-INF/MANIFEST.MF
```diff
@@ -38,6 +38,7 @@ Service-Component: OSGI-INF/component.xml
 Bundle-ClassPath: .,
  lib/commons-codec.jar,
  lib/commons-compress.jar,
+ lib/docker-java.jar,
```
nicolatimeus, line 41: The docker-java artifact seem to provide only the [DockerClientBuilder](https://github.com/docker-java/docker-java/blob/faa88e16460a8cb321c9695cdbc34cb7a662458e/docker-java/src/main/java/com/github/dockerjava/core/DockerClientBuilder.java#L11) class which does not seem to be used anywhere in the code. Are you sure that we need this dependency at all?

salvatore-coppola: no idea, I only fixed the classpath, maybe it is used by the other dependencies

nicolatimeus: What do you mean with fixed the classpath? Can you also try to remove the docker-java dependency?

PR #6134, opened 2025-12-24
build(target-platform-bom): update netty to 4.1.130.Final

MMaiero:
Updating to latest available version. Some security fixes available, for example for CVE-2025-67735


salvatore-coppola (approved):
I approve but we need to run artemis test relying on netty to confirm everything is ok (but since it is a patch release...)

PR #6136, opened 2026-01-07
chore: Updated AGENTS.md file

MMaiero:
Explicitly requiring the header on new file.
Updated the reference to the year removing s static year and requesting to set the current year


AGENTS.md
```diff
@@ -13,7 +13,37 @@
...
+        This program and the accompanying materials are made
+        available under the terms of the Eclipse Public License 2.0
+        which is available at https://www.eclipse.org/legal/epl-2.0/
+
+        SPDX-License-Identifier: EPL-2.0
+
+        Contributors:
+         {holder}
+    ```
+    Always update copyright year to the current year if editing files; add `// Content with portions generated by generative AI platform` after copyright header.
```
mattdibi: 🤔 I think we're asking too much out of the LLM here. It doesn't have a proper way to fill the `{holder}` information. I would just change the year as per our offline discussion and leave the rest as-is

MMaiero: I have asked codex to create a new java class and it was able to correctly fill the gaps. But it set 2025:

```
/******************************************************************************
 * Copyright (c) 2025 Eurotech and/or its affiliates and others
 *
 * This program and the accompanying materials are made
 * available under the terms of the Eclipse Public License 2.0
 * which is available at https://www.eclipse.org/legal/epl-2.0/
 *
 * SPDX-License-Identifier: EPL-2.0
 *
 * Contributors:
 *  Eurotech
// Content with portions generated by generative AI platform
package org.eclipse.kura;

public final class Template {

    private Template() {
    }
}

```

mattdibi: 😭

mattdibi: @MMaiero Let's leave it as-is. When 2027 will come the copyright check failures will remind us to update the agent file... hopefully

MMaiero: I am worried that the current version does not specify anything about the header for new files. So, yes, we could leave it as is now. But I believe that we will have to specify something like in this PR to specify that we want the copyright header for new files.

PR #6151, opened 2026-01-20
docs: Added section for ppp-based modems

pierantoniomerlino:
This pull request adds new documentation to clarify how NetworkManager handles network failover when using PPP-based cellular modems. It explains a known issue where NetworkManager adds a high metric to PPP interfaces, affecting failover behavior, and provides workarounds for using cellular connections as either primary or backup WAN interfaces.

**Documentation improvements for PPP-based modem failover:**

* Added a section describing how NetworkManager adds a value of 20000 to the metric for PPP-based modem interfaces, which can interfere with expected failover behavior. Included a reference to the upstream NetworkManager issue.
* Provided step-by-step workarounds for configuring priorities when using a cellular connection as either a primary or backup WAN interface, outlining limitations and configuration recommendations.

### DIFF
--- a/docs/gateway-configuration/network-failover.md
+++ b/docs/gateway-configuration/network-failover.md
@@ -77,7 +77,7 @@ Plese refer to the official NetworkManager [documentation](https://networkmanage
 ## Network Failover with PPP-based Modems
 
 When NetworkManager uses `ppp` technology to establish a cellular connection, a value of 20000 is always added to the modem interface causing an improper behavior of the failover feature.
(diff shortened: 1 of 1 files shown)


pierantoniomerlino:
@MMaiero @mattdibi Please take a look at the added section. Check if it's clear enought and if there are other cases to be covered.

mattdibi (approved):
LGTM

I think it covers pretty much everything. In summary, users needs to be aware of the fact that PPP-based interfaces will always have 20000 added to their metric.

I would probably add the fact that this behaviour is present only on specific versions of NM. Given that they recognised this as a bug they might fix it in a later version.

MMaiero (approved):

PR #6173, opened 2026-02-10
feat(core.system): Added logging and ping timeout to internet connection status task

salvatore-coppola:
Added logging and ping timeout

### DIFF
--- a/kura/org.eclipse.kura.core.system/src/main/java/org/eclipse/kura/core/system/SystemServiceImpl.java
+++ b/kura/org.eclipse.kura.core.system/src/main/java/org/eclipse/kura/core/system/SystemServiceImpl.java
@@ -1606,28 +1606,26 @@ public class SystemServiceImpl extends SuperSystemService implements SystemServi
 
             if (isPingable(StandardProtocolFamily.INET, getInternetConnectionStatusCheckHost())
                     || isPingable(StandardProtocolFamily.INET6, getInternetConnectionStatusCheckHost())) {
-                this.currentInternetStatus.set(InternetConnectionStatus.FULL);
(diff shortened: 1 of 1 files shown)


pierantoniomerlino (dismissed):

kura/org.eclipse.kura.core.system/src/main/java/org/eclipse/kura/core/system/SystemServiceImpl.java
```diff
@@ -1599,26 +1599,36 @@ private void checkInternetTask() {
...
-            this.currentInternetStatus.set(InternetConnectionStatus.UNAVAILABLE);
+            updateStatus(oldStatus, InternetConnectionStatus.UNAVAILABLE);
         } catch (Exception e) {
             logger.error("Error while checking internet connection status", e);
         }
     }
 
+    private void updateStatus(InternetConnectionStatus oldStatus, InternetConnectionStatus newStatus) {
+        this.currentInternetStatus.set(newStatus);
+        if (newStatus != oldStatus) {
```
MMaiero, line 1627: Don't remember if it is checked before writing in the log, but does it make sense to add to the if here the check if we are in debug? So that the write in the log is done only if we are in the actual debug mode for the logs

salvatore-coppola: Well, the debug level is designed for this purpose. Usually, isDebug is used to avoid expensive computations. 
Here, however, there are no computations of any kind.

pierantoniomerlino (approved):

MMaiero (approved):

PR #6179, opened 2026-02-16
chore(core.data.transport): Enhanced connection error log meesage

salvatore-coppola:
Enhanced connection error log meesage


salvatore-coppola:
it doesn't like

PR #6180, opened 2026-02-17
fix: Fixed error on sslmanagerservice not initialized

salvatore-coppola:
Fixed error during Kura bootstrap caused by SslManagerService not initialized in MqttDataTrasport

### DIFF
--- a/kura/org.eclipse.kura.cloud.base.provider/src/main/java/org/eclipse/kura/core/data/transport/mqtt/MqttDataTransport.java
+++ b/kura/org.eclipse.kura.cloud.base.provider/src/main/java/org/eclipse/kura/core/data/transport/mqtt/MqttDataTransport.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2011, 2023 Eurotech and/or its affiliates and others
+ * Copyright (c) 2011, 2026 Eurotech and/or its affiliates and others
  *
  * This program and the accompanying materials are made
  * available under the terms of the Eclipse Public License 2.0
@@ -16,6 +16,7 @@ package org.eclipse.kura.core.data.transport.mqtt;
 import java.nio.charset.StandardCharsets;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.Optional;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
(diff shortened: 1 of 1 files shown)


kura/org.eclipse.kura.cloud.base.provider/src/main/java/org/eclipse/kura/core/data/transport/mqtt/MqttDataTransport.java
```diff
@@ -155,8 +156,8 @@ public void setSslManagerService(SslManagerService sslManagerService) {
 
     public void unsetSslManagerService(SslManagerService sslManagerService) {
         synchronized (this.updateLock) {
-            if (sslManagerService == this.sslManagerService) {
-                this.sslManagerService = null;
+            if (Optional.of(sslManagerService).equals(Optional.of(this.sslManagerService))) {
```
marcellorinaldo: this.sslManagerService is already an `Optional`

kura/org.eclipse.kura.cloud.base.provider/src/main/java/org/eclipse/kura/core/data/transport/mqtt/MqttDataTransport.java
```diff
@@ -824,15 +825,20 @@ private MqttClientConfiguration buildConfiguration(Map<String, Object> propertie
...
         // SSL
-        if (isSSL(brokerUrl)) {
-            try {
-                SSLSocketFactory ssf = this.sslManagerService.getSSLSocketFactory();
-
-                conOpt.setSocketFactory(ssf);
-            } catch (Exception e) {
-                logger.error("SSL setup failed", e);
-                throw new IllegalStateException("SSL setup failed");
+        if (this.sslManagerService == null) {
```
marcellorinaldo, line 828: Can't we initialize it as empty and the use the `isPresent` instead of null-checking?

marcellorinaldo (changes requested):

kura/org.eclipse.kura.cloud.base.provider/src/main/java/org/eclipse/kura/core/data/transport/mqtt/MqttDataTransport.java
salvatore-coppola: they are different cases, null not initialized, empty, the service is not present

marcellorinaldo: Can't we treat them as they were the same case? Now if it is empty we do nothing and print nothing

salvatore-coppola: i'll add a log message

marcellorinaldo (dismissed):

marcellorinaldo (approved):

salvatore-coppola:
the issue reported by sonar is a false  flag, I'd go for a forcemerge

PR #6183, opened 2026-02-17
feat!: Java and JRE 21 support

salvatore-coppola:
Now the source code and the target code are set to java 21.

The required jre version is the 21th and tycho plugin has been upgraded to version 5.0.2.

The following dependencies have been upgraded:
* mockito: 5.21.0
* byte-buddy: 1.17.7
* maven-compiler: 3.15.0


kura/pom.xml
```diff
@@ -104,8 +104,8 @@
         <check.plugins.exist.remote>false</check.plugins.exist.remote>
         <kura.addons.url>https://artifactory.dev.everyware.io/artifactory/kura-addons</kura.addons.url>
 
-        <maven.compiler.source>17</maven.compiler.source>
-        <maven.compiler.target>17</maven.compiler.target>
+        <maven.compiler.source>21</maven.compiler.source>
```
marcellorinaldo: We should probably use
```
<maven.compiler.release>21</maven.compiler.release>
```
instead of the target and source options.
https://maven.apache.org/plugins/maven-compiler-plugin/examples/set-compiler-release.html

Remember to add the property also in `target-platform/pom.xml`, there the properties are still set at 17.

kura/test/pom.xml
```diff
@@ -162,6 +162,7 @@
                                     ${tycho.testArgLine}
                                     --add-opens=java.base/java.lang=ALL-UNNAMED
                                     --add-opens=java.base/java.util=ALL-UNNAMED
+                                    -javaagent:${settings.localRepository}/org/mockito/mockito-core/5.21.0/mockito-core-5.21.0.jar
```
marcellorinaldo: I would use a property to make the future updates of mockito easier
```
<dependency>
    <groupId>org.mockito</groupId>
    <artifactId>mockito-core</artifactId>
    <version>4.8.1</version>
    <version>${mockito.version}</version>
    <scope>test</scope>
</dependency>
```

```
-javaagent:${settings.localRepository}/org/mockito/mockito-core/5.21.0/mockito-core-${mockito.version}.jar
```

kura/distrib/pom.xml
```diff
@@ -38,7 +38,7 @@
     </modules>
 
     <properties>
-        <tycho-version>4.0.11</tycho-version>
+        <tycho-version>5.0.2</tycho-version>
```
marcellorinaldo: I think we can remove the property from here, there seems to be no tycho plugin in this project

kura/test/org.eclipse.kura.camel.test/META-INF/MANIFEST.MF
```diff
@@ -13,6 +13,6 @@ Import-Package: javax.activation,
  org.junit;version="[4.12.0,5.0.0)",
  org.junit.runner;version="[4.12.0,5.0.0)",
  org.junit.runners;version="[4.12.0,5.0.0)",
- org.mockito;version="[4.0.0,5.0.0)",
- org.mockito.invocation;version="[4.0.0,5.0.0)",
- org.mockito.stubbing;version="[4.0.0,5.0.0)"
+ org.mockito;version="5.0.0",
```
marcellorinaldo, line 16: Why not using a strict range like `;version="[5.0,6.0)"`?

marcellorinaldo (changes requested):

kura/test/org.eclipse.kura.camel.test/META-INF/MANIFEST.MF
salvatore-coppola: our policy usually is to use fixed version for non-kura packages

marcellorinaldo: It is not a fixed version, this means `[5.0.0, inf]` (no upper bound) https://docs.osgi.org/specification/osgi.core/7.0.0/framework.module.html#i3189032

salvatore-coppola: i know, with the fixed version i meant a fixed number, but I repeat myself, our policy for external dependencies is not rely on semantic version for projects outside kura space because we cannot be sure they adopt our same policy

kura/pom.xml
salvatore-coppola: OK, but I'll wait for the decision on whether we want to maintain retrocompatibility with version 17 or not

salvatore-coppola: added <maven.compiler.release>21</maven.compiler.release> in addition to others properties

marcellorinaldo (changes requested):
Please correct some missing references to Java 17:

- README.md https://github.com/eclipse-kura/kura/blob/develop/README.md
- AGENTS.md https://github.com/eclipse-kura/kura/blob/develop/AGENTS.md
- Release notes workflow https://github.com/eclipse-kura/kura/blob/8ee22db826ac319bab32217d45290830ecb0bf3d/.github/workflows/release-notes.yml#L35
- SBOMs workflows https://github.com/eclipse-kura/kura/blob/8ee22db826ac319bab32217d45290830ecb0bf3d/.github/workflows/target-platform-sbom.yml#L20 https://github.com/eclipse-kura/kura/blob/8ee22db826ac319bab32217d45290830ecb0bf3d/.github/workflows/kura-core-sbom.yml#L20

marcellorinaldo (approved):

PR #6188, opened 2026-02-20
fix: wrong keystore type in rest keystore response

sfiorani:
When calling the `/services/keystores/v1` api, the returned data could report a wrong keystore type, setting them all to PKCS12, even if the keystore was, for example a JKS. This PR fixes the problem and unifies the keystore type reported with all uppercase strings (there were some cases in which the type could be returned in lowecase, `pkcs12` instead of `PKCS12`).


MMaiero:
Also see: https://github.com/eclipse-kura/kura/pull/5752

salvatore-coppola (approved):

PR #6197, opened 2026-03-05
feat(linux.gpio.libgpiod): Added support to multiple pin listeners

pierantoniomerlino:
This is the backport of eclipse-kura/kura-gpio#6


pierantoniomerlino:
The Sonar analysis fails for a smell on test, but this is a false positive.

mattdibi (approved):

PR #6200, opened 2026-03-09
build(archetype): bnd migration

mattdibi:
This PR migrates the Eclipse Kura archetype from [Tycho](https://github.com/eclipse-tycho/tycho) to [bnd](https://bnd.bndtools.org/). This is the first step into migrating the whole project to bnd in our ongoing effort to improve the developer experience.

### Testing instructions

<details>
<summary> <a href="https://github.com/eclipse-kura/kura/pull/6200#issuecomment-4080457030">Outdated. See comments below</a> </summary>

#### Generating a project using the archetype

Install the archetype:

```
mvn install:install-file \
-Dfile=./kura-addon-archetype-6.0.0-SNAPSHOT.jar \
-DgroupId=org.eclipse.kura \
-DartifactId=kura-addon-archetype \
-Dversion="6.0.0-SNAPSHOT" \
-Dpackaging=jar \
-DgeneratePom=true
```

Start the generation of the archetype in interactive mode:

```
mvn archetype:generate \              
-DarchetypeArtifactId=kura-addon-archetype \      
-DarchetypeGroupId=org.eclipse.kura \
-DarchetypeVersion="6.0.0-SNAPSHOT"
```

Maven will ask for a few parameters on the command line:

- `groupId` : the Maven group id of the generated POM files. Example: `org.eclipse.kura`
- `artifactId`: the Maven artifact id of the generated parent pom file, and the name of the generated top level project folder. Example: `example`
- `package`: the base Java package and the main artifact of the project. Example: `org.eclipse.kura.example`.

Add the generated sources to `git`

```
git init
git add .
git commit -m "initial commit"
```

#### Building a project generated using the archetype

Perform a build skipping the tests

```
mvn clean install -DskipTests=true
```

Generate the bndrun file (you'll only need to run this once)

```
mvn -pl tests/<your-test-project> -am \
   bnd-indexer:index \
   bnd-indexer:index@test-index \
   bnd-resolver:resolve
```

Now you're all set and you can run the tests

```
mvn verify
```

</details>

#### Why are test dependencies not in the `test` scope?

Integration tests are treated as *main code* by Maven not test code and therefore don't have access to test-scoped dependencies. Therefore I had to remove the test dependencies scope.

This have a negative impact on SBOM computation since we'll see all Kura dependencies in there (which are required to run the Kura emulator in the integration tests). To avoid this we need to update the SBOM generation job to use the `-DexcludeTestProject` property. Example:

```
mvn org.cyclonedx:cyclonedx-maven-plugin:2.9.1:makeAggregateBom \
   -DprojectType=framework \
   -Dcyclonedx.skipNotDeployed=false \
   -DexcludeTestProject
```

Possible alternatives:
- Leverage the `scopes` property of the [bnd resolver plugin](https://github.com/bndtools/bnd/blob/master/maven-plugins/bnd-resolver-maven-plugin/README.md).

#### Why is the framework shut down from within the integration tests (see [ExampleComponentItTest.java‎](https://github.com/eclipse-kura/kura/pull/6200/changes#diff-babf586f6ec47f5f9db2c86592fb9a7ffe77c2a34fa39e3a31ce506e4a3e7de8))?

The framework ran by bnd doesn't shutdown after running the tests. This means we don't trip over `deactivate` method of the bundle under test. To overcome this I added the framework shutdown directly in the integration test code.

If we don't care about the framework shutting down after tests are ran we can remove that part of the code.

#### Integration tests not actually running doesn't trigger an error

There's an annoying issue when integration tests are not properly run: if you remove the testing bundle from the `runrequires` the framework is started but no error is thrown. The only way to notice the issue is by checking the coverage reports.

---

##### Resources

- Experimental project PR: https://github.com/mattdibi/kura-bnd-example-2/pull/1
- Maven plugin docs: https://github.com/bndtools/bnd/blob/master/maven-plugins/README.md
- OSGi enRoute Microservices tutorial: https://enroute.osgi.org/tutorial/030-tutorial_microservice.html
- Bnd Testing: https://bnd.bndtools.org/chapters/310-testing.html


kura/tools/kura-addon-archetype/src/main/resources/archetype-resources/pom.xml
```diff
@@ -138,80 +141,120 @@
...
+                    <executions>
+                        <execution>
+                            <goals>
+                                <goal>bnd-process</goal>
+                            </goals>
+                        </execution>
+                    </executions>
+                </plugin>
+
+                <!-- Required to make the maven-jar-plugin pick up the bnd
```
nicolatimeus, line 156: Have you tried configuring bnd-maven-plugin with `<extensions>true</extensions>`?

mattdibi: > When using this goal the `bnd-maven-plugin` must be configured with `<extensions>true</extensions>` in order to coordinate with the `maven-(jar|war)-plugin`.

https://github.com/bndtools/bnd/blob/master/maven-plugins/bnd-maven-plugin/README.md#jar-goal

🤔 I totally missed this.

mattdibi: Ok so, according to the docs, enabling extensions is not required given that we're using the `bnd-process` goal and not the `jar` goal. The `maven-jar-plugin` is required by https://github.com/bndtools/bnd/blob/master/maven-plugins/bnd-maven-plugin/README.md#important-note-about-maven-jarwar-plugin.

I tried setting `<extensions>true</extensions>` anyway but nothing changed. Removing `maven-jar-plugin` when extensions is `true` makes the integration tests fail and the resulting manifest is wrong.

kura/tools/kura-addon-archetype/src/main/resources/archetype-resources/distrib/pom.xml
```diff
@@ -34,7 +34,8 @@
         <dependency>
             <groupId>${groupId}</groupId>
             <artifactId>${package}</artifactId>
-            <type>eclipse-plugin</type>
+            <type>jar</type>
```
marcellorinaldo: The JAR packaging is the default, we can omit that here

kura/tools/kura-addon-archetype/src/main/resources/archetype-resources/__package__/pom.xml
```diff
@@ -25,7 +25,7 @@
     </parent>
 
     <artifactId>${package}</artifactId>
-    <packaging>eclipse-plugin</packaging>
+    <packaging>jar</packaging>
```
marcellorinaldo: Since now we are producing JARs and no more `eclipse-plugin` packages, we should update also the entry in the BOM: https://github.com/eclipse-kura/kura/blob/2d05dc062107223892de1c5c72d1c8171aba83d2/kura/tools/kura-addon-archetype/src/main/resources/archetype-resources/bom/pom.xml#L35

(we can remove the type altogether, since JAR is the default type)

kura/tools/kura-addon-archetype/src/main/resources/archetype-resources/tests/__package__.test/pom.xml
```diff
@@ -26,9 +26,18 @@
...
     <artifactId>${package}.test</artifactId>
-    <packaging>eclipse-test-plugin</packaging>
     <description>Tests of ${package}</description>
 
+    <!-- Insert your bundles as dependencies -->
+    <dependencies>
+        <dependency>
+            <groupId>${groupId}</groupId>
+            <artifactId>${package}</artifactId>
+            <type>jar</type>
```
marcellorinaldo: The JAR packaging is the default, we can omit that here

kura/tools/kura-addon-archetype/src/main/resources/archetype-resources/tests/pom.xml
```diff
@@ -38,13 +38,37 @@
...
+        </dependency>
+
+        <dependency>
+            <groupId>org.eclipse.kura</groupId>
+            <artifactId>kura-pde-deps</artifactId>
+            <version>6.0.0-SNAPSHOT</version>
+            <type>pom</type>
+        </dependency>
+
+        <dependency>
```
marcellorinaldo, line 67: Why do we need to import the pde deps?

kura/tools/kura-addon-archetype/src/main/resources/archetype-resources/tests/__package__.test/pom.xml
```diff
@@ -26,9 +26,18 @@
...
+            <version>${project.version}</version>
```
marcellorinaldo: Version should not be needed

kura/tools/kura-addon-archetype/src/main/resources/archetype-resources/distrib/pom.xml
```diff
@@ -34,7 +34,8 @@
...
+            <version>${project.version}</version>
```
marcellorinaldo: Version should not be needed here

kura/tools/kura-addon-archetype/src/main/resources/archetype-resources/tests/__package__.test/pom.xml
```diff
@@ -51,21 +60,46 @@
...
+                <configuration>
+                    <bndfile>integration-test.bndrun</bndfile>
+                </configuration>
             </plugin>
-
             <plugin>
                 <groupId>biz.aQute.bnd</groupId>
-                <artifactId>bnd-maven-plugin</artifactId>
+                <artifactId>bnd-indexer-maven-plugin</artifactId>
+                <version>${bnd.version}</version>
```
marcellorinaldo: We have already specified the version for this plugin the root project

kura/tools/kura-addon-archetype/src/main/resources/archetype-resources/tests/__package__.test/pom.xml
```diff
@@ -51,21 +60,46 @@
...
+                <configuration>
+                    <includeJar>true</includeJar>
+                </configuration>
+            </plugin>
+            <plugin>
+                <groupId>biz.aQute.bnd</groupId>
+                <artifactId>bnd-testing-maven-plugin</artifactId>
+                <version>${bnd.version}</version>
```
marcellorinaldo: We have already specified the version for this plugin the root project

kura/tools/kura-addon-archetype/src/main/resources/archetype-resources/pom.xml
```diff
@@ -106,6 +104,11 @@
             <name>Kura Addons Maven Repository</name>
             <url>https://artifactory.dev.everyware.io/artifactory/kura-addons</url>
         </repository>
+        <repository>
```
marcellorinaldo, line 107: Do we plan to use the moquette broker in every test? Does it make sense to have it in the archetype?

marcellorinaldo (commented):
One question, can we embed this required step into some maven phase?

```
mvn -pl tests/<your-test-project> -am \
   bnd-indexer:index \
   bnd-indexer:index@test-index \
   bnd-resolver:resolve
```

So that we can build the whole project with just `mvn clean install`?

kura/tools/kura-addon-archetype/src/main/resources/archetype-resources/tests/pom.xml
```diff
@@ -38,13 +38,37 @@
...
         </dependency>
         <dependency>
             <groupId>junit</groupId>
             <artifactId>junit</artifactId>
             <version>4.13.2</version>
-            <scope>test</scope>
+        </dependency>
+
+        <!-- This OSGi bundle wraps junit 4.13.2 jar file. -->
+        <dependency>
```
marcellorinaldo, line 49: From the enroute docs they are using:

```xml
<groupId>org.osgi</groupId>
<artifactId>osgi.enroute.junit.wrapper</artifactId>
<version>4.12.0</version>
```

Should we consider that one too?

kura/tools/kura-addon-archetype/src/main/resources/META-INF/maven/archetype-metadata.xml
marcellorinaldo: We should remove this too

marcellorinaldo (changes requested):

marcellorinaldo:
Should we take care of this? https://bnd.bndtools.org/chapters/310-testing.html. By default, BND uses the tester that has been deprecated https://bnd.bndtools.org/chapters/310-testing.html#default-tester---bizaqutetester

kura/tools/kura-addon-archetype/src/main/resources/archetype-resources/pom.xml
```diff
@@ -50,13 +50,11 @@
     <properties>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <maven.compiler.release>21</maven.compiler.release>
-        <tycho.version>5.0.2</tycho.version>
         <bnd.version>7.1.0</bnd.version>
```
marcellorinaldo, line 53: Is there a specific reason to why we are not using the latest version of the bnd plugins?

https://mvnrepository.com/artifact/biz.aQute.bnd/bnd-maven-plugin/7.2.1

mattdibi:
> Should we take care of this? https://bnd.bndtools.org/chapters/310-testing.html. By default, BND uses the tester that has been deprecated https://bnd.bndtools.org/chapters/310-testing.html#default-tester---bizaqutetester

The new tester requires JUnit 5 (correct me if I'm wrong). We're still running on JUnit 4.

kura/tools/kura-addon-archetype/src/main/resources/archetype-resources/pom.xml
mattdibi: I adopted the one we were already running.

marcellorinaldo:
> > Should we take care of this? https://bnd.bndtools.org/chapters/310-testing.html. By default, BND uses the tester that has been deprecated https://bnd.bndtools.org/chapters/310-testing.html#default-tester---bizaqutetester
> 
> The new tester requires JUnit 5 (correct me if I'm wrong). We're still running on JUnit 4.

Yes correct. I was wondering if this was the right time to update

mattdibi:
> > > Should we take care of this? https://bnd.bndtools.org/chapters/310-testing.html. By default, BND uses the tester that has been deprecated https://bnd.bndtools.org/chapters/310-testing.html#default-tester---bizaqutetester
> > 
> > 
> > The new tester requires JUnit 5 (correct me if I'm wrong). We're still running on JUnit 4.
> 
> Yes correct. I was wondering if this was the right time to update

I'm not against it but I wouldn't address the update in this PR. I would update JUnit in a separate PR.

kura/tools/kura-addon-archetype/src/main/resources/META-INF/maven/archetype-metadata.xml
mattdibi: Addressed in b836975d8a0355fc8157b7a25f8d778bb857c105

kura/tools/kura-addon-archetype/src/main/resources/archetype-resources/distrib/pom.xml
mattdibi: Addressed in c2802cc0fcb6d71b0ce4150be10e9c1f64be2e6c

kura/tools/kura-addon-archetype/src/main/resources/archetype-resources/__package__/pom.xml
mattdibi: Addressed in c2802cc0fcb6d71b0ce4150be10e9c1f64be2e6c

kura/tools/kura-addon-archetype/src/main/resources/archetype-resources/tests/__package__.test/pom.xml
mattdibi: Addressed in c2802cc0fcb6d71b0ce4150be10e9c1f64be2e6c

kura/tools/kura-addon-archetype/src/main/resources/archetype-resources/tests/__package__.test/pom.xml
mattdibi: Addressed in e3b5947001df9c0c95b808b8bd0f85bdd0d99a01

kura/tools/kura-addon-archetype/src/main/resources/archetype-resources/tests/__package__.test/pom.xml
mattdibi: Addressed in e3b5947001df9c0c95b808b8bd0f85bdd0d99a01

kura/tools/kura-addon-archetype/src/main/resources/archetype-resources/distrib/pom.xml
mattdibi: I distinctly remember the build failing due to the version missing 🤔  Must have had another issue somewhere because it is working correctly now.

Addressed in b68a67c8e0ad2b4ed807e63e49e58d2a36c57edc

kura/tools/kura-addon-archetype/src/main/resources/archetype-resources/tests/__package__.test/pom.xml
mattdibi: Addressed in b68a67c8e0ad2b4ed807e63e49e58d2a36c57edc

kura/tools/kura-addon-archetype/src/main/resources/archetype-resources/tests/pom.xml
mattdibi: Without them `bnd-resolver` is not able to resolve all the emulator dependencies. I get the following error:

```
[INFO] Resolved all Dependencies for org.eclipse.kura:org.eclipse.kura.pr.review.test:jar:1.0.0-SNAPSHOT. Checked: 238, Skipped(no need to revisit): 0, Failed: 0, Resulting Dependencies: 238
[ERROR] Resolution failed. Summary:
      ⇒ Bundle: org.eclipse.kura.core.configuration cannot be resolved
```

kura/tools/kura-addon-archetype/src/main/resources/archetype-resources/tests/pom.xml
mattdibi: Image

Image

🤔  they're both no longer maintained but the Apache wraps more versions and seems to be more popular. I would keep this one.

*Note*: As of JUnit 5.6, the JUnit jars already have the OSGi metadata and so can be used as bundles directly from Maven Central. Therefore, when we'll switch to newer versions of JUnit, we won't need the bundle wrappers anymore.

kura/tools/kura-addon-archetype/src/main/resources/archetype-resources/pom.xml
mattdibi: I was wondering the same. With the archetype we're providing a platform for developers to develop plugin. We don't know whether they'll need the moquette broker in their tests or not. I thought that providing *more* than they need is better than the opposite. 

If you disagree I'm ok with removing the broker from the archetype.

mattdibi:
> One question, can we embed this required step into some maven phase?
> 
> ```
> mvn -pl tests/<your-test-project> -am \
>    bnd-indexer:index \
>    bnd-indexer:index@test-index \
>    bnd-resolver:resolve
> ```
> 
> So that we can build the whole project with just `mvn clean install`?

This is quite an interesting point. Take into account that:
- In the various bnd tutorials I found they seem to keep that goal separate (see: https://enroute.osgi.org/tutorial/030-tutorial_microservice.html). Same goes for the [official documentation](https://github.com/bndtools/bnd/tree/master/maven-plugins/bnd-resolver-maven-plugin#executing-the-resolve-operation) but they do point out that it can be bound to a lifecycle phase.
- It only needs to be ran once (and when `runrequires` changes)

I'll give this a try.

kura/tools/kura-addon-archetype/src/main/resources/archetype-resources/tests/pom.xml
marcellorinaldo: Ok, great then!

kura/tools/kura-addon-archetype/src/main/resources/archetype-resources/pom.xml
marcellorinaldo: It makes sense to leave it there since the `org.eclipse.kura.core.testutil` is requiring it (see [here](https://github.com/eclipse-kura/kura/blob/89115aaa683e1e72f5f8022ac6c2886293484468/kura/test-util/org.eclipse.kura.core.testutil/META-INF/MANIFEST.MF#L56)) and that bundle is likely to be used in the tests

mattdibi:
> > One question, can we embed this required step into some maven phase?
> > ```
> > mvn -pl tests/<your-test-project> -am \
> >    bnd-indexer:index \
> >    bnd-indexer:index@test-index \
> >    bnd-resolver:resolve
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > So that we can build the whole project with just `mvn clean install`?
> 
> This is quite an interesting point. Take into account that:
> 
> * In the various bnd tutorials I found they seem to keep that goal separate (see: https://enroute.osgi.org/tutorial/030-tutorial_microservice.html). Same goes for the [official documentation](https://github.com/bndtools/bnd/tree/master/maven-plugins/bnd-resolver-maven-plugin#executing-the-resolve-operation) but they do point out that it can be bound to a lifecycle phase.
> * It only needs to be ran once (and when `runrequires` changes)
> 
> I'll give this a try.

Addressed in dd51c0922ce7da4b1530ca76cdd6be0bb865eaad

### Updated testing instructions

#### Generating a project using the archetype

Install the archetype:

```
mvn install:install-file \
-Dfile=./kura-addon-archetype-6.0.0-SNAPSHOT.jar \
-DgroupId=org.eclipse.kura \
-DartifactId=kura-addon-archetype \
-Dversion="6.0.0-SNAPSHOT" \
-Dpackaging=jar \
-DgeneratePom=true
```

Start the generation of the archetype in interactive mode:

```
mvn archetype:generate \              
-DarchetypeArtifactId=kura-addon-archetype \      
-DarchetypeGroupId=org.eclipse.kura \
-DarchetypeVersion="6.0.0-SNAPSHOT"
```

Maven will ask for a few parameters on the command line:

- `groupId` : the Maven group id of the generated POM files. Example: `org.eclipse.kura`
- `artifactId`: the Maven artifact id of the generated parent pom file, and the name of the generated top level project folder. Example: `example`
- `package`: the base Java package and the main artifact of the project. Example: `org.eclipse.kura.example`.

Add the generated sources to `git`

```
git init
git add .
git commit -m "initial commit"
```

#### Building a project generated using the archetype

```
mvn clean install
```

marcellorinaldo (dismissed):

mattdibi:
After some offline discussion we decided to drop the changes documented in https://github.com/eclipse-kura/kura/pull/6200#issuecomment-4074268635 given that it might be possible for the `bnd-resolver` to change the `runbundles` in the run file without dependency changes. This would add unnecessary changes to a file we plan on tracking under git (i.e. the `.bndrun` files).

Leveraging [`failOnChanges` and `writeOnChanges`](https://github.com/bndtools/bnd/tree/master/maven-plugins/bnd-resolver-maven-plugin#additional-configuration-properties-for-the-resolve-goal) wouldn't work for our intended use.

Given the above we decided to revert the changes introduces in d81da90 and give the user the control on when to resolve the `runbundles`. Possibly via a profile which is easier to handle than the command I documented in the initial version of the PR.

mattdibi:
As per my last comment, with dff8608 we introduced a new Maven profile responsible for enabling `bnd-resolver-maven-plugin` and resolving the `runbundles` before running the integration tests.

### Updated testing instructions

#### Generating a project using the archetype

Install the archetype:

```
mvn install:install-file \
-Dfile=./kura-addon-archetype-6.0.0-SNAPSHOT.jar \
-DgroupId=org.eclipse.kura \
-DartifactId=kura-addon-archetype \
-Dversion="6.0.0-SNAPSHOT" \
-Dpackaging=jar \
-DgeneratePom=true
```

Start the generation of the archetype in interactive mode:

```
mvn archetype:generate \              
-DarchetypeArtifactId=kura-addon-archetype \      
-DarchetypeGroupId=org.eclipse.kura \
-DarchetypeVersion="6.0.0-SNAPSHOT"
```

Maven will ask for a few parameters on the command line:

- `groupId` : the Maven group id of the generated POM files. Example: `org.eclipse.kura`
- `artifactId`: the Maven artifact id of the generated parent pom file, and the name of the generated top level project folder. Example: `example`
- `package`: the base Java package and the main artifact of the project. Example: `org.eclipse.kura.example`.

Add the generated sources to `git`

```
git init
git add .
git commit -m "initial commit"
```

#### Building a project generated using the archetype

```
mvn clean install -P resolve-integration-tests
```

> [!NOTE]
> After running the build with the `resolve-integration-tests` profile active you'll notice that the `.bndrun` files of the tests will be automatically updated. It is recommended to commit these changes.
>
> The `resolve-integration-tests` profile is only required to run when building the project for the first time and any time the `runrequires` field of the bndrun file changes.

After the first build you can build the project with:

```
mvn clean install
```

marcellorinaldo (commented):
Just a question, should we limit the bnd-indexer plugin execution to the profile as well? 

On a run without a profile it will be called and it outputs a message "no need to revisit": 
```
[INFO] 
[INFO] --- bnd-indexer:7.1.0:index (index) @ org.eclipse.kura.ciao.test ---
[INFO] Resolved all Dependencies for org.eclipse.kura:org.eclipse.kura.ciao.test:jar:1.0.0-SNAPSHOT. Checked: 313, Skipped(no need to revisit): 0, Failed: 0, Resulting Dependencies: 313
[INFO] 
[INFO] --- bnd-indexer:7.1.0:index (test-index) @ org.eclipse.kura.ciao.test ---
[INFO] Resolved all Dependencies for org.eclipse.kura:org.eclipse.kura.ciao.test:jar:1.0.0-SNAPSHOT. Checked: 313, Skipped(no need to revisit): 0, Failed: 0, Resulting Dependencies: 0
[INFO]
```

I was wondering if that plugin is able to modify the bndrun file as well. The indexer was mentioned in the previous tutorial:
```
mvn -pl tests/<your-test-project> -am \
   bnd-indexer:index \
   bnd-indexer:index@test-index \
   bnd-resolver:resolve
```

mattdibi:
> I was wondering if that plugin is able to modify the bndrun file as well. The indexer was mentioned in the previous tutorial:

Its only output is the `target/index.xml` file (and related archive) which is used by the `bnd-testing-maven-plugin` to run the framework from the bndrun file (which references the index). It doesn't affect the bndrun file directly.

> Just a question, should we limit the bnd-indexer plugin execution to the profile as well?

Given that it doesn't affect files under VCS I would keep it as-is. One optimisation we can perform is removing the `test-index` generation since we don't consume it (see d93e98a)

marcellorinaldo (approved):

mattdibi:
Given that the PR doesn't affect the build we're bypassing the CI check. For some reason tests keep randomly failing today...

PR #6201, opened 2026-03-16
build: updated nexus deployment link

sfiorani:
Alignment after eclipse migrated from `repo3` to `repo`


mattdibi:
LGTM

@sfiorani we also need to update the documentation since it still refers to the old repository.

mattdibi (approved):

PR #6249, opened 2026-05-26
ci: Fixed env.CHANGE_BRANCH expansion issue

salvatore-coppola:
This pull request updates the Jenkins pipeline to improve the handling of branch names in SonarQube analysis parameters by ensuring they are shell-escaped, which helps prevent issues with special characters in branch names.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,4 +1,5 @@
 import org.jenkinsci.plugins.pipeline.modeldefinition.Utils
+import org.apache.commons.text.StringEscapeUtils
 
 def boolean onlyDocumentationFilesChangedIn(String workDirectory) {
     if (!env.CHANGE_TARGET) {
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -106,8 +107,9 @@ node {
                     withSonarQubeEnv(credentialsId: 'sonarcloud-token') {
                         // Check if on primary branch
                         def analysisParameters = ""
+                        def changeBranch = StringEscapeUtils.escapeShellArg(env.CHANGE_BRANCH);
                         if (env.CHANGE_ID) {
-                            analysisParameters = "-Dsonar.pullrequest.branch=${env.CHANGE_BRANCH} -Dsonar.pullrequest.base=${env.CHANGE_TARGET} -Dsonar.pullrequest.key=${env.CHANGE_ID}"
+                            analysisParameters = "-Dsonar.pullrequest.branch=${changeBranch} -Dsonar.pullrequest.base=${env.CHANGE_TARGET} -Dsonar.pullrequest.key=${env.CHANGE_ID}"
```
mattdibi, line 112: I think we need to sanitize `CHANGE_TARGET` too.

> CHANGE_TARGET
> For a multibranch project corresponding to some kind of change request, this will be set to the target or base branch to which the change could be merged, if supported; else unset.

See: https://ci.eclipse.org/kura/job/multibranch/pipeline-syntax/globals

A malicious actor can create a new branch and create a PR that targets that branch thus code injecting the `CHANGE_TARGET` variable.

Jenkinsfile
mattdibi, line 1: If we want to proceed with this solution we would need to santise the variables used in `onlyDocumentationFilesChangedIn` too

Jenkinsfile
```diff
@@ -106,8 +107,9 @@ node {
...
                         } else {
                             analysisParameters = "-Dsonar.branch.name=${env.BRANCH_NAME}"
```
mattdibi, line 114: Same thing here probably

> BRANCH_NAME
> For a multibranch project, this will be set to the name of the branch being built, for example in case you wish to deploy to production from master but not from feature branches; if corresponding to some kind of change request, the name is generally arbitrary (refer to CHANGE_ID and CHANGE_TARGET).

mattdibi:
As discussed offline and in the original ticket: this doesn't prevent a malicious actor to modify our build and cause issues anyway. They don't need to resort to code injection to cause problems. I'm not convinced this PR achieves much honestly.

PR #6250, opened 2026-05-27
build: Updated io.netty version to 4.1.134.Final

pierantoniomerlino:
As per title.


marcellorinaldo (changes requested):
Please update the NOTICE file

marcellorinaldo (changes requested):
We're missing the update here: https://github.com/eclipse-kura/kura/blob/b90190b42f0170cafb93424cd48e7b37323929c1/kura/org.eclipse.kura.driver.opcua.provider/pom.xml#L34

marcellorinaldo (approved):

PR #6255, opened 2026-06-01
fix(core.keystore): prevent CRL downloads from blocking UI load

MMaiero:
Release the CRLManager instance lock before performing network I/O. The update cycle now splits into three phases: a short synchronized phase to collect pending downloads, an unsynchronized phase that waits on HTTP responses, and a short synchronized phase to apply results. This ensures callers of getCrls() are never blocked while a slow or unreachable CRL endpoint is being contacted.

Also adds connection (10 s) and read (30 s) timeouts on the HTTP connections opened by CRLUtil to bound the per-attempt wait time independently of the CompletableFuture outer timeout.


pierantoniomerlino (approved):

salvatore-coppola:
We can merge. By the way, I was unable to reproduce the issue.

salvatore-coppola (approved):

PR #6256, opened 2026-06-01
fix(web2): prevent APN password placeholder from being stored into configuration [backport release-5.6.0]

MMaiero:
The modem APN password placeholder was being written to ConfigurationAdmin and persisted into snapshots when the old config was not a GwtModemInterfaceConfig. Fixed the write path to skip storing the password when the placeholder is detected and the old config is not a modem config.


pierantoniomerlino (approved):

kura/org.eclipse.kura.web2/src/main/java/org/eclipse/kura/web/server/util/GwtServerUtil.java
```diff
@@ -10,6 +10,7 @@
  * Contributors:
  *  Eurotech
  *******************************************************************************/
+// Content with portions generated by generative AI platform
```
pierantoniomerlino, line 13: Is this comment needed?

pierantoniomerlino:
DO NOT MERGE THIS PR until the uptick to 5.6.3-SNAPSHOT is done in `release-5.6.0`.

PR #6264, opened 2026-06-04
fix(core.keystore): prevent CRL downloads from blocking UI load [backport release-5.6.0]

pierantoniomerlino:
Backport of #6255 

Release the CRLManager instance lock before performing network I/O. The update cycle now splits into three phases: a short synchronized phase to collect pending downloads, an unsynchronized phase that waits on HTTP responses, and a short synchronized phase to apply results. This ensures callers of getCrls() are never blocked while a slow or unreachable CRL endpoint is being contacted.

Also adds connection (10 s) and read (30 s) timeouts on the HTTP connections opened by CRLUtil to bound the per-attempt wait time independently of the CompletableFuture outer timeout.


salvatore-coppola (approved):

pierantoniomerlino:
The sonar analysis fails since it cannot detect an assertion in a test, even if it is present.

PR #6270, opened 2026-06-08
fix(test): fix intermittent failures in h2db provider tests due to shared state

salvatore-coppola:
## Summary

- **`DbDataStoreStorageTest`**: all tests were connecting to the same `jdbc:h2:mem:testdb` database, causing inter-test interference when run in any order. Each test now gets an isolated in-memory DB with a UUID-based name (`testdb_<uuid>`) and `DB_CLOSE_DELAY=-1`. A keep-alive connection is held open for the duration of the test via `@Before`/`@After`. JDBC connections opened inside `whenOverflowingIds` and `thenStoredMessageIs` are now properly closed via try-with-resources.
- **`H2DbServiceImplTest`**: `testUpdateUrlOccupied` was injecting an entry into the static `activeInstances` map without ever removing it, contaminating all subsequent tests. The entry is now removed in a `finally` block.

## Test plan

- [x] `mvn clean install` run 20 consecutive times on `org.eclipse.kura.db.h2db.provider.test` — all 20 passed (0 failures)

### DIFF
--- a/kura/test/org.eclipse.kura.db.h2db.provider.test/src/test/java/org/eclipse/kura/internal/db/h2db/provider/H2DbServiceImplTest.java
+++ b/kura/test/org.eclipse.kura.db.h2db.provider.test/src/test/java/org/eclipse/kura/internal/db/h2db/provider/H2DbServiceImplTest.java
@@ -49,7 +49,6 @@ public class H2DbServiceImplTest {
         H2DbServiceImpl svc = new H2DbServiceImpl();
         CryptoService csMock = mock(CryptoService.class);
         when(csMock.decryptAes(any(char[].class))).thenReturn(new char[0]);
-        when(csMock.decryptAes(encPass)).thenReturn(pass.toCharArray());
         svc.setCryptoService(csMock);
         svc.activate(Collections.emptyMap());
 
--- a/kura/test/org.eclipse.kura.db.h2db.provider.test/src/test/java/org/eclipse/kura/internal/db/h2db/provider/H2DbServiceImplTest.java
+++ b/kura/test/org.eclipse.kura.db.h2db.provider.test/src/test/java/org/eclipse/kura/internal/db/h2db/provider/H2DbServiceImplTest.java
@@ -1,12 +1,12 @@
 /*******************************************************************************
- * Copyright (c) 2017, 2024 Eurotech and/or its affiliates and others
- * 
+ * Copyright (c) 2017, 2026 Eurotech and/or its affiliates and others
+ *
  * This program and the accompanying materials are made
  * available under the terms of the Eclipse Public License 2.0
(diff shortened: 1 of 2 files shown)


pierantoniomerlino:
@salvatore-coppola The fix makes sense and indeed the tests passed. However, there are a lot of exceptions in the logs like this:

```
2026-06-22T09:19:23.723852Z pool-18-thread-1 ERROR Database initialization failed
java.lang.NullPointerException: Cannot invoke "org.eclipse.kura.crypto.CryptoService.decryptAes(char[])" because "this.cryptoService" is null
	at org.eclipse.kura.internal.db.h2db.provider.H2DbServiceImpl.decryptPassword(H2DbServiceImpl.java:530)
	at org.eclipse.kura.internal.db.h2db.provider.H2DbServiceImpl.updateInternal(H2DbServiceImpl.java:330)
	at org.eclipse.kura.internal.db.h2db.provider.H2DbServiceImpl.lambda$3(H2DbServiceImpl.java:152)
	at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:572)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:317)
	at java.base/java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:304)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1144)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:642)
	at java.base/java.lang.Thread.run(Thread.java:1583)
```

It happens also in my local environment, but with Eclipse in debug mode I cannot see any problem. Could you take a look? Maybe it's harmless and we can keep it as is, but it worths to check.

pierantoniomerlino:
It seems that the last update doesn't fix the issue.

kura/test/org.eclipse.kura.db.h2db.provider.test/src/test/java/org/eclipse/kura/internal/db/h2db/provider/H2DbServiceImplTest.java
```diff
@@ -46,12 +47,11 @@ public void testUpdate() throws KuraException, SQLException {
         String user = "USR";
 
         H2DbServiceImpl svc = new H2DbServiceImpl();
-        svc.activate(Collections.emptyMap());
-
         CryptoService csMock = mock(CryptoService.class);
-        svc.setCryptoService(csMock);
-
+        when(csMock.decryptAes(any(char[].class))).thenReturn(new char[0]);
```
pierantoniomerlino, line 51: The statement at line 52 is useless since the condition is included at line 51.

pierantoniomerlino (approved):

PR #6283, opened 2026-07-07
fix(web2): Manage empty passwords

pierantoniomerlino:
This PR add the support for empty passwords in the webUI. When a password is not set, it is applied and saved in the snapshot as an empty property instead of applying the default one. Moreover, when the webUI passes the `placeholder` word as a password, this is not applied nor saved in the snapshot.


pierantoniomerlino:
DO NOT MERGE THIS PR until the uptick to 5.6.3-SNAPSHOT is done in `release-5.6.0`.

MMaiero:
@pierantoniomerlino I have made some changes to the versions to make sure that we are distributing the correct bundle for web2

MMaiero (approved):

mattdibi (approved):

PR #6294, opened 2026-07-17
feat(container): pass container identity credentials via tmpfs token file

salvatore-coppola:
Container instances with identity integration enabled currently receive the temporary identity password through the `KURA_IDENTITY_PASSWORD` environment variable, which is exposed via `docker inspect`, `/proc/<pid>/environ` and container runtime metadata. This PR moves the credential to a tmpfs-backed file bind-mounted read-only into the container.

- The generated password is written to `<base>/kura-tokens/<uuid>/kura-token`, where `<base>` comes from the `kura.tmpfs.base` system property (default `/dev/shm`). The file has `400` permissions; both the shared `kura-tokens` directory and the per-token `<uuid>` directory are `700`. Everything is handled by the new package-private `TokenFileManager`.
- The file is bind-mounted read-only at `/run/secrets/kura-token` inside the container. The container is informed via the new `KURA_TOKEN_FILE=/run/secrets/kura-token` environment variable; `KURA_IDENTITY_NAME` and `KURA_REST_BASE_URL` are unchanged, `KURA_IDENTITY_PASSWORD` is removed completely (breaking change for containers relying on it).
- `ContainerOrchestrationServiceImpl` learns an internal `:ro` suffix convention on the container-path side of the volumes map to create read-only binds (`AccessMode.ro`). The suffix cannot be produced from user configuration (`ContainerInstanceOptions.parseVolume` drops 3-token entries), and user-configured volumes keep the read-write default.
- Token files are cleaned up (best-effort, warnings on failure) when the container is stopped/disabled, on configuration change, on startup failure and on bundle deactivation — all through the existing `cleanupTemporaryIdentity()` funnel.
- Error handling: a missing tmpfs base directory or an I/O failure during token file creation aborts container startup immediately (no retries) with an error message reporting the path. If POSIX permissions cannot be applied on a POSIX platform the partially created token is deleted and startup aborts, rather than handing out a world-readable secret; only on non-POSIX platforms (`UnsupportedOperationException`) is a warning logged and startup continued.

**Review feedback addressed:** POSIX permission failures now abort instead of being swallowed, the shared `kura-tokens` directory is also restricted to `700`, and the intentional cross-bundle duplication of the `:ro` suffix constant is documented.


salvatore-coppola:
## On-device smoke test report — tmpfs token passing

**Environment**: Raspberry Pi (aarch64, Debian, kernel 6.18), Kura 6.0.0-SNAPSHOT with the two bundles from this PR deployed, Docker 29.6.2, test container `nginx:latest` (runs as root) created via `ContainerInstance` with `container.identity.enabled=true`, `container.permissions=rest.system`, one user volume `/tmp/testdata:/data`.

### Results

| # | Check | Result | Evidence |
|---|-------|--------|----------|
| 1 | Token stored in tmpfs | ✅ PASS | `/dev/shm/kura-tokens/<uuid>/kura-token`, 32 bytes, file mode `400`, dir mode `700`, owner `kurad` |
| 2 | Environment variables | ✅ PASS | `KURA_IDENTITY_NAME`, `KURA_TOKEN_FILE=/run/secrets/kura-token`, `KURA_REST_BASE_URL=https://172.17.0.1:443/services` present; **no** `KURA_IDENTITY_PASSWORD` |
| 3 | Read-only mount | ✅ PASS | `HostConfig.Binds` = `.../kura-token:/run/secrets/kura-token:ro`; write attempt fails with "Read-only file system"; user volume stays `rw` |
| 4 | In-container read | ✅ PASS | `cat /run/secrets/kura-token` returns the password (container runs as root) |
| 5 | Credential validity | ✅ PASS | In-container `curl -u "$KURA_IDENTITY_NAME:$(cat $KURA_TOKEN_FILE)"` against `/services/system/v1/properties/framework` → **200**; wrong password → **401** |
| 6 | No credential leak via /proc | ✅ PASS | Token value found in **zero** `/proc/*/environ` on the host; `docker exec env` shows no password variable |
| 7 | Cleanup on config change | ✅ PASS | Old UUID dir deleted, new UUID dir created, container recreated with new bind |
| 8 | Cleanup on disable | ✅ PASS | `/dev/shm/kura-tokens` empty, container removed, temporary identity deleted (`Cleaned up temporary identity` in log) |
| 9 | Cleanup on framework stop | ⚠️ RESIDUAL | Platform limitation, see finding F1 (kept as documented residual, no code change) |
| 10 | Framework restart with surviving container | ✅ FIXED | Fixed in commit 7486b46, re-tested on device — see finding F2 |
| 11 | Reboot clears tmpfs | ⏭️ NOT RUN | To be verified manually; guaranteed by `/dev/shm` tmpfs semantics |

### Findings

#### F2 — Framework restart with a surviving container — FIXED (commit 7486b46)

Temporary identities are in-memory, so after a framework restart the old identity is gone; `startContainer` found the surviving container ACTIVE and reused it, so the container kept the old (now invalid → 401) token file mounted while the new instance's freshly written token dir stayed orphaned in `/dev/shm/kura-tokens`.

**Fix**: when identity integration is enabled, a surviving container is deleted and recreated on startup so it picks up the freshly minted temporary identity and its matching read-only token bind. Containers **without** identity integration keep surviving framework restarts unchanged.

On-device re-test:

| | Before fix | After fix |
|---|---|---|
| Container after framework restart | reused as-is (id unchanged) | recreated (new id), log: `Recreating container pr6294-test to apply refreshed identity credentials` |
| In-container `curl` with token | **401** (stale credential) | **200** |
| Token dirs in `/dev/shm/kura-tokens` | 2 (accumulating orphans) | 1, `-r--------`, matches the container's `:ro` bind |

New unit test: `recreatesSurvivingContainerToRefreshCredentialsOnStartup` (verifies stop+delete of the surviving container and recreation with a fresh token). All 103 provider tests + 47 orchestration tests pass.

#### F1 — `deactivate()` cleanup does not run on `systemctl stop kura` — verified, kept as documented residual

On-device verification (it is **not** a hard kill):

- The `kura.service` unit is `Type=forking`, `KillMode=process`, `KillSignal=15` (SIGTERM), `TimeoutStopSec=90s`, with **no graceful `ExecStop`**.
- On `systemctl stop kura`, the JVM receives SIGTERM and exits **cleanly with code 143** (`CLD_EXITED`, i.e. it is *not* SIGKILLed), in well under a second — but **without an orderly OSGi shutdown**: zero log lines are emitted during shutdown and no `deactivate()` runs. The DS `deactivate` method is correctly wired (`deactivate="deactivate"` in `OSGI-INF/containerinstance.xml`), so this is a **Kura platform shutdown behavior, not specific to this PR**: DS component deactivation is not guaranteed on `systemctl stop`.
- Consequence: the token directory of the running container survives the stop and is not reclaimed by the next start (new JVM has no reference to it) → one orphaned directory accumulates per ungraceful stop/start cycle.

**Impact is low and bounded**: the orphaned token lives in tmpfs (RAM, mode `400`, owner `kurad`), backs an in-memory temporary identity that **no longer exists** after the stop (so the leaked credential is unusable), and `/dev/shm` is cleared on reboot. The F2 fix was confirmed to still work across an ungraceful stop (container recreated on next start with a fresh token, HTTP 200).

**Decision**: not fixing this in this PR — the root cause is platform-level and the residual is harmless. A startup sweep of orphaned token directories is worth a **separate hardening ticket** rather than expanding this change.

### Configuration notes for reviewers/users

- `container.permissions` must reference permissions that exist on the device (e.g. `rest.system`); a non-existing name makes identity creation fail (`Invalid parameter. Permission does not exist`) and the container starts without credentials (pre-existing behavior).
- The REST basic authentication (`auth.basic.enabled` on `RestService`) must be enabled for containers to use the injected credentials; it is disabled by default on a fresh install.

kura/org.eclipse.kura.container.provider/src/main/java/org/eclipse/kura/container/provider/TokenFileManager.java
```diff
@@ -0,0 +1,126 @@
...
+
+    synchronized Optional<Path> getTokenFilePath() {
+        return Optional.ofNullable(this.tokenDirectory).map(dir -> dir.resolve(TOKEN_FILE_NAME));
+    }
+
+    private static void setPosixPermissions(final Path path, final String permissions) {
+        try {
+            Files.setPosixFilePermissions(path,
+                    java.nio.file.attribute.PosixFilePermissions.fromString(permissions));
+        } catch (final UnsupportedOperationException e) {
```
marcellorinaldo, line 115: I don't think we should swallow the exceptions here. What do you think? Is it better to abort the operation when `setPosixPermissions` fails?

kura/org.eclipse.kura.container.provider/src/main/java/org/eclipse/kura/container/provider/TokenFileManager.java
```diff
@@ -0,0 +1,126 @@
...
+                            + " set the \"" + TMPFS_BASE_PROPERTY
+                            + "\" system property to an existing tmpfs directory");
+        }
+
+        final Path directory = basePath.resolve(TOKENS_DIRECTORY_NAME).resolve(UUID.randomUUID().toString());
+        final Path tokenFile = directory.resolve(TOKEN_FILE_NAME);
+
+        try {
+            Files.createDirectories(directory);
+            setPosixPermissions(directory, "rwx------");
```
marcellorinaldo, line 70: `<uuid>` directory is tightened to 700; but the intermediate `kura-tokens` directory is left at umask? Is this ok?

kura/org.eclipse.kura.container.provider/src/main/java/org/eclipse/kura/container/provider/ContainerInstance.java
```diff
@@ -73,6 +74,8 @@ public class ContainerInstance implements ConfigurableComponent, ContainerOrches
     private static final String CONTAINER_IDENTITY_PREFIX = "container_";
     private static final int MAX_IDENTITY_NAME_LENGTH = 255;
     private static final int MAX_IDENTITY_NAME_GENERATION_ATTEMPTS = 10;
+    private static final String CONTAINER_TOKEN_PATH = "/run/secrets/kura-token";
+    private static final String READ_ONLY_VOLUME_SUFFIX = ":ro";
```
marcellorinaldo, line 82: I don't really like the fact that we keep `READ_ONLY_VOLUME_SUFFIX` duplicated in two separate places... If you can think of a better solution it would be great, otherwise nevermind (maybe comment it).

kura/org.eclipse.kura.container.provider/src/main/java/org/eclipse/kura/container/provider/TokenFileManager.java
salvatore-coppola: No it is required by design we should warn and continue, however i'll split the cases, UnsupportedOperationException and IOException

kura/org.eclipse.kura.container.provider/src/main/java/org/eclipse/kura/container/provider/TokenFileManager.java
salvatore-coppola: ok it is non necessary but doable

kura/org.eclipse.kura.container.provider/src/main/java/org/eclipse/kura/container/provider/ContainerInstance.java
salvatore-coppola: i'll let it as it is with a comment.

kura/org.eclipse.kura.container.orchestration.provider/src/main/java/org/eclipse/kura/container/orchestration/provider/impl/ContainerOrchestrationServiceImpl.java
```diff
@@ -769,24 +775,35 @@ private HostConfig containerVolumeMangamentHandler(ContainerConfiguration contai
...
+
+    }
 
+    private static Optional<Bind> volumeBind(String hostPath, String containerPathSpec) {
+        if (hostPath.isEmpty() || containerPathSpec.isEmpty()) {
+            return Optional.empty();
         }
 
-        return hostConfiguration;
+        String containerPath = containerPathSpec;
```
pierantoniomerlino: Why this assignement?

kura/org.eclipse.kura.container.orchestration.provider/src/main/java/org/eclipse/kura/container/orchestration/provider/impl/ContainerOrchestrationServiceImpl.java
```diff
@@ -769,24 +775,35 @@ private HostConfig containerVolumeMangamentHandler(ContainerConfiguration contai
...
+        AccessMode accessMode = AccessMode.DEFAULT;
+        if (containerPath.endsWith(READ_ONLY_VOLUME_SUFFIX)) {
+            containerPath = containerPath.substring(0, containerPath.length() - READ_ONLY_VOLUME_SUFFIX.length());
+            accessMode = AccessMode.ro;
+        }
+
+        if (containerPath.isEmpty()) {
+            return Optional.empty();
+        }
```
pierantoniomerlino, line 805: Is this check already performed at line 791?

kura/org.eclipse.kura.container.orchestration.provider/src/main/java/org/eclipse/kura/container/orchestration/provider/impl/ContainerOrchestrationServiceImpl.java
```diff
@@ -769,24 +775,35 @@ private HostConfig containerVolumeMangamentHandler(ContainerConfiguration contai
 
         List<Bind> bindsToAdd = new LinkedList<>();
 
-        if (containerDescription.getContainerVolumes() != null
-                && !containerDescription.getContainerVolumes().isEmpty()) {
+        for (Map.Entry<String, String> element : containerDescription.getContainerVolumes().entrySet()) {
```
pierantoniomerlino: Can the `containerDescription.getContainerVolumes()` return `null`?

marcellorinaldo (commented):
I think we should fix the tests to work also on MacOS

kura/org.eclipse.kura.container.provider/src/main/java/org/eclipse/kura/container/provider/ContainerInstance.java
```diff
@@ -618,6 +644,43 @@ private void startMicroservice(final ContainerInstanceOptions options) {
             logger.warn("Unable to start microservice...giving up");
         }
 
+        private void recreateSurvivingContainerForCredentialRefresh(final ContainerInstanceOptions options) {
```
pierantoniomerlino: Does this method only stop and delete the container?

salvatore-coppola: yes i'll rename it

kura/org.eclipse.kura.container.orchestration.provider/src/main/java/org/eclipse/kura/container/orchestration/provider/impl/ContainerOrchestrationServiceImpl.java
salvatore-coppola: it's a different path, i'll add a comment

kura/org.eclipse.kura.container.orchestration.provider/src/main/java/org/eclipse/kura/container/orchestration/provider/impl/ContainerOrchestrationServiceImpl.java
salvatore-coppola: i'll remove it

pierantoniomerlino (approved):

marcellorinaldo (approved):

PR #6298, opened 2026-07-22
feat!: added JWT REST authentication and token service

marcellorinaldo:
This PR aims to add a new authentication mechanism for REST calls using JSON Web Tokens (see https://www.jwt.io/ for tutorials).

Using dependency https://github.com/auth0/java-jwt (MIT license):
```
<groupId>com.auth0</groupId>
<artifactId>java-jwt</artifactId>
<version>4.6.0</version>
```

Without having to rewrite the password/certificate authentication endpoints, we simply added a new authentication provider and a new endpoint to issue a new token. The workflow is as follows:

1. Login using `/session/v1` (basic auth has no sense in this context)
2. Obtain the JSESSIONID cookie and the XSRF token
3. Issue a new JWT token by calling:
    ```shell
    curl --request GET \
        --url https://<DEVICE-IP>/services/jwt/v1/token \
        --header 'jsessionid: <JSESSIONID>' \
        --header 'x-xsrf-token: <XSRF-TOKEN>'
    ```
4. Perform following requests with:
    ```shell
    curl --request GET \
        --url https://<DEVICE-IP>/services/system/v1/properties/framework \
        --header 'authorization: Bearer <JWT-TOKEN>'
    ```
5. A caller from within the framework can simply call the `IdentityTokenService.issueTokenFor`

**Open points**:

- [ ] Should the `ttl` for the tokens issued with `services/jwt/v1/token` be configurable from the Kura Web UI in the `RestService` section? Now it is set fixed to 15 minutes. Should we put also a feature toggle in the same section?
- [ ] Should we restrict the token authorization provider to just SSL connections?
- [ ] Should we keep track of the issued tokens so that on reissuing for the same identity the old one expires?
- [ ] Should we separate this PR in two?
- [ ] The `JwtDTO` includes the expiration date. The same information is obtainable by base64 decoding the token. Should we keep this information duplicated in the returned DTO for convenience? The resolution reported in the token is seconds, the one in the DTO is in milliseconds.
- [ ] Issuer in the `IdentityJwtTokenServiceImpl` is set to `kura_<device-mac-address-without-:>`, is this OK?
- [ ] The JWT authorization is not really stateless, we perform a lookup in the identities store to check whether the identity exists. In that way, we can revoke a token by removing the identity. Should we instead be able to revoke the single token? Is it necessary if we shorthen the time-to-live of the token?


marcellorinaldo:
IP review request for the new dependency: https://gitlab.eclipse.org/eclipsefdn/emo-team/iplab/-/work_items/29896

PR #6308, opened 2026-08-07
fix: modem manager ip birth message

MMaiero:
connectionInterface/connectionIp were built solely from the legacy
NetworkService, which has no visibility on ModemManager-managed modem
interfaces (e.g. QMI/MBIM wwan*). Devices migrating from pppd-based
modems (ppp0 always appeared) lost the modem IP in the birth message.
Now also consult NetworkStatusService, when available, for connected
modem interfaces with a reported IP, deduplicated against interfaces
already reported by the legacy stack.


MMaiero:
The issue reported by Sonar and related to code duplication cannot be fixed. The PR will need to be force-merged if the contribution is ok code-wise

pierantoniomerlino (approved):

PR #6313, opened 2026-08-13
refactor!: remove the stress bundle from the framework

salvatore-coppola:
`org.eclipse.kura.stress` was a development-only heap stress tool that is not part of any distribution: it was meant to be installed by hand through the `stress.dp` deployment package committed in 2016, which no longer matches the sources and can no longer be rebuilt (its `stress_build.xml` is gone from the repository). It was still compiled on every build and consumed by nobody.

This removes the bundle, its test module and the references in `kura/pom.xml`, `kura/test/pom.xml` and `kura/kura-pde-deps/pom.xml`. No other module referenced it.

**Breaking change:** the `org.eclipse.kura.stress` artifact is no longer published and no longer part of `kura-pde-deps`, so it disappears from the target platforms generated for the add-on repositories.


salvatore-coppola:
@MMaiero can i remove this bundle ? It's totally useless

MMaiero (approved):

PR #6317, opened 2026-08-13
refactor!: move the CloudCat bundle to the kura-apps repository

salvatore-coppola:
CloudCat now lives in https://github.com/eclipse-kura/kura-apps (eclipse-kura/kura-apps#42, which has to merge first), next to the other example and add-on bundles, and is delivered as the `kura-cloudcat-example` Debian package instead of being built by the framework reactor.

Removes `org.eclipse.kura.misc.cloudcat` together with its module and `dependencyManagement` entries in `kura/pom.xml`, its `kura-pde-deps` dependency and the `misc-plugins` entry of the `core-plugins-4s` execution in distrib. The bundle exported no package, so nothing else in the framework depended on it.

**BREAKING CHANGE:** the `kura-core` package no longer ships `org.eclipse.kura.misc.cloudcat`. Install the `kura-cloudcat-example` package from kura-apps instead. The configuration PID is unchanged, so existing snapshots keep working once the package is installed.

**Manual Tests**: `mvn validate` on both the `kura` and the `kura/distrib` reactors. Not tested on hardware.


sfiorani (approved):

sfiorani:
Tested on latest kura develop aarch 64 at the moment while writing this comment

PR #6323, opened 2026-08-17
refactor!: move the S7 PLC Driver to the kura-s7 repository

salvatore-coppola:
The S7 PLC Communication Driver now lives in https://github.com/eclipse-kura/kura-s7 (eclipse-kura/kura-s7#1, which has to merge first) and is delivered as the `kura-s7` Debian package instead of being built by the framework reactor and shipped inside the core deployment package.

Removes `org.eclipse.kura.driver.s7plc.provider` and its test bundle together with their module and `dependencyManagement` entries in `kura/pom.xml` and `kura/test/pom.xml`, the `kura-pde-deps` dependency and the `core-dp` entry. The bundle exported no package, so nothing else in the framework depended on it.

`org.eclipse.kura.driver.block` stays here: it is published to Maven and kura-s7 consumes it from there, embedding it in the Driver bundle. The `org.moka7` target platform bundle stays as well, for the same reason.

Also drops the `org.eclipse.kura.driver.gpio` entry left behind in `core-dp` when the GPIO resources were removed from the codebase.

**BREAKING CHANGE:** the `kura-core` package no longer ships the S7 PLC Driver. Install the `kura-s7` package instead. The configuration PID and the channel descriptor are unchanged, so existing snapshots keep working once the package is installed.

**Manual Tests**: `mvn validate` on both the `kura` and the `kura/distrib` reactors. Not tested on hardware.


kura/distrib/core-dp/pom.xml
```diff
@@ -47,10 +47,8 @@
                         <configuration>
                             <includeArtifactIds>
                                 org.eclipse.kura.protocol.modbus,
-                                org.eclipse.kura.driver.s7plc,
                                 org.eclipse.kura.driver.eddystone,
                                 org.eclipse.kura.driver.ibeacon,
-                                org.eclipse.kura.driver.gpio,
```
mattdibi, line 53: Did we forget this from another PR?

mattdibi (approved):

kura/distrib/core-dp/pom.xml
salvatore-coppola: yep, i've added here. do you prefer a separate PR ?

PR #6326, opened 2026-08-17
refactor!: move the Filesystem log provider to its own repository

salvatore-coppola:
The `FilesystemLogProvider` now lives in https://github.com/eclipse-kura/kura-log-filesystem (eclipse-kura/kura-log-filesystem#1, which has to merge first) and is delivered as the `kura-log-filesystem` Debian package instead of being built by the framework reactor and installed by `kura-core` in `plugins/4s`.

Removes `org.eclipse.kura.log.filesystem.provider` and its test bundle together with their module and `dependencyManagement` entries in `kura/pom.xml` and `kura/test/pom.xml`, the `kura-pde-deps` dependency and the `core-plugins-4s` entry in distrib. The bundle exported no package, so nothing else in the framework depended on it.

Also drops the two provider instances the default snapshot used to pre-configure, `filesystem-kura-log` and `filesystem-kura-audit-log`: shipping configuration for a factory that is not installed would leave dangling entries in `snapshot_0.xml`. `kura.default.log.manager` keeps pointing at `filesystem-kura-log`, which is the pid to give the instance when recreating it.

`AGENTS.md` used this test bundle as the example of a single-module test run; it now points at `org.eclipse.kura.core.configuration.test`.

**BREAKING CHANGE:** the `kura-core` package no longer ships a `LogProvider`, and the default snapshot no longer configures one, so a fresh installation has no log provider until `kura-log-filesystem` is installed and an instance is created. Creating it with `kura.service.pid` `filesystem-kura-log` restores the previous behaviour, since that is what `kura.default.log.manager` still refers to. The addon README lists both instances.

**Manual Tests**: `mvn validate` on both the `kura` and the `kura/distrib` reactors, and a well-formedness check of the edited `snapshot_0.xml`. Not tested on hardware, so the snapshot has not been loaded by a running framework.


kura/distrib/src/main/resources/filtered/pkg/framework/kura.properties
```diff
@@ -57,6 +57,9 @@ os.distribution.version=${os.distribution.version}
 kura.command.user=kura
 kura.legacy.bluetooth.beacon.scan=false
 kura.legacy.ppp.logging.enabled=true
+# pid of the LogProvider instance the log consumers use by default. The
+# instance itself is not created by this package any more: it comes with the
+# kura-log-filesystem add-on.
```
mattdibi: What is this comment referring to?

salvatore-coppola: to the value of `kura.default.log.manager`

kura/distrib/src/main/resources/filtered/pkg/framework/kura.properties
```diff
@@ -57,6 +57,9 @@ os.distribution.version=${os.distribution.version}
...
 kura.default.log.manager=filesystem-kura-log
```
mattdibi: I would say that it's quite hard to read as-is. I propose the following:


```suggestion
kura.default.log.manager=filesystem-kura-log # pid of the LogProvider instance the log consumers use by default.
# NOTE: The kura.default.log.manager instance itself is now provided by the
# kura-log-filesystem add-on.
```

What do you think?

mattdibi (approved):

PR #6328, opened 2026-08-18
feat(linux.clock): expose NTP sync health via ClockService

Abe-AsYouWishEngineering:
Extends `ClockService` so consumers (first: the web UI Status page) can read
NTP sync **health** — synced / not synced / unknown, last sync, and the
active sync provider — instead of only `getLastSync()`.

Note: the base of this PR must be `fix/ntp-retry-interval-key` branch. I do not have rights to modify the PR to retarget the PR. 

### What

- New API in `org.eclipse.kura.clock` (org.eclipse.kura.api): tri-state
  `ClockSyncState` (`SYNCED` / `NOT_SYNCED` / `UNKNOWN`), immutable
  `ClockSyncStatus` DTO, and one **default** method
  `ClockService.getSyncStatus()` deriving a weak answer from
  `getLastSync()` (null → `NOT_SYNCED`).
- `org.eclipse.kura.linux.clock` overrides it with honest per-provider
  health:
  - **java-ntp**: state of the most recent *completed* sync attempt
    (retry exhaustion reports `NOT_SYNCED`; the internal refresh-window
    scheduling flip does not flap the reported state).
  - **chrony-advanced**: on-demand probe — chronyd active + `chronyc -c
    tracking` leap status `Normal` → `SYNCED`; `Not synchronised` →
    `NOT_SYNCED`; probe unavailable/unparseable → `UNKNOWN`, never an
    exception. The machine-readable CSV form (`-c`) is used deliberately:
    the human format varies across chrony versions/platforms (debian,
    rocky, docker, RPi, yocto).
- Tests: 17 new Gherkin unit scenarios plus the module's first in-OSGi
  integration tests (tycho-surefire enabled; DS lifecycle +
  `getSyncStatus()` end to end over mocked executor services).

### Reviewer callouts

- **Binary compatibility**: the addition is a default method, so every
  existing provider stays binary- and source-compatible — the UNMODIFIED
  emulator implementation (`org.eclipse.kura.emulator.clock`, zero Java
  changes) is the in-tree proof. Package export bumps 1.0.1 → 1.1.0 per
  OSGi semantic versioning (minor, new API); the two in-tree providers'
  import ranges move `[1.0,1.1)` → `[1.1,1.2)` (provider range), same
  pattern as the identity bump in #5965.
- **No new architectural surface**: the chronyc-tracking probing lives
  inside the chrony provider, consistent with its existing chronyc /
  systemctl / journalctl usage through `CommandExecutorService`.
- **Zero behavior change to sync/retry logic** — the API observes state,
  it never steers it. `getLastSync()` semantics are untouched (the
  tracking ref-time alternative was considered and rejected for exactly
  that reason). Slew-only healthy chrony systems will now report SYNCED
  while Last Sync stays "never" — intentional: health and last-step
  answer different questions.

### What

- **`FAILED` state** (`ClockSyncState`): entered ONLY on a positively
  detected failure condition, never inferred from time spent
  `NOT_SYNCED` (the javadoc states this guardrail). `NOT_SYNCED`
  remains the transient state.
- **`failureReason` + `retryCount`** (`ClockSyncStatus`, constructor
  grows to 5 params):
  - `getFailureReason()`: non-null only when `FAILED`; documents the
    detected failure.
  - `getRetryCount()`: consecutive failed sync attempts in the
    provider's current cycle; `null` where the provider doesn't track
    retries (chrony). The default `getSyncStatus()` passes null for
    both.
- **java-ntp** (`AbstractNtpClockSyncProvider`):
  - Give-up path (`clock.ntp.max-retry` > 0, retries exhausted) →
    `FAILED`, reason "Gave up after N failed synchronization
    attempts", appending the last exception message when the give-up
    attempt itself threw (see callouts).
  - Terminal one-shot (`clock.ntp.refresh-interval` = 0, the single
    attempt returns false, nothing re-armed) → `FAILED`, reason
    "Single synchronization attempt failed and no retry is
    scheduled". The throwing arm of the one-shot keeps rescheduling
    forever, so it honestly stays `NOT_SYNCED`; refresh-interval < 0
    (updates disabled) stays `UNKNOWN` — disabled is not failed.
  - The shipped default `clock.ntp.max-retry=0` retries forever, so
    **retry exhaustion can never carry hard failure** there. While retry is active,
    the surface is `NOT_SYNCED` plus a climbing `retryCount` (the live
    retry counter, read at call time).
  - `FAILED` clears when the refresh window re-arms attempts to match the existing cycle semantics.
- **chrony** (`ChronyClockSyncProvider`): the `systemctl is-active`
  stdout feeds the state: `failed` → `FAILED`, with the reason probed from
  `systemctl show <unit> -p Result` ("chrony service failed (Result:
  exit-code)"), degrading to a static "chrony service is in a failed
  state" on any probe trouble. `--value` is avoided for older-systemd
  compatibility; empty/unreadable `is-active` output degrades to
  `UNKNOWN`. Other non-active unit states still report
  `NOT_SYNCED`.
- Tests: the module unit suite grows to 53 Gherkin scenarios (all
  green; 1 pre-existing platform-gated skip), covering every FAILED
  trigger, the degradation paths, retry-count visibility, and the
  parse edge cases.

### Reviewer callouts

- **`FAILED` is positively detected**. Whether it
  self-recovers is provider- and configuration-dependent: NTP retry
  exhaustion self-clears at the next refresh re-arm (~1h at
  defaults), the NTP terminal one-shot does not recover without a
  reconfigure, and chrony re-reads the unit state on every call. It
  is not a permanent latch.
- **Exception-message attribution is read narrowly**: the last
  exception message is appended to the give-up reason only when the
  give-up attempt itself threw. A cycle where an earlier attempt
  threw and the final attempt merely returned false carries no
  exception text. This prevents false or misleading cached error messages for being shown on a failure that didn't cause them.


kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/clock/ClockSyncState.java
```diff
@@ -0,0 +1,39 @@
...
+// Content with portions generated by generative AI platform
+
+package org.eclipse.kura.clock;
+
+/**
+ * Represents the synchronization state of the clock, as reported by a {@link ClockService}.
+ *
+ * @since 3.0
+ */
+public enum ClockSyncState {
```
mattdibi, line 22: Would it make sense to have a "FAILURE"/"ERROR" state?

"NOT_SYNCED" could be a transient state: for example the device just booted and still needs to connect to the internet and grab the time from an NTP source. It is expected that the status is "NOT_SYNCED".

What if there's an error in the `chrony.conf` file and the `chronyd` daemon just fails? The status will not change unless the user does something. Are we able to retrieve this type of information?

Does it make sense to model this state?

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/clock/ClockSyncStatus.java
```diff
@@ -0,0 +1,80 @@
...
+import org.osgi.annotation.versioning.ProviderType;
+
+/**
+ * Represents the synchronization status of the clock, as reported by a {@link ClockService}.
+ *
+ * @since 3.0
+ * @noimplement This class is not intended to be extended by clients.
+ */
+@ProviderType
+public class ClockSyncStatus {
```
mattdibi, line 28: Following up on https://github.com/eclipse-kura/kura/pull/6328/changes#r3819335767, if we are able to retrive the "ERROR"/"FAILURE" status we might be able to also retrieve information about the _reason_ of the failure and therefore we could have a `String` representing that

kura/org.eclipse.kura.linux.clock/src/main/java/org/eclipse/kura/linux/clock/ChronyClockSyncProvider.java
```diff
@@ -239,6 +241,51 @@ public Date getLastSync() {
...
+
+            ByteArrayOutputStream chronycTrackingOutput = (ByteArrayOutputStream) chronycTrackingStatus
+                    .getOutputStream();
+
+            if (chronycTrackingOutput.size() == 0) {
+                return ClockSyncState.UNKNOWN;
+            }
+
+            String output = new String(chronycTrackingOutput.toByteArray(), StandardCharsets.UTF_8);
+            String firstLine = output.split("\\r?\\n", 2)[0];
```
mattdibi: Sonar is whining about this.

Fix this access on a collection that may throw an 'ArrayIndexOutOfBoundsException' when executed.

See: https://sonarcloud.io/project/issues?sinceLeakPeriod=true&issueStatuses=OPEN%2CCONFIRMED&pullRequest=6328&id=org.eclipse.kura%3Akura&open=AaAeHmK9bX8-Axx_o-e7

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/clock/ClockSyncState.java
Abe-AsYouWishEngineering: Agreed. Implemented as `ClockSyncState.FAILED`, with one guardrail
worth calling out: `FAILED` is entered only on a positively detected
failure condition. `FAILED` is never inferred from time spent `NOT_SYNCED`.

One wrinkle the default config forces: `clock.ntp.max-retry` defaults
to 0 ("retry forever"), so retry exhaustion alone can't carry hard
failure on default config. `ClockSyncStatus` also exposes
`getRetryCount()`, so a climbing count is visible on the status page for java-ntp to help the user interpret the `NOT_SYNCED` status.

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/clock/ClockSyncStatus.java
Abe-AsYouWishEngineering: Agreed. Added as a scoped `getFailureReason()` on `ClockSyncStatus`. NTP composes it from the give-up path ("Gave up after N failed synchronization attempts", with the exception message, if any, from the final attempt). Chrony probes `systemctl show <unit> -p Result` and reports e.g. "chrony service
failed (Result: exit-code)", degrading to a static reason string if that probe has any trouble.

kura/org.eclipse.kura.linux.clock/src/main/java/org/eclipse/kura/linux/clock/ChronyClockSyncProvider.java
Abe-AsYouWishEngineering: This should be resolved now.

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/clock/ClockSyncStatus.java
mattdibi, line 1: There's a lot of nullable values in this class, I think using `Optional` is preferrable here. `Optional` makes it explicit when a value might be missing, promoting better readability and safer code.

kura/org.eclipse.kura.linux.clock/src/main/java/org/eclipse/kura/linux/clock/AbstractNtpClockSyncProvider.java
```diff
@@ -143,10 +162,37 @@ public Date getLastSync() {
...
+    }
+
     // ----------------------------------------------------------------
     //
     // Private/Protected Methods
     //
     // ----------------------------------------------------------------
+    private void giveUp(KuraException cause) {
+        logger.error("Failed to synchronize System Clock. Exhausted retry attempts, giving up");
+        this.isSynced = true;
```
mattdibi, line 187: I think this is a little footgun waiting to go off.

We _failed_ to sync the clock and then set `isSynced` to `true`. Wouldn't it be better to rename `isSynced` to something like `done`/`isDone`?

kura/org.eclipse.kura.linux.clock/src/main/java/org/eclipse/kura/linux/clock/ClockServiceConfig.java
mattdibi, line 1: This file was already fixed in #6333 and generating a conflict

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/clock/ClockService.java
```diff
@@ -39,4 +41,24 @@ public interface ClockService {
...
+     * @return the current {@link ClockSyncStatus}.
+     * @throws KuraException
+     *             if the clock synchronization status cannot be retrieved.
+     * @since 3.0
+     */
+    default ClockSyncStatus getSyncStatus() throws KuraException {
+        Date lastSync = getLastSync();
+        return new ClockSyncStatus(lastSync != null ? ClockSyncState.SYNCED : ClockSyncState.NOT_SYNCED, lastSync,
+                null, null, null);
+    }
```
mattdibi, line 62: We usually don't provide default implementations for our APIs but leave everything to the implementation. I would remove this:


```suggestion
   ClockSyncStatus getSyncStatus() throws KuraException;
```

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/clock/ClockSyncStatus.java
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+    /**
+     * Returns the name of the synchronization provider that reported this status.
+     *
+     * @return the name of the synchronization provider, can be <code>null</code> if not known.
+     */
+    public String getSyncProvider() {
+        return this.syncProvider;
+    }
```
mattdibi, line 89: Is this really needed? One can simply check the ClockService configuration to see what's the provider currently set 🤔

I would also argue that this information is not part of the Clock _Synchronization_ Status but something else entirely.

kura/org.eclipse.kura.linux.clock/src/main/java/org/eclipse/kura/linux/clock/ChronyClockSyncProvider.java
```diff
@@ -239,6 +251,132 @@ public Date getLastSync() {
...
+            String firstLine = output.lines().findFirst().orElse("");
+            String leapStatus = firstLine.substring(firstLine.lastIndexOf(',') + 1).trim();
+
+            if ("Normal".equals(leapStatus)) {
+                return ClockSyncState.SYNCED;
+            } else if ("Not synchronised".equals(leapStatus)) {
+                return ClockSyncState.NOT_SYNCED;
+            } else {
+                return ClockSyncState.UNKNOWN;
+            }
```
mattdibi, line 304: This is not correct.

According to the [`chronyc` manual](https://chrony-project.org/doc/4.8/chronyc.html) the possible values are _Normal_, _Insert second_, _Delete second_ or _Not synchronised_.

According to [RFC 5905](https://www.rfc-editor.org/info/rfc5905/) (which correspond to the above-mentioned values)

```
   LI Leap Indicator (leap): 2-bit integer warning of an impending leap
   second to be inserted or deleted in the last minute of the current
   month with values defined in Figure 9.

           +-------+----------------------------------------+
           | Value | Meaning                                |
           +-------+----------------------------------------+
           | 0     | no warning                             |
           | 1     | last minute of the day has 61 seconds  |
           | 2     | last minute of the day has 59 seconds  |
           | 3     | unknown (clock unsynchronized)         |
           +-------+----------------------------------------+

                         Figure 9: Leap Indicator
```

This means that for Normal, Insert second, Delete second the `ClockSyncState` should be considered `SYNCED` since Chrony was able to retrieve the information about the leap second.

Only the case in which chrony reports the _Not synchronised_ values we can assume no synchronization took place.

mattdibi (changes requested):
I didn't have the time to take complete look at the PR but here's some notes for now

mattdibi:
@MMaiero would it make sense to expose this status via REST API? (In a separate PR/Task ofc)

kura/org.eclipse.kura.linux.clock/src/main/java/org/eclipse/kura/linux/clock/ChronyClockSyncProvider.java
mattdibi, line 304: Per https://github.com/eclipse-kura/kura/pull/6328/changes/BASE..863b32dfb1423b32246e05b96603fa1383fdec19#r3845198781 this should be:


```suggestion
            if ("Not synchronised".equals(leapStatus)) {
                return ClockSyncState.NOT_SYNCED;
            } else {
                return ClockSyncState.SYNCED;
            }
```

MMaiero:
> @MMaiero would it make sense to expose this status via REST API? (In a separate PR/Task ofc)

Yes, potentially even with a request handler for the cloud

kura/test/org.eclipse.kura.linux.clock.test/src/test/java/org/eclipse/kura/linux/clock/ClockServiceConfigRetryIntervalTest.java
mattdibi, line 1: I would rename this `ClockServiceConfigTest.java`. In the future it could be expanded to cover more stuff for the `ClockServiceConfig`, no reason to restrict its purpose.

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/clock/ClockSyncState.java
```diff
@@ -0,0 +1,49 @@
...
+
+    /**
+     * The clock synchronization failed. This state is reported only when a failure condition is positively
+     * detected, for example the time daemon unit is reported by systemd in a <code>failed</code> state, or a
+     * finite number of synchronization retries was configured and those retries have been exhausted. Where the
+     * synchronization provider is configured to retry indefinitely, retries alone never lead to this state. It
+     * is never inferred from how long the clock has been in the {@link #NOT_SYNCED} state: a device that has
+     * just booted and is still reaching its time source is honestly {@link #NOT_SYNCED}, not {@link #FAILED}.
+     */
+    FAILED,
```
mattdibi, line 42: Does the `FAILED` state implicitly mean that the clock is `NOT_SYNCED`? If so we should document that.

PR #6333, opened 2026-08-24
fix: read the NTP retry interval from its own configuration key

Abe-AsYouWishEngineering:
PROPERTY_NTP_RETRY_INTERVAL was declared on "clock.ntp.max-retry", a duplicate of PROPERTY_NTP_MAX_RETRIES, so a configured NTP retry interval was silently ignored and the max-retry value drove it instead. The metatype declares "clock.ntp.retry.interval" (min 1, default 5).

This changes effective behavior: with the shipped defaults the retry interval was resolving to the max-retry default of 0, which start() maps to a 1 second tick, so retries ran every second instead of every five. Any install that had set clock.ntp.retry.interval now gets the value it configured.


mattdibi (commented):
LGTM

Two things:
1. Requires rebase
2. Copyright header check is failing due to some whitespace issues. Double check

Abe-AsYouWishEngineering:
@mattdibi I rebased and I think I have the whitespace issues fixed now.

mattdibi (approved):

PR #6337, opened 2026-08-25
test: stabilize the flaky REST and clock tests

salvatore-coppola:
The REST and clock integration tests are flaky on this branch. Four CI runs on this PR failed for four different reasons, none of them the same, so this PR now collects the fixes for every cause that the logs actually showed.

## 1. `fix(rest.provider)`: audit context built from the wrong instance

`IncomingPortCheckFilter` runs at `Priorities.AUTHENTICATION - 100`, so before authentication and before any parameter validation. It called `initAuditContext`, which dereferenced the `HttpServletRequest` field of the *enclosing* `RestService` instance instead of the one injected into the filter itself.

`RestService` and `IncomingPortCheckFilter` are registered as two independent singletons, so the filter can run while the `@Context` injection of the enclosing instance has not happened yet. The resulting NPE was mapped to a generic 500 by `RestExceptionMapper`, so every request failed — including the ones expected to return 400 or 401:

```
ERROR o.e.k.i.r.p.RestExceptionMapper - Uncaught Exception while processing REST request
java.lang.NullPointerException
    at org.eclipse.kura.internal.rest.provider.RestService.initAuditContext(RestService.java:310)
    at org.eclipse.kura.internal.rest.provider.RestService$IncomingPortCheckFilter.filter(RestService.java:349)
```

`initAuditContext` now takes the servlet request from its caller, so each caller passes the instance injected into it, and the request IP is omitted from the audit context rather than throwing when no servlet request is available.

`develop` is not affected: the filters there are already top-level classes that pass their own injected request to `RestServiceUtils.initAuditContext`. This only realigns the 5.6.x code with that behaviour.

## 2. `test(core.testutil)`: transport did not wait for Jersey

`RestTransport.init` only waited for the TCP port to accept connections and then slept for a fixed second; it never waited for the JAX-RS publisher to finish building the Jersey container. On a loaded CI agent that is not enough, so the first requests of a test class were answered with `503` and a `{"message":"Jersey is not ready yet!"}` body. The transport now polls until the response is no longer that readiness error.

The probe stops at the first exception rather than swallowing it, times out after thirty seconds with a warning instead of throwing, and carries an explicit connect and read timeout — it checked its deadline only between attempts, so a request that never got a response would have blocked indefinitely. Ordinary transport requests keep their previous unbounded behaviour.

## 3. `test(linux.clock)`: clock sync providers leaked between scenarios

`ClockServiceImplTest` activated the service under test in seven scenarios but deactivated it in only three, so each of the others left a live clock sync provider running for the rest of the JVM, holding that scenario's `EventAdmin` mock as its listener.

`testClockUpdateBasic`, `testClockUpdateErrors` and `testClockUpdate` activated the service only to populate its configuration, then asserted an exact `postEvent` count. A sync from the provider that `activate` starts calls `onClockUpdate` on the same instance, and that method posts the event unconditionally when the offset is zero, so the assertion could see an extra event. This is what failed run #3 of this PR:

```
NeverWantedButInvoked: eventAdmin.postEvent(isA(org.eclipse.kura.clock.ClockEvent));
Never wanted here:  -> ClockServiceImplTest.testClockUpdateErrors(...:265)
But invoked here:   -> ClockServiceImpl.onClockUpdate(ClockServiceImpl.java:238)
```

Those three scenarios now inject the configuration directly, and a teardown deactivates whatever the other scenarios activated.

`JavaNtpClockSyncProviderTest` created a scheduler per scenario, started the provider and stopped neither, leaving three tasks ticking every second for the rest of the JVM, one of them throwing on every tick. It now keeps both and stops them in a teardown, since `AbstractNtpClockSyncProvider.stop` only cancels the future and never shuts the injected scheduler down.


salvatore-coppola:
The first CI run failed on a **different** test than the one this PR started from: `org.eclipse.kura.rest.system.provider.test.EndpointsTest`, 11 tests, all answered with `503` and a `{"message":"Jersey is not ready yet!"}` body rather than an assertion mismatch.

That is a second, independent cause in the same family, and it lives in the shared test transport. `RestTransport.init` only waited for the TCP port to accept connections (`waitPortOpen`) and then slept for a fixed second; it never waited for the JAX-RS publisher to finish building the Jersey container. On a loaded CI agent one second is not enough, so the first requests of a test class hit the readiness error.

Added a second commit that makes the transport poll until the response is no longer that readiness error, with a two minute timeout. Since every REST integration test goes through this transport, it should remove this whole class of flakes.

Verified locally (JDK 8, as CI), all with a clean build: `EndpointsTest` 24/24, `DependenciesTest` 5/5, `RestServiceTest` 60/60, `ServiceListingEndpointsTest` 56/56, `TamperDetectionRemoteServiceTest` 16/16.

mattdibi (approved):

PR #6343, opened 2026-08-26
fix(core.configuration): keep the Password type of properties written to a snapshot before their metatype is tracked

salvatore-coppola:
A `Password` property can end up in a snapshot as an unencrypted `String` holding the CryptoService ciphertext, and the next boot then loads it as a plain string that is never decrypted.

ConfigurationAdmin can only hold plain strings, so the `Password` type is rebuilt from the metatype on read back. `ConfigurationServiceImpl.activate()` pushes the latest snapshot into ConfigurationAdmin *before* opening the metatype bundle tracker, while the `ConfigurableComponent` service reference registers a pid in `allActivatedPids` independently of that tracking. Any snapshot written in that window — at first boot the keystore instances with `randomize.password=true` force one — loses the type.

- Record the properties holding a `Password` when they are handed over to ConfigurationAdmin, and use them to restore the type of the properties the metatype does not describe yet; an attribute definition, when available, still wins over the tracked names.
- Replace the tracked names on every write instead of accumulating them, so they always describe what ConfigurationAdmin currently holds and a rollback adopts the types of the snapshot it restores.
- Track and read them back on the update path too, so an update neither loses the type of a password it leaves untouched nor stores a new one as a plain string.
- Extend `ConfigurationServiceTest` with scenarios covering the password type across update, snapshot and rollback, including a property the metatype does not declare, which is the only way to exercise the fallback end to end. Its two ignored scenarios are retargeted at components this module already provides, so they run again.


kura/org.eclipse.kura.core.configuration/src/main/java/org/eclipse/kura/core/configuration/ConfigurationServiceImpl.java
```diff
@@ -1618,6 +1630,18 @@ private Tocd getOCDForPid(String pid) {
         return ocd;
     }
 
+    private synchronized void trackPasswordProperties(String pid, Map<String, Object> properties) {
+        Set<String> names = CollectionsUtil.passwordPropertyNames(properties);
+        if (!names.isEmpty()) {
+            this.passwordPropertiesByPid.computeIfAbsent(pid, key -> new HashSet<>()).addAll(names);
```
nicolatimeus: Is addAll() correct for the case of the rollback?

nicolatimeus:
do we need to store the password property type also in the case of configuration update?

kura/org.eclipse.kura.core.configuration/src/main/java/org/eclipse/kura/core/configuration/ConfigurationServiceImpl.java
salvatore-coppola: no we need to remove the pid from the map if names is empty

nicolatimeus: yes you should replace the tracked property names with the ones from the rolled back snapshot

salvatore-coppola:
@nicolatimeus made several change to cover rollback and update

nicolatimeus (approved):

PR #6344, opened 2026-08-26
fix(distrib): replace org.eclipse.kura.emulator.net with org.eclipse.kura.linux.net in nn profiles

MMaiero:
## Summary

- Replaces `org.eclipse.kura.emulator.net` with `org.eclipse.kura.linux.net` as the low-level `NetworkService` implementation shipped in the "-nn" ("no network[-manager]") distrib build profiles (`generic-arm32-nn`, `generic-x86_64-nn`, `generic-aarch64-nn`, `intel-up2-ubuntu-20-nn`, `docker-x86_64-nn`, `docker-ubi8-x86_64-nn`, `docker-alpine-x86_64-nn`, `nvidia-jetson-nano-nn`).
- `emulator.net` relies on `java.net` APIs (e.g. `NetworkInterface.getNetworkInterfaces()`) that can fail to report interfaces without an established link, which can prevent Kura from resolving the primary interface's MAC address. `linux.net` is the same, already-tested implementation used by every non-`-nn` profile.
- Scope is limited to this one swap: `org.eclipse.kura.emulator` and its GPIO/USB/clock/watchdog/position siblings, the IDE's `Kura_Emulator.launch` dev configs, and the `net.admin`/`nm` wiring are all untouched. A single edit to `kura/distrib/src/main/ant/build_equinox_distrib.xml` covers every "-nn" profile since they share the same `is.nn` Ant condition.

## Verification done

- Diff-reviewed against the plan; static verification confirmed: identical SCR component name/PID between `emulator.net` and `linux.net` (so no snapshot/config migration is implicated), `linux.net`'s full `Import-Package` set resolves against the bundles already shipped unconditionally in every "-nn" build, `org.eclipse.kura.linux.net` is already staged as a build artifact for every profile, and the Ant `depends`-reachability mechanism that makes `linux-config`/`linux-jars` run for "-nn" builds without any new `antcall` was confirmed against the file's actual behavior (empirically, via the existing 6x-duplicated `linux.net` entry already present in a pre-change non-`-nn` build).

## Still needed before merge (not run in this environment)

- [ ] `mvn -PbuildAll,generic-x86_64-nn package` and `mvn -PbuildAll,generic-x86_64 package` (or your CI equivalent), then confirm `org.eclipse.kura.linux.net` is present and `org.eclipse.kura.emulator.net` is absent from the `-nn` profile's `config.ini` and output zip, and that the non-`-nn` profile's output is unchanged.
- [ ] Boot-test at least one container "-nn" profile (`docker-alpine-x86_64-nn`, `docker-ubi8-x86_64-nn`) — `linux.net` shells out to native tools (`ip`, `ifconfig`, `iw`, `ethtool`) that `emulator.net` never needed, and minimal container base images may lack them.

## Test plan

- [ ] CI build passes for all affected `-nn` and non-`-nn` profiles
- [ ] `-nn` profile `config.ini`/zip contain `linux.net`, not `emulator.net`
- [ ] Non-`-nn` profile output unchanged
- [ ] At least one container `-nn` profile boots and its `NetworkService` activates without errors


MMaiero:
Verified as working on a raspberry pi using the aarch64_nn profile

image

nicolatimeus (dismissed):

kura/distrib/src/main/ant/build_equinox_distrib.xml
```diff
@@ -1818,12 +1819,10 @@ fi]]>
             file="${project.build.directory}/${build.output.name}/config.ini">
             <entry key="osgi.bundles" operation="+"
                 value=", reference:file:${kura.install.dir}/${kura.symlink}/${plugins.folder}/org.eclipse.kura.emulator_${org.eclipse.kura.emulator.version}.jar@4" />
```
nicolatimeus: Is this still needed?

MMaiero: No it is not. Removed

nicolatimeus (approved):

PR #6345, opened 2026-08-26
fix(watchdog): remove stale enabled-marker file when watchdog is disabled

MMaiero:
## Summary
- `WatchdogServiceImpl` writes a temporary reference file (`/tmp/watchdog`) while the watchdog is enabled, indicating which device is armed so that shutdown tooling can keep petting it while the framework stops.
- When the watchdog was later disabled again, this file was never removed, so it kept pointing at a device the service no longer manages. On a subsequent restart, shutdown tooling would treat the watchdog as still active based on the stale file, which could lead to an unwanted reboot.
- `doUpdate()` now deletes this file when the watchdog config is applied as disabled.

## Test plan
- [x] Added `temporaryReferenceFileRemovedWhenWatchdogDisabledAfterBeingEnabled` to `WatchdogServiceImplTest`, verified it fails without the fix and passes with it.
- [x] Ran the full `org.eclipse.kura.linux.watchdog.test` suite locally (all green).


kura/org.eclipse.kura.linux.watchdog/src/test/java/org/eclipse/kura/linux/watchdog/WatchdogServiceImplTest.java
```diff
@@ -188,6 +204,15 @@ public void testDisable() throws Throwable {
         svc.deactivate();
     }
 
+    @Test
+    public void temporaryReferenceFileRemovedWhenWatchdogDisabledAfterBeingEnabled() throws Throwable {
```
nicolatimeus, line 211: Please add a test that checks that temporary reference file is not deleted if the WatchdogService is deactivated with enabled watchdog

MMaiero: Added `temporaryReferenceFileNotRemovedWhenServiceDeactivatedWhileWatchdogEnabled` in 312317e1b, which asserts the temporary reference file still exists after `deactivate()` is called while the watchdog is enabled.

nicolatimeus (approved):

PR #6348, opened 2026-08-28
feat(api): report primary SIM eUICC (EID) in the birth certificate

MMaiero:
## Summary
- Add a `modem_eid` metric to `KuraBirthPayload` (new `getModemEid()`/`withModemEid()`, `@since 3.0`), bumping the `org.eclipse.kura.message` package version 1.5.0 → 1.6.0 to satisfy the OSGi API baseline check.
- Capture `Sim.getEid()` from the active/primary SIM in both the Kapua and Eclipse IoT cloud connectors and forward it into the birth payload, mirroring the existing ICCID/IMSI handling exactly.
- Only the primary SIM is reported, matching current ICCID/IMSI scope — no multi-SIM enumeration.

The device birth certificate needs to carry the SIM's eUICC so it can be correlated with the SIM management platform.

## Test plan
- [x] `KuraBirthPayloadTest` (unit) — 10/10 pass
- [x] `LifeCyclePayloadBuilderTest` (unit) — 2/2 pass
- [x] `CloudServiceTest` (in-OSGi MQTT integration test via `verify`) — 13/13 pass, including `shouldPublishBirthMessageWithModemInfoWhenNetworkStatusServiceIsPresent` and `shouldPublishBirthMessageWithConnectedModemInfoWhenMultipleModems`, both asserting `modem_eid` in the published birth payload
- [x] Full reactor build (`org.eclipse.kura.api` baseline check + all touched bundles) compiles clean


kura/org.eclipse.kura.cloudconnection.kapua.mqtt.provider/src/main/java/org/eclipse/kura/core/cloud/LifeCyclePayloadBuilder.java
```diff
@@ -120,6 +111,11 @@ public KuraBirthPayload buildBirthPayload() {
             birthPayloadBuilder.withModemImsi(this.cloudServiceImpl.imsi);
         }
 
+        if (this.cloudServiceImpl.eid != null && this.cloudServiceImpl.eid.length() > 0
+                && !this.cloudServiceImpl.eid.equals(ERROR)) {
```
pierantoniomerlino, line 124: Why ERROR? The eid is initialized to NA.

PR #6360, opened 2026-09-02
test(eclipseiot.mqtt.provider): fix wrong coverage path

mattdibi:
Title

### DIFF
--- a/kura/org.eclipse.kura.cloudconnection.eclipseiot.mqtt.provider/pom.xml
+++ b/kura/org.eclipse.kura.cloudconnection.eclipseiot.mqtt.provider/pom.xml
@@ -30,8 +30,7 @@
 
     <properties>
         <kura.basedir>${project.basedir}/..</kura.basedir>
-        <sonar.coverage.jacoco.xmlReportPaths>
-            ${project.basedir}/../test/org.eclipse.kura.cloudconnection.eclipseiot.mqtt.provider.test/target/site/jacoco-aggregate/jacoco.xml</sonar.coverage.jacoco.xmlReportPaths>
+        <sonar.coverage.jacoco.xmlReportPaths>${project.build.directory}/site/jacoco-aggregate/jacoco.xml</sonar.coverage.jacoco.xmlReportPaths>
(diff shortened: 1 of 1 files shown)


mattdibi:
@salvatore-coppola In the https://github.com/eclipse-kura/kura/tree/develop/kura/org.eclipse.kura.event.publisher bundle POM I see:

https://github.com/eclipse-kura/kura/blob/002978fe3fc3826a67798e0be58db99d83c87e06/kura/org.eclipse.kura.event.publisher/pom.xml#L31

which points to the integration tests... but that bundle also has unit tests (see https://github.com/eclipse-kura/kura/tree/develop/kura/org.eclipse.kura.event.publisher/src/test/java/org/eclipse/kura/event/publisher/test). Are they picked up correctly?

Shouldn't this have **two paths**? One for the integration tests one for the unit?

---

Looking at the coverage reported by Sonar it looks ok-ish... needs further investigation

salvatore-coppola (approved):

mattdibi:
> @salvatore-coppola In the https://github.com/eclipse-kura/kura/tree/develop/kura/org.eclipse.kura.event.publisher bundle POM I see:
> 
> https://github.com/eclipse-kura/kura/blob/002978fe3fc3826a67798e0be58db99d83c87e06/kura/org.eclipse.kura.event.publisher/pom.xml#L31
> 
> which points to the integration tests... but that bundle also has unit tests (see https://github.com/eclipse-kura/kura/tree/develop/kura/org.eclipse.kura.event.publisher/src/test/java/org/eclipse/kura/event/publisher/test). Are they picked up correctly?
> 
> Shouldn't this have **two paths**? One for the integration tests one for the unit?
> 
> Looking at the coverage reported by Sonar it looks ok-ish... needs further investigation

As per our offline discussion: current code is correct. The unit test coverage report is aggregated to the integration test coverage report by the integration tests bundle, therefore we only need to specify the integration tests coverage report path.

---

The build passes but it was stopped by a rebase. Force merging to avoid spending 45 minutes waiting uselessly

PR #6370, opened 2026-09-03
ci(Jenkinsfile): exclude duplicated cloud connection classes from Sonar CPD

salvatore-coppola:
This change adds those classes to `sonar.cpd.exclusions` so their known duplication no longer counts against the new-code gate. It only affects the copy-paste detection: the classes are still analysed for issues, coverage and security.


salvatore-coppola:
we decided to wait the 30 days lifetime of new code block

PR #6372, opened 2026-09-03
test: bridge the log4j API into slf4j-simple for the unit tests

salvatore-coppola:
The unit tests hosted in the bundles run with `slf4j-simple` as their only logging provider, but `org.eclipse.kura.api` depends on `log4j-api` and `KuraException`, `KuraRuntimeException`, `Cloudlet` and `CommConnectionImpl` log through it. Every surefire run therefore prints `Log4j API could not find a logging provider` and whatever those classes log is lost.

This adds a test-scoped `log4j-to-slf4j` bridge next to `slf4j-simple` in `kura/pom.xml` (version managed in the bom at `${log4j.version}`), so the log4j 2 API output reaches the same sink. Test scope only: nothing changes in the bundle manifests or in the bnd-testing OSGi runtime, which keeps logging through log4j2.

### DIFF
--- a/bom/pom.xml
+++ b/bom/pom.xml
@@ -169,6 +169,11 @@
                 <version>${log4j.version}</version>
             </dependency>
             <dependency>
+                <groupId>org.apache.logging.log4j</groupId>
+                <artifactId>log4j-to-slf4j</artifactId>
+                <version>${log4j.version}</version>
+            </dependency>
+            <dependency>
                 <groupId>org.apache.commons</groupId>
                 <artifactId>commons-fileupload2-jakarta-servlet5</artifactId>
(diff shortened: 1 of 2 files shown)


marcellorinaldo (approved):

salvatore-coppola:
Closing: the log4j-to-slf4j bridge conflicts with log4j-slf4j2-impl, which the test modules carry at compile scope for the OSGi index (`log4j-slf4j2-impl cannot be present with log4j-to-slf4j`, see the CI failures in test/org.eclipse.kura.core.inventory.test). Without a provider log4j-api falls back to its SimpleLogger at ERROR level, so nothing important is lost in the unit tests; the extra complexity is not worth it.

PR #6374, opened 2026-09-04
docs(archetype): update the add-on archetype page to the kura-bom based project

salvatore-coppola:
The Kura Addon Archetype page still described the archetype as a jar shipped with the Kura release artifacts and a generated project based on the `kura` and `target-platform` BOMs, the PDE dependency aggregators and manifests to maintain by hand. None of that matches the archetype after the Kura migration to bnd and eclipse-kura/kura-archetype#4.

The page now documents the eclipse-kura/kura-archetype repository and how to install the archetype from it, the one-shot generation command with a table of the properties, the generated project structure, the `kura-bom` based dependency management (Kura dependencies declared without version, extra libraries managed in the root pom), the explicit Kura runtime indexed for the bnd integration tests and how to require more bundles in the `.bndrun`, the `bundle.under.test` requirement, the unit and OSGi test reports location, and the current Debian control file (`kura-core` dependency bounds) and package naming. The IDE and architecture-specific sections are kept, with the outdated `Depends: kura` and installer naming fixed.

The interactive generation screenshot is regenerated from the current prompt, which also shows the `year` and `archetypeVersion` properties.


docs/java-application-development/kura-addon-archetype.md
```diff
@@ -1,72 +1,102 @@
...
+- **JDK 21**: the generated project sets `maven.compiler.release=21`.
+- **Maven 3.9.x**
+- **git**: required to build a generated project, the `git-commit-id-maven-plugin` reads the commit hash that goes into the Debian package version.
+
+## Installing the archetype
+
+Clone the repository and install the archetype in the local Maven repository:
 
 ```shell
-mvn install:install-file \
```
marcellorinaldo, line 12: The archetype should be available on Maven central. This steps should not be required

marcellorinaldo (changes requested):

docs/java-application-development/kura-addon-archetype.md
salvatore-coppola: Right, removed the installation section: the page now states that the archetype is consumed from Maven Central with no installation step, the generate commands no longer pass -DarchetypeCatalog=local and use <kura-version>, and a note covers the snapshots repository for -SNAPSHOT versions. Screenshot regenerated accordingly.

marcellorinaldo (approved):

PR #6377, opened 2026-09-04
refactor(api): log through SLF4J instead of the log4j 2 API in api and core.comm

salvatore-coppola:
`KuraException`, `KuraRuntimeException`, `Cloudlet` and `CommConnectionImpl` were the only framework classes logging through `org.apache.logging.log4j` directly (`LogManager.getLogger`); every other bundle uses SLF4J. Besides the inconsistency, a consumer whose unit tests have an SLF4J provider but no log4j 2 provider on the classpath gets `Log4j API could not find a logging provider` as soon as a `KuraException` is created, and cannot fix it with the `log4j-to-slf4j` bridge when `log4j-slf4j2-impl` is also present (log4j refuses the combination). This is what the ESF unit tests hit after the bnd migration.

### Changes
- `Cloudlet` and `CommConnectionImpl` switch to SLF4J (`LoggerFactory.getLogger`). The serial-payload debug logs keep their lazy formatting through the SLF4J fluent API (`logger.atDebug().addArgument(() -> …).log(…)`). In `Cloudlet`, log calls that fed a trailing exception into a `{}` placeholder are corrected to pass it as the SLF4J throwable, so the stack trace is logged instead of a literal `{}`.
- `KuraException` and `KuraRuntimeException` lose their logger (its only uses were one unreachable warning and one for a missing bundle/code, a case the generic `Generic Error - <code>` pattern already reports to the caller). The lookup/formatting they carried in identical copies now lives in a package-private `KuraExceptionMessages` helper: it checks the key with `ResourceBundle.containsKey` and still falls back to the generic pattern (a retained `try/catch` covers a missing bundle). Each exception keeps only two one-line `getMessage`/`getLocalizedMessage` overrides, with the locale selection centralized in the helper so the block is no longer duplicated.
- The api bundle has no test module, so the message-building unit test lives in the core integration test suite (`org.eclipse.kura.core.test`), whose JaCoCo report already backs the api bundle's Sonar coverage.
- The `org.apache.logging.log4j` import and the `log4j-api` dependency leave the `Import-Package`/poms of `org.eclipse.kura.api` and `org.eclipse.kura.core.comm`. No other module relied on the log4j API arriving transitively from the api bundle.


marcellorinaldo (dismissed):

kura/org.eclipse.kura.api/src/test/java/org/eclipse/kura/KuraExceptionMessageTest.java
```diff
@@ -0,0 +1,91 @@
...
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *  Eurotech
+ *******************************************************************************/
+package org.eclipse.kura;
```
marcellorinaldo: We do not have a test module for the APIs-only. I won't introduce it with this PR.

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/cloud/Cloudlet.java
marcellorinaldo: Be careful on how the log4j2 and the slf4j APIs differ (log4j consumes a trailing Throwable as a parameter when placeholders == args).

Using slf4j here will print a trailing `{}`. Check also in the other places.

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/KuraExceptionMessages.java
```diff
@@ -0,0 +1,57 @@
...
+            final StringBuilder allArguments = new StringBuilder();
+            for (final Object argument : arguments) {
+                allArguments.append(" - ").append(argument);
+            }
+            messageArguments = new Object[] { allArguments.toString() };
+        }
+        return MessageFormat.format(pattern, messageArguments);
+    }
+
+    static String messagePattern(final Locale locale, final KuraErrorCode code) {
```
marcellorinaldo: This could be private

kura/org.eclipse.kura.core.comm/src/main/java/org/eclipse/kura/core/comm/CommConnectionImpl.java
marcellorinaldo: Invoke lazily like the others in this file

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/KuraExceptionMessages.java
marcellorinaldo: We are dropping the try/catch for the `MissingResourceException`, is this still safe?

marcellorinaldo (changes requested):
<del>One further question: why is this a stacked PR?</del>

Nevermind, it is written in the description

kura/org.eclipse.kura.api/src/test/java/org/eclipse/kura/KuraExceptionMessageTest.java
salvatore-coppola: Removed the test and the api-only test setup (junit/mockito, jacoco, coverage path) in 23a7b4f924.

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/cloud/Cloudlet.java
salvatore-coppola: Good catch. Fixed in 5a8539e143: dropped the extra `{}` so the exception is passed as the SLF4J throwable (logged with its stack trace). Checked the other places too — the same pattern was in the 'Unexpected throwable' catch and in 'Error publishing response for topic', both fixed; the two `logger.warn` calls in CommConnectionImpl pass a String (not a Throwable) so they are fine.

kura/org.eclipse.kura.api/src/main/java/org/eclipse/kura/KuraExceptionMessages.java
salvatore-coppola: Done in 765140acc6 — `messagePattern` is now private (only used by `localizedMessage`).

salvatore-coppola: Good point — it was not fully safe. `containsKey` covers a missing key, but `ResourceBundle.getBundle` still throws `MissingResourceException` if the whole bundle is absent (the old code caught that). Restored the fallback in 765140acc6: the lookup is wrapped in try/catch and a missing bundle now falls back to the generic pattern.

kura/org.eclipse.kura.core.comm/src/main/java/org/eclipse/kura/core/comm/CommConnectionImpl.java
salvatore-coppola: Done in 4cc71aadaf — switched to `logger.atDebug().addArgument(() -> getBytesAsString(command)).log(SEND_MESSAGE)`, matching the other two sendMessage logs.

marcellorinaldo (dismissed):

marcellorinaldo (dismissed):

sfiorani (approved):

PR #6160, opened 2026-01-22
feat(api)!: Updated GPIO APIs

pierantoniomerlino:
This pull request introduces significant enhancements and refactoring to the GPIO API and emulator implementation to modernize GPIO pin access and description in the Kura project. The main focus is on replacing the legacy terminal-based pin access with a more descriptive, property-based approach using the new `KuraGPIODescription` class, and updating the emulator to support these changes. Several legacy methods are marked as deprecated in favor of the new API.

**API modernization and new features:**

* Introduced the new `KuraGPIODescription` class, which encapsulates GPIO pin properties and provides a standardized way to describe and identify pins. (`KuraGPIODescription.java`)
* Extended the `GPIOService` interface to add new methods for retrieving pins and pin descriptions using property maps, and marked legacy terminal-based methods as deprecated. (`GPIOService.java`) [[1]](diffhunk://#diff-2116eab2fe95af1019246ef5e394940c88b8f9a2816c8ecadefc05c40024a648R15-R22) [[2]](diffhunk://#diff-2116eab2fe95af1019246ef5e394940c88b8f9a2816c8ecadefc05c40024a648R31-R134)

**Emulator and implementation updates:**

* Refactored `EmulatedPin` to use controller and line numbers instead of a single terminal index, and implemented the `getDescription()` method to return a `KuraGPIODescription`. (`EmulatedPin.java`) [[1]](diffhunk://#diff-cef6196f5ccad2f90357d84b2b43d0e66d671e4cdc59eabf6f35e09c72a01cbeL34-R38) [[2]](diffhunk://#diff-cef6196f5ccad2f90357d84b2b43d0e66d671e4cdc59eabf6f35e09c72a01cbeL47-R58) [[3]](diffhunk://#diff-cef6196f5ccad2f90357d84b2b43d0e66d671e4cdc59eabf6f35e09c72a01cbeL60-R81) [[4]](diffhunk://#diff-cef6196f5ccad2f90357d84b2b43d0e66d671e4cdc59eabf6f35e09c72a01cbeL119-R168)
[...]

PR #5734, opened 2025-02-20
feat: adding encryption and serialization through input/output stream

sfiorani:
This PR introduces new APIs for the CryptoService and the Marshalling service: these new APIs allow to encrypt/decrypt and marshal/unmarshal through Input and Output streams.

Before this change, the available apis forced the storage of inputs and ouputs of operation in memory: this could be a problem for example during snapshot encryption, because it was necessary to store different copies of the entire snapshot in memory to encrypt and marshalling it; if the snapshot was huge, the memory was filled of copies of the same object.

This PR introduces an optimization, exploiting input and output streams to lighten these processes to avoid filling the memory.

**Related Issue:** This PR fixes/closes {issue number}

**Description of the solution adopted:** A more detailed description of the changes made to solve/close one or more issues. If the PR is simple and easy to understand this section can be skipped

**Screenshots:** If applicable, add screenshots to help explain your solution

**Manual Tests**: Optional description of the tests performed to check correct functioning of changes, useful for an efficient review

**Any side note on the changes made:** Description of any other change that has been made, which is not directly linked to the issue resolution [e.g. Code clean up/Sonar issue resolution]

PR #6380, opened 2026-09-07
fix: correct the exported package versions and shrink the exported surface

salvatore-coppola:
Follow-up of #6379. Every `Export-Package` of the repository was compared with `KURA_5.6.2_RELEASE`, then the `bnd-baseline-maven-plugin` was enabled on all the exporting bundles to keep the result honest, and finally the exported surface itself was reduced to what is actually consumed.

## 1. Versions that did not follow the API changes

| Package | Was | Now | Reason |
|---|---|---|---|
| `org.eclipse.kura.core.configuration` | 2.0.0 | 3.0.0 | public `unset*` methods removed with the DS annotations (#6275), `marshal`/`unmarshal` moved to streams (#5734), `deactivate` signature changed |
| `org.eclipse.kura.core.data` | 1.3.0 | 2.0.0 | `openMessageStore`, `withConnection`, H2 setters and listener methods removed (#5343, #5344) |
| `org.eclipse.kura.core.inventory` | 1.1.0 | 1.1.1 | implementation changes only |
| `org.eclipse.kura.core.keystore.util` | 1.1.0 | 2.0.0 | `KeystoreRemoteService` removed (#5341) |
| `org.eclipse.kura.core.ssl` | 1.0.0 | 2.0.0 | `activate` replaced by a public constructor (#5703) |
| `org.eclipse.kura.rest.auth` | 1.0.0 | 2.0.0 | the signature of `AuthenticationProvider.authenticate` moved from `javax` to `jakarta` |
| `org.eclipse.kura.rest.configuration.api` | 1.1.0 | 2.0.0 | `PropertyDTO.scalarFormSingletonClass`/`scalarFromClass` removed, `javax.ws.rs` → `jakarta.ws.rs` in the signatures (#5478, #5624) |
| `org.eclipse.kura.util.validation` | 1.0.0 | 2.0.0 | a method was added to the `PasswordStrengthValidators.Messages` interface (#5949), which carries no `@ProviderType`, so its implementors break |
| `org.eclipse.kura.util.zip` | 1.0.0 | 1.0.1 | `UnZip` rewritten with the same API (#6327) |
| `org.eclipse.kura.emulator.gpio` | 1.0.0 | 1.1.0 | new public methods (#6160) |
| `org.eclipse.kura.emulator.net` | 1.0.0 | 1.1.0 | three new public classes |
| `org.eclipse.kura.emulator.clock`/`.position`/`.watchdog` | 1.0.0 | 1.1.0 | new public metatype types added with the migration |
| `org.eclipse.kura.emulator`/`.usb` | 1.0.0 | 1.0.1 | implementation changes only |

The `Import-Package` ranges of the consuming bundles are widened accordingly. Three of them imported
`org.eclipse.kura.configuration` with `[1.2,1.3)`, a range that excludes the next minor of an API package by
construction: they are widened to `[1.2,2.0)` as well.

`org.eclipse.kura.configuration` itself is left at 1.2.0. The audit first raised it to 1.3.0 for the `equals`
and `hashCode` added to `Password` by #2534, but #6399 reverted that addition on `develop`, so the package is
unchanged since 5.6.2 again and this pull request no longer touches the manifest of the API bundle. The
baseline report agrees, it lists the package as `UNCHANGED`.

## 2. Baselining every exporting bundle

The `bnd-baseline-maven-plugin` was configured in the root POM but bound only in `org.eclipse.kura.api`, so the versions of every other bundle were never compared with the previous release. It is now enabled in the bundles that declare an `Export-Package`, with `failOnMissing` turned off so that `cloud.base.provider` and `core.testutil`, introduced in 6.0.0 and without a released baseline, are skipped rather than failing the build.

Running it disagreed with five of the versions this branch had set by hand, and the table above already carries the corrected values: `util.validation` and `rest.auth` needed a major rather than the minor or no bump they had, and `core.inventory`, `emulator` and `emulator.usb` needed a micro they were not getting. From here on, a package whose API moves without its version following fails the build instead of reaching a release.

## 3. Shrinking the exported surface

Only the packages of `org.eclipse.kura.api` are API. The 54 packages exported by the other bundles were checked against every repository of the `eclipse-kura` and `eurotech` organisations, both Java sources and manifests.

| | Packages | What changed |
|---|---|---|
| Consumed by a submodule or an add-on | 46 | keep their export, now all declared `;x-internal:=true` |
| Consumed by nobody | 8 | no longer exported |

The eight that go private:

| Package | Evidence |
|---|---|
| `org.eclipse.kura.core.crypto` | only the bundle is referenced, as a Maven test dependency, never the package |
| `org.eclipse.kura.emulator.clock`/`.gpio`/`.net`/`.position`/`.usb`/`.watchdog` | referenced by bundle name in target definitions only; the emulator bundles are wired by `osgi.identity` |
| `org.eclipse.kura.internal.rest.cloudconnection.provider.dto` | reachable only because its test module was a plain bundle instead of a fragment |

That last one is fixed rather than worked around: the test module becomes a fragment of the bundle it tests, like every other test module, which also gives it access to the private packages. Its 28 integration tests pass unchanged.

Two packages that look unused inside this repository are kept because add-ons consume them: `org.eclipse.kura.core.ssl`, used in Java by an ESF add-on, and `org.eclipse.kura.core.linux.util`, declared in two add-on manifests. `org.eclipse.kura.request.handler.jaxrs.consumer` has no external consumer either, but its types appear in the public signatures of the exported `request.handler.jaxrs` package, so it has to stay reachable.

## 4. Shrinking the `org.eclipse.kura.core.configuration` export

The same idea applied to the one package whose major bump started all this. It exported the ConfigurationService implementation (`ConfigurationServiceImpl`, `ConfigurationServiceAuditFacade`, `CloudConfigurationHandler` and the two trackers) next to the DTOs the other bundles consume, and no bundle outside this one uses those classes: they move to the private `org.eclipse.kura.core.configuration.internal` package, together with the test classes that reach their package-private members. What is left exported is `ComponentConfigurationImpl`, `ConfigurationChangeEvent` and the `Xml*` types, all unchanged since 5.6.2, and the `uses:` constraint of the package drops from 14 packages to 4.

The Declarative Services component names are declared explicitly in the annotations, so the generated descriptors, the component names and the service PIDs are unchanged; the integration test suite of the bundle (162 tests) passes unmodified. The unused `Password` class, a duplicate of `org.eclipse.kura.configuration.Password` that nothing referenced, is removed.


salvatore-coppola:
### Why `org.eclipse.kura.core.configuration` goes `2.0.0` → `3.0.0`

The exported package `org.eclipse.kura.core.configuration` contains the `public` class `ConfigurationServiceImpl`, whose **binary API broke since `KURA_5.6.2_RELEASE`**. Per OSGi semantic versioning these are **major** changes, so bnd-baseline mandates the major bump.

**1. Eight `public` DS unbind methods removed** (removing a `public` member is binary-incompatible):

```diff
- public void unsetConfigurationAdmin(ConfigurationAdmin configAdmin)
- public void unsetMetaTypeService(MetaTypeService metaTypeService)
- public void unsetSystemService(SystemService systemService)
- public void unsetCryptoService(CryptoService cryptoService)
- public void unsetScrService(ServiceComponentRuntime scrService)
- public void unsetXmlMarshaller(final Marshaller marshaller)
- public void unsetXmlUnmarshaller(final Unmarshaller unmarshaller)
- public void unsetEventAdmin(EventAdmin eventAdmin)
```

(explicit unbind methods are no longer needed after the move to DS annotation / field references)

**2. `protected` signatures changed** (part of the subclass contract → also incompatible):

```diff
- protected void deactivate(ComponentContext componentContext)
+ protected void deactivate()

- protected <T> T unmarshal(final String string, final Class<T> clazz)
+ protected <T> T unmarshal(final InputStream input, final Class<T> clazz)

- protected String marshal(final Object object)
+ protected void marshal(final OutputStream out, final Object object)
```

The snapshot (de)serialization moved from `String` to streams (`FileInputStream` / `encryptedStream` / `decryptedStream`), so it no longer loads the whole snapshot into memory.

Any one of these already forces a major bump; here there are several, so `3.0.0` is correct.

> Not counted: `public synchronized void rollback(long)` → `public void rollback(long)` — `synchronized` is not part of the binary signature seen by callers, so it is not an API change.

The accompanying `Import-Package` change (`org.eclipse.kura.configuration;version="[1.2,1.3)"` → `"[1.2,2.0)"`) just widens an over-tight upper bound to the standard `[x.y,2.0)` range used by the other kura imports in this bundle.

salvatore-coppola:
I'm not sure on org.eclipse.kura.core.configuration @nicolatimeus  what do you think ?

kura/org.eclipse.kura.core.configuration/bnd.bnd
```diff
@@ -28,9 +28,9 @@ Import-Package: \
     org.slf4j;version="1.6.4",\
     *
 Export-Package: \
-    org.eclipse.kura.core.configuration;version="2.0.0",\
-    org.eclipse.kura.core.configuration.metatype;version="1.0.0",\
-    org.eclipse.kura.core.configuration.util;version="2.1.0"
+    org.eclipse.kura.core.configuration;version="3.0.0";x-internal:=true,\
```
marcellorinaldo, line 31: I'm not sure about this. We removed the OSGi hook methods and changed the deactivate signature but in my opinion those were not part of the intended APIs. The mentioned `marshal/unmarshal moved to streams` (https://github.com/eclipse-kura/kura/pull/5734) was not a move but an API addition. I would just bump the minor here.

kura/org.eclipse.kura.core/bnd.bnd
```diff
@@ -25,9 +24,9 @@ Import-Package: \
     org.slf4j;version="1.6.4",\
     *
 Export-Package: \
-    org.eclipse.kura.core.linux.executor;version="1.0.0",\
-    org.eclipse.kura.core.linux.util;version="1.2.0",\
-    org.eclipse.kura.core.ssl;version="1.0.0",\
+    org.eclipse.kura.core.linux.executor;version="1.0.0";x-internal:=true,\
+    org.eclipse.kura.core.linux.util;version="1.2.0";x-internal:=true,\
+    org.eclipse.kura.core.ssl;version="2.0.0";x-internal:=true,\
```
marcellorinaldo, line 29: The mentioned `activate replaced by a public constructor (https://github.com/eclipse-kura/kura/pull/5703)` is not an API intended to be used outside of the OSGi SCR. I wouldn't do a major version bump here

kura/org.eclipse.kura.rest.identity.provider/bnd.bnd
```diff
@@ -6,13 +6,13 @@ Import-Package: \
...
-    org.eclipse.kura.core.configuration;version="[2.0,3.0)",\
+    org.eclipse.kura.core.configuration;version="[3.0,4.0)",\
     org.eclipse.kura.crypto;version="[1.3,2.0)",\
     org.eclipse.kura.identity;version="[1.1,2.0)",\
     org.eclipse.kura.request.handler.jaxrs;version="[2.0,3.0)",\
-    org.eclipse.kura.rest.configuration.api;version="[1.0,2.0)",\
+    org.eclipse.kura.rest.configuration.api;version="[2.0,3.0)",\
     org.eclipse.kura.util.useradmin;version="[1.0,2.0)",\
-    org.eclipse.kura.util.validation;version="[1.0,2.0)",\
+    org.eclipse.kura.util.validation;version="[2.0,3.0)",\
```
marcellorinaldo: I would tackle this separately

marcellorinaldo (commented):
Lets discuss about these points

kura/org.eclipse.kura.core.configuration/bnd.bnd
nicolatimeus: This package was exporting the `ConfigurationServiceImpl` implementation, this PR moves that to an internal (non exported) package, so a major version bump can be ok in my opinion. This PR should also enable the `bnd-baseline-plugin` on this bundle, so the major version bump will probably be enforced.

marcellorinaldo: Oh okay, I missed that!

kura/org.eclipse.kura.core/bnd.bnd
salvatore-coppola: Regardless of the intended behaviour, the binary compatibility has been broken. We need to update the version to a major one, as the baseline plugin confirms.

kura/org.eclipse.kura.api/bnd.bnd
```diff
@@ -41,7 +41,7 @@ Export-Package: \
     org.eclipse.kura.cloudconnection.subscriber.listener;version="1.0.0",\
     org.eclipse.kura.comm;version="1.1.0",\
     org.eclipse.kura.command;version="1.2.0",\
-    org.eclipse.kura.configuration;version="1.2.0",\
```
nicolatimeus, line 44: Is this bump needed?

nicolatimeus: The changes to the passowrd.hashCode and equals methods have been rolled back

salvatore-coppola: reverted

nicolatimeus (approved):

PR #6384, opened 2026-09-10
fix(core.configuration): register SelfConfiguringComponents by kura.service.pid

salvatore-coppola:
A `SelfConfiguringComponent` was tracked only when its service reference exposed a `service.pid` property.

These components declare an OPTIONAL configuration policy and usually have no ConfigurationAdmin configuration, and Felix SCR publishes `service.pid` only when it comes from a configuration or from an explicit component property, so the component was silently discarded: it never showed up in the snapshot and `updateConfiguration` reported it as not tracked.

The registration now keys on `kura.service.pid`, the identity every other lookup already uses, and falls back to it when `service.pid` is missing.

**Description of the solution adopted:** `addSelfConfiguringComponent` and `removeSelfConfiguringComponent` read `kura.service.pid` and no longer bail out when `service.pid` is absent; the kura pid is used as the ConfigurationAdmin target in that case, the same fallback `updateWithDefaultConfiguration` already applied. The null handling is left to `registerSelfConfiguringComponent` and `unregisterComponentConfiguration`, so a reference without `kura.service.pid` is still rejected with the same warning as before.

`makeString` was also hardened. With a DS 1.3 or later descriptor, a component that declares `service.pid` as a property while a configuration exists for the same pid makes SCR publish `service.pid` as a two element list. That value was turned into its string representation (`[pid, pid]`) and used as a ConfigurationAdmin target, so the real configuration was never updated and the component was never reconfigured. Collections and arrays of any type holding a single distinct pid are now unwrapped, and ambiguous ones are rejected with a warning.


salvatore-coppola:
Verified on a Raspberry Pi running Kura 6.0.0-SNAPSHOT, comparing the same boot with and without this change.

The test bundle is a `SelfConfiguringComponent` whose DS 1.3 descriptor declares only `kura.service.pid`, with `configuration-policy="optional"` and a real OCD, i.e. the shape a component ends up with once the redundant `service.pid` property is dropped.

| | without the fix | with the fix |
| --- | --- | --- |
| Bundle / component state | ACTIVE, activate() runs | ACTIVE, activate() runs |
| Pid in `configuration/v2/configurableComponents` | absent | present |
| `getConfiguration()` called while taking a snapshot | never | yes |
| Snapshot size | 49773 bytes | 50213 bytes |

So without the change the component is up and registered as an OSGi service, but the ConfigurationService silently ignores it: no configuration, and nothing in the snapshot.

One side observation worth recording. After taking a snapshot with the fix in place, reverting to the unpatched bundle still showed the component as tracked: the snapshot had recreated a ConfigurationAdmin configuration for that pid, and an existing configuration is enough for SCR to publish `service.pid` and hide the problem. Clearing those snapshots restored the failing behaviour. This is why a component that gets a seeded configuration at install time never shows the bug, while one that does not is invisible.

The DS 1.3 duplicated `service.pid` part of the change is covered by unit tests only, it was not exercised on the device.

kura/org.eclipse.kura.core.configuration/src/main/java/org/eclipse/kura/core/configuration/ConfigurationServiceImpl.java
```diff
@@ -278,27 +278,25 @@ protected void removeConfigurableComponent(final ServiceReference<ConfigurableCo
             unbind = "removeSelfConfiguringComponent")
     protected void addSelfConfiguringComponent(final ServiceReference<SelfConfiguringComponent> reference) {
 
-        final String servicePid = makeString(reference.getProperty(Constants.SERVICE_PID));
+        final String kuraPid = makeString(reference.getProperty(ConfigurationService.KURA_SERVICE_PID));
 
-        if (servicePid == null) {
+        if (kuraPid == null) {
```
nicolatimeus: This is probably a breaking change.. components that set service.pid but not kura.service.pid were accepted before of this change and now silently dropped

salvatore-coppola: I changed something and added warnings but a component exposing `service.pid` but no `kura.service.pid` was not accepted before either: `addSelfConfiguringComponent` passed the null kura pid to `registerSelfConfiguringComponent`, which has rejected it with `PID (kura.service.pid) is null` since the initial commit, and `testRegisterSelfConfiguringComponentNull` in `ConfigurationServiceJunitTest` documents exactly that ("just abort").

Such a component could not have worked anyway, since every SelfConfiguringComponent lookup (`getSelfConfiguringComponentConfiguration`, the `(kura.service.pid=...)` filter in `getSelfConfiguringComponentDefaultConfiguration`, `getComponentConfigurations(Filter)`) keys on `kura.service.pid`.

kura/org.eclipse.kura.core.configuration/src/main/java/org/eclipse/kura/core/configuration/ConfigurationServiceImpl.java
```diff
@@ -1688,9 +1675,28 @@ private static String makeString(Object value) {
         if (value instanceof String) {
             return (String) value;
         }
+        if (value instanceof Collection) {
+            return makeStringFromCollection((Collection<?>) value);
+        }
+        if (value instanceof Object[]) {
```
nicolatimeus: Will this match arrays of primitive types (int[] ecc)?

salvatore-coppola: No, it did not: a primitive array is not an `Object[]`, so an `int[]` fell through to `value.toString()`.

For the keys read through `makeString` (`service.pid`, `kura.service.pid`, `service.factoryPid`) the OSGi spec only allows `String`, `String[]` or `Collection<String>`, and Felix SCR publishes the duplicated pid as a `List<String>`, so the case should not occur in practice. 

Still, the branch is now generic in bd0bed21af: any array is read through `java.lang.reflect.Array` and handled like a collection, with two test scenarios for the primitive case.

nicolatimeus (approved):

PR #6388, opened 2026-09-15
fix(core.configuration): close snapshot file streams before rewriting the snapshot

salvatore-coppola:
## Summary

`ConfigurationServiceImpl` leaked two `FileInputStream`s on the snapshot file:

- `loadEncryptedSnapshotFileContent` passed `new FileInputStream(fSnapshot)` to `CryptoServiceImpl.aesDecryptingStream`, which throws `KuraException` when the snapshot is not encrypted **without closing the wrapped stream**; on the success path `unmarshal` never closes its input either.
- `encryptPlainSnapshots` re-read the same snapshot through another unclosed `FileInputStream` and then immediately replaced the file with `Files.move(..., ATOMIC_MOVE)` while both handles were still open.

Both reads now use try-with-resources so the snapshot file is released before it is rewritten. The behaviour of the method is unchanged (decrypt failure still propagates to the plain-snapshot fallback, XML parse failure still logs and returns `null`).


kura/org.eclipse.kura.core.configuration/src/main/java/org/eclipse/kura/core/configuration/ConfigurationServiceImpl.java
```diff
@@ -1389,23 +1391,20 @@ XmlComponentConfigurations loadEncryptedSnapshotFileContent(long snapshotID) thr
...
-        try {
-            decryptedStream = this.cryptoService.aesDecryptingStream(new FileInputStream(fSnapshot));
+        try (InputStream fileStream = new FileInputStream(fSnapshot)) {
+            InputStream decryptedStream = this.cryptoService.aesDecryptingStream(fileStream);
+            try {
+                return unmarshal(decryptedStream, XmlComponentConfigurations.class);
+            } catch (KuraException e) {
+                logger.warn("Error parsing xml", e);
+                return null;
+            }
```
pierantoniomerlino: I don't like the nested try. Is it possible to use only one, moving the internal one to the external and adding the appropriate catches?

kura/org.eclipse.kura.core.configuration/src/main/java/org/eclipse/kura/core/configuration/ConfigurationServiceImpl.java
```diff
@@ -1389,23 +1391,20 @@ XmlComponentConfigurations loadEncryptedSnapshotFileContent(long snapshotID) thr
                     fSnapshot != null ? fSnapshot.getAbsolutePath() : "null");
         }
 
-        InputStream decryptedStream = null;
-        try {
-            decryptedStream = this.cryptoService.aesDecryptingStream(new FileInputStream(fSnapshot));
+        try (InputStream fileStream = new FileInputStream(fSnapshot)) {
+            InputStream decryptedStream = this.cryptoService.aesDecryptingStream(fileStream);
```
pierantoniomerlino: Does it make sense to move this inputstream in the try-with-resources?

kura/org.eclipse.kura.core.configuration/src/main/java/org/eclipse/kura/core/configuration/ConfigurationServiceImpl.java
salvatore-coppola: Done in efd3a24930: the parsing moved to a helper (`unmarshalSnapshot`), so the method has a single try-with-resources. A single `catch` was not enough because the two `KuraException`s have different outcomes: from `aesDecryptingStream` it must propagate so that the caller falls back to the plain-snapshot migration, from `unmarshal` it is logged and `null` is returned (as before the change).

kura/org.eclipse.kura.core.configuration/src/main/java/org/eclipse/kura/core/configuration/ConfigurationServiceImpl.java
salvatore-coppola: I'd rather keep it out of the resources: it is an AES/GCM `CipherInputStream`, whose `close()` runs `doFinal()`; on a partially read stream (the parse-error case) that throws an `IOException` (`AEADBadTagException`), turning "log and return null" into a `KuraIOException`. Closing the underlying `FileInputStream` is what releases the file handle.

pierantoniomerlino (approved):

salvatore-coppola:
Superseded by #6393: the Jenkins pr-merge check does not run on PRs opened from forks, so the same commits were re-submitted from a branch on this repository.

PR #6393, opened 2026-09-15
fix(core.configuration): close snapshot file streams before rewriting the snapshot

salvatore-coppola:
## Summary

`ConfigurationServiceImpl` leaked two `FileInputStream`s on the snapshot file:

- `loadEncryptedSnapshotFileContent` passed `new FileInputStream(fSnapshot)` to `CryptoServiceImpl.aesDecryptingStream`, which throws `KuraException` when the snapshot is not encrypted **without closing the wrapped stream**; on the success path `unmarshal` never closes its input either.
- `encryptPlainSnapshots` re-read the same snapshot through another unclosed `FileInputStream` and then immediately replaced the file with `Files.move(..., ATOMIC_MOVE)` while both handles were still open.

Both reads now use try-with-resources so the snapshot file is released before it is rewritten. The behaviour of the method is unchanged (decrypt failure still propagates to the plain-snapshot fallback, XML parse failure still logs and returns `null`).


salvatore-coppola:
@pierantoniomerlino this replaces #6388, which you had already approved: the required `continuous-integration/jenkins/pr-merge` check never ran there because the Jenkins multibranch job does not build PRs opened from forks, so the same commits were pushed to a branch on this repository and re-submitted. Nothing changed in the content, apart from addressing your two review comments (parsing extracted to a helper, the decrypting stream kept out of the try-with-resources as discussed). Could you re-approve?

pierantoniomerlino (approved):

PR #6397, opened 2026-09-16
build: fix packaging warnings left by the bnd migration

salvatore-coppola:
`mvn clean install -DskipTests` emitted 28 warnings on `develop`. This PR addresses the ones that are real packaging defects, bringing the count down to 19.

- **Dead embedded protobuf** in the Eclipse IoT and Kapua MQTT cloud connection providers. Both declared `Bundle-ClassPath: .,lib/protobuf-java.jar`, but `-includeresource` listed the jar as a bare path, which bnd includes under its basename only. The jar landed in the bundle root and the classpath entry pointed nowhere, so bnd promoted protobuf to an external dependency and both bundles already resolve against the standalone `com.google.protobuf` bundle. The embedded copy was never loaded: 1.8 MB of dead weight each. Removed, keeping the `Import-Package` wiring in use today.
- **Invalid package names** in `org.usb4java`: `org.usb4java.linux-aarch64` and `org.usb4java.linux-x86-64` are resource directories, not packages. Declaring them as exports also duplicated each `libusb4java.so` into the bundle root. The libraries remain reachable through the nested `libusb4java-linux-*.jar` entries in `Bundle-ClassPath`, which is where `org.usb4java.Loader` looks them up via `Class.getResource`.
- **Unusable export** of `...sparkplug.mqtt.transport`, whose API exposes `org.eclipse.paho.client.mqttv3` types while paho is embedded as a private package. The only in-tree consumer is a test fragment, which shares the host class loader.
- **Redundant `org.newsclub.net.unix.*`** instruction in the container orchestration provider, never matched because the exact entry precedes it.
- **Platform-dependent encoding** in `checkstyle-resources`, which has no parent pom and so did not inherit `project.build.sourceEncoding`.
- **`KuraErrorCode.INTERNAL_ERROR`** has carried an `@deprecated` javadoc tag since 2016 with no matching annotation; added `@Deprecated`.

**Manual Tests:** `mvn clean install -DskipTests` succeeds; warnings go from 28 to 19. Verified on the produced bundles that the two MQTT providers no longer ship protobuf and keep importing `com.google.protobuf`, and that `org.usb4java` still exports `org.usb4java` with both native libraries present in the nested jars on `Bundle-ClassPath`.


kura/org.eclipse.kura.cloudconnection.eclipseiot.mqtt.provider/pom.xml
```diff
@@ -102,42 +102,13 @@
 
     <build>
         <plugins>
-            <plugin>
```
marcellorinaldo, line 105: Please give it a try on a real device to rule out that this was intentionally there

salvatore-coppola: Tested on a Raspberry Pi 4 (aarch64, Debian 13) running Kura 6.0.0-SNAPSHOT, against `test.mosquitto.org` as broker.

First on the build currently installed there, i.e. with the embedded jar still in place: I created an Eclipse IoT cloud connection and connected it, and the BIRTH certificate was published normally. That alone shows the embedded copy was already dead weight, since `Bundle-ClassPath` pointed at `lib/protobuf-java.jar` while the jar sat in the bundle root, so the encoder was already resolving `com.google.protobuf` against the standalone 4.30.2 bundle.

Then I installed the jars built from this branch (eclipseiot 1857825 → 126949 bytes, kapua 1879188 → 148323 bytes) and restarted the framework. Both providers come up ACTIVE, with no unresolved requirement, `NoClassDefFoundError` or `UnsatisfiedLinkError` in the log. I reconnected and captured the messages from the broker with a plain Paho client: the death certificate arrives raw, 66 bytes, and parses as a `KuraPayload`; the birth certificate arrives gzipped, 623 bytes expanding to 1051, and parses with 27 metrics, among them `payload_encoding = KURA_PROTOBUF`, `kura_version = KURA_6.0.0-SNAPSHOT` and `os_arch = aarch64`. The mixed compression is expected: `CloudPayloadGZipEncoder` only keeps the compressed form when it is shorter.

I also checked the `org.usb4java` change on the same device, since it is the other one with a runtime risk. The bundle now exports only `org.usb4java`, the two nested jars carrying the native libraries are still on its `Bundle-ClassPath`, and loading the native code straight from the new jar reports libusb 1.0.22.11312 and enumerates the three devices `lsusb` lists.

marcellorinaldo (approved):

PR #6404, opened 2026-09-19
feat: openapi generation

MMaiero:


pom.xml
```diff
@@ -94,6 +96,7 @@
         <!-- Dependency versions -->
         <mockito.version>5.21.0</mockito.version>
         <perfmark-api.version>0.26.0</perfmark-api.version>
+        <swagger.version>2.2.55</swagger.version>
```
salvatore-coppola: The comment a few lines below says that all dependency versions live in `bom/pom.xml` and are imported through `dependencyManagement`. Could `swagger.version` and the `swagger-annotations-jakarta` entry move there, so that the root pom keeps following its own rule?

openapi/pom.xml
```diff
@@ -0,0 +1,217 @@
...
+    <name>Kura OpenAPI documentation</name>
+    <dependencies>
+        <dependency>
+            <groupId>org.eclipse.kura</groupId>
+            <artifactId>org.eclipse.kura.rest.provider</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>io.swagger.parser.v3</groupId>
+            <artifactId>swagger-parser-v3</artifactId>
+            <version>2.1.48</version>
```
salvatore-coppola: Same as for the root pom: `swagger-parser-v3` 2.1.48 and `json-schema-validator` 1.5.6 are pinned here instead of in the BOM. Moving them to `bom/pom.xml` keeps the version management in one place.

openapi/pom.xml
```diff
@@ -0,0 +1,217 @@
...
+            <groupId>org.eclipse.kura</groupId>
+            <artifactId>org.eclipse.kura.rest.tamper.detection.provider</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.eclipse.kura</groupId>
+            <artifactId>org.eclipse.kura.rest.inventory.provider</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>io.swagger.core.v3</groupId>
+            <artifactId>swagger-jaxrs2-jakarta</artifactId>
```
salvatore-coppola, line 102: swagger-core/jaxrs2 2.2.55, swagger-parser 2.1.48 and networknt json-schema-validator 1.5.6 (plus Jackson, snakeyaml and classgraph transitively) are new third-party content. They are build/test-only so they most likely fall under "works-with", but they are not listed in `NOTICE.md` and there is no Dash run in the PR. Worth a quick check against the Eclipse IP policy before merging.

bom/pom.xml
```diff
@@ -11,6 +11,7 @@
     Contributors:
      Eurotech
 -->
+<!-- // Content with portions generated by generative AI platform -->
```
salvatore-coppola, line 14: This is the only change in `bom/pom.xml`: the file has no other modification, so this marker line looks like a leftover and could be dropped. Also, `<!-- // ... -->` is an unusual way to write it in XML.

openapi/src/main/java/org/eclipse/kura/openapi/KuraOpenApiReader.java
```diff
@@ -0,0 +1,114 @@
...
+                            && m.getAnnotation(SecurityRequirements.class).value().length == 0)
+                    .filter(m -> m.isAnnotationPresent(Operation.class))
+                    .map(m -> m.getAnnotation(Operation.class).operationId())
+                    .filter(operationId -> !operationId.isBlank())
+                    .collect(Collectors.toSet());
+            result.getPaths().forEach((path, item) -> item.readOperationsMap().forEach((method, operation) -> {
+                if (publicOperations.contains(operation.getOperationId())) {
+                    operation.setSecurity(java.util.Collections.emptyList());
+                }
+                if (path.startsWith("/security/") || path.startsWith("/keystores/")
```
salvatore-coppola: The operationId rewrite is keyed on the hard-coded `/security/` and `/keystores/` prefixes and also overrides explicitly declared ids. The next service that inherits operations from a shared base class will need a reader change to avoid the duplicate-id failure. Two alternatives: detect duplicates generically and rewrite only those, or declare explicit operationIds in the V1/V2 subclasses (which the `GenerateOpenApiTest` uniqueness check would then enforce).

kura/org.eclipse.kura.rest.provider/bnd.bnd
```diff
@@ -1,8 +1,12 @@
+# Copyright (c) 2026 Eurotech and/or its affiliates and others
+# SPDX-License-Identifier: EPL-2.0
+# Content with portions generated by generative AI platform
 Bundle-SymbolicName: ${project.artifactId}
 Bundle-Name: Kura Rest Provider
 Bundle-ActivationPolicy: lazy
 Require-Bundle: org.eclipse.osgitech.rest.servlet.whiteboard;bundle-version="1.2.3"
 Import-Package: \
+    io.swagger.v3.oas.annotations*;resolution:=optional,\
```
salvatore-coppola, line 9: The annotations jar is not shipped in the distribution, so this optional import is never wired at runtime and the JVM silently drops the unresolvable annotations on reflection. That should work, but since it touches every REST bundle it would be good to confirm on a real installation that the bundles resolve and Jersey still registers all resources, and to mention the result in the PR description.

openapi/README.md
```diff
@@ -0,0 +1,82 @@
...
+mvn -pl openapi -am verify
+```
+
+The generated documents are `openapi/target/openapi/openapi.json` and
+`openapi/target/openapi/openapi.yaml`. Both are attached to
+`org.eclipse.kura:org.eclipse.kura.openapi` with classifier `openapi` and their
+respective file extensions during `package`, so `install` and `deploy` publish
+them with the Kura version. Generated files are not committed.
+
+Generation runs at `process-classes`, including with `-DskipTests`. That flag
```
salvatore-coppola, line 21: Worth adding that the tests read `target/openapi/openapi.json`, so running them from an IDE requires a previous `mvn process-classes` (or `verify`) on this module.

kura/org.eclipse.kura.rest.system.provider/src/main/java/org/eclipse/kura/rest/system/SystemRestService.java
```diff
@@ -49,6 +58,14 @@
 import org.osgi.service.component.annotations.Reference;
 import org.osgi.service.component.annotations.ReferenceCardinality;
 import org.osgi.service.component.annotations.ReferencePolicy;
+@Tag(name = "System", description = "Requires rest.system or kura.admin permission.")
```
salvatore-coppola, line 62: Nit: missing blank line between the imports and the class-level annotations.

openapi/src/main/resources/swagger-config.yaml
```diff
@@ -0,0 +1,172 @@
...
+  - org.eclipse.kura.internal.rest.service.listing.provider.RestServiceListingProvider
+  - org.eclipse.kura.internal.rest.cloudconnection.provider.CloudConnectionRestService
+  - org.eclipse.kura.internal.rest.keystore.provider.KeystoreRestServiceV1
+  - org.eclipse.kura.internal.rest.keystore.provider.KeystoreRestServiceV2
+  - org.eclipse.kura.internal.rest.tamper.detection.TamperDetectionRestService
+  - org.eclipse.kura.internal.rest.inventory.InventoryRestService
+readerClass: org.eclipse.kura.openapi.KuraOpenApiReader
+readAllResources: true
+openAPI:
+  openapi: 3.0.3
```
salvatore-coppola, line 21: Not for this PR: OpenAPI 3.1 has a `mutualTLS` security scheme, which would remove the need to describe certificate authentication in prose, and swagger-core 2.2.x can emit 3.1 via the `openapi31` flag. Might be worth a follow-up.

salvatore-coppola (commented):
Overall this looks solid: no functional code is touched in the REST bundles (only annotations, `provided` deps and optional imports), the reflection-based endpoint coverage check plus the embedded HTTP contract tests against the real auth/authorization/audit filters and the Gson serializer make the generated spec trustworthy, and the Jenkins build and Sonar gate are green.

A few non-blocking remarks inline. One more general point: the PR description is still the empty template. Please add the purpose, how to regenerate the spec and where the artifact ends up, plus the result of a check on a real installation (see the comment on `bnd.bnd`).

salvatore-coppola (commented):
I ran a contract test of the generated spec against a real device (Raspberry Pi, Kura 6.0.0-SNAPSHOT from `develop`): built `openapi/target/openapi/openapi.json` from this branch, generated a Python client with openapi-generator-cli 7.25.0 (`-g python`), then called every operation through the client and validated status code and body against the spec (schema plus undocumented-property check, same idea as the embedded tests).

Result: 101/103 operations exercised (the two `security-policy/apply*` ones were skipped on purpose), 104 calls pass, 22 fail. The failures cluster into the six points below, all reproducible on any device except the tamper ones. Inline comments point at the places to fix.

1. `inventory/v1/containers/_start`, `containers/_stop` and `images/_delete` reject the documented bodies with 400; the device unmarshals `{"name", "version"}` (the same shape the GET endpoints return).
2. Inventory GET schemas do not match the JSON marshaller output: `inventory` (not `resources`), `systemPackages` (not `packages`), container items are `{name, version, type, state}`.
3. Keystore entries are polymorphic: certificate entries carry `subjectDN`, `subjectAN`, `issuer`, `startDate`, `expirationDate`, which `EntryInfo` does not declare.
4. 401/403 are documented as "Empty body". That is true only when the client sends no `Accept` header. With `Accept: application/json` (which every generated client sends) the Felix HTTP bridge returns a JSON page `{"servlet", "message", "url", "status"}`. The same page is returned for 404s thrown without an entity (unknown pid in `servicePids/satisfyingReference`, unknown alias in `keystores/*/entries/entry`, unrouted paths), so those do not match the `Error` schema either. The embedded tests do not see this because they set `jersey.config.server.response.setStatusOverSendError=true`, which the real whiteboard does not.
5. The 409 responses of identity v2 have an `{"message": ...}` body, the spec documents them without content.
6. Identity v2 constraints are not documented: identity names must be alphanumeric runs separated by `.` or `_` (3..255 chars), and `configurationComponents` accepts only `AdditionalConfigurations`, `AssignedPermissions`, `PasswordConfiguration` (an unknown name currently yields a 500).

Everything else matched on the device, including the full session flow (login, XSRF token, password change, logout), snapshots with rollback, factory components, the cloud endpoint/pubsub lifecycle, all keystore write operations and identity v1/v2. Happy to share the harness if useful.

openapi/src/main/resources/swagger-config.yaml
```diff
@@ -0,0 +1,172 @@
...
+    schemas:
+      InventoryBundleReference:
+        type: object
+        required: [name]
+        properties:
+          name:
+            type: string
+          version:
+            type: string
+      InventoryContainerReference:
```
salvatore-coppola, line 79: On the device `containers/_start` and `containers/_stop` answer 400 to `{"containerName": ...}`. `InventoryHandlerV1` unmarshals the payload into `DockerContainer` through the JSON marshaller, whose format is `{"name": "<container name>", "version": "<image:tag>"}` (same as the GET output). `{"name":"pr6294-test","version":"nginx:latest"}` is accepted.

openapi/src/main/resources/swagger-config.yaml
```diff
@@ -0,0 +1,172 @@
...
+        type: object
+        required: [containerName]
+        properties:
+          containerName:
+            type: string
+      InventoryImageReference:
```
salvatore-coppola, line 89: Same for `images/_delete`: `{"imageName","imageTag"}` gives 400, `{"name":"hello-world","version":"latest"}` deletes the image (200).

openapi/src/main/resources/swagger-config.yaml
```diff
@@ -0,0 +1,172 @@
...
+            description: Updated or expired session cookie; preserve the cookie jar.
+            schema:
+              type: string
+      BadRequest:
+        description: Invalid JSON, missing required fields, invalid values or a service-level configuration error.
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/Error'
+      Unauthenticated:
```
salvatore-coppola, line 246: "Empty body" holds only when the request has no `Accept` header. With `Accept: application/json` the Felix HTTP bridge (Jetty error handler) answers 401/403 with `{"servlet":"...BridgeActivator$1-...","message":"Unauthorized","url":"/services/...","status":"401"}` and `Content-Type: application/json`. Generated clients always send that header, so the description (and `Forbidden` below) should say that the body is container-generated and unspecified. Alternatively the production whiteboard could set `jersey.config.server.response.setStatusOverSendError=true`, which is what the embedded tests rely on.

openapi/src/main/resources/swagger-config.yaml
```diff
@@ -0,0 +1,172 @@
...
+            description: Present when the Basic authentication provider is enabled.
+            schema:
+              type: string
+      Unauthorized:
+        description: Invalid login credentials or no usable principal/session for this session operation.
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/Error'
+      NotFound:
```
salvatore-coppola, line 268: For 404s thrown without an entity (unknown pid in `servicePids/satisfyingReference`, unknown alias in `keystores/v1|v2/entries/entry`, and unrouted paths) the body is the same Jetty JSON page as for 401/403 when `Accept: application/json` is sent, and empty `text/html` otherwise. Neither matches `Error`. Either document it like the 401 case or give `Error` `additionalProperties: true`.

kura/org.eclipse.kura.rest.inventory.provider/src/main/java/org/eclipse/kura/internal/rest/inventory/InventoryRestService.java
```diff
@@ -79,6 +101,10 @@ public void setInventoryHandlerV1(InventoryHandlerV1 inventoryHandlerV1) {
     @RolesAllowed("inventory")
     @Path("/inventory")
     @Produces(MediaType.APPLICATION_JSON)
+    @Operation(operationId = "get_inventory_v1_inventory", summary = "Read inventory summary",
+            description = "Returns available system, bundle and container resources.")
+    @ApiResponse(responseCode = "200", description = "Requested inventory data.",
+            content = @Content(mediaType = MediaType.APPLICATION_JSON, schema = @Schema(implementation = SystemResourcesInfo.class)))
```
salvatore-coppola: Actual payload is `{"inventory":[{"name","version","type"}]}`; the generated `SystemResourcesInfo` schema says `resources`. The inventory responses go through Kura's JSON marshaller, so the field names come from its format, not from the Java classes.

kura/org.eclipse.kura.rest.inventory.provider/src/main/java/org/eclipse/kura/internal/rest/inventory/InventoryRestService.java
```diff
@@ -172,6 +216,10 @@ public Response getDeploymentPackages() {
     @RolesAllowed("inventory")
     @Path("/systemPackages")
     @Produces(MediaType.APPLICATION_JSON)
+    @Operation(operationId = "get_inventory_v1_systemPackages", summary = "List system packages",
+            description = "Returns installed system packages.")
+    @ApiResponse(responseCode = "200", description = "Requested inventory data.",
+            content = @Content(mediaType = MediaType.APPLICATION_JSON, schema = @Schema(implementation = SystemPackages.class)))
```
salvatore-coppola: Actual payload key is `systemPackages`, the schema says `packages`.

kura/org.eclipse.kura.rest.inventory.provider/src/main/java/org/eclipse/kura/internal/rest/inventory/InventoryRestService.java
```diff
@@ -191,6 +239,10 @@ public Response getSystemPackages() {
     @RolesAllowed("inventory")
     @Path("/containers")
     @Produces(MediaType.APPLICATION_JSON)
+    @Operation(operationId = "get_inventory_v1_containers", summary = "List containers",
+            description = "Returns containers; may be unavailable when orchestration is not installed.")
+    @ApiResponse(responseCode = "200", description = "Requested inventory data.",
+            content = @Content(mediaType = MediaType.APPLICATION_JSON, schema = @Schema(implementation = DockerContainers.class)))
```
salvatore-coppola: Actual container items are `{"name":"pr6294-test","version":"nginx:latest","type":"DOCKER","state":"uninstalled"}`; the generated `DockerContainer` schema lists `containerName`, `containerState`, ports, etc. and has no `state`.

kura/org.eclipse.kura.rest.keystore.provider/src/main/java/org/eclipse/kura/internal/rest/keystore/provider/KeystoreRestService.java
```diff
@@ -67,6 +78,10 @@ public List<KeystoreInfo> listKeystores() {
     @Path("/entries")
     @RolesAllowed("keystores")
     @Produces(MediaType.APPLICATION_JSON)
+    @Operation(summary = "List keystore entries", description = "Returns entries, optionally filtered by keystore PID or alias.")
+    @ApiResponse(responseCode = "200", description = "Requested entries.",
+            content = @Content(mediaType = MediaType.APPLICATION_JSON,
+                    array = @ArraySchema(schema = @Schema(implementation = EntryInfo.class))))
```
salvatore-coppola: Entries are polymorphic: certificate entries on the device carry `subjectDN`, `subjectAN`, `issuer`, `startDate`, `expirationDate` (and key pairs / private keys their own fields), none of which `EntryInfo` declares. A `oneOf` over `CertificateInfo`, `KeyPairInfo` and `PrivateKeyInfo` (or `@Schema(anyOf = ...)`) would describe the real payload; same for the single-entry endpoint below.

kura/org.eclipse.kura.rest.identity.provider/src/main/java/org/eclipse/kura/internal/rest/identity/provider/IdentityRestServiceV2.java
```diff
@@ -122,6 +140,12 @@ public void bindPasswordStrengthVerificationService(
     @RolesAllowed(REST_ROLE_NAME)
     @Path("/identities")
     @Consumes(MediaType.APPLICATION_JSON)
+    @Operation(operationId = "post_identity_v2_identities", summary = "Create an identity",
+            description = "Creates an identity with the supplied name.",
+            requestBody = @RequestBody(required = true, content = @Content(mediaType = MediaType.APPLICATION_JSON,
+                    schema = @Schema(implementation = IdentityDTO.class))))
+    @ApiResponse(responseCode = "200", ref = "#/components/responses/EmptySuccess")
+    @ApiResponse(responseCode = "409", description = "Identity or permission already exists.")
```
salvatore-coppola: The 409 does have a body: `{"message":"Identity already exists"}` (and `Permission already exists` for `/permissions`). It should reference `#/components/schemas/Error` like the other error responses.

kura/org.eclipse.kura.rest.identity.provider/src/main/java/org/eclipse/kura/internal/rest/identity/provider/IdentityRestServiceV2.java
```diff
@@ -165,6 +194,12 @@ public Response updateIdentity(final IdentityConfigurationDTO identityConfigurat
     @Path("/identities/byName")
     @Consumes(MediaType.APPLICATION_JSON)
     @Produces(MediaType.APPLICATION_JSON)
+    @Operation(operationId = "post_identity_v2_identities_byName", summary = "Read an identity configuration",
+            description = "Returns selected configuration components for the named identity.",
+            requestBody = @RequestBody(required = true, content = @Content(mediaType = MediaType.APPLICATION_JSON,
+                    schema = @Schema(implementation = IdentityConfigurationRequestDTO.class))))
+    @ApiResponse(responseCode = "200", description = "Requested data.",
+            content = @Content(mediaType = MediaType.APPLICATION_JSON, schema = @Schema(implementation = IdentityConfigurationDTO.class)))
     public IdentityConfigurationDTO getIdentityByName(
```
salvatore-coppola, line 204: Two constraints worth documenting here and on `IdentityDTO.name`: identity names are validated by `ValidationUtil` as alphanumeric runs separated by `.` or `_`, 3 to 255 characters (`oapi-v2` gets 400, `oapi_v2` works), and `configurationComponents` accepts only `AdditionalConfigurations`, `AssignedPermissions`, `PasswordConfiguration` (`IdentityDTOUtils`), so an `enum` would help; an unknown name currently ends in a 500 rather than a 400.

openapi/src/main/resources/swagger-config.yaml
```diff
@@ -0,0 +1,289 @@
...
+      retain the JSESSIONID cookie, obtain session/v1/xsrfToken, then send both the cookie and
+      X-XSRF-Token on subsequent requests, including GET. Only xsrfToken is exempt from the token requirement.
+      A locked session permits only xsrfToken and changePassword until the password is changed.
+      Stateless client-certificate authentication and certificate session login are also supported
+      when configured on a mutual-TLS port. OpenAPI 3.0 cannot express a mutualTLS security scheme;
+      the security requirements below describe Basic and cookie authentication.
+      The cookie name shown is the servlet default and can depend on the deployment.
+      Missing or invalid authentication returns a 401 after the audit filter and missing permission
+      a 403; these responses carry no error entity, so the body is generated by the servlet container
+      and is unspecified (empty when the client does not accept JSON).
```
salvatore-coppola, line 40: With the filter every error carries an `Error` body.
```suggestion
      Missing or invalid authentication returns a 401 after the audit filter and missing permission
      a 403. Every error response carries a JSON Error body; when an operation provides no specific
      message, the error entity filter sets the HTTP reason phrase.
```

openapi/src/main/resources/swagger-config.yaml
```diff
@@ -0,0 +1,289 @@
...
+            schema:
+              type: string
+      BadRequest:
+        description: Invalid JSON, missing required fields, invalid values or a service-level configuration error.
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/Error'
+      Unauthenticated:
+        description: Missing or invalid authentication, locked or expired session, or missing or invalid XSRF token. No error entity is attached; the body is generated by the servlet container and is unspecified (empty when the client does not accept JSON).
```
salvatore-coppola, line 247:
```suggestion
        description: Missing or invalid authentication, locked or expired session, or missing or invalid XSRF token.
        content:
          application/json:
            schema:
              $ref: '#/components/schemas/Error'
```

openapi/src/main/resources/swagger-config.yaml
```diff
@@ -0,0 +1,289 @@
...
+        headers:
+          WWW-Authenticate:
+            description: Present when the Basic authentication provider is enabled.
+            schema:
+              type: string
+      Forbidden:
+        description: Authenticated identity lacks the required permission. No error entity is attached; the body is generated by the servlet container and is unspecified (empty when the client does not accept JSON).
```
salvatore-coppola, line 254:
```suggestion
        description: Authenticated identity lacks the required permission.
        content:
          application/json:
            schema:
              $ref: '#/components/schemas/Error'
```

openapi/src/main/resources/swagger-config.yaml
```diff
@@ -0,0 +1,289 @@
...
+            schema:
+              type: string
+      Unauthorized:
+        description: Invalid login credentials or no usable principal/session for this session operation.
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/Error'
+      EntitylessNotFound:
+        description: Resource not found and no error entity attached; the body is generated by the servlet container and is unspecified (empty when the client does not accept JSON).
```
salvatore-coppola, line 267: No longer needed: the keystore, tamper detection and service-listing 404s can reference `NotFound` again (four `@ApiResponse` refs, see the patch in the review body).
```suggestion
```

openapi/src/main/resources/swagger-config.yaml
```diff
@@ -0,0 +1,289 @@
...
+      NotFound:
+        description: Resource or snapshot not found. Requests to unrouted paths also fail with a container-generated 404 that carries no error entity.
```
salvatore-coppola, line 269:
```suggestion
        description: Resource or snapshot not found, or no operation matches the request path.
```

openapi/src/main/resources/swagger-config.yaml
```diff
@@ -0,0 +1,289 @@
...
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/Error'
+      Disabled:
+        description: Session management or the requested login mechanism is disabled. No error entity is attached; the body is generated by the servlet container and is unspecified (empty when the client does not accept JSON).
```
salvatore-coppola, line 275:
```suggestion
        description: Session management or the requested login mechanism is disabled.
        content:
          application/json:
            schema:
              $ref: '#/components/schemas/Error'
```

openapi/src/test/java/org/eclipse/kura/openapi/AbstractHttpContractTest.java
```diff
@@ -0,0 +1,352 @@
...
+        final SessionRestService sessionResource = new SessionRestService(this.identities, sessions,
+                mock(ConfigurationAdmin.class), passwords, banners);
+        final RestServiceOptions restOptions = new RestServiceOptions(this.options);
+        sessionResource.setOptions(restOptions);
+        final SessionAuthProvider sessionAuth = new SessionAuthProvider(sessions,
+                Set.of("/session/v1/changePassword", "/session/v1/xsrfToken"), Set.of("/session/v1/xsrfToken"));
+        sessionAuth.setOptions(restOptions);
+
+        final ResourceConfig application = new ResourceConfig();
+        application.property("jersey.config.server.response.setStatusOverSendError", true);
```
salvatore-coppola, line 175: The device cannot set this flag (see the review body), so the embedded server should not either: drop it and register `ErrorEntityFilter` with the other response filters (line 193), so the tests reproduce the real error bodies.
```suggestion
```

openapi/src/test/java/org/eclipse/kura/openapi/AbstractHttpContractTest.java
```diff
@@ -0,0 +1,352 @@
...
+        if (restOptions.isSessionManagementEnabled()) {
+            this.authentication.registerAuthenticationProvider(sessionAuth);
+        }
+        application.register(systemResource);
+        application.register(configurationResource);
+        application.register(sessionResource);
+        registerAdditionalResources(application);
+        application.register(this.authentication);
+        application.register(AuthorizationFilter.class);
+        application.register(AuditFilter.class);
```
salvatore-coppola, line 193:
```suggestion
        application.register(AuditFilter.class);
        application.register(ErrorEntityFilter.class);
```
(plus `import org.eclipse.kura.internal.rest.provider.ErrorEntityFilter;`)

salvatore-coppola (changes requested):
Requesting one change, about the servlet-container error page that the entity-less 401/403/404 responses carry on a real device. **This does not have to land here:** since this PR otherwise ships no runtime behaviour change, the fix below could equally go into a separate follow-up PR (e.g. `fix(rest.provider): attach an Error body to entity-less error responses`), with this PR keeping the current "unspecified body" wording until then. Your call.

**Problem.** When a JAX-RS response has an error status and no entity, Jersey calls `HttpServletResponse.sendError()` and Jetty's `ErrorHandler` renders a page in the format negotiated by `Accept`. With `Accept: application/json`, which every generated client sends, the body is:

```json
{"servlet":"org.apache.felix.http.bridge.internal.BridgeActivator$1-52fc356a","message":"Unauthorized","url":"/services/system/v1/properties/kura","status":"401"}
```

This affects the 401/403 aborted by the auth filters, 404s thrown without an entity, unrouted paths and 405s. The body depends on the container and the `Accept` header, it does not match `Error`, and it leaks the internal bridge servlet class name.

**Why `setStatusOverSendError` cannot be used.** The embedded tests set `jersey.config.server.response.setStatusOverSendError=true`, but on the device there is no way to pass it to Jersey. I tried both options on a Pi:
- *Whiteboard runtime configuration* (`configureDefaultWhiteboard`): the property is stored, but osgitech 1.2.3 builds the implicit `.default` application from `new Application()` with fixed properties (`JerseyServiceRuntime`), so runtime configuration properties never reach the `ResourceConfig`.
- *A `Feature` extension* calling `context.property(...)`: it is registered (`Added content ... to application .default`) but has no effect, because `WebComponent` reads the flag from the `ResourceConfig` it was constructed with, not from the runtime configuration that features modify.

**Proposed fix.** Give every entity-less error an `Error` body, so Jersey writes the response itself and never calls `sendError`. This lives in a dedicated response filter in `org.eclipse.kura.rest.provider`. Its priority is `Priorities.USER - 1`, so it runs after the other response filters, once an unauthenticated 403 has already been turned into a 401, because response filters run in descending priority order:

```java
@Priority(Priorities.USER - 1)
public class ErrorEntityFilter implements ContainerResponseFilter {

    private static final String DEFAULT_MESSAGE = "Request failed";

    @Override
    public void filter(final ContainerRequestContext requestContext, final ContainerResponseContext responseContext) {
        final int status = responseContext.getStatus();

        if (status < 400 || responseContext.hasEntity()) {
            return;
        }

        final Status knownStatus = Status.fromStatusCode(status);
        final String message = knownStatus != null ? knownStatus.getReasonPhrase() : DEFAULT_MESSAGE;

        responseContext.setEntity(Collections.singletonMap("message", message), null, MediaType.APPLICATION_JSON_TYPE);
    }
}
```

It is registered in `RestService.activate()` together with the other response filters:

```java
        this.registeredServices.add(
                bundleContext.registerService(ContainerResponseFilter.class, new ErrorEntityFilter(), serviceProperties));
```

The spec then documents `Error` on `Unauthenticated`, `Forbidden` and `Disabled` (inline suggestions), `EntitylessNotFound` goes away and its four references go back to `NotFound`, and the embedded tests drop `setStatusOverSendError` and register the filter, so they now reproduce the device configuration.

**Verification.**
- Device (Raspberry Pi, `rest.provider` with the filter; the rest of the platform is `develop` of 2026-08-31): 401 with and without `Accept`, 403 from an identity without permission, 404 for an unknown pid and an unrouted path, and 405 all return `application/json` with `{"message": "<reason phrase>"}`. Errors that already carry an entity (e.g. `{"message":"Identity does not exist"}`) are unchanged, and so are 2xx responses and headers.
- `mvn -pl openapi -am verify` on 674fc72ae1 plus the patch below: 84/84 tests green.
- Behaviour change to be aware of: a 401 without `Accept` used to have an empty `text/html` body and now has a JSON body. Worth a quick check with the web console and other known clients before merging, wherever this ends up.

<details><summary>Full patch against 674fc72ae1 (applies with <code>git apply</code>)</summary>

```diff
diff --git a/kura/org.eclipse.kura.rest.keystore.provider/src/main/java/org/eclipse/kura/internal/rest/keystore/provider/KeystoreRestServiceV1.java b/kura/org.eclipse.kura.rest.keystore.provider/src/main/java/org/eclipse/kura/internal/rest/keystore/provider/KeystoreRestServiceV1.java
index 54b0a24188..bb8e826901 100644
--- a/kura/org.eclipse.kura.rest.keystore.provider/src/main/java/org/eclipse/kura/internal/rest/keystore/provider/KeystoreRestServiceV1.java
+++ b/kura/org.eclipse.kura.rest.keystore.provider/src/main/java/org/eclipse/kura/internal/rest/keystore/provider/KeystoreRestServiceV1.java
@@ -26,7 +26,7 @@ import org.osgi.service.component.annotations.Component;
         @ApiResponse(responseCode = "400", ref = "#/components/responses/BadRequest"),
         @ApiResponse(responseCode = "401", ref = "#/components/responses/Unauthenticated"),
         @ApiResponse(responseCode = "403", ref = "#/components/responses/Forbidden"),
-        @ApiResponse(responseCode = "404", ref = "#/components/responses/EntitylessNotFound"),
+        @ApiResponse(responseCode = "404", ref = "#/components/responses/NotFound"),
         @ApiResponse(responseCode = "500", ref = "#/components/responses/InternalError")
 })
 @Path("/keystores/v1")
diff --git a/kura/org.eclipse.kura.rest.keystore.provider/src/main/java/org/eclipse/kura/internal/rest/keystore/provider/KeystoreRestServiceV2.java b/kura/org.eclipse.kura.rest.keystore.provider/src/main/java/org/eclipse/kura/internal/rest/keystore/provider/KeystoreRestServiceV2.java
index 8b44e9a196..64bf47dafc 100644
--- a/kura/org.eclipse.kura.rest.keystore.provider/src/main/java/org/eclipse/kura/internal/rest/keystore/provider/KeystoreRestServiceV2.java
+++ b/kura/org.eclipse.kura.rest.keystore.provider/src/main/java/org/eclipse/kura/internal/rest/keystore/provider/KeystoreRestServiceV2.java
@@ -39,7 +39,7 @@ import org.osgi.service.component.annotations.Component;
         @ApiResponse(responseCode = "400", ref = "#/components/responses/BadRequest"),
         @ApiResponse(responseCode = "401", ref = "#/components/responses/Unauthenticated"),
         @ApiResponse(responseCode = "403", ref = "#/components/responses/Forbidden"),
-        @ApiResponse(responseCode = "404", ref = "#/components/responses/EntitylessNotFound"),
+        @ApiResponse(responseCode = "404", ref = "#/components/responses/NotFound"),
         @ApiResponse(responseCode = "500", ref = "#/components/responses/InternalError")
 })
 @Path("/keystores/v2")
diff --git a/kura/org.eclipse.kura.rest.provider/src/main/java/org/eclipse/kura/internal/rest/provider/ErrorEntityFilter.java b/kura/org.eclipse.kura.rest.provider/src/main/java/org/eclipse/kura/internal/rest/provider/ErrorEntityFilter.java
new file mode 100644
index 0000000000..91383a66fb
--- /dev/null
+++ b/kura/org.eclipse.kura.rest.provider/src/main/java/org/eclipse/kura/internal/rest/provider/ErrorEntityFilter.java
@@ -0,0 +1,44 @@
+/*******************************************************************************
+ * Copyright (c) 2026 Eurotech and/or its affiliates and others
+ *
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *  Eurotech
+ *******************************************************************************/
+// Content with portions generated by generative AI platform
+package org.eclipse.kura.internal.rest.provider;
+
+import java.util.Collections;
+
+import jakarta.annotation.Priority;
+import jakarta.ws.rs.Priorities;
+import jakarta.ws.rs.container.ContainerRequestContext;
+import jakarta.ws.rs.container.ContainerResponseContext;
+import jakarta.ws.rs.container.ContainerResponseFilter;
+import jakarta.ws.rs.core.MediaType;
+import jakarta.ws.rs.core.Response.Status;
+
+@Priority(Priorities.USER - 1)
+public class ErrorEntityFilter implements ContainerResponseFilter {
+
+    private static final String DEFAULT_MESSAGE = "Request failed";
+
+    @Override
+    public void filter(final ContainerRequestContext requestContext, final ContainerResponseContext responseContext) {
+        final int status = responseContext.getStatus();
+
+        if (status < 400 || responseContext.hasEntity()) {
+            return;
+        }
+
+        final Status knownStatus = Status.fromStatusCode(status);
+        final String message = knownStatus != null ? knownStatus.getReasonPhrase() : DEFAULT_MESSAGE;
+
+        responseContext.setEntity(Collections.singletonMap("message", message), null, MediaType.APPLICATION_JSON_TYPE);
+    }
+}
diff --git a/kura/org.eclipse.kura.rest.provider/src/main/java/org/eclipse/kura/internal/rest/provider/RestService.java b/kura/org.eclipse.kura.rest.provider/src/main/java/org/eclipse/kura/internal/rest/provider/RestService.java
index 9e3af9813d..0c475dc24c 100644
--- a/kura/org.eclipse.kura.rest.provider/src/main/java/org/eclipse/kura/internal/rest/provider/RestService.java
+++ b/kura/org.eclipse.kura.rest.provider/src/main/java/org/eclipse/kura/internal/rest/provider/RestService.java
@@ -142,6 +142,8 @@ public class RestService implements ConfigurableComponent {
                 serviceProperties));
         this.registeredServices.add(
                 bundleContext.registerService(ContainerResponseFilter.class, new AuditFilter(), serviceProperties));
+        this.registeredServices.add(
+                bundleContext.registerService(ContainerResponseFilter.class, new ErrorEntityFilter(), serviceProperties));
         this.registeredServices.add(bundleContext.registerService(
                 new String[] { MessageBodyReader.class.getName(), MessageBodyWriter.class.getName() },
                 new GsonSerializer<Object>(), serviceProperties));
diff --git a/kura/org.eclipse.kura.rest.service.listing.provider/src/main/java/org/eclipse/kura/internal/rest/service/listing/provider/RestServiceListingProvider.java b/kura/org.eclipse.kura.rest.service.listing.provider/src/main/java/org/eclipse/kura/internal/rest/service/listing/provider/RestServiceListingProvider.java
index 5c4edcc9da..d792090d08 100644
--- a/kura/org.eclipse.kura.rest.service.listing.provider/src/main/java/org/eclipse/kura/internal/rest/service/listing/provider/RestServiceListingProvider.java
+++ b/kura/org.eclipse.kura.rest.service.listing.provider/src/main/java/org/eclipse/kura/internal/rest/service/listing/provider/RestServiceListingProvider.java
@@ -235,7 +235,7 @@ public class RestServiceListingProvider {
                     schema = @Schema(implementation = RefDTO.class))))
     @ApiResponses({
             @ApiResponse(responseCode = "400", ref = "#/components/responses/BadRequest"),
-            @ApiResponse(responseCode = "404", ref = "#/components/responses/EntitylessNotFound")
+            @ApiResponse(responseCode = "404", ref = "#/components/responses/NotFound")
     })
     @ApiResponse(responseCode = "200", description = "Matching PIDs.",
             content = @Content(mediaType = MediaType.APPLICATION_JSON, schema = @Schema(implementation = PidSet.class)))
diff --git a/kura/org.eclipse.kura.rest.tamper.detection.provider/src/main/java/org/eclipse/kura/internal/rest/tamper/detection/TamperDetectionRestService.java b/kura/org.eclipse.kura.rest.tamper.detection.provider/src/main/java/org/eclipse/kura/internal/rest/tamper/detection/TamperDetectionRestService.java
index 41f1f13957..0d7a45c8e8 100644
--- a/kura/org.eclipse.kura.rest.tamper.detection.provider/src/main/java/org/eclipse/kura/internal/rest/tamper/detection/TamperDetectionRestService.java
+++ b/kura/org.eclipse.kura.rest.tamper.detection.provider/src/main/java/org/eclipse/kura/internal/rest/tamper/detection/TamperDetectionRestService.java
@@ -47,7 +47,7 @@ import org.osgi.service.component.annotations.Reference;
 @ApiResponses({
         @ApiResponse(responseCode = "401", ref = "#/components/responses/Unauthenticated"),
         @ApiResponse(responseCode = "403", ref = "#/components/responses/Forbidden"),
-        @ApiResponse(responseCode = "404", ref = "#/components/responses/EntitylessNotFound"),
+        @ApiResponse(responseCode = "404", ref = "#/components/responses/NotFound"),
         @ApiResponse(responseCode = "500", ref = "#/components/responses/InternalError")
 })
 @Path("/tamper/v1")
diff --git a/openapi/README.md b/openapi/README.md
index f44e4aa699..15bcc441be 100644
--- a/openapi/README.md
+++ b/openapi/README.md
@@ -51,9 +51,9 @@ authenticated principal without a specific REST role. Identity permission
 and password-requirement discovery endpoints have no identity-role restriction.
 Security debug-mode status requires an authenticated principal but no security
 role. Missing or invalid authentication returns a 401 after the audit filter,
-and a missing permission returns a 403; these responses carry no error entity,
-so the body is generated by the servlet container and is unspecified (empty
-when the client does not accept JSON).
+and a missing permission returns a 403. Every error response carries a JSON
+`Error` body: when an operation provides no specific message, the audit filter
+sets the HTTP reason phrase.
 
 The documentation-only reader uses private fields instead of getters to match
 Kura's Gson serialization. Java `Object` values remain unconstrained, including
diff --git a/openapi/src/main/resources/swagger-config.yaml b/openapi/src/main/resources/swagger-config.yaml
index e49511cff6..b2e0c09688 100644
--- a/openapi/src/main/resources/swagger-config.yaml
+++ b/openapi/src/main/resources/swagger-config.yaml
@@ -36,8 +36,8 @@ openAPI:
       the security requirements below describe Basic and cookie authentication.
       The cookie name shown is the servlet default and can depend on the deployment.
       Missing or invalid authentication returns a 401 after the audit filter and missing permission
-      a 403; these responses carry no error entity, so the body is generated by the servlet container
-      and is unspecified (empty when the client does not accept JSON).
+      a 403. Every error response carries a JSON Error body; when an operation provides no specific
+      message, the audit filter sets the HTTP reason phrase.
       Protected operations require their corresponding rest.system, rest.configuration,
       rest.identity, rest.security, rest.cloudconnection, rest.keystores,
       rest.tamper.detection or rest.inventory permission; kura.admin also grants access.
@@ -244,14 +244,22 @@ openAPI:
             schema:
               $ref: '#/components/schemas/Error'
       Unauthenticated:
-        description: Missing or invalid authentication, locked or expired session, or missing or invalid XSRF token. No error entity is attached; the body is generated by the servlet container and is unspecified (empty when the client does not accept JSON).
+        description: Missing or invalid authentication, locked or expired session, or missing or invalid XSRF token.
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/Error'
         headers:
           WWW-Authenticate:
             description: Present when the Basic authentication provider is enabled.
             schema:
               type: string
       Forbidden:
-        description: Authenticated identity lacks the required permission. No error entity is attached; the body is generated by the servlet container and is unspecified (empty when the client does not accept JSON).
+        description: Authenticated identity lacks the required permission.
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/Error'
         headers:
           WWW-Authenticate:
             description: Present when the Basic authentication provider is enabled.
@@ -263,16 +271,18 @@ openAPI:
           application/json:
             schema:
               $ref: '#/components/schemas/Error'
-      EntitylessNotFound:
-        description: Resource not found and no error entity attached; the body is generated by the servlet container and is unspecified (empty when the client does not accept JSON).
       NotFound:
-        description: Resource or snapshot not found. Requests to unrouted paths also fail with a container-generated 404 that carries no error entity.
+        description: Resource or snapshot not found, or no operation matches the request path.
         content:
           application/json:
             schema:
               $ref: '#/components/schemas/Error'
       Disabled:
-        description: Session management or the requested login mechanism is disabled. No error entity is attached; the body is generated by the servlet container and is unspecified (empty when the client does not accept JSON).
+        description: Session management or the requested login mechanism is disabled.
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/Error'
       InternalError:
         description: Internal service error.
         content:
diff --git a/openapi/src/test/java/org/eclipse/kura/openapi/AbstractHttpContractTest.java b/openapi/src/test/java/org/eclipse/kura/openapi/AbstractHttpContractTest.java
index 6acd590ff6..4e2ba7b716 100644
--- a/openapi/src/test/java/org/eclipse/kura/openapi/AbstractHttpContractTest.java
+++ b/openapi/src/test/java/org/eclipse/kura/openapi/AbstractHttpContractTest.java
@@ -67,6 +67,7 @@ import org.eclipse.kura.internal.rest.configuration.ConfigurationRestService;
 import org.eclipse.kura.internal.rest.provider.AuditFilter;
 import org.eclipse.kura.internal.rest.provider.AuthenticationFilter;
 import org.eclipse.kura.internal.rest.provider.AuthorizationFilter;
+import org.eclipse.kura.internal.rest.provider.ErrorEntityFilter;
 import org.eclipse.kura.internal.rest.provider.GsonSerializer;
 import org.eclipse.kura.internal.rest.provider.RestExceptionMapper;
 import org.eclipse.kura.internal.rest.provider.RestServiceOptions;
@@ -172,7 +173,6 @@ public abstract class AbstractHttpContractTest {
         sessionAuth.setOptions(restOptions);
 
         final ResourceConfig application = new ResourceConfig();
-        application.property("jersey.config.server.response.setStatusOverSendError", true);
         final BundleContext bundle = mock(BundleContext.class);
         when(bundle.registerService(eq(ContainerResponseFilter.class), any(ContainerResponseFilter.class), isNull()))
                 .thenAnswer(invocation -> {
@@ -191,6 +191,7 @@ public abstract class AbstractHttpContractTest {
         application.register(this.authentication);
         application.register(AuthorizationFilter.class);
         application.register(AuditFilter.class);
+        application.register(ErrorEntityFilter.class);
         application.register(GsonSerializer.class);
         application.register(RestExceptionMapper.class);
 
diff --git a/openapi/src/test/java/org/eclipse/kura/openapi/KeystoreHttpContractTest.java b/openapi/src/test/java/org/eclipse/kura/openapi/KeystoreHttpContractTest.java
index 2c263c0509..4b51310528 100644
--- a/openapi/src/test/java/org/eclipse/kura/openapi/KeystoreHttpContractTest.java
+++ b/openapi/src/test/java/org/eclipse/kura/openapi/KeystoreHttpContractTest.java
@@ -120,14 +120,13 @@ public class KeystoreHttpContractTest extends AbstractHttpContractTest {
     }
 
     @Test
-    public void unknownEntryYieldsEntitylessNotFound() throws Exception {
+    public void unknownEntryReturnsTheErrorBody() throws Exception {
         givenKeystorePermission();
         givenServer();
 
         whenRequest("GET", "/keystores/v1/entries/entry?keystoreServicePid=" + KEYSTORE_PID + "&alias=missing", null);
 
         thenResponseMatchesContract(404);
-        assertTrue(this.response.body().isEmpty());
     }
 
     private void givenKeystorePermission() {
diff --git a/openapi/src/test/java/org/eclipse/kura/openapi/RemainingServicesContractTest.java b/openapi/src/test/java/org/eclipse/kura/openapi/RemainingServicesContractTest.java
index ce145a541c..6cabefe640 100644
--- a/openapi/src/test/java/org/eclipse/kura/openapi/RemainingServicesContractTest.java
+++ b/openapi/src/test/java/org/eclipse/kura/openapi/RemainingServicesContractTest.java
@@ -69,12 +69,12 @@ public class RemainingServicesContractTest {
     }
 
     @Test
-    public void entitylessFailuresDoNotPromiseAnErrorBody() throws Exception {
+    public void notFoundFailuresReferenceTheErrorResponse() throws Exception {
         givenGeneratedPaths();
 
-        whenEntitylessNotFoundOperationsAreRead();
+        whenNotFoundOperationsAreRead();
 
-        thenEntitylessOperationsReferenceTheContainerGeneratedResponse();
+        thenNotFoundOperationsReferenceTheErrorResponse();
     }
 
     @Test
@@ -105,7 +105,7 @@ public class RemainingServicesContractTest {
         assertFalse(this.paths.path("/keystores/v1/entries").isMissingNode());
     }
 
-    private void whenEntitylessNotFoundOperationsAreRead() {
+    private void whenNotFoundOperationsAreRead() {
         assertFalse(this.paths.path("/keystores/v1/entries/entry").isMissingNode());
     }
 
@@ -168,12 +168,12 @@ public class RemainingServicesContractTest {
         }
     }
 
-    private void thenEntitylessOperationsReferenceTheContainerGeneratedResponse() {
+    private void thenNotFoundOperationsReferenceTheErrorResponse() {
         for (String operation : new String[] { "get:/keystores/v1/entries/entry",
                 "get:/keystores/v2/entries/entry", "get:/tamper/v1/pid/{pid}",
                 "post:/serviceListing/v1/servicePids/satisfyingReference" }) {
             final String[] parts = operation.split(":", 2);
-            assertEquals(operation, "#/components/responses/EntitylessNotFound",
+            assertEquals(operation, "#/components/responses/NotFound",
                     this.paths.path(parts[1]).path(parts[0]).path("responses").path("404").path("$ref").asText());
         }
     }
```

</details>

openapi/pom.xml
salvatore-coppola: Not needed: these are build/test-only dependencies of the `openapi` module and are not shipped in the distribution.

kura/org.eclipse.kura.rest.provider/bnd.bnd
salvatore-coppola: Checked on a Raspberry Pi: the annotated `rest.provider` resolves and serves the REST API with the Swagger packages absent, the optional imports are simply left unwired. (At the current head the bundle only needs a platform built after #6380 for `org.eclipse.kura.util.validation`, unrelated to this change.)

openapi/src/main/resources/swagger-config.yaml
salvatore-coppola: Moved to #6419, which implements the filter and the matching spec changes.

salvatore-coppola (approved):
All review points are addressed: the spec matches the device contract test (103/103 operations exercised on a Raspberry Pi), and the error-body normalization continues in #6419. Approving at c51f8af905.

salvatore-coppola:
@MMaiero one last thought, not blocking: it may be worth exploring whether we could target the latest OpenAPI version directly, [3.2.1](https://spec.openapis.org/oas/v3.2.1.html), instead of 3.0.3. Among other things, it would let us describe certificate authentication with the `mutualTLS` security scheme instead of in prose. This could be a follow-up once this PR is merged.

PR #6406, opened 2026-09-21
fix(rest.configuration.provider): unescape the metatype default values exposed by the REST API

salvatore-coppola:
`AdDTO` copied the `default` attribute of the metatype verbatim. Kura parses that attribute with `StringUtil.splitValues`, which honours the backslash escapes of the Metatype specification, but the REST API exposed the raw string. Nobody noticed while every metatype was a hand-written XML without escapes: the bundles migrated to `@ObjectClassDefinition` now get their metatype generated by bnd, which escapes every space, so `configuration/v2` started returning defaults such as `0\ 0\ 0\ ?\ *\ *\ *` for the connection schedule CRON expression, `Sample\ Banner\ Content` for the login banners or `-tcpPort\ 9123\ -tcpAllowOthers\ -ifExists` for the H2 command line. The `driverOCDs` endpoint of kura-wires goes through the same DTO, which is what makes `WireGraphRestServiceTest` fail in eclipse-kura/kura-wires#23.

The DTO now exposes the default the way `ComponentUtil` applies it: the single unescaped value for scalar attributes, and the unescaped values joined by commas, with literal commas kept escaped, for arrays. Three scenarios in `ConfigurationRestServiceTest` cover the escaped space, the escaped comma in a scalar and the array case.

The Web UI has the same issue on its own path (`GwtServerUtil` and `GwtComponentServiceInternal` pass `ad.getDefault()` to the client), to be handled in kura-management-ui.


salvatore-coppola:
@marcellorinaldo could you prioritize this review ? Thanks

kura/org.eclipse.kura.rest.configuration.provider/src/main/java/org/eclipse/kura/rest/configuration/api/AdDTO.java
```diff
@@ -96,4 +99,22 @@ public boolean isRequired() {
         return isRequired;
     }
 
+    private static String unescapeDefault(final String rawDefault, final int cardinality) {
```
marcellorinaldo: This seems pretty similar to https://github.com/eclipse-kura/kura/blob/3523c1cd09201a3321d419e39352b8dfab25f29d/kura/org.eclipse.kura.core.configuration/src/main/java/org/eclipse/kura/core/configuration/util/ComponentUtil.java#L356

Should we make that public and use it here? I wouldn't mirror the logic that is already avaialble in a utility class

test/org.eclipse.kura.rest.configuration.provider.test/src/main/java/org/eclipse/kura/rest/configuration/provider/test/ConfigurationRestServiceTest.java
```diff
@@ -1093,6 +1094,58 @@ public void testADDefaultValue() throws KuraException {
                 .arrayItem(0).field("defaultValue"));
     }
 
+    @Test
+    public void testADDefaultValueWithEscapedSpacesIsUnescaped() throws KuraException {
```
marcellorinaldo, line 1098: Can we add some tests to cover the other cardinality cases?

kura/org.eclipse.kura.rest.configuration.provider/src/main/java/org/eclipse/kura/rest/configuration/api/AdDTO.java
```diff
@@ -96,4 +99,22 @@ public boolean isRequired() {
...
+        if (rawDefault == null) {
+            return null;
+        }
+
+        final String[] values = StringUtil.splitValues(rawDefault);
+
+        if (values.length == 0) {
```
marcellorinaldo: Is this code reachable?

```
if (values.length == 0) {
```

We have that:
```
if (rawDefault == null) {
    return null;
}
```

Then:
```
final String[] values = StringUtil.splitValues(rawDefault);
```

Can values here be of length 0?

kura/org.eclipse.kura.rest.configuration.provider/src/main/java/org/eclipse/kura/rest/configuration/api/AdDTO.java
salvatore-coppola: Good point, the duplicated part is now in `ComponentUtil`: `getDefaultValues(AD)` splits the raw default and `isSingleValued(AD)` holds the cardinality rule, and both `getDefaultValue` and `AdDTO` use them. I didn't make `getDefaultValue` itself public because it converts the values to the typed objects of the attribute and, for passwords, encrypts them with the `CryptoService` taken from the `ComponentContext`; the DTO needs the plain unescaped string and has no component context.

test/org.eclipse.kura.rest.configuration.provider.test/src/main/java/org/eclipse/kura/rest/configuration/provider/test/ConfigurationRestServiceTest.java
salvatore-coppola: Added scenarios for cardinality `1` and `-1` (single unescaped value), `-3` (vector, separators and escaped commas kept), an array with a single element, and a scalar whose default is an unescaped list (only the first token is exposed, as `ComponentUtil` does when it applies it).

kura/org.eclipse.kura.rest.configuration.provider/src/main/java/org/eclipse/kura/rest/configuration/api/AdDTO.java
salvatore-coppola: You're right, `StringUtil.splitValues` always returns at least one token for a non-null input, so that branch was dead. Removed.

marcellorinaldo (approved):

